PR #82, opened 2018-10-10
Introduced option for berry_phase(...) to not close the integration contour

tfrederiksen:
In certain situations it may be relevant to allow for integrations of the Berry connection over _open contours_.

For instance, the **Zak phase** [PRL 62, 2747 (1989)], relevant to classify inversion-symmetric one-dimensional topological insulators, is computed as a one-dimensional integral of the Berry connection over the first Brillouin zone, i.e., for _k_ over the interval [-0.5, 0.5].

I thus propose an option to skip closing the integration contour.

### DIFF
--- a/sisl/physics/electron.py
+++ b/sisl/physics/electron.py
@@ -713,7 +713,7 @@ def _inv_eff_mass_tensor_ortho(state, ddHk, degenerate, as_matrix):
     return M * _inv_eff_mass_const
 
 
-def berry_phase(bz_loop, sub=None, eigvals=False, _gauge='r'):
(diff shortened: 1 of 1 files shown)


tfrederiksen:
A related question: How is spin handled in `berry_phase()`? For a spin-polarized Hamiltonian, shouldn't the spin channel not be a keyword?

zerothi:
@tfrederiksen thanks for the PR.

As for the spin component, yes currently it only handles the first spin component. However, it should be easy to split up the Hamiltonian in two and perform it twice. I would be hesitant to put the `spin` keyword as an argument to `berry_phase` because it isn't really an inherent property of the calculation of the Berry phase. I haven't thought this through yet ;)

PR #84, opened 2018-10-11
Zak phase from closed-loop integration

tfrederiksen:
The discretized Zak phase over [0, 2\pi/a] (1BZ) can be written as
```
\phi = - Im \ln [<0|1><1|2>...<N-1|N>]
= - Im \ln [<0|1><1|2>...<N-1|exp(i 2\pi/a x)|0>]
```
The Bloch phases in the latter form, which explicitly introduce a position-dependence of the orbitals in the unit cell, allows to close the integration loop (even in 1D). This route is advantageous over an "open-loop integration" because it avoids possible arbitrary phases associated with the diagonalization for the states |0> and |N>. In other words, arbitrary phases cancel for pairs |i><i|, but this is not necessarily guaranteed for the combination |N><0|.

### DIFF
--- a/sisl/physics/electron.py
+++ b/sisl/physics/electron.py
@@ -714,7 +714,7 @@ def _inv_eff_mass_tensor_ortho(state, ddHk, degenerate, as_matrix):
     return M * _inv_eff_mass_const
 
 
-def berry_phase(contour, sub=None, eigvals=False, closed=True):
+def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
     r""" Calculate the Berry-phase on a loop using a predefined path
 
     The Berry phase for a single Bloch state is calculated using the discretized formula:
@@ -735,7 +735,12 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True):
     eigvals : bool, optional
        return the eigenvalues of the product of the overlap matrices
     closed : bool, optional
-       whether or not to include the connection of the last and first points in the loop (unwanted for the Zak phase)
+       whether or not to include the connection of the last and first points in the loop
+    method : string, optional
+       'berry' will return the usual integral of the Berry connection over the specified contour
+       'zak' will compute the Zak phase for 1D systems by performing a closed loop integration but
+       taking into account the Bloch factor exp(-i 2\pi/a x) accumulated over a Brillouin zone,
+       see J. Zak, "Berry's phase for energy bands in solids" PRL 62, 2747 (1989).
 
     Notes
     -----
@@ -780,6 +785,9 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True):
         # When the user has the contour points closed, we don't need to do this in the below loop
         closed = False
 
+    if method == 'zak':
+        closed = True
+
     # Whether we should calculate the eigenvalues of the overlap matrix
     if eigvals:
         # We calculate the final eigenvalues
@@ -808,6 +816,14 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True):
                 prd = _process(prd, prev.inner(second, False))
                 prev = second
 
+            # Insert Bloch phase for 1D integral?
+            if method == 'zak':
+                g = contour.parent.geometry
+                axis = contour.k[1] - contour.k[0]
+                axis /= np.linalg.norm(axis)
+                phase = dot(g.xyz[g.o2a(_a.arangei(g.no)), :], dot(axis, g.rcell)).reshape(1, -1)
+                prev.state *= np.exp(-1j*phase)
+
             # Complete the loop
             if closed:
                 # Include last-to-first segment
@@ -825,8 +841,13 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True):
                 second.change_gauge('r')
                 prd = _process(prd, prev.inner(second, False))
                 prev = second
+            if method == 'zak':
+                g = contour.parent.geometry
+                axis = contour.k[1] - contour.k[0]
+                axis /= np.linalg.norm(axis)
+                phase = dot(g.xyz[g.o2a(_a.arangei(g.no)), :], dot(axis, g.rcell)).reshape(1, -1)
+                prev.state *= np.exp(-1j*phase)
             if closed:
-                # Include last-to-first segment
                 prd = _process(prd, prev.inner(first, False))
             return prd
 
--- a/sisl/physics/electron.py
+++ b/sisl/physics/electron.py
@@ -736,10 +736,10 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
        return the eigenvalues of the product of the overlap matrices
     closed : bool, optional
        whether or not to include the connection of the last and first points in the loop
-    method : string, optional
+    method : {'berry', 'zak'}
        'berry' will return the usual integral of the Berry connection over the specified contour
        'zak' will compute the Zak phase for 1D systems by performing a closed loop integration but
-       taking into account the Bloch factor exp(-i 2\pi/a x) accumulated over a Brillouin zone,
+       taking into account the Bloch factor :math:`e^{-i2\pi/a x}` accumulated over a Brillouin zone,
        see J. Zak, "Berry's phase for energy bands in solids" PRL 62, 2747 (1989).
 
     Notes
@@ -820,7 +820,7 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
             if method == 'zak':
                 g = contour.parent.geometry
                 axis = contour.k[1] - contour.k[0]
-                axis /= np.linalg.norm(axis)
+                axis /= (axis ** 2).sum() ** 0.5
                 phase = dot(g.xyz[g.o2a(_a.arangei(g.no)), :], dot(axis, g.rcell)).reshape(1, -1)
                 prev.state *= np.exp(-1j*phase)
 
@@ -844,7 +844,7 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
             if method == 'zak':
                 g = contour.parent.geometry
                 axis = contour.k[1] - contour.k[0]
-                axis /= np.linalg.norm(axis)
+                axis /= (axis ** 2).sum() ** 0.5
                 phase = dot(g.xyz[g.o2a(_a.arangei(g.no)), :], dot(axis, g.rcell)).reshape(1, -1)
                 prev.state *= np.exp(-1j*phase)
             if closed:
--- a/sisl/physics/tests/test_hamiltonian.py
+++ b/sisl/physics/tests/test_hamiltonian.py
@@ -615,6 +615,19 @@ class TestHamiltonian(object):
         bz2 = BandStructure.param_circle(H, 20, 0.01, [0, 0, 1], [1/3] * 3, loop=True)
         assert np.allclose(elec.berry_phase(bz1), elec.berry_phase(bz2))
 
+    def test_berry_phase_zak(self):
+        # SSH model, topological cell
+        g = Geometry([[-.6, 0, 0], [0.6, 0, 0]], Atom(1, 1.001), sc=[2, 10, 10])
+        g.set_nsc([3, 1, 1])
+        H = Hamiltonian(g)
+        H.construct([(0.1, 1.0, 1.5), (0, 1., 0.5)])
+        # Contour
+        k = np.linspace(0.0, 1.0, 101)
+        K = np.zeros([k.size, 3])
+        K[:, 0] = k
+        bz = BrillouinZone(H, K)
+        assert np.allclose(np.abs(elec.berry_phase(bz, sub=0, method='zak')), np.pi)
+
     def test_gauge_inv_eff(self, setup):
         R, param = [0.1, 1.5], [1., 0.1]
         g = setup.g.tile(2, 0).tile(2, 1).tile(2, 2)
--- a/sisl/physics/tests/test_hamiltonian.py
+++ b/sisl/physics/tests/test_hamiltonian.py
@@ -5,7 +5,7 @@ import pytest
 import warnings
 import numpy as np
 
-from sisl import Geometry, Atom, SuperCell, Hamiltonian, Spin, BandStructure, MonkhorstPack
+from sisl import Geometry, Atom, SuperCell, Hamiltonian, Spin, BandStructure, MonkhorstPack, BrillouinZone
 from sisl import Grid, SphericalOrbital, SislError
 from sisl import electron as elec
 


zerothi:
I am not exactly sure what this means in all cases. ;) I am no expert in Berry-phases, nor Zak...

I have now amended some corrections for berry-phase and in particular when performing inner products, please update sisl, and rebase this branch to the master.
This should rotate states to align, however there are still "problems" in some cases. This will only remove sign-changes on states.
In my snippet you'll also notice how to do the xyz phase for arbitrary number of orbitals.

I have created this small snippet which should demonstrate a few details that I think needs ironing out.
```python
import numpy as np
import sisl as si

_a = si._array
dot = np.dot
svd_destroy = si.linalg.svd_destroy
det_destroy = si.linalg.det_destroy
angle = np.angle

def zak_phase(contour, sub=None, eigvals=False, closed=True):
    Hamiltonian = si.Hamiltonian
    # Currently we require the Berry phase calculation to *only* accept Hamiltonians
    if not isinstance(contour.parent, Hamiltonian):
        raise si.SislError('berry_phase: requires the Brillouin zone object to contain a Hamiltonian!')

    if not contour.parent.orthogonal:
        raise si.SislError('berry_phase: requires the Hamiltonian to use an orthogonal basis!')

    if np.allclose(contour.k[0, :], contour.k[-1, :]):
        # When the user has the contour points closed, we don't need to do this in the below loop
        if closed:
            si.info('changing')
            closed = False

    # Whether we should calculate the eigenvalues of the overlap matrix
    if eigvals:
        # We calculate the final eigenvalues
        def _process(prd, ovr):
            U, _, V = svd_destroy(ovr)
            return dot(prd, dot(U, V))
    else:
        # We calculate the final angle from the determinant
        _process = dot

    if sub is None:
        def _berry(eigenstates):
            # Grab the first one to be able to form a loop
            first = next(eigenstates)
            first.change_gauge('r')
            # Create a variable to keep track of the previous state
            prev = first

            # Initialize the consecutive product
            # Starting with the identity matrix!
            prd = 1

            # Loop remaining eigenstates
            for second in eigenstates:
                second.change_gauge('r')
                prd = _process(prd, prev.inner(second, False))
                prev = second

            # Complete the loop
            if closed:
                # Include last-to-first segment
                g = contour.parent.geometry
                phase = dot(g.xyz[g.o2a(_a.arangei(g.no)), :], dot([1, 0, 0], g.rcell)).reshape(1, -1)
                prev.state *= phase
                prd = _process(prd, prev.inner(first, False))
            return prd

    else:
        def _berry(eigenstates):
            first = next(eigenstates).sub(sub)
            first.change_gauge('r')
            prev = first
            prd = 1
            for second in eigenstates:
                second = second.sub(sub)
                second.change_gauge('r')
                prd = _process(prd, prev.inner(second, False))
                prev = second
            if closed:
                # Include last-to-first segment
                g = contour.parent.geometry
                phase = dot(g.xyz[g.o2a(_a.arangei(g.no)), :], dot([1, 0, 0], g.rcell)).reshape(1, -1)
                prev.state *= phase
                prd = _process(prd, prev.inner(first, False))
            return prd

    # Do the actual calculation of the final matrix
    d = _berry(contour.asyield().eigenstate())

    # Correct return values
    if eigvals:
        ret = -angle(eigvals_destroy(d))
        ret = sort(ret)
    else:
        ret = -angle(det_destroy(d))

    return ret


def calc(k):
    H = create_H()
    K = np.zeros([k.size, 3])
    K[:, 0] = k
    bz = si.BrillouinZone(H, K)
    eig = bz.eigh()

    closed = 1 - (k[-1] - k[0]) > (k[1] - k[0]) * 0.5
    print('  [{}, {}] {}'.format(k[0], k[-1], k[1] - k[0]), closed)
    print(si.electron.berry_phase(bz, sub=0, closed=closed))
    print(zak_phase(bz, sub=0, closed=closed))
    H._geometry = H.geometry.translate([0.1, 0, 0])
    bz = si.BrillouinZone(H, K)
    print(si.electron.berry_phase(bz, sub=0, closed=closed))
    print(zak_phase(bz, sub=0, closed=closed))

def create_H():
    g = si.Geometry([0]*3, si.Atom(1, 1.001), sc=[1, 10, 10]).tile(2, 0)
    g.set_nsc([3, 1, 1])
    H = si.Hamiltonian(g)
    H.construct([(0.1, 1.1), (0, 1.)])
    H[0, 0] = 0.1
    H[1, 1] = -H[0, 0]
    return H


N = 101
calc(np.linspace(-0.5, 0.5, N))
calc(np.linspace(-0.5, 0.5, N-1, endpoint=False))
calc(np.linspace(0, 1, N))
calc(np.linspace(0, 1, N-1, endpoint=False))
```

The snippet shows that by translating a simple system one can get *very* different results. Could you comment on this? Also using a closed vs. a non-closed contour does not give the same result?
Is my Hamiltonian even applicable? ;)

Some general considerations before merging:
1. It seems to me that the phase accummulation should only be along the 1D path chosen, i.e. `k=[1, 0, 0]` and not `k=[1, 1, 1]` for the first lattice vector?
2. Instead of using `zak=...`, I would prefer `method='berry'|'zak'` such that additional methods may be added, if so be.
3. Add a test where calculating the Zak phase is done in `sisl/physics/tests/test_hamiltonian.py`, here it would be nice if there is a calculation of the closed (with end-points) compared to the closed (without end-point), they should be equivalent, no (see however the above snippet).
4. A reference to the Zak paper would probably be nice ;)
5. There is no check when `zak=True` whether the contour actually holds the entire BZ? I.e. should there be a check against `contour.k[0, :] ~ contour.k[-1, :] - 1`, or something similar?

tfrederiksen:
Please hang on, trying to learn how to rebase my branch! This is not the merge I want to propose now.

tfrederiksen:
Here a modified snippet to test the implementation on the classic SSH model, where the Zak phase should be 0 or pi (mod 2pi) when the cell origo is centered at the inversion symmetric point:
```
import numpy as np
import sisl as si

def calc(k):
    print('Trivial:')
    calcd(k, 0.1)
    print('Topological:')
    calcd(k, -0.1)

def calcd(k, d, axis=1):
    H = create_H(d, axis=axis)
    K = np.zeros([k.size, 3])
    K[:, axis] = k
    bz = si.BrillouinZone(H, K)
    eig = bz.eigh()
    closed = 1 - (k[-1] - k[0]) > (k[1] - k[0]) * 0.5
    print('  [{}, {}] {}'.format(k[0], k[-1], k[1] - k[0]), closed)
    if not closed:
        print('berry:', si.electron.berry_phase(bz, sub=0, closed=closed))
    print('zak', si.electron.berry_phase(bz, sub=0, method='zak'))
    dx = [0, 0, 0]
    dx[axis] = 0.1
    H._geometry = H.geometry.translate(dx)
    bz = si.BrillouinZone(H, K)
    if not closed:
        print('berry-transl:', si.electron.berry_phase(bz, sub=0, closed=closed))
    print('zak-transl:', si.electron.berry_phase(bz, sub=0, method='zak'))

def create_H(d=0.1, axis=0):
    x = np.zeros([2, 3])
    x[:, axis] = [-.5+d, 0.5-d]
    sc = [10, 10, 10]
    sc[axis] = 2
    g = si.Geometry(x, si.Atom(1, 1.001), sc=sc)
    nsc = [1, 1, 1]
    nsc[axis] = 3
    g.set_nsc(nsc)
    H = si.Hamiltonian(g)
    H.construct([(0.1, 1.0, 1.5), (0, 1., 0.5)])
    return H

N = 101
calc(np.linspace(-0.5, 0.5, N))
calc(np.linspace(-0.5, 0.5, N-1, endpoint=False))
calc(np.linspace(0, 1, N))
calc(np.linspace(0, 1, N-1, endpoint=False))
calc(np.linspace(-.2, 0.8, N))
calc(np.linspace(-0.5, 0.3, N))
```
which gives me
```
Trivial:
('  [-0.5, 0.5] 0.01', False)
('berry:', 4.1957250168020507e-16)
('zak', 3.9181692606457616e-16)
('berry-transl:', -0.31415926535897937)
('zak-transl:', -0.31415926535897937)
Topological:
('  [-0.5, 0.5] 0.01', False)
('berry:', 3.141592653589793)
('zak', -3.141592653589793)
('berry-transl:', 2.8274333882308142)
('zak-transl:', 2.8274333882308142)
Trivial:
('  [-0.5, 0.49] 0.01', True)
('zak', 2.538589926132333e-16)
('zak-transl:', -0.31415926535897953)
Topological:
('  [-0.5, 0.49] 0.01', True)
('zak', -3.141592653589793)
('zak-transl:', 2.8274333882308142)
Trivial:
('  [0.0, 1.0] 0.01', False)
('berry:', -6.699056749515879e-17)
('zak', -2.7738256028064404e-17)
('berry-transl:', -0.31415926535897903)
('zak-transl:', -0.314159265358979)
Topological:
('  [0.0, 1.0] 0.01', False)
('berry:', -3.1415926535897927)
('zak', -3.1415926535897927)
('berry-transl:', 2.8274333882308142)
('zak-transl:', 2.8274333882308142)
Trivial:
('  [0.0, 0.99] 0.01', True)
('zak', -4.759856111498124e-17)
('zak-transl:', -0.3141592653589791)
Topological:
('  [0.0, 0.99] 0.01', True)
('zak', -3.1415926535897927)
('zak-transl:', 2.8274333882308142)
Trivial:
('  [-0.2, 0.8] 0.01', False)
('berry:', 3.2084956010839176e-16)
('zak', 3.7636071133964964e-16)
('berry-transl:', -0.3141592653589794)
('zak-transl:', -0.3141592653589795)
Topological:
('  [-0.2, 0.8] 0.01', False)
('berry:', -3.1415926535897927)
('zak', -3.1415926535897927)
('berry-transl:', 2.8274333882308134)
('zak-transl:', 2.827433388230814)
Trivial:
('  [-0.5, 0.3] 0.008', True)
('zak', 4.795710769449784e-16)
('zak-transl:', -0.31415926535898003)
Topological:
('  [-0.5, 0.3] 0.008', True)
('zak', -3.141592653589793)
('zak-transl:', 2.827433388230814)
```
Note that the result is unchanged if we change the orientation of the 1D system (setting `axis=0` or `axis=2`).

sisl/physics/electron.py
```diff
@@ -735,7 +735,12 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True):
     eigvals : bool, optional
        return the eigenvalues of the product of the overlap matrices
     closed : bool, optional
-       whether or not to include the connection of the last and first points in the loop (unwanted for the Zak phase)
+       whether or not to include the connection of the last and first points in the loop
+    method : string, optional
```
zerothi: when string options are in play one should typically do:

`method : {'berry', 'zak'}`

where the first entry is the default value. This is according to the numpy doc-standards.

sisl/physics/electron.py
```diff
@@ -735,7 +735,12 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True):
...
+       'berry' will return the usual integral of the Berry connection over the specified contour
+       'zak' will compute the Zak phase for 1D systems by performing a closed loop integration but
+       taking into account the Bloch factor exp(-i 2\pi/a x) accumulated over a Brillouin zone,
```
zerothi: Here you may change exp(...) to
```
:math:`e^{-i2\pi/a x}`
```
to show in LaTeX format.

sisl/physics/electron.py
```diff
@@ -780,6 +785,9 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True):
         # When the user has the contour points closed, we don't need to do this in the below loop
         closed = False
 
+    if method == 'zak':
+        closed = True
```
zerothi: Is it necessary to check that the `k[-1] - k[0] ~ 1` because otherwise the closed contour doesn't make sense?

sisl/physics/electron.py
```diff
@@ -808,6 +816,14 @@ def _berry(eigenstates):
                 prd = _process(prd, prev.inner(second, False))
                 prev = second
 
+            # Insert Bloch phase for 1D integral?
+            if method == 'zak':
+                g = contour.parent.geometry
+                axis = contour.k[1] - contour.k[0]
+                axis /= np.linalg.norm(axis)
```
zerothi: You don't have to do this, I will change it after merging, but `np.linalg.norm` is horrendously slow, therefore I always do: `(axis ** 2).sum() ** 0.5`.

The `norm` function was really a performance bottleneck in `Hamiltonian.construct` for very large systems.

sisl/physics/tests/test_hamiltonian.py
```diff
@@ -615,6 +615,19 @@ def test_berry_phase_loop(self, setup):
         bz2 = BandStructure.param_circle(H, 20, 0.01, [0, 0, 1], [1/3] * 3, loop=True)
         assert np.allclose(elec.berry_phase(bz1), elec.berry_phase(bz2))
 
+    def test_berry_phase_zak(self):
```
zerothi: Great!

tfrederiksen: How do I run this test?

sisl/physics/electron.py
tfrederiksen: OK, good to know! But axis is just a 3-element vector so performance is probably not affected by this.

sisl/physics/tests/test_hamiltonian.py
zerothi: You install sisl locally, then open a new terminal (pytest doesn't work while running in the sisl directory).

Then do:
```bash
pytest --pyargs sisl -m hamiltonian
```
Which will only run the tests marked with `hamiltonian`. Alternatively, to test only one test I typically do:

```python
@pytest.mark.only
def test_berry_phase_zak(self):
    ...
```
Then you can do:
```bash
pytest --pyargs sisl -m only
```
to only run that test (or if you mark more as `only`).

sisl/physics/electron.py
tfrederiksen: I think it is not necessary. If you check my snippet above one gets also the correct results by sampling just the interval `[-0.5, 0.3]`. A bit of magic here...

sisl/physics/electron.py
zerothi: Agreed, no need to change (I'll do it after merge)

sisl/physics/electron.py
zerothi: Ok, but is that a coincidence or? It truly seems like magic ;)

sisl/physics/electron.py
tfrederiksen: OK, I will change this.

sisl/physics/electron.py
tfrederiksen: Yes, much better.

sisl/physics/electron.py
tfrederiksen: I will change this too

sisl/physics/electron.py
zerothi: What is your idea about this? Do you think it generically holds, or?

tfrederiksen: Not a coincidence, I think. It seems that the discrete formulation does not rely on having linearly-spaced contour points. And with the trick of the phase factor + going back to first point, one ends up integrating the full 1BZ.

Further, I find that the computed phases do not really change if I reduce the contour points down to only `N=4` in the snippet above. Going lower, the code fails to execute for other reasons.

In fact, in R. Resta, J. Phys.: Condens. Matter 12, R107 (2000) [http://iopscience.iop.org/article/10.1088/0953-8984/12/9/201/meta](url) points out relevance of the extreme case of _single-point Berry's phase_. This could perhaps be an interesting feature for later.

sisl/physics/tests/test_hamiltonian.py
tfrederiksen: Thanks. Running the test I found a namespace issue and fixed it.

sisl/physics/electron.py
zerothi: Yes, ok, they also seem to discuss why the path need not be regularly spaced.

PR #94, opened 2018-11-30
Added option to plot atom indices in sisl.plot(Geometry)

jonaslb:
Added the `atom_indices` keyword to `Geometry.__plot__`. It puts little numbers on the atoms to indicate their index. 
It is a simple way of ensuring that the indices are as expected when creating a structure with sisl.
It is also only three lines, but I thought it'd be a useful addition still.

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -2827,7 +2827,8 @@ class Geometry(SuperCellChild):
         """
         return self.sc.offset(self.o2isc(o))
 
-    def __plot__(self, axis=None, supercell=True, axes=False, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


sisl/geometry.py
```diff
@@ -2881,6 +2882,10 @@ def __plot__(self, axis=None, supercell=True, axes=False, *args, **kwargs):
         else:
             axes.scatter(xyz[:, axis[0]], xyz[:, axis[1]], s=area, c=colors, alpha=0.8)
 
+        if atom_indices:
```
zerothi: It should be move up into the `else` clause

Secondly, it needs to be duplicated for the 3D plot case.

PR #99, opened 2018-12-10
Convert `sc` to `cell`

### LINKED ISSUE #95: sc arguments should be named cell
**Describe the bug**
`sc` keyword arguments are used in a few places.

All these should be named `cell` to make it clear what it is.

The main problem is that `sc` has been used since the beginning, so probably we need to have it allowed (until it truly becomes deprecated)

The following should be changed:

- [x] Change `SuperCell` to `Cell`
- [x] Change `sc` arguments to `cell`
- [x] Allow `sc` arguments via `**kwargs`` for fall-back and backward compatibility

Now I have a problem with the attached supercell instances. E.g. `Geometry.sc` (and all the other classes) is now a bit misleading.

How should this object be renamed (**HELP WANTED**):
- `obj.abc` for a-b-c lattice vectors
- `obj.uc` for unit-cell
- `obj.c` for cell
- `obj.cell`, this is the only thing that won't work because `obj.cell` already refers to the cell-vectors
- `obj.*` other suggestions?

zerothi:
This fixes #95


zerothi:
This will be superseeded by a change of sc to lattice, as per discussion.

Issue #87, opened 2018-10-24
Append Hamiltonians


zerothi:
**Describe the feature**
We want to be able to append sparse matrices, and for the more specialized ones one could imagine taking the supercell connection for one and using that as the $V$ between the attached version.

PR #103, opened 2019-01-15
Initial implementation of the sparse-orbital.append

zerothi:
It currently works as expected. However, it re-uses the elements from both sparse geometries and thus one cannot ensure a Hermitian matrix.

For this the `make_hermitian` method should be implemented (the old one did not work)


zerothi:
Fixes #87

PR #105, opened 2019-01-23
Minor fix on rotate documentation: degrees/rads

jonaslb:
This makes the documentation clear on whether degrees or radians are used. (yes, I had a small 'accident' because of it)


zerothi:
Thanks!!!

PR #113, opened 2019-02-26
Added some noisy files to gitignore

jonaslb:
Eg. when using `setup.py develop`, a number of files will be laying around.

On a side note: Should the cython-generated C-files not also be in gitignore?


zerothi:
Great thanks!

No, currently I ship the sources together with the source since this removes the Cython dependency for users. So for now I also have them in git. :)

Issue #112, opened 2019-02-24
Add tests for pathlib


zerothi:
**Describe the issue**

Add tests for pathlib (in `sisl/io/tests/test_object.py`)

zerothi:
@jonaslb

PR #114, opened 2019-02-26
WIP: All siles now supports pathlib paths; tests added

jonaslb:
At least they support it as far as the tests show. I guess that this should be tested on older Pythons (3.5, 2.7) to ensure it works there as well, but it should work, as far as I can see.

### DIFF
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -3,11 +3,7 @@ from __future__ import print_function, division
 from functools import wraps
 from os.path import splitext, isfile, dirname, join, abspath, basename
 import gzip
-try:
-    from pathlib import Path
-except ImportError:  # Ancient Python
-    class Path:
-        pass
+from pathlib import Path
 
 import numpy as np
 
@@ -338,7 +334,7 @@ class BaseSile(object):
         """ File of the current `Sile` """
         if filename is None:
             filename = Path(self._file).name
-        return Path(self._directory)/filename
+        return Path(self._directory) / filename
 
     def exist(self):
         """ Query whether the file exists """
@@ -511,11 +507,10 @@ class Sile(BaseSile):
         self._base_setup(*args, **kwargs)
 
     def _open(self):
-        file = str(self.file)
-        if file.endswith('gz'):
-            self.fh = gzip.open(file)
+        if self.file.suffix == ".gz":
+            self.fh = gzip.open(str(self.file))
         else:
-            self.fh = open(file, self._mode)
+            self.fh = self.file.open(self._mode)
         self._line = 0
 
     def __enter__(self):
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -337,8 +337,8 @@ class BaseSile(object):
     def dir_file(self, filename=None):
         """ File of the current `Sile` """
         if filename is None:
-            filename = basename(self._file)
-        return join(self._directory, filename)
+            filename = Path(self._file).name
+        return Path(self._directory)/filename
 
     def exist(self):
         """ Query whether the file exists """
@@ -408,12 +408,12 @@ class BaseSile(object):
         base = kwargs.get('base', None)
         if base is None:
             # Extract from filename
-            self._directory = dirname(self._file)
+            self._directory = Path(self._file).parent
         else:
             self._directory = base
-        if len(self._directory) == 0:
+        if not str(self._directory):
             self._directory = '.'
-        self._directory = abspath(self._directory)
+        self._directory = Path(self._directory).resolve()
 
         self._setup(*args, **kwargs)
 
@@ -455,7 +455,7 @@ class BaseSile(object):
 
     def __str__(self):
         """ Return a representation of the `Sile` """
-        return ''.join([self.__class__.__name__, '(', self.base_file, ', base=', self._directory, ')'])
+        return "{0}({1!s}, base={2!s})".format(self.__class__.__name__, self.base_file, self._directory)
 
 
 def sile_fh_open(from_closed=False):
@@ -497,7 +497,7 @@ class Sile(BaseSile):
     """
 
     def __init__(self, filename, mode='r', comment=None, *args, **kwargs):
-        self._file = filename
+        self._file = Path(filename)
         self._mode = mode
         if isinstance(comment, (list, tuple)):
             self._comment = list(comment)
@@ -681,7 +681,7 @@ class SileCDF(BaseSile):
     """
 
     def __init__(self, filename, mode='r', lvl=0, access=1, *args, **kwargs):
-        self._file = filename
+        self._file = Path(filename)
         # Open mode
         self._mode = mode
         # Save compression internally
@@ -924,7 +924,7 @@ class SileBin(BaseSile):
     """
 
     def __init__(self, filename, mode='r', *args, **kwargs):
-        self._file = filename
+        self._file = Path(filename)
         # Open mode
         self._mode = mode.replace('b', '') + 'b'
 
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -203,7 +203,7 @@ def get_sile_class(filename, *args, **kwargs):
 
     # Split filename into proper file name and
     # the Specification of the type
-    tmp_file, fcls = str_spec(filename)
+    tmp_file, fcls = str_spec(str(filename))
 
     if cls is None and not fcls is None:
         # cls has not been set, and fcls is found
@@ -290,10 +290,8 @@ def get_sile(file, *args, **kwargs):
        function returns a random one.
     """
     cls = kwargs.pop('cls', None)
-    if isinstance(file, Path):
-        file = str(file)
     sile = get_sile_class(file, *args, cls=cls, **kwargs)
-    return sile(str_spec(file)[0], *args, **kwargs)
+    return sile(Path(str_spec(str(file))[0]), *args, **kwargs)
 
 
 def get_siles(attrs=None):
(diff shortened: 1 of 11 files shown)


jonaslb:
So it turns out py2.7 compat is not too great still, I must have forgotten it in a few places. Will look into it later.

zerothi:
Once done, this will fix #112

jonaslb:
This should be ready for merging when py 2.7 support is dropped, as I will not go through the effort to support that. I tested it on 3.6 and 3.7, but it should also work on 3.5.

zerothi:
Great work. I'll review this :)

sisl/io/tbtrans/tbt.py
```diff
@@ -2742,7 +2743,7 @@ def copy_attr(f, t):
                 v[:] = dvg[:]
 
         # Update the source attribute to signal the originating file
-        self.setncattr('source', 'k-average of: ' + tbt._file)
+        self.setncattr('source', 'k-average of: ' + str(tbt._file))
```
zerothi: casting to `str` should be implicit, I don't think it is needed?

sisl/io/tests/test_object.py
```diff
@@ -3,6 +3,12 @@
 import pytest
 
 import numpy as np
+import sys
+import os
+try:
+    from pathlib import Path
```
zerothi: delete `try...except` since pathlib is a requirement

sisl/io/tests/test_object.py
```diff
@@ -24,7 +30,9 @@ def _my_intersect(a, b):
 
 
 def _fnames(base, variants):
-    return [base + '.' + v if len(v) > 0 else base for v in variants]
+    files = [base + '.' + v if len(v) > 0 else base for v in variants]
```
zerothi: `base` is not a `Path`? Should it?

zerothi: It seems so...

sisl/io/tests/test_object.py
```diff
@@ -56,6 +64,24 @@ class TestObject(object):
     def test_siesta_sources(self):
         pytest.importorskip("sisl.io.siesta._siesta")
 
+    def test_direct_path_instantiation(self):
+        fp = Path("/tmp/shouldnotexist.1234567")
```
zerothi: use `Path("tmp") / Path("...")`, since we may support windows in the not too far future. `osp.join` or `Path` intrinsic should be used

sisl/io/tests/test_object.py
```diff
@@ -56,6 +64,24 @@ class TestObject(object):
...
+        if fp.exists():
+            os.remove(str(fp))
+        for Sile in get_siles():
+            sile = Sile(fp, _open=False)
+            assert isinstance(sile, Sile)
+            assert not fp.exists()
+
+    def test_direct_string_instantiation(self):
+        fp = "/tmp/shouldnotexist.1234567"
```
zerothi: same as above

sisl/io/tests/test_table.py
```diff
@@ -19,8 +19,8 @@ def test_tbl1(sisl_tmp):
     io0.write_data(dat0, dat1)
     io1.write_data((dat0, dat1))
 
-    F0 = open(io0.file).readlines()
-    F1 = open(io1.file).readlines()
+    F0 = open(str(io0.file)).readlines()
```
zerothi: use `Path.open`

sisl/io/tests/test_table.py
```diff
@@ -33,8 +33,8 @@ def test_tbl2(sisl_tmp):
...
```
zerothi: same

sisl/io/wannier90/seedname.py
```diff
@@ -53,14 +54,14 @@ class winSileWannier90(SileWannier90):
     def _setup(self, *args, **kwargs):
         """ Setup `winSileWannier90` after initialization """
         self._comment = ['!', '#']
-        self._seed = self.file.replace('.win', '')
+        self._seed = str(self.file).replace('.win', '')
```
zerothi: perhaps `_seed` should also be a `Path`, then `_seed.with_suffix('.win')` would work

sisl/io/wannier90/seedname.py
```diff
@@ -53,14 +54,14 @@ class winSileWannier90(SileWannier90):
...
 
     def _set_file(self, suffix=None):
         """ Update readed file """
         if suffix is None:
-            self._file = self._seed + '.win'
+            self._file = Path(self._seed + '.win')
         else:
-            self._file = self._seed + suffix
+            self._file = Path(self._seed + suffix)
```
zerothi: see above

sisl/io/xsf.py
```diff
@@ -378,10 +378,8 @@ class axsfSile(Sile):
     """
 
     def _setup(self, *args, **kwargs):
+        kwargs["_md_steps"] = kwargs.get("_md_steps", 1)
```
zerothi: I don't think this belongs in this PR, this PR is already quite extensive and this should go in another PR.

sisl/io/sile.py
```diff
@@ -699,7 +698,7 @@ def __init__(self, filename, mode='r', lvl=0, access=1, *args, **kwargs):
             # The CDF file can easily open the file
         if kwargs.pop('_open', True):
             _import_netCDF4()
-            self.__dict__['fh'] = _netCDF4.Dataset(self.file, self._mode,
+            self.__dict__['fh'] = _netCDF4.Dataset(str(self.file), self._mode,
```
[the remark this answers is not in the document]
jonaslb: Yes, but I looked it up: Py3.5 was released 2015/9 and pathlib support was merged (no idea about released) in netcdf in 2015/11 (see https://github.com/Unidata/netcdf4-python/pull/490 ). It is possible that if someone out there is using Py3.5, that they might also use an older version of netcdf4.

sisl/io/tbtrans/tbt.py
```diff
@@ -2537,7 +2537,8 @@ def __call__(self, parser, ns, value, option_string=None):
                 else:
                     ns._tbt.write_tbtav(value)
         p.add_argument('--tbt-av', action=AVOut, nargs='?', default=None,
-                       help='Create "{0}" with the k-averaged quantities of this file.'.format(self.file.replace('TBT.nc', 'TBT.AV.nc')))
+                       help='Create "{0}" with the k-averaged quantities of this file.'.format(
+                        str(self.file).replace('TBT.nc', 'TBT.AV.nc')))
```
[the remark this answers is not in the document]
jonaslb: Unfortunately it's not so easy with multiple suffixes, there isn't a `.with_suffixes` method :( It could be `self.file.with_name(self.file.name.replace('TBT.nc', 'TBT.AV.nc'))` but that's hardly more elegant...

sisl/io/tbtrans/tbt.py
jonaslb: I can change it to a string with `format`, but `+` only works if both are strings.

sisl/io/wannier90/seedname.py
jonaslb: I don't know if it every happens, but can't the seed contain a dot? In that case we'd be overwriting the last part.

sisl/io/sile.py
```diff
@@ -513,10 +511,11 @@ def __init__(self, filename, mode='r', comment=None, *args, **kwargs):
         self._base_setup(*args, **kwargs)
 
     def _open(self):
```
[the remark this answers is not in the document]
jonaslb: It would be simpler but it's not a common practice, as far as I know. So it might be easier to write the code, but it'd be harder to read.

sisl/io/xsf.py
jonaslb: It is necessary because otherwise, the later access to `_md_steps` causes it to try to open the file for reading, and then it fails the tests (where it's given a nonexistent file).

jonaslb: Actually come to think of it the precise line you commented on is not necessary. Just the deleted ones below were necessary to delete. Does the `_md_steps` even matter?

jonaslb: I think it should inherit from xsfSile instead of Sile and then it would matter to pass the `steps` parameter instead of `_md_steps`. But like you said that might be for another PR. Right now I guess it should just pass the tests.

sisl/io/tests/test_object.py
```diff
@@ -1,14 +1,12 @@
 from __future__ import print_function, division
 
+import tempfile
```
zerothi: Instead of using `tempfile` you should add an argument `sisl_tmp` to the test which requires a temporary file.
That would ensure you have a temporary file based on the same test-directory. See below

sisl/io/tests/test_object.py
```diff
@@ -65,22 +63,24 @@ def test_siesta_sources(self):
...
 
     def test_direct_path_instantiation(self):
-        fp = Path("/tmp/shouldnotexist.1234567")
-        if fp.exists():
-            os.remove(str(fp))
-        for Sile in get_siles():
-            sile = Sile(fp, _open=False)
-            assert isinstance(sile, Sile)
-            assert not fp.exists()
+        with tempfile.TemporaryDirectory() as dir:
```
zerothi: Change to:

```python
    def test_direct_path_instantiation(self, sisl_tmp):
fp = sisl_tmp.file('sisl', 'shouldnotexist.1234657')
```
which will create a temporary file for you (and automatically delete it once the test is done.

For details, see `sisl/conftest.py`, I would suggest you to add `pathlib` to that file as well. :)

Thanks!
￼

sisl/io/sile.py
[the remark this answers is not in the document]
zerothi: yeah, probably that is correct. It is just annoying to call a function all the time... Lets keep it like this for now.

sisl/io/wannier90/seedname.py
zerothi: Doing 
```python
a = pl.Path('soaethuoase.')
a.with_suffix('.win')
PosixPath('soaethuoase..win')
```
yields the correct thing. :)

sisl/io/xsf.py
zerothi: Agreed, I have fixed the inheritance in `axsf`. So if you rebase it should be fixed.
Weird I haven't had errors before?

zerothi:
This has now been merged (I did a rebase to check conflicts step-by-step)

PR #115, opened 2019-02-26
Fancier indexing on geometry

jonaslb:
This allows slighty easier-to-read code when using Geometry:
```
geom = si.get_sile("tano_struct.fdf").read_geometry()
geom[geom.xyz[:, 2] > 12.6] = "tip"  # previously, (geom.xyz[:, 2] > 12.6).nonzero()[0] was needed
tip = geom.sub("tip")  # previously geom.sub(geom.names["tip"])
```
I didn't implement the same for Hamiltonians (or others?) because it is a daunting task to figure out where to do it. If you throw a tip I could look at it. Perhaps other functions in Geometry could use the improvement as well.


zerothi:
Regarding sparse matrices, then no, those codes needs some revisions to make fancy indexing unified.

PR #116, opened 2019-02-26
Adding geometries preserves named regions

jonaslb:
`geom1+geom2` now preserves named regions

### DIFF
--- a/sisl/_namedindex.py
+++ b/sisl/_namedindex.py
@@ -4,7 +4,7 @@ This module implements the base-class which allows named indices
 
 >>> nidx = NamedIndex('hello', [1, 2])
 """
-from collections import Counter
+from collections import Counter, defaultdict
 import numpy as np
 
 from ._indices import indices_only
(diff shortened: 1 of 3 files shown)


zerothi:
I am not too happy about the add function.  
In principle I think adding two `NamedIndex` should raise an error if the same name occurs twice. Since renaming is non-obvious and the user may expect something else.

It would however be nice to implement some kind of
```
with NamedIndex.option.add(merge=True, rename_equal=True):
    namedindex1.add(namedindex2)
```
or the like. This would allow fine-grained control.

jonaslb:
Yea, I follow that renaming is maybe not a very good solution. But I think that raising an exception could be surprising as well. Effectively you'd need to manually check your geometries for names before you do any addition, and then rename them.

The previous behaviour for `geom1+geom2` has been to take the union of atoms. So perhaps the most correct behaviour for names would be, by default, to union them as well? So that, if "region1" is in both `geom1` and `geom2`, `(geom1+geom2).sub("region1") == (geom1.sub("region1") + geom2.sub("region1"))`? 

I like the context manager idea, but perhaps could be another PR and another time? :) A sensible default is the only 'must have', I think.

zerothi:
Agreed that indices could be merged would be a good idea... But consider e.g. a device you wan't to `tile`, then you would have a very weird electrode.

I think raising an error is not too surprising since the user will suspect one of two cases 1) a merge of indices, 2) a separation of names.

With a non-clear answer it _has_ to raise an exception/error.

PS. Agreed the context manager is out of scope now.  ;)

zerothi:
Many thanks for the update! I am now working through this!

Would you mind me deleting the `duplicate_names` in Geometry.add? Would you use it?

Primarily because I would prefer not to have cascading arguments... (otherwise they need to go to `.prepend` and `.append` as well ;)

Also I am going to change the name `NamedIndex.add` to `NamedIndex.merge`, I think it makes more sense and is less confusing given that `NamedIndex.add_name` exists.

PS. Your idea of allowing dicts as namedindex is nice, I have added that to the constructor of `NamedIndex`.

jonaslb:
It's not (yet) super critical, but I am trying to be organized about putting some different heterostructure geometries together. So I wanted a 'layer index' (ie a name, like 'bottom_graphene') to use throughout. I'm not actually (yet) using that index though, but I guessed I might in the future. 

I have a small helper function named `geometry_union(geoms)` that I use to unify the different structures. It relies on repeated use of geom.add(other, duplicate_names='union'), so that's really where I used all this. But perhaps a static method `Geometry.union(*geometries)` should be added to sisl instead? (similar to `set.union(*sets)`)

PR #117, opened 2019-02-26
Geometry.sub preserves names

jonaslb:


sisl/_namedindex.py
```diff
@@ -106,6 +106,20 @@ def __contains__(self, name):
         """ Check whether a name exists in this group a named group """
         return name in self._name
 
+    def sub(self, index):
```
zerothi: I should revise what NamedIndex does in these cases.

Currently the `remove` function works in-place. But your sub does not.
I'll have to think about this. Also, I think that a named-index should be allowed, even if it has no entries.

sisl/geometry.py
```diff
@@ -111,7 +111,7 @@ class Geometry(SuperCellChild):
     Atom : contained atoms are each an object of this
     """
 
-    def __init__(self, xyz, atom=None, sc=None):
+    def __init__(self, xyz, atom=None, sc=None, names=None):
```
zerothi: Hmm... probably this should be `namedindex` but I need to think if the NamedIndex functionality could be extended, and if so, perhaps having them in the constructor is bad... :( ?
Well...

jonaslb: I named it like so because the named index is currently accesible on `geom.names`. I think it's good to have them be named similarly?

sisl/_namedindex.py
jonaslb: In-place or not, whichever you like -- it doesn't really matter I think. I don't think there's a usecase for sub outside of when it's used in geometry.

I personally don't like 'leftovers'. I am similarly annoyed that there can be (and often is) an Atom in the Atoms which is not present in the geometry. But 'empty' names could be left behind for consistency with this behaviour. (I might send a PR with a `.clean()` or similar function later then)

(regarding consistency: Atoms.sub is not in-place either, I believe?)

zerothi: The leftover in `Atoms` has its use cases (primarily targetted Siesta). And in principle it could be nice to have assignments of names that "disappear" and then you at least know that you *had* some name associated, but have (un)-intentionally removed it.

`Atoms.reduce` removes un-used atoms, so it is there.

Regarding consistency, `Atoms.sub` is not in-place, just as I wrote above. `sub`/`remove` should create new instances while `delete` should do it *in-place*.

sisl/geometry.py
zerothi: Yes, I know, but I haven't decided whether `names` is the appropriate inner name (also it is currently "hidden" so hence I haven't cleaned it.

jonaslb:
I added the `NamedIndex.sub_*`-methods, as well as a reduce-function to compensate for the new behaviour. I did not add/change the other methods you mentioned.

zerothi:
Thanks!!

I changed many of your list-comprehensions to explicit for loops, removes one loop and improves readability (especially with nested `if ... else`).

Lastly, this PR and (one of the others) contained the same change. This is not good ;)

PR #118, opened 2019-03-05
Added tiling and repeating to NamedIndex

jonaslb:
And Geometry tile and repeat makes use of it.
Tests were added as well.

### DIFF
--- a/sisl/_namedindex.py
+++ b/sisl/_namedindex.py
@@ -262,3 +262,26 @@ class NamedIndex(object):
         for i in range(len(self))[::-1]:
             if len(self._index[n]) == 0:
                 del self[self.names[i]]
+
+    def tile(self, reps, offset):
+        """ Returns a new NamedIndex corresponding to the case where the overlying
+        geometry was tiled a number of times."""
+        new_names = self._name.copy()
+        new_indices = []
+        offset = np.arange(reps).reshape(reps, 1) * offset
+        for i, index in enumerate(self._index):
+            n_idx = np.tile(index, reps).reshape(reps, -1) + offset
+            new_indices.append(n_idx.flatten())
+        return self.__class__(new_names, new_indices)
+
+    def repeat(self, reps, offset):
+        """ Returns a new NamedIndex corresponding to the case where the overlying
+        geometry was repeated a number of times."""
+        new_names = self._name.copy()
+        new_indices = []
+        areps = np.arange(reps).reshape(1, -1)
+        for i, index in enumerate(self._index):
+            new = index * reps
+            new = new.reshape(-1, 1) + areps
+            new_indices.append(new.reshape(-1))
+        return self.__class__(new_names, new_indices)
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -1206,7 +1206,7 @@ class Geometry(SuperCellChild):
 
         # Create the geometry and return it (note the smaller atoms array
         # will also expand via tiling)
-        return self.__class__(xyz, atom=self.atoms.tile(reps), sc=sc)
+        return self.__class__(xyz, atom=self.atoms.tile(reps), sc=sc, names=self.names.tile(reps, len(self)))
 
     def repeat(self, reps, axis):
         """ Create a repeated geometry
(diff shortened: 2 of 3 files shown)


zerothi:
Similar to `NamedIndex.merge` the tiling and repeating could have some way of handling this.
There are various things that have not been cleared out in terms of `NamedIndex`, for instance if the names always follow a sub/remove etc. from a `Geometry`, then imagine doing:
```python
device = geom.cut(...).cut(...)
# at this point 'Left' is all atoms in device
device = device.tile(...).tile(...)
```
how should the final name-space look like? It isn't obvious what to do.
Again, I think that one user might expect one thing, while another might expect another. So I would also prefer some context handler in this case.

Also, I want to re-think the NamedIndex so it has an intuitive handling from the `Geometry` perspective. Currently it is highly irregular (as you have found! ;) )

jonaslb:
I think as long as it is documented and consistent, users can (and should) adjust expectations. As long as naming isn't clearly misleading, but that's not the case here, I think.

And I'd rather be able to tile NamedIndex than not, which is the case currently. And I also think it'd be inelegant if Geometry didn't handle it.

Regarding a re-think: Perhaps one possibility would be to take some inspiration from CAD programs. How things are created is always a 'list of instructions' eg. "cylinder here, hole in it here, ...". So perhaps a `tile` operation could somehow result in a GeometryUnion object that contained references to its building blocks. Anyway, just ideas.

zerothi:
> I think as long as it is documented and consistent, users can (and should) adjust expectations. As long as naming isn't clearly misleading, but that's not the case here, I think.

I don't think this is always the case. ;)
From experience people will be troubled if they get a result that is different from their expectations, regardless of documentation. Again, in this case there may be several *useful* outcomes. And choosing one or the other as the baseline in `Geometry.tile` is dangerous. Since 1) one user may be happy with the solution, 2) other users may not expect this and thus needs to revert what is happening, thus adding code, just as you need! In either case it is loose-loose.

This, I think, is a good example when not to add something to the code. Instead there could be a function implementing your said idea for the `Geometry` and then once we have a context handler which can deal with *all* expected situations it should be hard-coded.

As another example consider why `dict` does not have `__add__` implemented? It could easily be equal to `update`. But it isn't because of ambiguity, same applies here. They could have documented it and selected method, but decided not to since it creates ambiguity.

I really think that methods should be chosen with care. This I have learn over the years, and probably there are quite a few methods that should be removed from various classes!

> 
> And I'd rather be able to tile NamedIndex than not, which is the case currently. And I also think it'd be inelegant if Geometry didn't handle it.

I like your `NamedIndex` tile! I am just opposed to the `Geometry.tile` handling. For instance it would be very easy to have a method employing what you did:

```python
def geometry_name_tile(geometry, reps, axis):
    geom = geometry.tile(reps, axis)
    geom._names = geometry.names.tile(reps, len(geometry))
    return geom
```
Similarly with `repeat`.
> 
> Regarding a re-think: Perhaps one possibility would be to take some inspiration from CAD programs. How things are created is always a 'list of instructions' eg. "cylinder here, hole in it here, ...". So perhaps a `tile` operation could somehow result in a GeometryUnion object that contained references to its building blocks. Anyway, just ideas.

This is a great idea, and would solve lots of problems related to name-spaces of indices etc.

jonaslb:
I don't think the comparison to dicts are on-point because sisl geometries are not equally generalized (they only handle geometries of atoms).

Regardless, I can remove the integration into `Geometry` to get this merged. What should the default behaviour be then? The previous behaviour was to delete the names, but was that by intension or an omission?

zerothi:
> I don't think the comparison to dicts are on-point because sisl geometries are not equally generalized (they only handle geometries of atoms).

I have to disagree ;)
> 
> Regardless, I can remove the integration into `Geometry` to get this merged. What should the default behaviour be then? The previous behaviour was to delete the names, but was that by intension or an accidental omission?

It was by omission since they where never really intended for these use cases. What I think is the least invasive and expected output is that the indices are kept to retain the same ordering for `tile` and `repeat`.

jonaslb:
I will close this because I'm not working on it any more.

PR #119, opened 2019-03-05
Added geometry union

jonaslb:
Comments?

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -5,6 +5,7 @@ from numbers import Integral, Real
 from six import string_types
 from math import acos
 from itertools import product
+from functools import reduce
 
 import numpy as np
 from numpy import ndarray, int32, bool_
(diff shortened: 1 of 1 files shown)


zerothi:
I am not too happy about this method. :(
1) This shouldn't be an object method, if anything, it should be a `@classmethod`
2) This method will be superfluous if we get a context to handle the `dublicate` keyword.

I would prefer not to have this method associated with the geometry to not clutter the namespace. Secondly, one could equally imagine additional methods `append_union` etc. so to reduce number of methods I think we need to go for some kind of context handler.

FYI I am currently re-thinking what to do with the `NamedIndex` class. It needs a thorough clean-up in terms of method names and how it should be *attached* to the geometry etc.

jonaslb:
1. The inspiration is the builtin `set.union`, which is used in precisely the same way as this, and is an object method. I could change the first argument to `self`, since `geoms[0]` is used particularly anyways and is always required. I wouldn't mind changing it to a classmethod if you insist, but I must say that I think the discrepancy in usage compared to `set.union` would be confusing.

2. Yes, but you removed that keyword in my PR for add? This was the other solution

jonaslb:
Ah nevermind on 2., you want a context manager? I must say I don't think it would be elegant compared to this.

zerothi:
I stand corrected on the 1. point.  
However, I still see this method as problematic as we would then need to add a `union_add`, `union_append`, `union_insert`.

I think actually your recent PR's opens up another approach. Here's an idea.
Offload some of the specialised methods into functions.  
I.e. the methods are not particular querying data from a `Geometry` but rather handling geometries. Thus functions for this would probably be better since the `Geometry.*` namespace is already quite cluttered (already around 100 methods)... :(

But again, I think this needs some revisions since the `NamedIndex` needs careful thoughts. My original idea for the `NamedIndex` was not to complicate things. But it turns out it has... Not good...

jonaslb:
I will close this because I'm not working on it any more.

PR #131, opened 2019-06-20
bugfix: inner product in berry phase

sofiasanz:
Previously in berry phase calculation inner product of states acquired an uncontrolled and unwanted phase due to the lack of setting `align=False`. With this change the phases seems to come out correctly.

### DIFF
--- a/sisl/physics/electron.py
+++ b/sisl/physics/electron.py
@@ -905,7 +905,7 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
             # Loop remaining eigenstates
             for second in eigenstates:
                 second.change_gauge('r')
-                prd = _process(prd, prev.inner(second, False))
+                prd = _process(prd, prev.inner(right=second, diagonal=False, align=False))
                 prev = second
 
             # Complete the loop
@@ -919,7 +919,7 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
                     prev.state *= np.exp(-1j * phase)
 
                 # Include last-to-first segment
-                prd = _process(prd, prev.inner(first, False))
+                prd = _process(prd, prev.inner(right=first, diagonal=False, align=False))
             return prd
 
     else:
@@ -931,7 +931,7 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
             for second in eigenstates:
                 second = second.sub(sub)
                 second.change_gauge('r')
-                prd = _process(prd, prev.inner(second, False))
+                prd = _process(prd, prev.inner(right=second, diagonal=False, align=False))
                 prev = second
             if closed:
                 if method == 'zak':
@@ -940,7 +940,7 @@ def berry_phase(contour, sub=None, eigvals=False, closed=True, method='berry'):
                     axis /= (axis ** 2).sum() ** 0.5
                     phase = dot(g.xyz[g.o2a(_a.arangei(g.no)), :], dot(axis, g.rcell)).reshape(1, -1)
                     prev.state *= np.exp(-1j * phase)
-                prd = _process(prd, prev.inner(first, False))
+                prd = _process(prd, prev.inner(right=first, diagonal=False, align=False))
             return prd
 
     # Do the actual calculation of the final matrix


zerothi:
Thanks for the PR!

It is however, unclear to me whether the align is needed in all but the last call?
I believe this is the case. Do you have a test case I can try out?

sofiasanz:
I have tested it calculating the Zak phase on te SSH model. This simple system has two phases, one topological (zak=pi) an another trivial one (zak=0) depending on the relation between `t1` and `t2`. 
With the previous implementation, if the system had no band crossings the code seemed to work fine. However I found some numerical instabilities (if the system had band crossings) when changing the number of k-points to sample the integration path (the Zak phase oscillated between zero and pi for the same system but different integration paths). 
A way to have this band crossing in the SSH model is by taking the unit cell as double the original cell to induce band folding (this model originally has no band crossings in the occupied bands).
With this new implementation the code seems to work as expected with and without band crossings and the result does not depend on the integration path.

```
import sisl

xyz = [[0.0,0.0,0.0],[1.0,0,0]]
geom = sisl.Geometry(xyz, sc=[2.5,10,10])
geom = geom.tile(2, axis=0)
geom = geom.move(-geom.center(what='xyz'))
geom.set_nsc([3, 1, 1])
H = sisl.Hamiltonian(geom)
# Trivial when |t1| > |t2| (zak=0)
# Topological when |t2| > |t1| (zak=pi)
t1 = -1.0
t2 = -0.5
for ia in geom:
    idx = geom.close(ia, R=[0.1, 1.1, 1.6])
    H[ia, idx[0]] = 0.
    H[ia, idx[1]] = t1 
    H[ia, idx[2]] = t2

def func(sc, frac):
        return [-0.5+frac, 0, 0]

# Compute Zak phase for all occ. bands for different integration paths
occ_bands = range(len(H)/2)
for nx in range(50,151):
    bz = sisl.BrillouinZone(H).parametrize(H, func, nx)
    zak = sisl.electron.berry_phase(bz, sub=occ_bands, closed=True, method='zak')
    print('Zak =', zak)
```

zerothi:
Thanks for the example.

However, I am not sure the Berry phase calculation would *ever* work when you have band-crossings? (in the documentation I also note this!)
Since then you start mixing the phases between two different states?

This actually brings up another idea: Changing `align` to `align_phase` and adding a new method `align_norm` which re-orders the states of two eigenstates such that their orbital norms are closer together, this should handle most band-crossings (except at points of degeneracies).
Once this is in place then Berry phase should work for band-crossings.

zerothi:
On a side note:
```
    bz = sisl.BrillouinZone(H).parametrize(H, func, nx)
```
should be changed to:
```
    bz = sisl.BrillouinZone.parametrize(H, func, nx)
```
if performance is an issue.

zerothi:
Do you think the above would work in your case?

sofiasanz:
I have found in the literature that band crossings shouldn't prevent us from obtaining good results when performing calculations of the Berry connection (check the review of R. Resta ref. [J. Phys.: Condens. Matter 12 (2000) R107–R143. [http://www.theochem.unito.it/didattica/tec-comp_sdm/note1.pdf](url)] ). In this review there is a comment on band crossings (sec. 4.4, page R122) and the conclusion is that this does not affect the result provided that the occupied bands do not cross the empty ones (the system must be an insulator).

About the `align` issue I don't see why should it be included in the inner product (even in the middle terms), because in this calculation the second wavefunction _needs_ to be the same as the previous one (in each step of the iteration along the integration path) so arbitrary phases disappear. With `align=True` an arbitrary phase appears which is not guaranteed to be compensated in the next iteration, that's why I think it should be set to `False`.

tfrederiksen:
@zerothi, people argue that band crossings is not a problem for computing the global phase (sum over bands) as in your implementation (but the phases of individual bands are ill-defined). See for instance the review by [Resta, JPCM 12, R107 (2000)](https://iopscience.iop.org/article/10.1088/0953-8984/12/9/201): 

> It may happen that occupied bands are energy degenerate [...] This fact does not cause any problem for the Berry’s phase, which is uniquely defined once the projector on the occupied spin orbitals is provided as a function of `x`. A sufficient condition for having a well defined Berry’s phase is that the highest occupied orbital is separated by a finite gap from the lowest unoccupied one at any point of the path. (page R121)

In light of this, I do not really follow your warning in the code? One could also argue that the contribution from band crossings to the discretized integral diminishes as the `k`-mesh is made denser?

In practice, Sofia's patch now gives good results many different systems with lots of band crossings. You can for instance try the SSH model with a repeated cell containing `n` copies of the primitive cell (and translated to have its inversion symmetry at origo). Then you get band foldings but no change in the computed Berry/Zak phase.

zerothi:
Great, thanks for this! I'll merge, and remove the notes. And then I'll probably add the other things anyway :)

@sofiasanz @tfrederiksen I adopted this align phases according to pythtb, but it seems that I am not doing it exactly the same way (that code also enforces a continuous phase but also with phase-bounds).

@sofiasanz as for the align issue, when you have band crossings then suddenly `i` and `i+1` swaps place along a k-direction. It was for this reason I had always thought that Berry-phases did not work for band-crossings, but I see that the discussion says it will work as long as you select a full contained window of bands.

tfrederiksen:
@zerothi, by the way, in your implementation for the Berry phase you choose to work in the `r` gauge. I think that one could equally well use the `R` gauge. This choice should make no difference for a closed loop integration.

However, for the Zak phase, which is an open 1D integral over the BZ, the `initial` and `final` states, separated by a reciprocal lattice vector `G`, are in fact _identical_ in the `R` gauge because of periodicity. Hence the phase `exp(-1j G.r)` we add to the last segment is only required in the `r` gauge. The code would therefore look much simpler in the `R` gauge (just closing the loop, that's all).

zerothi:
Ha, I saw what you did there ;) :o: 

I must admit, I am no expert in this. I read: http://www.physics.rutgers.edu/pythtb/formalism.html where it is discussed that the `r` gauge is more natural for Berry phase calculations, hence I forcefully used this. My initial tests also indicated that it did not matter, but I refrained from doing anything since PythTB suggested otherwise. :)

zerothi:
@sofiasanz could you please try the latest commit? I have implemented your changes and fixed some things here and there.
In your example I now only get ~0.

tfrederiksen:
Commit 412c0bd0a3c64e048f50379a5c9fb7b022d022a3 works fine and as expected on my machine for different systems.

sofiasanz:
@zerothi do you get `zak=0` for both phases (topological and trivial)? I have tested the example from above with the last version (commit 412c0bd0a3c64e048f50379a5c9fb7b022d022a3) and it still works fine. I have tested it with both python2.7 and python3.. I am not sure why are you getting this result :-S

zerothi:
Oh, I had only tested trivial (which was correct).

Also testing the topological I get pi or -pi.
So all good!

PR #132, opened 2019-06-27
xyz file parsing works for ase gui-exports

jonaslb:
The comment line in an ASE-gui exported xyz file looks like so:

```
Lattice="34.56 0.0 0.0 0.0 24.94153163 0.0 0.0 0.0 69.13138595" Properties=species:S:1:pos:R:3 pbc="F F F"
```

And attempting to read such an xyz file with sisl resulted in a crash. I changed the xyzSile code to allow parsing this type of xyz file as well.

### DIFF
--- a/sisl/io/xyz.py
+++ b/sisl/io/xyz.py
@@ -4,6 +4,7 @@ Sile object for reading/writing XYZ files
 from __future__ import print_function
 
 from collections import defaultdict
+import re
 import numpy as np
 
 # Import sile objects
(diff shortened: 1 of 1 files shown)


jonaslb:
Actually, I just realized that the minimum Python version is 3.5 (perhaps even still 2.7?). So I should change the f-strings.

zerothi:
Thanks, I'll have a go at it for ASE exports. I will probably reformat the xyz completely to make it deterministic towards special sisl keywords!

zerothi:
Merged, thanks for bringing this to my attention! Added tests now.

PR #143, opened 2019-09-11
bug: ignore geometry constraints in POSCAR/CONTCAR

tfrederiksen:
The `VASP` geometry files `POSCAR/CONTCAR` generally contain 6 columns, corresponding to first the coordinates (3 floats) and then the geometry constraints (3 boolean values).

This PR fixes the error associated with reading the last three columns.


zerothi:
Thanks for the PR! :)

PR #145, opened 2019-09-27
Dynamical matrix from gulp works for norbs%12!=0

jonaslb:
An error would cause `gotSileGULP.read_dynamical_matrix()` to crash if the number of orbitals was not divisible by 12 (the number of forces gulp outputs per line).

### DIFF
--- a/sisl/io/gulp/got.py
+++ b/sisl/io/gulp/got.py
@@ -222,7 +222,7 @@ class gotSileGULP(SileGULP):
         # default range
         dat = np.empty([no], dtype=dtype)
         i, j = 0, 0
-        while True:
+        while i < no:
             l = self.readline().strip()
             if len(l) == 0:
                 break
@@ -240,8 +240,6 @@ class gotSileGULP(SileGULP):
                     i += 1
                     # reset column
                     j = 0
-                    if i >= no:
-                        break
             else:
                 # add the values (12 values == 3*4)
                 # for atoms on each line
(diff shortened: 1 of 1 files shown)


sisl/io/gulp/got.py
```diff
@@ -255,8 +253,7 @@ def _r_dynamical_matrix_got(self, geom, **kwargs):
 
                         i += 1
                         j = 0
-                        if i >= no:
-                            break
+                        break
```
zerothi: I think this is wrong. This should happen when a new row of *orbitals* occurs in the middle of the 12 entries. Say for 5 atom structures. In that case the `dat` array needs to be filled in the new row.

jonaslb: Gulp puts a line break when you have a new row - it doesn't go to the next row in the middle of a 12 entries wide line. So it needs to go the next line. That was actually why it was crashing for me.

jonaslb: I was considering putting a test for it (ie just read a .gout file with norbs%12!=0 without crashing), but I don't think I can do it in one PR, right? It has to go to the other repo as well?

zerothi:
Great ok, this clarifies it. If you could also make a pr for the test repo it would be great!
I'll try and merge tonight!

sisl/io/gulp/got.py
zerothi: Yes and no, either send me the test file and I'll do it, or.
1) create a pr with the test file in the other repo
2) amend this pr with the test (and also added the new hash for the files repo) once I have merged your other pr with the file

Then both things will happen simultaneously.

zerothi:
I'll just merge now. But please send the test ;)

PR #149, opened 2019-11-11
Take into account double occupancy in fermi_level

sofiasanz:
I think that there should be a factor of `0.5` to account for the double occupancy in non-spin-polarized Hamiltonians.

### DIFF
--- a/sisl/physics/hamiltonian.py
+++ b/sisl/physics/hamiltonian.py
@@ -505,7 +505,10 @@ class Hamiltonian(SparseOrbitalBZSpin):
         bz.asarray()
 
         if q is None:
-            q = self.geometry.q0
+            if self.spin.is_unpolarized:
+                q = self.geometry.q0 * 0.5
+            else:
+                q = self.geometry.q0
         # Ensure we have an "array" in case of spin-polarized calculations
         q = np.asarray(q, dtype=np.float64)
 


sisl/physics/hamiltonian.py
```diff
@@ -505,7 +505,7 @@ def fermi_level(self, bz=None, q=None, distribution='fermi_dirac', q_tol=1e-10):
         bz.asarray()
 
         if q is None:
-            q = self.geometry.q0
+            q = self.geometry.q0 * 0.5 # Account for double occupancy
```
zerothi: This will not work for non-collinear and SOC calculations.

Instead I would do something like:
```python
if q is None:
    if self.is_unpolarized:
        q = self.geometry.q0 * 0.5
    else:
        q = self.geometry.q0
```
or?

sisl/physics/hamiltonian.py
```diff
@@ -551,7 +551,7 @@ def _Ef(q, eig):
             Ef[1] = _Ef(q[1], bz.eigh(spin=1))
         else:
             # Ensure a single charge
-            q = q.sum()
+            q = q.sum() * 0.5 # Account for double occupancy
```
zerothi: In cases where the user supplies the correct number of electrons, then this won't work, no?

If a user supplies the number of electrons, shouldn't we expect it to be correct, or?

sisl/physics/hamiltonian.py
zerothi: I think the above should be the only change, actually?

sofiasanz: True, this 'fix' only refers to strictly non-polarized Hamiltonians, where there is spin degeneracy. This change makes more sense probably, since is more specific.

sisl/physics/hamiltonian.py
sofiasanz: Also true, although if the user supplies `q = [q[0], q-q[0]]`, and the Hamiltonian is not spin polarized, then `q = N` and we have the same problem as above, no?

zerothi: Agreed, but then it is a question of what we expect the user to input. Should they always input the # of electrons based on spin, or should it be the number of electrons in terms of the basis functions...

sofiasanz: OK, I think I wasn't seeing the case where a list of `q=[q0, q1]` is passed but the Hamiltonian is non-spin-polarized as something that would make sense... But I guess that for some calculations this might be the situation?

sofiasanz:
I've changed my commit to only apply this factor of `0.5` for only non-polarized Hmailtonians.

PR #152, opened 2019-12-17
Fixing spgeom.set_nsc when new cell is larger

### LINKED ISSUE #151: Increasing the number of supercells in an spgeom with set_nsc causes a loss of matrix elements
**Describe the bug**

When increasing the number of supercells on a spgeom, I would expect a simple 'padding' of the spgeom with zeros. What happens in practice is some odd loss of matrix elements.

**Reproducable code**
```python
In [1]: import sisl as si

In [2]: g = si.geom.graphene()

In [3]: H = si.Hamiltonian(g)

In [4]: H.nsc
Out[4]: array([3, 3, 1], dtype=int32)

In [5]: H.construct([[0.1, 1.5],[0, -2.7]])

In [6]: H.tocsr().todense()
Out[6]: 
matrix([[ 0. , -2.7,  0. ,  0. ,  0. , -2.7,  0. ,  0. ,  0. , -2.7,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ],
        [-2.7,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
         -2.7,  0. ,  0. ,  0. , -2.7,  0. ,  0. ,  0. ]])

In [7]: Hbig = H.copy()

In [8]: Hbig.set_nsc((5, 5, 1))

In [9]: Hbig.tocsr().todense()
Out[9]: 
matrix([[ 0. , -2.7,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. , -2.7,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ],
        [-2.7,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ,  0. ]])
```

(of course the supercell connecting elements can move around as a consequence of the extra zeros, but as you can see, it seems that half the elements were lost in this case)

**Version details**
Run the below code and add to bug-report:
```
3.7.5 (default, Nov 20 2019, 09:21:52) 
[GCC 9.2.1 20191008]
0.9.7+15
80f9c57f6e6f5c891c6fce65ae70bdf9f25a6491
```

jonaslb:
Fixes #151.

* Don't clean on every column translation to avoid deleting new columns
* Don't do the check that all supercells are accounted for (this will happen when expanding along one axis while shrinking another)

I wasn't 100% sure that not using clean in the sparsecsr would work, but I tested it on some phonon bandstructures and it seems ok. ~~Perhaps some tests should be added before merging?~~ I added some tests.

### DIFF
--- a/sisl/sparse_geometry.py
+++ b/sisl/sparse_geometry.py
@@ -266,7 +266,7 @@ class _SparseGeometry(object):
                 old.append(j)
                 new.append(i)
                 deleted[j] = False
-            except:
+            except ValueError:
                 # Not found, i.e. new, so no need to translate
                 pass
 
@@ -307,12 +307,11 @@ class _SparseGeometry(object):
             new = array_arange(new * size, n=n)
 
             # Move data to new positions
-            self._csr.translate_columns(old, new)
+            self._csr.translate_columns(old, new, clean=False)
 
             max_n = new.max() + 1
         else:
             max_n = 0
-
         # Make sure we delete all column values where we have put fake values
         delete = _a.arangei(sc.n_s * size, max(max_n, self.shape[1]))
         if len(delete) > 0:
@@ -322,6 +321,7 @@ class _SparseGeometry(object):
         shape = list(self._csr.shape)
         shape[1] = size * sc.n_s
         self._csr._shape = tuple(shape)
+        self._csr._clean_columns()
 
         self.geometry.set_nsc(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


zerothi:
Thanks for the bug and fix!!! :)

PR #154, opened 2019-12-19
Fixed reading overlaps where geometry=None may be passed

### LINKED ISSUE #153: Using read_overlap from an fdf-sile in a "onlyS"-type calculation causes AttributeError
**Describe the bug**
Take a siesta calculation where only the overlap is desired, ie. `TS.S.Save True` is in the fdf file. Then using `fdfSileSiesta.read_overlap()` will fail, although `onlysSileSiesta.read_overlap` works fine. The "source" is in `fdfSileSiesta._r_overlap_onlys` where the geometry is loaded *only* from either an nc or TSHS file (which doesn't exist in this case). Therefore `geometry=None` is passed to `onlysSileSiesta.read_overlap` and eventually this causes the error in the method `_geometry_align`. 

**Reproducable code**
```python
# (ensure you're in a dir with an "onlyS"-siesta calculation and no nc or tshs)
import sisl as si
si.get_sile("main.fdf").read_overlap()
```

Traceback
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-26899188ee5e> in <module>()
----> 1 si.get_sile("RUN_1.fdf").read_overlap()

/home/jolub/git/sisl/sisl/io/siesta/fdf.py in read_overlap(self, *args, **kwargs)
   1604         for f in order:
   1605             print("fdf", f, kwargs)
-> 1606             v = getattr(self, '_r_overlap_{}'.format(f.lower()))(*args, **kwargs)
   1607             if v is not None:
   1608                 return v

/home/jolub/git/sisl/sisl/io/siesta/fdf.py in _r_overlap_onlys(self, *args, **kwargs)
   1646                 # to ensure we get the correct orbital count
   1647                 kwargs['geometry'] = self.read_geometry(True, order=['nc', 'TSHS'])
-> 1648             S = onlysSileSiesta(f).read_overlap(*args, **kwargs)
   1649         return S
   1650 

/home/jolub/git/sisl/sisl/io/siesta/binaries.py in read_overlap(self, **kwargs)
    172         """ Returns the overlap matrix from the TranSiesta file """
    173         tshs_g = self.read_geometry()
--> 174         geom = _geometry_align(tshs_g, kwargs.get('geometry', tshs_g), self.__class__, 'read_overlap')
    175 
    176         # read the sizes used...

/home/jolub/git/sisl/sisl/io/siesta/binaries.py in _geometry_align(geom_b, geom_u, cls, method)
     65         return geom.copy(), True
     66 
---> 67     if geom_b.na != geom.na:
     68         # we have no way of solving this issue...
     69         raise SileError("{cls}.{method} could not use the passed geometry as the "

AttributeError: 'NoneType' object has no attribute 'na'
```

**Version details**
```
3.7.3 (default, Jun 27 2019, 02:40:43) 
[GCC 9.1.0]
0.9.7+24
5108ac4b6ca4788057589132b3684219815af45c
```

jonaslb:
Fixes #153

I was a little unsure how to approach that issue because it could be fixed in several places.
However where I fixed it is definitely the simplest place to do it. 
That's not the only motivation though. `kwargs` often have default values of None, but this read_overlap function didn't handle the case were it was None at all. So I thought it was also more appropriate to handle the situation there.

### DIFF
--- a/sisl/io/siesta/binaries.py
+++ b/sisl/io/siesta/binaries.py
@@ -171,7 +171,7 @@ class onlysSileSiesta(SileBinSiesta):
     def read_overlap(self, **kwargs):
         """ Returns the overlap matrix from the TranSiesta file """
         tshs_g = self.read_geometry()
-        geom = _geometry_align(tshs_g, kwargs.get('geometry', tshs_g), self.__class__, 'read_overlap')
+        geom = _geometry_align(tshs_g, kwargs.get('geometry', None) or tshs_g, self.__class__, 'read_overlap')
 
         # read the sizes used...
         sizes = _siesta.read_tshs_sizes(self.file)


zerothi:
This was fixed in another commit since it could be generalized for other SiestaSiles.

PR #156, opened 2020-01-03
Deterministic ordering of tests for parallel runs

jonaslb:
When running the tests with pytest-xdist, failure could occur because
a set intersection was used to generate them, which is not
deterministically ordered, so different processes did not have the same
(ordering of) tests. This changes the _my_intersect function to iterate
over a list rather than a set.


zerothi:
Thanks!

PR #159, opened 2020-01-22
Fix crash in Overlap.fromsp due to overridden init

jonaslb:
Using Overlap.fromsp would crash with the message "__init__() takes from
2 to 4 positional arguments but 5 were given". This was due to the class
instantiation call in super().fromsp which used positional arguments
that coincided with that class's kwargs, but not with the Overlap
object's kwargs. Putting the `dim` kwarg into **kwargs was also
necessary to allow it being passed (but still overridden), similar to
the `orthogonal` keyword.

### DIFF
--- a/sisl/io/siesta/siesta_nc.py
+++ b/sisl/io/siesta/siesta_nc.py
@@ -162,7 +162,7 @@ class ncSileSiesta(SileCDFSiesta):
 
         # Now create the tight-binding stuff (we re-create the
         # array, hence just allocate the smallest amount possible)
-        C = cls(geom, dim, nnzpr=1)
+        C = cls(geom, dim=dim, nnzpr=1)
 
         C._csr.ncol = np.array(sp.variables['n_col'][:], np.int32)
         # Update maximum number of connections (in case future stuff happens)
--- a/sisl/physics/densitymatrix.py
+++ b/sisl/physics/densitymatrix.py
@@ -572,7 +572,7 @@ class DensityMatrix(_realspace_DensityMatrix):
 
     def __init__(self, geometry, dim=1, dtype=None, nnzpr=None, **kwargs):
         """ Initialize density matrix """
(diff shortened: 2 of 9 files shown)


zerothi:
You have done many changes to ensure all calls to `__init__` uses keyword arguments. I would refrain from implementing those changes.

It seems only your first commit is necessary?

jonaslb:
The first commit doesn't pass the tests. I don't recall exactly why exactly but I think it was also to do with the kwargs.

The changes are mostly in sparse geometry which the overlap inherits from. I had an error in the tests when using the `copy` function because it was making a call with positional args where it should have been keywords. The other functions (tile, repeat, etc) are not tested in the overlap object, but they will fail if used unpatched. The alternative is to create overrides for all of them in the overlap object, which I don't think is a better solution.

There are also a few changes in hamiltonian, densitymatrix, etc. These are more of a principal nature, not strictly required. 
Hypothetically it could happen if someone made a FrankenClass inheriting both eg. DynamicalMatrix and Overlap -- then calling `super().__init__(...)` would fail because the super-init in `DynamicalMatrix` would try to call `Overlap.__init__` and fail due to the positional/kwarg mixup. But quite hypothetical of course, I see no reason someone would do that.

zerothi:
I agree it ain't pretty... :(

Hmm. Probably it would simply make sense to have the init method for `Overlap` also have the `dim` keyword. Forcing an overlap matrix to only have 1 dimension may also be somewhat counter intuitive if one wishes to use different overlap values in some cases... 

What would you say to simply adding `dim=1` to the `__init__` of `Overlap`, that would solve all problems and also add flexibility to `Overlap` that I don't think would be needed, but then it is there and does not mess up the interfaces for other matrices.

The above change just forces us to **ALWAYS** have `dim` in the keyword argument list.

zerothi:
I think the `dim` argument is the way to go. I have just added a new commit which fixes this and also passes your new tests (thanks for them!!!).

Could you check if everything fits you? Otherwise let me know!

PR #162, opened 2020-02-05
WIP: removing py2 support

### LINKED ISSUE #101: INFO: sisl will be Python 3 "only" from 2019
The sisl toolbox will be Py3 *only* from 2019.

This is to follow the development of the scientific community which already establishes this relation ship with Py3: https://python3statement.org/

From 2019, sisl will be Python >=3.6 only. Just before switching I will release a Py2 compatible version which will be the last one to enable Py2.

This should release some code. 

- [x] remove need for 'six'
- [x] remove metaclass workaround 
- [ ] add type annotations where applicable
- [x] remove `object` inheritance
- [x] remove py 2.7 from test suite
- [x] remove all `_zip`/`izip` usages (and similar for other)
- [ ] merge #114 to accommodate `pathlib`

zerothi:
Fixes #101


jonaslb:
See also [pyupgrade](https://github.com/asottile/pyupgrade). Running it as `pyupgrade --py36-plus **/*.py` yields a 3500 line git diff!

zerothi:
Thanks, however, I would prefer to do this in steps until the *diff* ain't that large. I would like to be clear on what is happening, also typically these are minor things in my opinion (and probably quite a bit of them are related to `autopep`. Some of the conventions used there I don't particularly like. So for now I'll wait ;)

Initially I thought that I could stick with >=3.5, but I now see that numpy and scipy will change to >=3.6. So I have to adjust...

zerothi:
After checking what it actually did I am quite happy with using it, it was basically only 300 lines without f-strings.

Once I have made the final 2 release I will use the upgrade functionality to change f-strings and complete the transfer. 

Thanks!

PR #163, opened 2020-02-13
Reading geometry from fdf-file when reading a binary requires to know correct number of orbitals.

juijan:
I suggest, adding 'fdf' as an option to read the geometry when `fdfSile.read_hamiltonian` and similar functions. 

These functions rely on knowing the correct number of orbitals.  However, when neither SystemLabel.nc nor SystemLabel.TSHS was generated, an attempt at guessing the geometry is being made.

While reading from the fdf file does not under all circumstances guarantee to return the correct number of orbitals, I would argue it is worth attempting.

### DIFF
--- a/sisl/io/siesta/bands.py
+++ b/sisl/io/siesta/bands.py
@@ -66,7 +66,7 @@ class bandsSileSiesta(SileSiesta):
             k = np.empty([nk, 3], np.float64)
             for ik in range(nk):
                 l = [float(x) for x in self.readline().split()]
-                k[ik, :] = l[0:2]
+                k[ik, :] = l[0:3]
                 del l[2]
                 del l[1]
                 del l[0]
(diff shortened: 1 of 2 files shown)


zerothi:
Yeah, it is probably ok to do this. My main concern has always been that if one does an MD simulation and then tries to read the binary stuff, then a fallback to the fdf file would result in the wrong geometry... :(

And there is no way to check it... :(

zerothi:
This has been merged

juijan:
I see. Maybe it is not worth doing then since there won't be consistency either way.

Do you think it would be better to raise an error, when the geometry was not retrieved? 
Warn the user that the geometry could not be read.

Currently, the routine just crashes:
`AttributeError: 'NoneType' object has no attribute 'no'`

zerothi:
@juijan could you please clean your master (re-fork).
Also, when you do PR's please create new branches. Doing merges together with PR's is a mess ;)

juijan:
Okay will do

zerothi:
No, I think it is fine. Probably it would just be wise to check if it was an MD simulation or not.

PR #166, opened 2020-02-18
Fix sgeom's behavior if --origin flag is set

juijan:
Previously, sgeom would shift the geometry such that all cartesian coordinates are greater than 0 and not like the description state all atoms such that one atom will be at the origin.

Now it should always shift the atom which is closest to the origin to the origin.


zerothi:
Thanks for the PR. However, the intent was actually what it did. Move atoms such that each coordinate is aligned to the minimum of each cartesian coordinate.

I'll amend the help docstring

juijan:
Sounds good.

PR #167, opened 2020-02-19
Geometry collections

pfebrer:
I though it could be useful to have a class to store multiple geometries.

This came to my mind when I needed to treat an animation, therefore I have also implemented reading multiple geometries from an xyzSile as an example.

My idea here is that you can operate simultaneously on all the geometries, not only independently but also based on some reference geometry (e.g. the first one). You could also use this class to easily track how the geometry is evolving or interpolate steps between geometries, which would be useful for NEB or to render animations with higher fps/resolution, for example. 

**This is just a sketch to know your opinion about it.**

**Note:**
If you believe this is a good idea, I need to find the best way use the methods from the `Geometry` class on all geometries of the collection, which are stored in a list.
One option would be to make `GeometryCollection` inherit from `Geometry` and decorate all the methods in the `Geometry` class so that they check whether they have been called from a geometry instance or from a collection and act accordingly. But I'm not sure if there is a better way to do it.

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -3879,6 +3879,7 @@ class GeometryCollection:
             self.geometries = [
                 Geometry(xyzs, atom=atom, sc=sc, names=names)
                 for xyz, sc in zip(xyz, sc)
+            ]
 
 def sgeom(geometry=None, argv=None, ret_geometry=False):
     """ Main script for sgeom.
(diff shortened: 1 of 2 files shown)


zerothi:
Thanks for the MR, I think this is a good idea. I will give this some thought and probably implement it in another way (to generalize for *all* objects).

pfebrer96:
Nice, that would definitely be the best scenario. It would be helpful for plots to make animations too (I already implemented it with my own approacb, but if there is a general way that would be great).

zerothi:
Superseeded by #545 thanks for comments and ideas!

PR #169, opened 2020-02-28
Fixed bug in Geometry.rotate_miller

pfebrer96:
rotate_miller was passing an angle in radians to Geometry.rotate without the rad keyword set to true and therefore it was not working properly.


zerothi:
@pfebrer96 thanks for PR!

PR #171, opened 2020-03-05
Blender support

pfebrer96:
*This is the beggining of an era of beautiful slide presentations.*

Here I have implemented the first basic functionalities that will allow an easy interaction between scientific data manipulated by sisl and beautiful 3D image rendering by blender (https://www.blender.org/). The aim of it is to **introduce people into 3D image rendering** and provide a set of automatic tools that will require little to no interaction of the user with blender's GUI. However, it is meant to also **serve people who is already familiar with blender by automating some repetitive tasks** and while still leaving room for the user to display their creativity.

How to use it
------

A function has been implemented under sisl.blender that will take care of everything if you pass it a renderable object *(at the moment, only Geometry is renderable)*. The easiest way to use it is by building a script that looks something like this:
```python
import sisl

geom = sisl.get_sile(<your path to a geometry file here>).read_geometry()

sisl.blender(geom, camera_settings={}, lighting_settings={}, render_settings={})
```
Then, you can run this python script within blender in many ways, one of them being:
```
blender -P path/to/your/script.py
```
from the terminal, which will open blender, run your script and then leave the GUI open. Use the *-b* flag if you want to run blender in background mode.

*Check the docs of the sisl.blender function under _blender.py, and specially the docs on the `camera()`, `lighting()` and `render()` methods of the `BlenderScene` class for more info on how to use it. I believe they are quite explanatory. If not, don't hesitate to ask questions here.*

How to set it up
-----
For this tools to work, you need a python environment that contains both the sisl package and blender modules (bpy, mathutils, addon_utils). There are three main ways (that I know) to set up an environment that satisfies these requirements:
- **Make blender "aware" of your environment (recommended):** This option is the best one as you don't need to modify blender's filesystem and therefore you can do it with the snap installation (from the Ubuntu Store). Follow the instructions here https://blender.stackexchange.com/a/76124, it's very clear. 

    Basically you just need to have a python environment with the same python version as blender (run `import sys; print(sys.version_info)` inside blender, for blender 2.83 is python 3.7.4) and install sisl there. Then add the path of the environment to blender's python. Note that you can make blender automatically add the path on start up (last point of the stack exchange comment)

- **Install sisl in blender's python environment:** Download blender from https://www.blender.org/download/ (*NOT FROM THE UBUNTU SOFTWARE STORE, because it won't let you modify the filesystem and we need to*). Then find the python binary under 2.82/python/bin. You then need to install pip for that python and with that install sisl:
```
<path/to/blender/python> -m ensurepip
<path/to/blender/python> -m pip install sisl
```
If you can't install sisl because of an error that is something like: *Python.h no such file or directory*, it is because you are missing some headers. If blender's python is python3.7, they can be installed in linux like so:
```
sudo apt-get install python3.7-dev
```
However, it seems like the python for blender it is unable to find them (at least for me), so you will probably need to specify their location like so:

```
CFLAGS="-I/usr/include/python3.7" <path/to/blender/python> -m pip install sisl
```

as suggested by Nick.

- **Install blender as a package in your sisl environment:** This should be possible either by building blender yourself as a python package (https://wiki.blender.org/wiki/Building_Blender/Other/BlenderAsPyModule,https://blender.stackexchange.com/questions/117200/how-to-build-blender-as-a-python-module) or using prebuilt versions that can be installed via pip (https://pypi.org/project/bpy/, https://pypi.org/project/blender/). *I have not been able to accomplish any of the two.*

**Let me know your opinion about it, suggest changes in the structure of the code or request further functionality that you think would be necessary.** Thanks!


.env
```diff
@@ -0,0 +1 @@
+/home/pfebrer/.venvs/sisl/
```
zerothi: Delete this file, it should be in the repo

docs/index.rst
```diff
@@ -1,172 +0,0 @@
-.. sisl documentation master file, created by
```
zerothi: I guess this is an accident, we really need this file ;), please undelete it

sisl/_blender.py
```diff
@@ -0,0 +1,501 @@
...
+
+except ModuleNotFoundError as BLENDER_IMPORT_EXCEPTION:
+    BLENDER_AVAIL = False
+    BLENDER_ASPACKAGE = False
+    INSIDE_BLENDER = False
+
+__all__ = ['BlenderScene', 'blender']
+
+class BlenderScene:
+
```
zerothi: Add some documentation here.

sisl/_blender.py
```diff
@@ -0,0 +1,501 @@
...
+        '''
+        self.bounding_box = np.array([
+            np.min([np.min(obj.bound_box, axis=0 ) for obj in self.data.objects], axis = 0),
+            np.max([np.max(obj.bound_box, axis=0 ) for obj in self.data.objects], axis = 0),
+        ]).T
+
+        return self
+
+    def lim(self, min_max = None, axis = None):
+
```
zerothi: add comments, possibly extending this name to `set_limits` or just `limits` (for completeness sake).

sisl/_blender.py
```diff
@@ -0,0 +1,501 @@
...
+    def blender(obj, *args, **kwargs):
+        raise BLENDER_IMPORT_EXCEPTION
+
+# Clean up
+del BLENDER_AVAIL
+del INSIDE_BLENDER
+del BLENDER_ASPACKAGE
+
+if __name__ == "__main__":
+
```
zerothi: This should go in a file in a test directory to disconnect it from the actual code.

docs/index.rst
pfebrer96: Yes, I had it in .gitignore by mistake.

sisl/_blender.py
pfebrer96: Done

sisl/_blender.py
pfebrer96: I have renamed it to `get_limits`

sisl/_blender.py
pfebrer96: I forgot to remove it, I wasn't using it anymore anyway.

How could I implement tests for this file? It needs the right environment to run and this will only happen when the methods are being run inside blender.

pfebrer96:
I'm not sure if it will actually be that large, because most things could be implemented as a blender add-on that uses sisl and use this just as an interface.

However, I'm not sure if blender add-ons are accessible when blender is built as a python module (probably yes).

> Message to Nick: Is there a way that the headers' location can be specified so that it finds the ones installed by apt-get in system's python?

Do you know if it is possible to solve this? It would make it much easier for people to install sisl in blender's python.

zerothi:
> * **Install sisl in blender's python environment:** Download blender from https://www.blender.org/download/ (_NOT FROM THE UBUNTU SOFTWARE STORE, because it won't let you modify the filesystem and we need to_). Then find the python binary under 2.82/python/bin. You then need to install pip for that python and with that install sisl:
> 
> ```
> <path/to/blender/python> -m ensurepip
> <path/to/blender/python> -m pip install sisl
> ```
> 
> If you can't install sisl because of an error that is something like: _Python.h no such file or directory_, it is because you are missing some headers. If blender's python is python3.7, they can be installed in linux like so:
> 
> ```
> sudo apt-get install python3.7-dev
> ```
> 
> However, it seems like the python for blender it is unable to find them (at least for me), so you may want to follow the instructions in the second alternative of this answer: https://blender.stackexchange.com/a/107381. _It worked for me._
> 
> **Message to Nick:** _Is there a way that the headers' location can be specified so that it finds the ones installed by apt-get in system's python?_

Perhaps doing
```bash
CFLAGS="-I/usr/include/python3.7" <path/to/blender/python> -m pip install sisl
```
would do?


I am currently working through this PR, first need to install blender, ;)

pfebrer96:
Thanks for checking it!

> CFLAGS="-I/usr/include/python3.7" <path/to/blender/python> -m pip install sisl

Yes, it works! It's much easier than downloading the headers yourself :)

I haven't touched this PR since I submitted it, I would say it still needs some things to be improved, it is not ready for production yet. It does some simple stuff, but probably there are lots of things that would be nice to consider. I asked a PhD student that was supervising me last year who is a beast using blender to help me with his expertise and he was down. So, I would say we could keep this as a draft, you don't need to bother to check it now :)

zerothi:
Ok, I will create a MR for your branch for somethings I have done now (un-tested) but I would like this to enter at some point.

Perhaps you can add some discussion points here: https://github.com/jonaslb/siutil/issues/1

pfebrer96:
> Ok, I will create a MR for your branch for somethings I have done now (un-tested) but I would like this to enter at some point.

Did you submit any changes already? Should I see them here in the PR or elsewhere?

zerothi:
I have now created the branch 171-blender-updates. Don't merge it since I have changed (and broken) some things.

But perhaps you can see what my intend is ;)

pfebrer96:
Nice, I've checked the branch. Sorry for the cleanup that was needed :)

What is it that is broken?

zerothi:
> What is it that is broken?

I don't know, I didn't test it, so potentially everything ;)
I re-arranged the directories etc. so maybe ;)

pfebrer96:
By the way, one key thing is that the atomic_blender add on (https://docs.blender.org/manual/en/dev/addons/import_export/mesh_atomic.html), which is what I'm using to import the geometry into blender, is not very flexible. I'm saying this because:

- You can not pass a python object representing the geometry. The geometry needs to be written to a file and then imported. 
- If it imports from an xyz file, it can not draw bonds.
- The only other way to import a geometry is with a pdb file, which sisl doesn't have a working parser for.

Maybe it is worth creating our own implementation of drawing an atom and a bond and then we would be free of these constrains. I believe there is only one important thing to take into account:

- To reduce the computational resources needed, the atomic blender add on creates each atom as a dupliverts, which basically means that there is only one instance of the sphere that is displayed at multiple locations. We can get inspiration from their implementation (https://github.com/sobotka/blender-addons/tree/master/io_mesh_atomic) or others (https://github.com/smparker/molecular-blender, https://github.com/floaltvater/molblend). I think this fits nicely with the structure of the `Atoms` object in sisl. Maybe `Atom` could store their blender representation that `Atoms` would make use of. At the end `Geometry` would ask `Atoms` to represent the atoms and then add the bonds.

the rest should be easy, because you can draw a sphere with just one command (`bpy.ops.mesh.primitive_uv_sphere_add`).

By the way, these dupliverts can easily be made independent afterwards, which I think would be a nice feature to have automatized to, for example, color atoms other than by species or highlight a given atom by giving it a special texture/color/etc. This would go in sync with the `BaseGeometryPlot` that I have implemented in the viz module, which allows you to provide different parameters for each atom/bond with the same objective in mind. In this way, any plot inheriting from `BaseGeometryPlot` would have the possibility to do:

```python
plot.render()
```

to use the blender support. Even new plots developed by users would be able to do this without them worrying about it, which I think is pretty cool :)

zerothi:
> By the way, one key thing is that the atomic_blender add on (https://docs.blender.org/manual/en/dev/addons/import_export/mesh_atomic.html), which is what I'm using to import the geometry into blender, is not very flexible. I'm saying this because:
> 
> * You can not pass a python object representing the geometry. The geometry needs to be written to a file and then imported.

Agreed, I also figured this out. This was why I had `SISL_BLENDER_TMP` at the top, I imagine having a temp-directory (through `os.environ`) to store the geometry like you did. Then it is hidden for the user.

As for the rest, for now, lets concentrate on getting it to work, I have too much on the table to figure out all the implications ;)

pfebrer96:
Hahaha ok, then should we use the pdb parser from ASE?

> Agreed, I also figured this out. This was why I had SISL_BLENDER_TMP at the top, I imagine having a temp-directory (through os.environ) to store the geometry like you did. Then it is hidden for the user.

I was just worried that if these directories are not removed for some reason (errors or unexpected closing) and the user filesystem might end up with lots of useless undeleted temporal folders. Can python's `tempfile` be used to avoid this? I don't know exactly how it works.

pfebrer96:
Just wanted to say that when this is in, we will be able to say that sisl supports virtual reality, because blender now officially supports virtual reality, starting from `2.83` (https://wiki.blender.org/wiki/Reference/Release_Notes/2.83/Virtual_Reality).

I think this is a very cool thing that we will have for free, just wanted to share it :) (specially cool if we find a good representation for sisl grids)

PR #172, opened 2020-03-06
Getting and setting atoms by tag in sisl.Atoms

pfebrer96:
Hi!

I just needed this and since there was no functionality implemented yet for getting atoms by tag in the `Atoms` class I thought it doesn't do harm to have it there just in case someone might find it useful.

Basically:

```python
atoms = sisl.Atoms(["C","C","Au")
atoms["C"]
```
will get you the atom object with tag "C".

and

```python
atoms["C"] = "N"
```
acts as a shortcut for `replace_atom` and will replace all "C" atoms by "N".

I also implemented some tests for it.

Sorry for submitting so much pull requests, I thought since I need these things, someone in the future might find it handy :)

### DIFF
--- a/sisl/atom.py
+++ b/sisl/atom.py
@@ -1106,6 +1106,9 @@ class Atom(metaclass=AtomMeta):
             return [self.orbital[o] for o in range(ol[0], ol[1], ol[2])]
         elif isinstance(key, Integral):
             return self.orbital[key]
+        elif isinstance(key, str):
+            orbs = [orb for orb in self.orbital if key.lower() in orb.name().lower()]
+            return orbs if len(orbs) != 1 else orbs[0]
         return [self.orbital[o] for o in np.asarray(key).ravel()]
 
     def maxR(self):
--- a/sisl/atom.py
+++ b/sisl/atom.py
@@ -1683,17 +1683,23 @@ class Atoms:
             return [self.atom[self._specie[s]] for s in range(sl[0], sl[1], sl[2])]
         elif isinstance(key, Integral):
             return self.atom[self._specie[key]]
+        elif isinstance(key, str):
+            return [at for at in self.atom if at.tag == key][0]
         return [self.atom[i] for i in self._specie[_a.asarrayi(key).ravel()]]
 
     def __setitem__(self, key, value):
         """ Overwrite an `Atom` object corresponding to the key(s) """
+        #If key is a string, we replace the atom that matches 'key'
+        if isinstance(key, str):
+            return self.replace_atom(self[key], value)
+
         # Convert to array
         if isinstance(key, slice):
             sl = key.indices(len(self))
             key = _a.arangei(sl[0], sl[1], sl[2])
         else:
             key = _a.asarrayi(key).ravel()
-
+        
         if len(key) == 0:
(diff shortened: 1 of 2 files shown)


pfebrer96:
There are two commits from the branch that you already merged and I don't understand why, sorry.

zerothi:
Yeah, exactly. You probably just continued working on that branch.

Always do a new branch from master (after pulling it)

pfebrer96:
It's a different branch, but I probably did something wrong.

I merged the old branch to my master, and then from my master I started this new one.

zerothi:
Alternatively do:

```bash
git diff HEAD^.. > new.patch
git checkout master
git pull
git branch -D getatom
git checkout -b getatom
patch -p1 < new.patch
git commit ...
git push --force
```
This should reset everything (assuming the patch generation is for the latest commit).

pfebrer96:
Done, thanks!

pfebrer96:
I did the same for `Atom` so that you can go like:

```python
atom["2pz"]
```

and get all 2pz orbitals.

Again this is because I need to use it right now, but if you don't think it's useful I can cancel it :)

sisl/atom.py
```diff
@@ -1106,6 +1106,9 @@ def __getitem__(self, key):
             return [self.orbital[o] for o in range(ol[0], ol[1], ol[2])]
         elif isinstance(key, Integral):
             return self.orbital[key]
+        elif isinstance(key, str):
+            orbs = [orb for orb in self.orbital if key.lower() in orb.name().lower()]
```
zerothi: I will change this to `key in orb.name()` since if one queries `p` orbitals one will also find polarization orbitals.
So we have to be explicit.

Also, in case none match, `None` will be returned.

sisl/atom.py
```diff
@@ -1683,17 +1686,23 @@ def __getitem__(self, key):
             return [self.atom[self._specie[s]] for s in range(sl[0], sl[1], sl[2])]
         elif isinstance(key, Integral):
             return self.atom[self._specie[key]]
+        elif isinstance(key, str):
```
zerothi: Minor change to ensure one returns `None` in case it ain't found.

pfebrer96:
Nice, thanks for taking the time!

sisl/atom.py
pfebrer96: I didn't notice that, great!

zerothi:
Thanks for the MR!

pfebrer96:
It just came to my mind that this would be useful for geometries too to get the indices of all atoms from a certain species. So that you could do, for example:

```python
geom = sisl.Geometry(...)
geom.xyz[geom["C"]]
```

to get the coordinates of all C atoms. I thought the indices would be a nice thing to get since many methods ask for atomic indices, but maybe there's some other behavior that would help more.

What do you think?

zerothi:
Actually this is not a good idea ;) I already have name association for clusters of atoms. So that would confuse users.

I. E. One can do

```python
Geom["hello" ] =[1, 2,4]
```

pfebrer96:
I didn't know that! That is definitely useful. I guess I'll start to make use of it from now on :)

Issue #11, opened 2016-10-28
Add vector object to abstract geometry and forces into a similar feature


zerothi:
Instead of having a geometry object with similar features, it would be nice to have a vector object which takes into account the basic routines for duplicating stuff, etc.

I.e. the geometry should inherit the vector class.

PR #174, opened 2020-03-15
Enhancing the siesta outSile

pfebrer96:
*In the visualization module, plots have the ability to listen to changes in files*. The main idea I had in mind for providing this feature was to **monitorize simulations in real time**. The outSile is clearly a main target for this type of application and I would like to enhance it so that we can benefit from it.

**With this in mind:**
- In the `read_force()` method, I added the possibility to read *total forces* and *max force* apart from the atomic forces which were already implemented. I added also some tests for it.
- The `read_scf()` method is awesome for this application! :)
- I did not want to touch anything because the way in which you manage file handles seems very well organized and I did not want to mess up with it, but: is there a possibility that the `_job_completed` attribute could be determined on `__init__()` by reading the last line of the file? It seems weird that you need to read a property that reads all the file to know whether the job is completed or not.

### DIFF
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -295,7 +295,7 @@ class outSileSiesta(SileSiesta):
             if last:
                 return Fs[-1]
             if self.job_completed:
-                return Fs[:-1]
+                return np.array(Fs[:-1])
             return np.array(Fs)
 
         return next_force()
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -272,6 +272,9 @@ class outSileSiesta(SileSiesta):
                 - atomic (default): (nMDsteps, nAtoms, 3)
                 - total: (nMDsteps, 3)
                 - maxF: (nMDsteps, )
+            
+            If `all` is `False`, the first dimension does not exist. In the case of maxF, the returned value
+            will therefore be just a float, not an array.
         """
         if all:
             last = False
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -269,7 +269,10 @@ class outSileSiesta(SileSiesta):
 
             # Now read data
             F = []
-            line = self.readline()
+            line = self.readline() # This line may be the first separator
+            if '---' in line:
+                line = self.readline()
+
             while '---' not in line:
                 line = line.split()
                 F.append([float(x) for x in line[-3:]])
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -296,7 +296,7 @@ class outSileSiesta(SileSiesta):
                 return Fs[-1]
             if self.job_completed:
                 return Fs[:-1]
-            return Fs
+            return np.array(Fs)
 
         return next_force()
 
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -325,7 +325,7 @@ class outSileSiesta(SileSiesta):
                 Fs.append(F)
 
             if last:
-                return Fs[-1]
+                return Fs[-1] if not maxF else Fs[-2]
             if self.job_completed:
                 return np.array(Fs[:-1])
             return np.array(Fs)
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -239,7 +239,7 @@ class outSileSiesta(SileSiesta):
         return next_geom()[1]
 
     @sile_fh_open()
-    def read_force(self, last=True, all=False):
+    def read_force(self, last=True, all=False, total=False, maxF=False):
         """ Reads the forces from the Siesta output file
 
         Parameters
@@ -249,6 +249,13 @@ class outSileSiesta(SileSiesta):
         all: bool, optional
            return a list of all forces (like an MD)
            If `True` `last` is ignored
+        maxF: bool, optional
+            whether only the maximum atomic force should be returned for each step.
+            Note that this is not the same as doing `max(outSile.read_force(total=True))` since
+            the forces returned in that case are averages on each axis.
+
+            Setting it to `True` is equivalent to `max(outSile.read_force())` in case atomic forces
+            are written in the output file (`WriteForces .true.` in the fdf file)
 
         Returns
         -------
@@ -261,6 +268,7 @@ class outSileSiesta(SileSiesta):
 
         # Read until forces are found
         def next_force():
+
             line = self.readline()
             while not 'siesta: Atomic forces' in line:
                 line = self.readline()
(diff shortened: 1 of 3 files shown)


zerothi:
Thanks for the PR.

However, I think reading those variables are redundant in the interface.
Instead of your additions one can do:
```python
total = out.read_force().sum(axis=0)
total_md = out.read_force(all=True).sum(axis=1)
fmax = np.abs(out.read_force()).max()
fmax_md = np.abs(out.read_force(all=True)).max(axis=1).max(axis=1)
```
As for your comments, thanks!
I would rather work on #11 which abstracts away a geometry related quantity. Once the *vector* object is there, then it would be very easy to amend the data extraction:
```python
geom_data = out.read_force()
fmax = geom_data.abs().max()
total = geom_data.sum(axis=0)
```
or something similar.

Another problem with your implementation is that if one wishes to read both `fmax` and `total` one have to read the full file twice, using the above one can simply read the complete data once, and post-process it :)
With this in mind I would prefer not to put this PR into sisl, hope this is ok?
I am still behind due to the COVID-19, I don't have much spare time... But I'll get there eventually! Thanks for all your work!!!

Regarding your last point.
It isn't so easy to read a file backwards (it seems like it is, but it ain't). :)
See [here](https://stackoverflow.com/questions/3568833/how-to-read-lines-from-a-file-in-python-starting-from-the-end) for instance. I don't know how to do it exactly, so I refrained. If it were to be added it should be added in the regular `Sile` object since that would be generally useful.
So until this functionality is there we can't really do anything (I would rather not read the entire file just to initialize a variable.

pfebrer96:
Hi Nick, thanks for the answer!

> However, I think reading those variables are redundant in the interface.
Instead of your additions one can do:

Yes, sure you can do that. The problem is that SIESTA's default is not to write atomic forces. Therefore, by default the output file only contains the total force on each axis and the maximum force. I believe in most cases the only relevant quantity for monitoring is the max force and therefore users will not have the `WriteForces` flag set to true. Actually that was my case, and this is why I did it. Knowing this o you still think that most people will have the atomic forces in their output files so there's no need to implement it?

> Another problem with your implementation is that if one wishes to read both fmax and total one have to read the full file twice, using the above one can simply read the complete data once, and post-process it :)

We could definitely add an argument to get both :)

> It isn't so easy to read a file backwards

Well, in this case we only need to check if the file ends with that string. I don't know if there is any other case that could benefit from this, but a method that does this could be:

```python
def ends_with(self, string, bytes_per_character=1):
    
    with open(self.basefile, "rb") as fh:
    
        #Position the file handle -<length of the string + 2 (\n)> bytes from the end of the file (the 2 means end)
        fh.seek( -(len(string) + 2) * bytes_per_character, 2)
        
        #Check if the string appears from that point to the end
        return string in str(fh.read())
```
And then, in `__init__()` go like:

```python
self._job_completed = self.ends_with("Job completed")
```

Or alternatively use the tail function if you don't want to support windows, like:

```python
last_line = subprocess.check_output(['tail', '-1', filename])
```

> I would rather work on #11 which abstracts away a geometry related quantity. Once the vector object is there, then it would be very easy to amend the data extraction:

With your examples I can't see the difference between that Vector object and a numpy array. If you expand more what this should look like maybe I can work on it.

Cheers!

pfebrer96:
By the way, do you think it makes sense to have the possibility of returning an `xarray` DataArray from the `read_scf` method so that it is explicit what each value means?

zerothi:
> The problem is that SIESTA's default is not to write atomic forces. Therefore, by default the output file only contains the total force on each axis and the maximum force. I believe in most cases the only relevant quantity for monitoring is the max force and therefore users will not have the WriteForces flag set to true. 

Agreed, totally forgot about this.

Good with  your snippet, yet another thing on the table I need to look into :)
Agreed that `tail` is a bad choice.

> With your examples I can't see the difference between that Vector object and a numpy array. If you expand more what this should look like maybe I can work on it.

Well, the point I want with the `Vector` object, or more likely `GeometryData` (to clarify) would be to attach some value to atoms. Also, it would mean one could do additional data-handling on groups of data.

> By the way, do you think it makes sense to have the possibility of returning an xarray DataArray from the read_scf method so that it is explicit what each value means?

Yes, that would definitely be a good idea!

sisl/io/siesta/out.py
```diff
@@ -239,7 +239,7 @@ def next_geom():
         return next_geom()[1]
 
     @sile_fh_open()
-    def read_force(self, last=True, all=False):
+    def read_force(self, last=True, all=False, total=False, maxF=False):
```
zerothi: I would prefer `max` rather than `maxF`, it is implicit is about forces :)

sisl/io/siesta/out.py
```diff
@@ -249,18 +249,39 @@ def read_force(self, last=True, all=False):
         all: bool, optional
            return a list of all forces (like an MD)
            If `True` `last` is ignored
+        total: bool, optional
+            return the total forces instead of the atomic forces.
+        maxF: bool, optional
+            whether only the maximum atomic force should be returned for each step.
+            If `True`, `total` is ignored.
```
zerothi: Allow both `total` and `max` at the same time.

sisl/io/siesta/out.py
```diff
@@ -249,18 +249,39 @@ def read_force(self, last=True, all=False):
...
         Returns
         -------
         numpy.ndarray or None
             returns ``None`` if the forces are not found in the
             output, otherwise forces will be returned
+
+            The shape of the array will be different depending on the type of forces requested:
+                - atomic (default): (nMDsteps, nAtoms, 3)
+                - total: (nMDsteps, 3)
+                - maxF: (nMDsteps, )
```
zerothi: If both are queried then it should probably be `(nMDsteps, 4)`, or what do you think?

sisl/io/siesta/out.py
pfebrer96: Yes, I did it not to overwrite the `max()` builtin function, but if we don't ever plan to use it, I guess it's safe to use max.

sisl/io/siesta/out.py
pfebrer96: Will do!

sisl/io/siesta/out.py
zerothi: I thought so ;)
One can easily do `from numpy import amax as ...` or `_max = max` outside the function. So I think we could bypass this :)

sisl/io/siesta/out.py
pfebrer96: I don't know, do you think there will be a case in which it will be needed to do the same operation on both at the same time?

Otherwise we could just return [ total, maxF] (or the opposite). The user could zip them in a loop and still kind of have them separately. I mean

```python
for totalF, max in zip(total, maxF):
     # totalF[0] to get Fx, etc...
```
vs

```python
for Fx, Fy, Fz, max in forces:
```

I don't know what's best. Whatever you prefer.

pfebrer96: I just realized that you can go like 

```python
for *totalF, max in forces:
```
so my point is useless (unless you consider that begginers in python will not come up with this probably). But I think it's more a matter of the two things not quite belonging to the same property that confuses me if you put them together in an array. Idk

zerothi: yeah, probably return a tuple, yeah do that :)

pfebrer96:
> By the way, do you think it makes sense to have the possibility of returning an xarray DataArray from the read_scf method so that it is explicit what each value means?

I've been looking into this and it should be an `xarray.Dataset` instead of an `xarray.DataArray`, because there is more than 1 dependent variable. In fact, there are at least 6:  `["Eharris","E_KS", "FreeEng", "dDmax", "Ef", "dHmax"]`. In theory, there might be 2 independent variables if all MD steps and all iscf are requested ( `["MDstep", "iscf"]` ), but otherwise there is only one. In the case there is 1, I think it is much more convenient to use a `pandas.DataFrame`. Even in the case with 2 independent variables, "iscf" could really be made dependent so that you can have a reguar `pandas.DataFrame`. Since different MDsteps have different numbers of iscf you end up with a very empty `xarray.Dataset` anyway, so I don't know what's the difference in memory.

All in all, I know that `pandas` would be an extra dependency, but I don't see much advantage with `xarray` unless the data is in a grid (in that situation there are no empty slots).

I'm just writing this to keep this thought here for further possible discussion :)

zerothi:
@pfebrer96 could you please check the latest commit, I have just fixed the method for clarity and your test example. :)

PR #179, opened 2020-03-23
Possibility to return bands as dataarray

pfebrer96:
I thought it would be nice to have this feature for bands too.

The calculation of the gap probably should not be there. If the returned array was a custom object, `gap` would be a property. But if not, I think it's a nice thing to pass to the users for convenience. I can definitely remove it from there if you want.

### DIFF
--- a/sisl/io/siesta/bands.py
+++ b/sisl/io/siesta/bands.py
@@ -89,7 +89,7 @@ class bandsSileSiesta(SileSiesta):
 
             ticks = {"tick_vals": xlabels, "tick_labels": labels} if band_lines else {}
             
-            arr = DataArray(
+            return DataArray(
                 name="Energy",
                 data=b,
                 coords=[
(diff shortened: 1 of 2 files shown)


pfebrer96:
Done, I removed it :)

zerothi:
Merged, please note I have changed names for coords and attributes :)

Thanks!

pfebrer96:
Good! By the way, do you think it makes sense in these cases to have the converter to xarray/pandas as a static method so that other parts of the code can access it?

For example, let's say I got the bands values from BandStructure.eigh() but I want them as a dataarray. I would have to rewrite all the code, assuming that I know how to build a DataArray properly, or copy/paste it in case I knew where the code that I need is located (unlikely to happen with users).

I guess there should be a way to unify these different cases, don't you think? If it's not a static method in the sile class, it could be a class for each important/common property, such as PDOS or bands.

PR #180, opened 2020-03-26
Fixed some bugs in the out sile

pfebrer96:
Some things were wrong with my previous implementation, sorry.

- The logic for reading atomic forces was `not total or max` when it should be `not (total or max)`.
- I've noticed (the hard way, i.e. because the method fails for one of my files), that the forces summary is not always written even if the job is completed. I don't know which versions of SIESTA show this behavior. I've added checking if `'siesta:'` is in the line with the forces value to dismiss the summary if it is there, as it does not provide new information. Do you think this is safe to do?
-  When `max and not all`, Fs was an adimensional numpy array and we were trying to retrieve the first item, which results in an error.

### DIFF
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -295,7 +295,7 @@ class outSileSiesta(SileSiesta):
             # First, we encounter the atomic forces
             while '---' not in line:
                 line = line.split()
-                if not total or max:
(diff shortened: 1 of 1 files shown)


zerothi:
Thanks, I'll post-fix a few things :)

PR #181, opened 2020-03-26
Twisted bilayers

tfrederiksen:
I implemented the construction of twisted bilayer commensurate cells according to Laissardiere et al., Nano Lett. 10, 804-808 (2010).

This implementation first grows two layers, rotates one with respect to the other, and then remove excess atoms outside the unit cell. The latter step clearly adds an overhead to the construction. Maybe this can be improved.

### DIFF
--- a/sisl/geom/bilayer.py
+++ b/sisl/geom/bilayer.py
@@ -5,8 +5,8 @@ from sisl import Atom, geom, Cuboid
 __all__ = ['bilayer']
 
 
-def bilayer(bond, atom=None, twist=(5, 6), separation=3.35, return_angle=False, layer='both'):
-    """ Commensurate unit cell of a twisted (hexagonal) bilayer structure.
+def bilayer(bond=1.42, bottom_atom=None, top_atom=None, stacking='AB', twist=(0, 0), separation=3.35, ret_angle=False, layer='both'):
+    """ Commensurate unit cell of a hexagonal bilayer structure, possibly with a twist angle.
 
     This routine follows the prescription of twisted bilayer graphene found in
     Laissardiere et al., Nano Lett. 10, 804-808 (2010).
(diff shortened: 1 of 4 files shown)


sisl/geom/bilayer.py
```diff
@@ -0,0 +1,95 @@
...
+        m, n = n, m
+
+    if not (isinstance(n, int) and isinstance(m, int)):
+        raise RuntimeError('twist coordinates need to be integers!')
+
+    cos_theta = (n ** 2 + 4 * n * m + m ** 2) / (2 * (n ** 2 + n * m + m ** 2) )
+    theta = np.arccos(cos_theta) * 360 / (2 * np.pi)
+
+    # Build structure
+    graphene = geom.graphene(bond=bond, atom=atom)
```
zerothi: May I suggest doing `geom.honeycomb` to clarify that this is not merely graphene, but potentially also hBN?
And then say `honeycomb = geom.honeycomb` for further clarification.

sisl/geom/bilayer.py
```diff
@@ -0,0 +1,95 @@
...
+    bond : float
+       length between atoms in nano-tube
+    atom : Atom(6)
+       bilayer atoms
+    twist : (int, int)
+       coordinates (m, n) defining the twist angle (theta)
+    separation : float
+       distance between the two layers
+    return_angle : bool
+       computed twist angle is returned in addition to the geometry instance
```
zerothi: In what unit is `angle`?

sisl/geom/bilayer.py
```diff
@@ -0,0 +1,95 @@
...
+    if layer.lower() == 'bottom':
+        bilayer = bottom
+    elif layer.lower() == 'top':
+        bilayer = top
+    else:
+        bilayer = bottom.add(top)
+        natoms *= 2
+
+    # Remove atoms outside cell
+    cell_box = Cuboid(bilayer.cell[:], center=bilayer.sc.center())
```
zerothi: Using `Cuboid` is pretty efficient, so that seems like a good idea.

But `within` accepts multiple coordinates.
So simply do:
```python
inside_idx = cell_box.within_index(bilayer.xyz)
bilayer = bilayer.sub(inside_idx)
```
That should greatly speed it up :)

sisl/geom/bilayer.py
```diff
@@ -0,0 +1,95 @@
...
+    # Rotate whole cell
+    vec = bilayer.cell[0] + bilayer.cell[1]
+    vec_costh = np.dot([1, 0, 0], vec) / np.dot(vec, vec) ** .5
+    vec_th = np.arccos(vec_costh) * 360 / (2 * np.pi)
+    bilayer = bilayer.rotate(vec_th, [0, 0, 1])
+
+    # Sanity check
+    assert len(bilayer) == natoms
+
+    if return_angle:
```
zerothi: I have generally used `ret_*` for keywords, for consistency it would be nice if you changed to `ret_angle` :)

sisl/geom/nanotube.py
```diff
@@ -20,7 +20,7 @@ def nanotube(bond, atom=None, chirality=(1, 1)):
        chirality of nanotube (n, m)
     """
     if atom is None:
-        atom = Atom[6]
+        atom = Atom(Z=6, R=bond * 1.01)
```
zerothi: Great, thanks!

tfrederiksen:
Sorry for messing with the `master` branch, still learning the proper workflow with `git` (next time I will use a separate branch). I incorporated your comments in f5695c859bb205eae43694de62b3b4433b435b9e and also made some further improvements
- one can set the atom types for each layer
- when the indices are identical `m == n` one always gets the simple 4-atom unit cell
- more tests added

sisl/geom/bilayer.py
tfrederiksen: Good idea. Now used in f5695c859bb205eae43694de62b3b4433b435b9e

sisl/geom/bilayer.py
tfrederiksen: Fixed in f5695c859bb205eae43694de62b3b4433b435b9e

sisl/geom/bilayer.py
tfrederiksen: Used in f5695c859bb205eae43694de62b3b4433b435b9e

sisl/geom/bilayer.py
tfrederiksen: OK

sisl/geom/bilayer.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Compute twist angle
+    m, n = twist
+    m, n = abs(m), abs(n)
+    if m > n:
+        # Set m as the smaller integer
+        m, n = n, m
+
+    if not (isinstance(n, int) and isinstance(m, int)):
+        raise RuntimeError('twist coordinates need to be integers!')
```
zerothi: This should be a `ValueError`.

sisl/geom/bilayer.py
```diff
@@ -0,0 +1,116 @@
...
+    separation : float
+       distance between the two layers (in Angstrom)
+    ret_angle : bool
+       return the twist angle (in degrees) in addition to the geometry instance
+    layer : {'both', 'bottom', 'top'}
+       control which layer(s) to include in geometry
+    """
+    if bottom_atom is None:
+        bottom_atom = Atom(Z=6, R=bond * 1.01)
+    if top_atom is None:
```
zerothi: Wouldn't it make sense to do:
```python
if top_atom is None:
    top_atom = bottom_atom
```
?

sisl/geom/bilayer.py
```diff
@@ -0,0 +1,116 @@
...
+    elif layer.lower() == 'top':
+        bilayer = top
+    else:
+        bilayer = bottom.add(top)
+        natoms *= 2
+
+    if rep > 1:
+        # Remove atoms outside cell
+        center = 1e-3 * np.array([1, 1, 1])
+        cell_box = Cuboid(1.0 * bilayer.cell[:], center=center)
```
zerothi: I'll have a quick look at this. I think for clarity `center` should be shifted slightly negatively (to be consistent with the first atom of the bottom layer.

sisl/geom/bilayer.py
tfrederiksen: I get the point, but what if the user only specifies `top_atom`? Maybe one should rather get two identical layers, if only one of the two `atom` parameters is given?

sisl/geom/bilayer.py
tfrederiksen: OK, thanks

sisl/geom/bilayer.py
tfrederiksen: Maybe it would be better instead to select the atoms in each layer in the final cell before joining the two into one geometry? In order to have all bottom atoms appearing before the top ones (not sure if this ordering is currently respected)?

zerothi: I am pretty sure this is respected. The indices are in ascending order, so first bottom, then top.
I guess for extremely big systems it would be better to first create bottom, remove atoms, then create top, then remove atoms, then add.
But I don't think it is too important, the algorithms should be more or less the same. :)

sisl/geom/bilayer.py
zerothi: Agreed, good idea.

PR #185, opened 2020-03-27
enh: enable selective dynamics when writing POSCAR files

tfrederiksen:
For geometry optimizations with `VASP` I find the "Selective Dynamics" format very handy: one simply sets each degree of freedom to either `True` or `False`. I propose that `sisl` should use this format as default for writing `POSCAR` files.


zerothi:
Thanks for the PR, I took the liberty of extending the functionality to accept `fixed` keyword to selectively specify atoms dynamic setting.

PR #186, opened 2020-03-27
Bilayer again

tfrederiksen:
Some further polishing:
- keep the rotation axis in the middle of the cell (generates visually more intuitive structures in my opinion)
- make layers of identical atom types if only one layer (but either of them) is provided 
- added BA stacking (different from AB for stacked heterolayers)

### DIFF
--- a/sisl/geom/bilayer.py
+++ b/sisl/geom/bilayer.py
@@ -39,6 +39,8 @@ def bilayer(bond=1.42, bottom_atom=None, top_atom=None, stacking='AB',
     .. [1] G. Trambly de Laissardiere, D. Mayou, L. Magaud, "Localization of Dirac Electrons in Rotated Graphene Bilayers", Nano Letts. 10, 804-808 (2010)
     """
     if bottom_atom is None:
+        bottom_atom = top_atom
+    if bottom_atom is None:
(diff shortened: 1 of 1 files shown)


sisl/geom/bilayer.py
```diff
@@ -104,7 +108,7 @@ def bilayer(bond=1.42, bottom_atom=None, top_atom=None, stacking='AB',
     if rep > 1:
         # Remove atoms outside cell
         cell_box = Cuboid(bilayer.cell)
-        cell_box.set_origo([-0.0001] * 3)
+        cell_box.set_center([-0.0001] * 3)
```
zerothi: I am not too happy about this. With origo all atoms are inside the unit cell, with center only 1/4 of the atoms are in the unit cell. This was why I changed it.

Also, with origo the atoms in the middle are on top, with center, the atoms at origo is on top of each other?
If the geometry looks better then the geometry should be shifted back.

zerothi: I just had a look between yours and mine. To be honest I don't really have a preference on the appearance. I'll shift everything to the primary unit cell and keep your implementation.

zerothi:
This is now merged! Thanks!

PR #177, opened 2020-03-21
Support for xarray in read_scf outSile

pfebrer96:
The pull request that comes from https://github.com/zerothi/sisl/issues/175.

PR #188, opened 2020-03-28
Pandas support for outSile.read_scf

pfebrer96:
Copy of https://github.com/zerothi/sisl/pull/177

### DIFF
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -534,7 +534,7 @@ class outSileSiesta(SileSiesta):
         return val
 
     @sile_fh_open()
-    def read_scf(self, key='scf', iscf=-1, imd=None):
+    def read_scf(self, key='scf', iscf=-1, imd=None, as_dataframe=False):
         r""" Parse SCF information and return a table of SCF information depending on what is requested
 
         Parameters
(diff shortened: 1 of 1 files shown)


pfebrer96:
Can you check the way that a single iscf index is handled? I'm not sure it is the desired behavior.

If you ask for iscf=5 and all steps of the MD have that iscf, I've set iscf = 5 to the uppermost level index. However, if that iscf is not found in all the MD steps, iscf is just a column because it varies with the MD step.

Try for example:
```python
# First behavior
outSile.read_scf(iscf=1, as_dataframe=True)
# Second behavior
outSile.read_scf(iscf=-1, as_dataframe=True)
```
Probably they both should return the same structure instead of what I've done. What's your opinion on that?

pfebrer96:
By the way, the `props` variable will probably be different for transiesta. Can you check that? I've barely used tranSIESTA so I don't have any example or knowledge.

zerothi:
This is now merged with some minor edits.

For instance, I don't think we should change indices in DataFrames depending on what is present in the output. I.e. now the indices are consistent regardless of the content of the output file.

Thanks for the PR!

pfebrer96:
Great, thanks! I can now play with it to build a kind of scf monitoring plot :)

PR #190, opened 2020-03-29
enh: enabled writing without selective dynamics

zerothi:
Added new default of POSCAR file geometry writes to not add "Selective Dynamics" flags.

Thus *CAR files can still write selective dynamic flags, or not.

@tfrederiksen would you please have a look.

### DIFF
--- a/sisl/io/vasp/car.py
+++ b/sisl/io/vasp/car.py
@@ -23,7 +23,7 @@ class carSileVASP(SileVASP):
         self._scale = 1.
 
     @sile_fh_open()
-    def write_geometry(self, geometry, dynamic=None, sort=False):
+    def write_geometry(self, geometry, dynamic=True, group_species=False):
         r""" Writes the geometry to the contained file
 
         Parameters
@@ -34,9 +34,9 @@ class carSileVASP(SileVASP):
            define which atoms are dynamic in the VASP run (default is True,
            which means all atoms are dynamic).
            If None, the resulting file will not contain any dynamic flags
-        sort : bool, optional
+        group_species: bool, optional
            before writing `geometry` first re-order species to
-           have species in consecutive blocks (see `geometry_sort`)
+           have species in consecutive blocks (see `geometry_group`)
 
         Examples
         --------
@@ -48,13 +48,13 @@ class carSileVASP(SileVASP):
 
         See Also
         --------
-        geometry_sort: method used to sort atoms in geometry
+        geometry_group: method used to group atoms together according to their species
         """
         # Check that we can write to the file
         sile_raise_write(self)
 
-        if sort:
-            geometry, idx = self.geometry_sort(geometry, ret_index=True)
+        if group_species:
+            geometry, idx = self.geometry_group(geometry, ret_index=True)
         else:
             # small hack to allow dynamic
             idx = _a.arangei(len(geometry))
--- a/sisl/io/vasp/car.py
+++ b/sisl/io/vasp/car.py
@@ -50,11 +50,14 @@ class carSileVASP(SileVASP):
         --------
         geometry_sort: method used to sort atoms in geometry
         """
-        if sort:
-            geometry = self.geometry_sort(geometry)
-
         # Check that we can write to the file
         sile_raise_write(self)
+
+        if sort:
+            geometry, idx = self.geometry_sort(geometry, ret_index=True)
+        else:
+            # small hack to allow dynamic
+            idx = _a.arangei(len(geometry))
 
         # LABEL
         self._write('sisl output\n')
@@ -107,7 +110,7 @@ class carSileVASP(SileVASP):
 
         fmt = '{:18.9f}' * 3
         for ia in geometry:
-            self._write(fmt.format(*geometry.xyz[ia, :]) + todyn(dynamic[ia]))
+            self._write(fmt.format(*geometry.xyz[ia, :]) + todyn(dynamic[idx[ia]]))
 
     @sile_fh_open(True)
     def read_supercell(self):
--- a/sisl/io/vasp/car.py
+++ b/sisl/io/vasp/car.py
@@ -23,7 +23,7 @@ class carSileVASP(SileVASP):
         self._scale = 1.
 
     @sile_fh_open()
-    def write_geometry(self, geometry, dynamic=None):
+    def write_geometry(self, geometry, dynamic=None, sort=False):
         r""" Writes the geometry to the contained file
 
         Parameters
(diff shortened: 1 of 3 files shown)


tfrederiksen:
Yes, this looks good. (But I am not aware of good reasons **not** to work with "Selective dynamics" as default.)

In any case I was wondering if it would make sense to include more generally info about "dynamic or static atoms" in the geometry class? Such that the user does not have to pass around this info from read/write calls? Say in the workflow `read -> manipulation -> write` it is natural to expect that all info (i.e. including constraints) is carried over to the final file.

tfrederiksen:
By the way, another thing is the ordering of atoms in the VASP format. It is convenient to group each element in a block. Otherwise the combined `POTCAR` file needs to contain multiple copies of the same elemental `POTCAR`s.

I would suggest to have a keyword for the user to ask for this compact grouping of atoms. Perhaps called `compact=False`?

tfrederiksen:
A method for reordering atoms into groups of identical elements (species) could perhaps be a method in the `Geometry` class?

zerothi:
> Yes, this looks good. (But I am not aware of good reasons **not** to work with "Selective dynamics" as default.)

Mainly because LOCPOT and other output files contains somewhat the same as a POSCAR file, but also without some information. When [reading documentation about POSCAR files](https://cms.mpi.univie.ac.at/vasp/guide/node59.html) it seems that `Selective dynamics` is optional, and my interpretation is that it is not the default. In any case, what would an MD simulation without Selective dynamics in the POSCAR do? Wouldn't it be the same as `write_geometry(..., dynamic=True)` or?
 
> In any case I was wondering if it would make sense to include more generally info about "dynamic or static atoms" in the geometry class? Such that the user does not have to pass around this info from read/write calls? Say in the workflow `read -> manipulation -> write` it is natural to expect that all info (i.e. including constraints) is carried over to the final file.

Yes, my problem is how to convey this. I have been given this some thought over the past weeks, and I am starting to think about a stable and reliable implementation that has additional valuable information.
It

tfrederiksen:
> it seems that `Selective dynamics` is optional, and my interpretation is that it is not the default.

I think there is no "default" as such. The user needs to supply a POSCAR file either with or without the "Selective dynamics" line. (I always use the latter).

zerothi:
> By the way, another thing is the ordering of atoms in the VASP format. It is convenient to group each element in a block. Otherwise the combined `POTCAR` file needs to contain multiple copies of the same elemental `POTCAR`s.
> 
> I would suggest to have a keyword for the user to ask for this compact grouping of atoms. Perhaps called `compact=False`?

Yes, good idea.

> A method for reordering atoms into groups of identical elements (species) could perhaps be a method in the Geometry class?

Hmm.. I am a little reluctant for this. The `Geometry` contains already too many methods and I would really try and limit it.
What I would do is add a method to VASP siles for ordering species? Then one could do:

```python
geometry = sisl.io.vasp.carSile.geometry_sort(geometry)
# or simply
poscarSile(...).write_geometry(geom, sort=True)
```
which orders it.

zerothi:
Or should it be named `sort_specie` or something?

tfrederiksen:
> The `Geometry` contains already too many methods and I would really try and limit it.

I understand, but I still think it could be convenient to have some general methods for reordering geometries, such as
- ascending/descending along lattice vectors
- along axes
- species
- group labeling
- combinations (e.g. first species, then axes)

zerothi:
> > it seems that `Selective dynamics` is optional, and my interpretation is that it is not the default.
> 
> I think there is no "default" as such. The user needs to supply a POSCAR file either with or without the "Selective dynamics" line. (I always use the latter).

But if it is not there, what would VASP do? I would rather not write optional stuff when it isn't needed. 

I just scoured VASP sources. By default all POSCAR files does _not_ write out such information. Only the `CONTCAR` file contains the selective dynamics info. As such I would prefer not to do it?

I.e. if the default behaviour of VASP is equivalent to `dynamic=True` then I think the default should either be `dynamic=None` or `dynamic=True`.

zerothi:
> > The `Geometry` contains already too many methods and I would really try and limit it.
> 
> I understand, but I still think it could be convenient to have some general methods for reordering geometries, such as
> 
> * ascending/descending along lattice vectors
> * along axes
> * species
> * group labeling
> * combinations (e.g. first species, then axes)

Yes, a single method doing all/combinations of this would be acceptable. The current `sort` function could be extended to do the above. :)

tfrederiksen:
> what would an MD simulation without Selective dynamics in the POSCAR do

All atoms will then be considered dynamic in VASP.

> I just scoured VASP sources. By default all POSCAR files does _not_ write out such information. Only the `CONTCAR` file contains the selective dynamics info.

I do not understand this. VASP only reads (never writes) POSCAR and only writes (never reads) CONTCAR. It is therefore up to the user to decide between the two formats.

zerothi:
> > what would an MD simulation without Selective dynamics in the POSCAR do
> 
> All atoms will then be considered dynamic in VASP.

Ok, what would be the benefit of then writing `Selective dynamics` vs. nothing? :)
Then we should either do `dynamic=None` or  `dynamic=True` as the default. I would ok both, what would you advice?

> 
> > I just scoured VASP sources. By default all POSCAR files does _not_ write out such information. Only the `CONTCAR` file contains the selective dynamics info.
> 
> I do not understand this. VASP only reads (never writes) POSCAR and only writes (never reads) CONTCAR. It is therefore up to the user to decide between the two formats.

Sorry, it is because VASP uses the POSCAR file format for nearly all outputs, CHG, LOCPOT, etc.

zerothi:
Ok, now we just need to clear out the behaviour.

Should the default be:
- [ ] `dynamic=None`
- [ ] `dynamic=True`

Also, do you like the name `sort`? I thought `compact` was not too specific, perhaps `sort_species`? Or what do you think?
Perhaps `sort` is fine enough since users of VASP knows the _problem_ of `POTCAR` and geometries? Well, what do you think?

tfrederiksen:
I understand the logic if `dynamic=None` were to be the default. However, if one uses `sgeom` from the command line, one would end up with a POSCAR that is not so easily adapted to "Selective dynamics". It would be more convenient to have `dynamic=True` - then the user can easily adapt the constraints by editing the output.

Regarding the keyword `sort` maybe this meaning should be reserved to the planned `Geometry` class function? Here `sort_species`, `group_elements`, or `group_species` would be more descriptive for what it desirable when working with VASP?

zerothi:
Ok, `dynamic=True` and `sort -> group_species` it is (I like that). :)

PR #195, opened 2020-04-01
Ensure that orbital is pickleable

pfebrer96:
Hey Nick!

If you could merge this or a similar fix to avoid problems when an orbital doesn't have a "f" attribute or it is not a function it would make me very happy, because it was messing with the plots' pickling abilities :)

Thaaanks!


zerothi:
I have now made another commit which ensures it is always set.

I think that is a more stringent behaviour. I.e. if it isn't set, it will always return 0.

Thanks for bringing this to my attention.

zerothi:
@pfebrer96 could you check if it works now?

pfebrer96:
Yes, it works, thanks!

Issue #191, opened 2020-03-30
Geometry.sort


zerothi:
I understand, but I still think it could be convenient to have some general methods for reordering geometries, such as
- [x] ascending/descending along lattice vectors
- [x] along axes
- [ ] species
- [ ] group labeling
- [x] combinations (e.g. first species, then axes), this is particularly easy since Python >=3.6 uses ordered dicts for `**kwargs`, so `sort` should change interface to `**kwargs`
- [x] one of the above on a subset of atoms

These should be added to `Geometry.sort`

_Originally posted by @tfrederiksen in https://github.com/zerothi/sisl/pull/190#issuecomment-605926535_

zerothi:
@tfrederiksen for group labelling, do you mean the `geometry['name']` thingy? Or by *tags* for atoms?

tfrederiksen:
> @tfrederiksen for group labelling, do you mean the `geometry['name']` thingy? Or by _tags_ for atoms?

I had the first thingy in mind, but perhaps both possibilities are reasonable?

As a thought example, what would be a convenient way for the user to reorder a geometry into (i) first a molecule sorted according to ascending atomic numbers, then (ii) the substrate atoms ordered according to lattice vectors (but not atomic numbers).

By the way, another useful feature could be to also translate atoms by lattice vectors to ensure they are in the primitive cell (before/after sorting). I know this could be rather easily done by the user querying `Geometry.fxyz`, but some simple functionality could be cool to have.

zerothi:
> By the way, another useful feature could be to also translate atoms by lattice vectors to ensure they are in the primitive cell (before/after sorting). I know this could be rather easily done by the user querying `Geometry.fxyz`, but some simple functionality could be cool to have.

Yes, this would be ideal. I don't think this functionality should go in `sort`.

tfrederiksen:
You are right, a `move_inside_cell`-type function is different from `sort` as it would not changing the atom ordering in the geometry. But the result could be different if called before or after a `sort`.

zerothi:
I'll close this as it is tracked in #197

PR #197, opened 2020-04-01
WIP: improved Geometry.sort, for #191

zerothi:
This change allows very complicated sort calls.

- [x] ascending/descending along lattice vectors
- [x] ascending/descending along Cartesian coordinates
- [x] ascending/descending along a vector
- [x] external function to sort indices
- [x] species
- [x] group labelling
- [x] combinations (e.g. first species, then axes), this is particularly easy since Python >=3.6 uses ordered dicts for `**kwargs`, so `sort` should change interface to `**kwargs`
- [x] the above on a subset of atoms

All sorts are now chained (much like lexsort).
One can now do

    geom.sort(axis=0, lattice=0, axis1=1)

to first sort Cartesian x, then for each subgroup in this
sort according to first lattice vector, then for each subgroup
sort according to Cartesian y.

Note how axis1 allows numeric flags to allow multiple same
sorting algorithms for the same (to achieve different order
of sort).
One can also add ascending/descending

    geom.sort(axis=0, ascend=False, lattice=1, ascend=True, axis=1)

to ascend-sort x, descend-sort fa, ascend-sort y.

Also added preliminary support for sort in sgeom.

@tfrederiksen could you please have a look

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 0.9.9
 =====
 
+- Changed Geometry.sort to be more diverse (this may break old code)
+  This new way of sorting is way more flexible and allows very fine
+  control, partially fixes #191
+
 - Added a bilayer geometry which can create twisted bilayers #181, #186
 
 - Enabled VASP *CAR files to write/read dynamic specifications #185
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,7 +3,7 @@
 
 - Changed Geometry.sort to be more diverse (this may break old code)
   This new way of sorting is way more flexible and allows very fine
-  control, partially fixes #191
+  control, fixes #191, #197
 
 - Added a bilayer geometry which can create twisted bilayers #181, #186
 
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -8,7 +8,7 @@ import numpy as np
 from numpy import ndarray, int32, bool_
 from numpy import dot, square, sqrt, diff
 from numpy import floor, ceil
-from numpy import argsort, split, concatenate
+from numpy import argsort, split, isin, concatenate
 
 from . import _plot as plt
 from . import _array as _a
@@ -1028,6 +1028,7 @@ class Geometry(SuperCellChild):
         - by Cartesian coordinates, `axis`
         - by lattice vectors, `lattice`
         - by user defined vectors, `vector`
+        - by grouping atoms, `group`
         - by a user defined function, `func`
 
         - a combination of the above in arbitrary order
@@ -1062,6 +1063,17 @@ class Geometry(SuperCellChild):
            sort along a user defined vector, similar to `lattice` but with a user defined
            direction. Note that `lattice` sorting and `vector` sorting are *only* equivalent
            when the lattice vector is orthogonal to the other lattice vectors.
+        group : {'Z', 'symbol', 'tag', 'species'} or (str, ...), optional
+           group together a set of atoms by various means.
+           `group` may be one of the listed strings.
+           For ``'Z'`` atoms will be grouped in atomic number
+           For ``'symbol'`` atoms will be grouped by their atomic symbol.
+           For ``'tag'`` atoms will be grouped by their atomic tag.
+           For ``'species'`` atoms will be sorted according to their specie index.
+           If a tuple/list is passed the first item is described. All subsequent items are a
+           list of groups, where each group comprises elements that should be sorted on an
+           equal footing. If one of the groups is None, that group will be replaced with all
+           non-mentioned elements. See examples.
         func : callable, optional
            pass a sorting function which should have an interface like ``func(geometry, atom, **kwargs)``.
            The first argument is the geometry to sort. The 2nd argument is a list of indices in
@@ -1180,6 +1192,18 @@ class Geometry(SuperCellChild):
         [0, 1, 2, 3]
 
         This would split the sorting into sections ``a[:1]``, ``a[1:2]``, ``a[2:3]``, ``a[3:4]`` and ``a[4:5]``
+
+        Sort by atomic numbers
+
+        >>> geom.sort(group='Z') # 5, 6, 7
+
+        One may group several elements together on an equal footing (``None`` means all non-mentioned elements)
+        The order of the groups are important (the first two are _not_ equal, the last three _are_ equal)
+
+        >>> geom.sort(group=('symbol', 'C', None)) # C, [B, N]
+        >>> geom.sort(group=('symbol', None, 'C')) # [B, N], C
+        >>> geom.sort(group=('symbol', ['N', 'B'], 'C')) # [B, N], C (B and N unaltered order)
+        >>> geom.sort(group=('symbol', ['B', 'N'], 'C')) # [B, N], C (B and N unaltered order)
         """
         # We need a way to easily handle nested lists
         # This small class handles lists and allows appending nested lists
@@ -1202,7 +1226,7 @@ class Geometry(SuperCellChild):
                     if sort:
                         self._idx.append(np.sort(idx))
                     else:
-                        self._idx.append(idx)
+                        self._idx.append(np.asarray(idx))
             def __iter__(self):
                 yield from self._idx
             def __len__(self):
@@ -1220,18 +1244,13 @@ class Geometry(SuperCellChild):
             ascend = kwargs['ascend']
             atol = kwargs['atol']
 
-            if len(nl) == 0:
-                # Loop reverse and digitize individually
-                jdx = argsort(val)
-                if ascend:
-                    d = diff(val[jdx]) > atol
-                else:
-                    jdx = jdx[::-1]
-                    d = diff(val[jdx]) < -atol
-                return NestedList(split(jdx, d.nonzero()[0] + 1), sort=True)
-
             new_nl = NestedList()
             for idx in nl:
+                if len(idx) <= 1:
+                    # no need for complexity
+                    new_nl.append(idx)
+                    continue
+
                 # Sort values
                 jdx = idx[argsort(val[idx])]
                 if ascend:
@@ -1286,9 +1305,6 @@ class Geometry(SuperCellChild):
             """
             User defined function
             """
-            if len(atom) == 0:
-                # Nothing has been sorted...
-                atom = [_a.arangei(len(self))]
             nl = NestedList()
             for a in atom:
                 # TODO add check that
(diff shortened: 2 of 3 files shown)


zerothi:
species/groups to come

sisl/geometry.py
```diff
@@ -1019,27 +1021,340 @@ def overlap(self, other, eps=0.1, offset=(0., 0., 0.), offset_other=(0., 0., 0.)
...
+        >>> assert geom.sort(axis2=1, axis0=0, axis1=2) == geom.sort(axis=(1, 0, 2))
+
+        A too high `atol` may have unexpected side-effects. This is because of the way
+        the sorting algorithm splits the sections for nested sorting.
+        So for coordinates with a continuous displacement the sorting may break and group
+        a large range into 1 group.
+
+        >>> a = np.arange(5) * 0.1
+        >>> (np.diff(a) > 0.05).nonzero()[0]
+        [0, 1, 2, 3]
```
tfrederiksen, line 1168: Sorry, I do not follow this example with the `a` array. Can you be more explicit how it relates to the `atol` discussion?

sisl/geometry.py
```diff
@@ -1019,27 +1021,268 @@ def overlap(self, other, eps=0.1, offset=(0., 0., 0.), offset_other=(0., 0., 0.)
...
+           This is different behaviour than `numpy.lexsort`!
+        lattice : int or tuple of int, optional
+           sort coordinates according to lattice vectors, if a tuple of
+           ints is passed it will be equivalent to ``sort(lattice0=lattice[0], lattice1=lattice[1])``.
+           This is different behaviour than `numpy.lexsort`!
+           Note that before sorting we multiply the fractional coordinates by the length of the
+           lattice vector. This ensures that `atol` is meaningful for both `axis` and `lattice` since
+           they will be on the same order of magnitude.
+        ascend : bool, optional
+            control ascending or descending sorting, default ``True``
```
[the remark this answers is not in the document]
tfrederiksen: Also, add `descend=False` to the description here?

sisl/geometry.py
```diff
@@ -3810,6 +4124,38 @@ def __call__(self, parser, ns, value, option_string=None):
...
+                    else:
+                        # it must be a range/tuple
+                        val = lstranges(strmap(int, val))
+                    # parse `val` to appropriate values
+                    # we always add integers to allow users to use the same keywords on commandline
+                    kwargs[opt.strip() + str(i)] = val
+                ns._geometry = ns._geometry.sort(**kwargs)
+        p.add_argument(*opts('--sort'), nargs=1, metavar='SORT',
+                       action=Sort,
+                       help='Semi-colon separated options for sort [axis=0;descend;lattice=(1, 2)].')
```
tfrederiksen, line 4157: Maybe stress that the CLI argument needs to be given with quotation marks, as in `--sort "axis=0;descend;lattice=(1, 2)"`.

These calls are invalid: `--sort axis=0;descend;lattice=(1,2)` and `--sort [axis=0;descend;lattice=(1, 2)]` (and the error message not too helpful to localize the mistake).

sisl/geometry.py
```diff
@@ -1019,27 +1021,340 @@ def overlap(self, other, eps=0.1, offset=(0., 0., 0.), offset_other=(0., 0., 0.)
...
+
+        Notes
+        -----
+        The order of arguments is also the sorting order. ``sort(axis=0, lattice=0)`` is different
+        from ``sort(lattice=0, axis=0)``
+
+        All arguments may be suffixed with integers. This allows multiple keyword arguments
+        to control sorting algorithms
+        in different order. It also allows changing of sorting settings between different calls.
+        Note that the integers have no relevance to the order of execution!
```
tfrederiksen, line 1092: Good explanation!

tfrederiksen:
Note that in your opening comment this line

> geom.sort(axis=0, ascend=False, lattice=1, ascend=True, axis=1)

has a clash of the repeated argument `axis`.

sisl/geometry.py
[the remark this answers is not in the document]
zerothi: Done!

sisl/geometry.py
```diff
@@ -1019,27 +1021,268 @@ def overlap(self, other, eps=0.1, offset=(0., 0., 0.), offset_other=(0., 0., 0.)
...
+
+        >>> geom.sort(axis=2, lattice=0)
+
+        Sort according to :math:`z` (ascending), then first lattice vector (descending)
+
+        >>> geom.sort(axis=2, ascend=False, lattice=0)
+
+        Sort according to :math:`z` (descending), then first lattice vector (ascending)
+
+        >>> geom.sort(ascend0=False, axis=2, ascend1=True, lattice=0)
```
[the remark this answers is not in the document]
zerothi: I have your example later down ;)

sisl/geometry.py
zerothi: I have tried to add a more reasonable explanation...

sisl/geometry.py
zerothi: thanks

zerothi:
Could you have a look again, the `atol` is a bit difficult to explain, for instance it would be really difficult for the sort algorithm to do a splitting of groups in twisted bilayers since there coordinates are continuous.
Using numpy `lexsort` forces values to be *unique*, my implementation does not impose this. The reason is that typically when rotating geometries one will get very tiny differences ~1e-15 which then breaks `lexsort`. I try to avoid this by not doing anything to coordinates that are continuously closer than `atol`.

For coordinates `[0.1, 0.2, 0.3]` and a tolerance of `0.05` one would have to split all into groups, whereas for `0.15` one could not split. One could then argue that one should split into either `[0.1, 0.2] + [0.3]` or `[0.1] + [0.2, 0.3]` but this is rather non-trivial...

tfrederiksen:
Why is `sgeom in.xyz --sort "axis=2;axis=1" out.xyz` allowed? It appears to me that it is not generating a clash of repeated argument?

Also, with `sgeom in.xyz --sort "descend;axis=2;axis=1" out.xyz` it seems that both axes are sorted descending (not just the first).

zerothi:
> Why is `sgeom in.xyz --sort "axis=2;axis=1" out.xyz` allowed? It appears to me that it is not generating a clash of repeated argument?

This is because I always append an integer to keywords (so users can do what you do ;))

> 
> Also, with `sgeom in.xyz --sort "descend;axis=2;axis=1" out.xyz` it seems that both axes are sorted descending (not just the first).

Yes, this is by design. Any keyword has precedence until changed. The updated documentation should reflect this. So one have to _disable_ it again, if one does not want this.

tfrederiksen:
Very cool! Now I notice that indeed you write "sorting _for all subsequent sorting methods_" in the documentation. So the user can also do this in a script `geom.sort(descend=True, axis=2, axis1=1)` and achieve the same as in my `sgeom` command above.

zerothi:
> Very cool! Now I notice that indeed you write "sorting _for all subsequent sorting methods_" in the documentation. So the user can also do this in a script `geom.sort(descend=True, axis=2, axis1=1)` and achieve the same as in my `sgeom` command above.

Yes :)

zerothi:
@tfrederiksen I have now added the remaining sortings, currently you cannot do the complicated groupings on the CLI, then one have to parse to a tuple...

You can however do the simple stuff, like doing `sgeom ... --sort "group=Z"`, but not with optional arguments...

sisl/geometry.py
zerothi: @tfrederiksen does it make sense now?

sisl/geometry.py
```diff
@@ -1019,27 +1021,451 @@ def overlap(self, other, eps=0.1, offset=(0., 0., 0.), offset_other=(0., 0., 0.)
...
+           as ``atom``.
+        ascend, descend : bool, optional
+            control ascending or descending sorting for all subsequent sorting methods.
+            Default ``ascend=True``.
+        atol : float, optional
+            absolute tolerance when sorting numerical arrays for subsequent sorting methods.
+            When a selection of sorted coordinates are grouped via `atol`, we ensure such
+            a group does not alter its indices. I.e. the group is *always* ascending indices.
+            Note this may have unwanted side-effects if `atol` is very large.
+            Default ``1e-9``.
```
tfrederiksen, line 1095: What is `very large`? Compared to what?

sisl/geometry.py
```diff
@@ -1019,27 +1021,451 @@ def overlap(self, other, eps=0.1, offset=(0., 0., 0.), offset_other=(0., 0., 0.)
...
+        >>> (np.diff(a) > 0.02).nonzero()[0]
+        [2]
+
+        This would split the sorting into sections ``a[:3]`` and ``a[3:]``
+
+        >>> (np.diff(a) > 0.002).nonzero()[0]
+        [0, 1, 2, 3]
+
+        This would split the sorting into sections ``a[:1]``, ``a[1:2]``, ``a[2:3]``, ``a[3:4]`` and ``a[4:5]``
+
```
tfrederiksen, line 1195: Clearer now, thanks. Could you add a more "realistic" example with relevance of changing (raising) `atol`?

Minor stuff: I would suggest to move this `atol` discussion to after the `sort(group=...)`.

CHANGELOG
```diff
@@ -1,6 +1,10 @@
 0.9.9
 =====
 
+- Changed Geometry.sort to be more diverse (this may break old code)
+  This new way of sorting is way more flexible and allows very fine
+  control, partially fixes #191
```
tfrederiksen, line 6: Fully fixes #191 and #197?

sisl/geometry.py
```diff
@@ -1019,27 +1021,451 @@ def overlap(self, other, eps=0.1, offset=(0., 0., 0.), offset_other=(0., 0., 0.)
...
+                # create numpy array
+                strs = np.array(list(map(lambda a: a.tag, self.atoms)))
+                return _group_vals(strs, groups, atom, **kwargs)
+            elif method == 'symbol':
+                strs = np.array(list(map(lambda a: a.symbol, self.atoms)))
+                return _group_vals(strs, groups, atom, **kwargs)
+            else:
+                raise ValueError(f"{self.__class__.__name__}.sort group only supports ['Z','species','tag','symbol'] for sorting")
+        funcs["group"] = _group
+
```
tfrederiksen, line 1403: Not sure why, but the command `sgeom in.xyz --sort "group=Z" out.xyz` fails for me with a `ValueError: invalid literal for int() with base 10: 'Z'`

sisl/geometry.py
zerothi: I have added ` is very large compared
            to the difference between atomic coordinates.`

sisl/geometry.py
zerothi: Moved it to be the last entry. Hmm... I did this small thing:

```python
An example would be a linear chain with a middle point with a much shorter distance.

        >>> x = np.arange(5) * 0.1
        >>> x[3:] -= 0.095
        y = z = np.zeros(5)
        geom = si.Geometry(np.stack((x, y, z), axis=1))
        >>> geom.xyz[:, 0]
        [0.    0.1   0.2   0.205 0.305]

        In this case a high tolerance (``atol>0.005`) would group atoms 2 and 3
        together

        >>> geom.sort(atol=0.01, axis=0, ret_atom=True)[1]
        [[0], [1], [2, 3], [4]]

        However, a very low tolerance will not find these two as atoms close
        to each other.

        >>> geom.sort(atol=0.001, axis=0, ret_atom=True)[1]
        [[0], [1], [2], [3], [4]]
```

sisl/geometry.py
zerothi: Thanks, now I fixed it :)

zerothi:
It is in ;)

PR #201, opened 2020-04-02
Nanoribbon geometry

tfrederiksen:
Easy construction of nanoribbon geometries
- armchair and zigzag type
- w/wo edge saturation

### DIFF
--- a/sisl/geom/nanoribbon.py
+++ b/sisl/geom/nanoribbon.py
@@ -5,7 +5,7 @@ from sisl import geom, Atom, Geometry
 __all__ = ['nanoribbon']
 
 
-def nanoribbon(width, kind='armchair', bond=1.42, atom=None, sat_bond=1.09, sat_atom=None):
+def nanoribbon(width, kind='armchair', bond=1.42, atom=None):
     r""" Construction of a nanoribbon unit cell of type armchair or zigzag.
     The geometry is oriented along the x-axis.
 
@@ -19,11 +19,6 @@ def nanoribbon(width, kind='armchair', bond=1.42, atom=None, sat_bond=1.09, sat_
        bond length between atoms in the honeycomb lattice. Defaults to 1.42
     atom : Atom, optional
        atom (or atoms) in the honeycomb lattice. Defaults to ``Atom(6)``
-    sat_bond : float, optional
-       bond length to the edge saturation atoms. A negative value will return 
-       a nanoribbon without edge decoration. Defaults to 1.09
-    sat_atom : Atom, optional
-       atom (or atoms) for the edge decoration. Defaults to ``Atom(1)``
 
     See Also
     --------
@@ -32,8 +27,6 @@ def nanoribbon(width, kind='armchair', bond=1.42, atom=None, sat_bond=1.09, sat_
     """
     if atom is None:
         atom = Atom(Z=6, R=bond * 1.01)
-    if sat_atom is None:
-        sat_atom = Atom(Z=1, R=sat_bond * 1.01)
 
     # Width characterization
     if not isinstance(width, int):
@@ -63,34 +56,15 @@ def nanoribbon(width, kind='armchair', bond=1.42, atom=None, sat_bond=1.09, sat_
         ribbon.xyz[:, 1] *= -1
         # Set lattice vectors strictly orthogonal
         ribbon.cell[:] = np.diag([ribbon.cell[1, 0], -ribbon.cell[0, 1], ribbon.cell[2, 2]])
+        # Sort along x-axis
+        ribbon = ribbon.sort()
     else:
         raise ValueError("nanoribbon: kind must be armchair or zigzag")
 
     # Separate ribbons along y-axis
     ribbon.cell[1, 1] += 10.
 
-    # Edge saturation?
-    if sat_bond > 0:
-        edge_atoms = []
-        # Find atoms with just two neighbors
-        for ia in ribbon:
-            idx = ribbon.close(ia, R=(0.1, bond + 0.1))
-            if len(idx[1]) == 2:
-                a, b = idx[1]
-                v1 = ribbon.axyz(a) - ribbon.xyz[ia, :]
-                v2 = ribbon.axyz(b) - ribbon.xyz[ia, :]
-                # Compute normalized vector p pointing away from the two neighbors
-                p = -1 * (v1 + v2)
-                p /= p.dot(p) ** .5
-                edge_atoms.append(ribbon.xyz[ia, :] + sat_bond * p)
-        if len(edge_atoms) > 0:
-            edge_atoms = Geometry(edge_atoms, atom=sat_atom)
-            ribbon = ribbon.add(edge_atoms)
-
-    # Sort along x-axis
-    ribbon = ribbon.sort()
-
-    # Place ribbon inside unit cell
+    # Movie inside unit cell
     ribbon = ribbon.move([-np.min(ribbon.xyz[:, 0]), -np.min(ribbon.xyz[:, 1]), 0])
 
     return ribbon
--- a/sisl/geom/nanoribbon.py
+++ b/sisl/geom/nanoribbon.py
@@ -5,25 +5,25 @@ from sisl import geom, Atom, Geometry
 __all__ = ['nanoribbon']
 
 
-def nanoribbon(width, kind='armchair', bond=1.42, atom=None):
+def nanoribbon(bond, atom, width, kind='armchair'):
     r""" Construction of a nanoribbon unit cell of type armchair or zigzag.
     The geometry is oriented along the x-axis.
 
     Parameters
     ----------
+    bond : float
+       bond length between atoms in the honeycomb lattice. Defaults to 1.42
+    atom : Atom
+       atom (or atoms) in the honeycomb lattice
     width : int
        number of atoms in the transverse direction
     kind : {'armchair', 'zigzag'}
        type of ribbon
-    bond : float, optional
-       bond length between atoms in the honeycomb lattice. Defaults to 1.42
-    atom : Atom, optional
-       atom (or atoms) in the honeycomb lattice. Defaults to ``Atom(6)``
 
     See Also
     --------
-    honeycomb: honeycomb lattices
-    graphene: graphene geometry
+    honeycomb : honeycomb lattices
+    graphene : graphene geometry
     """
     if atom is None:
         atom = Atom(Z=6, R=bond * 1.01)
--- a/sisl/geom/nanoribbon.py
+++ b/sisl/geom/nanoribbon.py
@@ -61,7 +61,7 @@ def nanoribbon(bond, atom, width, kind='armchair'):
         # Set lattice vectors strictly orthogonal
         ribbon.cell[:] = np.diag([ribbon.cell[1, 0], -ribbon.cell[0, 1], ribbon.cell[2, 2]])
         # Sort along x-axis
-        ribbon = ribbon.sort()
+        ribbon = ribbon.sort(axis=(0, 1))
     else:
         raise ValueError("nanoribbon: kind must be armchair or zigzag")
 
--- a/sisl/geom/nanoribbon.py
+++ b/sisl/geom/nanoribbon.py
@@ -1,7 +1,6 @@
 import numpy as np
 
-from math import acos, pi
-from sisl import geom, Atom, Cuboid
+from sisl import geom, Atom, Geometry
 
 __all__ = ['nanoribbon']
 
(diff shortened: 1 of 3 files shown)


sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,96 @@
+import numpy as np
+
+from sisl import geom, Atom, Geometry
+
+__all__ = ['nanoribbon']
+
+
+def nanoribbon(width, kind='armchair', bond=1.42, atom=None, sat_bond=1.09, sat_atom=None):
```
zerothi, line 8: I would prefer the order `bond, atom, width, kind='armchair', saturate/saturation=None`.

Then I would do  
`saturation` may be either an `Atom` or `tuple(Atom, bond)`
If only an atom I would read the `calculated` radius from the `PeriodicTable` for the nano-ribbon atom and the saturation atom and use that as the bond.
Something like:
```python
sat_bond = sisl.PeriodicTable.radius(atom.Z) + sisl.PeriodicTable(saturation.Z)
```
For C and H it would be `1.2 A`. For a tuple, the bond length supplied will be used.

Also, since H does not really play a big role it should perhaps default to *not* add saturation atoms if `saturation` is None? I.e. for TB one would not use the hydrogens?

sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,96 @@
...
+    r""" Construction of a nanoribbon unit cell of type armchair or zigzag.
+    The geometry is oriented along the x-axis.
+
+    Parameters
+    ----------
+    width : int
+       number of atoms in the transverse direction
+    kind : {'armchair', 'zigzag'}
+       type of ribbon
```
zerothi, line 17: Perhaps add a comment about which direction the resulting geometry would have, i.e. `x` vs `y` direction.

sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,96 @@
...
+    # Width characterization
+    if not isinstance(width, int):
+        raise ValueError("nanoribbon: the width needs to be a postive integer!")
+    width = max(width, 1)
+    n, m = width // 2, width % 2
+
+    ribbon = geom.honeycomb(bond, atom, orthogonal=True)
+
+    kind = kind.lower()
+    if kind.startswith('a'):
```
zerothi, line 47: probably we should force `kind=='armchair'` to be sure users do it correctly.
For instance with the nano-porous graphene GNR the `kind` could also be added.

Actually one could add an optional parameter `length=1` which automatically tiles along length of ribbon. Then `width` and `length` could be tuples/lists where one appends the nanoribbons of those individual ribbons together? That would make one able to do rather complicated nano-ribbons with very little complexity (append would work seamlessly and a nested call would be really easy ;)

What do you think?
If

sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,96 @@
...
+            ribbon = ribbon.tile(n + 1, 0)
+            ribbon = ribbon.remove(-1).remove(-1)
+        else:
+            ribbon = ribbon.tile(n, 0)
+        # Invert y-coordinates
+        ribbon.xyz[:, 1] *= -1
+        # Set lattice vectors strictly orthogonal
+        ribbon.cell[:] = np.diag([ribbon.cell[1, 0], -ribbon.cell[0, 1], ribbon.cell[2, 2]])
+    else:
+        raise ValueError("ribbon: kind must be armchair or zigzag")
```
zerothi, line 67: typo, `nanoribbon`

pfebrer96:
Hey! Nice work :)

I need lots of different ribbons for a project that I am doing and I built something similar. There are two things that I "discovered" and I think they could be useful for this pull request, if you don't mind me suggesting them:

- The saturation part would be very nice as a standalone function. If there is only one missing bond, the calculation of the direction of the missing bond works exactly the same for all the coordinations that I can think of (linear, trigonal, tetrahedral, quadroplanar, octahedral...), it's always the sum of all bonds and then change the sign. Adding Nick suggestions about possibility to saturate with any atom it would be pretty general. I guess you would only need an extra argument `min_neighbours`, and below that number, you would saturate. The only thing that I don't know how to build easily is when more than one bond is missing, but I believe it's worth giving it a thought to provide the saturating functionality for any geometry that it's fairly normal, because doing it manually is a pain :)

- This second one is less important, but I'm studying specifically ribbons that don't have a constant width and it is impossible to find a way of programming them (there are endless possibilities on how to join them). Therefore I wrote an interactive ribbon builder that displays your current ribbon (with Geometry.__plot__()) and asks for input to add a section with a different width, tile the current section, move it up and down or offset its unit cell by one half (you can build any ribbon basically). Do you think it's meaningful to make it available here also making use of Thomas nanoribbon builder or is it a specific thing that I use but nobody will need? 

Cheers!

pfebrer96:
Forget the second part, sorry, but having saturation as a function would be great!

zerothi:
> * The saturation part would be very nice as a standalone function. If there is only one missing bond, the calculation of the direction of the missing bond works exactly the same for all the coordinations that I can think of (linear, trigonal, tetrahedral, quadroplanar, octahedral...), it's always the sum of all bonds and then change the sign. Adding Nick suggestions about possibility to saturate with any atom it would be pretty general. I guess you would only need an extra argument `min_neighbours`, and below that number, you would saturate. The only thing that I don't know how to build easily is when more than one bond is missing, but I believe it's worth giving it a thought to provide the saturating functionality for any geometry that it's fairly normal, because doing it manually is a pain :)

I don't get this? ;)
Do you mean if you saturate with more than 1 atom at the edges, or?

> * This second one is less important, but I'm studying specifically ribbons that don't have a constant width and it is impossible to find a way of programming them (there are endless possibilities on how to join them). Therefore I wrote an interactive ribbon builder that displays your current ribbon (with Geometry.**plot**()) and asks for input to add a section with a different width, tile the current section, move it up and down or offset its unit cell by one half (you can build any ribbon basically). Do you think it's meaningful to make it available here also making use of Thomas nanoribbon builder or is it a specific thing that I use but nobody will need?

Hmm, actually with my suggestion wouldn't attaching different ribbons be unique with 3 different integers: `width`, `length` and `offset`

E.g. something like this:
```python
def shift(r, offset=0):
    # shift to center
    xyz = np.average(r.xyz, axis=0)
    xyz[0] = 0.
    return r.move(-xyz)

ribbon = shift(nanoribbon(6))
for w, l, o in [(10, 2, 0), (6, 2, 0), (10, 2, 1), (6, 1, 1)]:
    ribbon = ribbon.append(shift(nanoribbon(w), o).tile(l, 0), 0)
```
Yes, there are some problems with the above for offsets and widths that don't play well. But one *must* be able to figure out how the commensurate stuff would work ;)

> 
> Cheers!

Thanks for chipping in Pol!

pfebrer96:
> I don't get this? ;)
> Do you mean if you saturate with more than 1 atom at the edges, or?

Not at the edges of a ribbon, of course, because that be weird :) The easiest example I can think of (maybe with little relevance in the objective of sisl) is a terminal carbon of an organic molecule, it can need either 1, 2 or 3 saturating Hs. Is it possible that some surfaces might also have more than one dangling bond per atom? Idk. But if not, this means less problems then!

The saturate function could accept a function to decide if an atom needs to be saturated, apart from `min_neighbors`. Maybe it is useful for someone. I'm thinking of, for example :

```python
saturate(geom, min_neighbours = 3, sat_atom=O, should_saturate=lambda at: random())
```

to randomly oxidize a surface.

> Hmm, actually with my suggestion wouldn't attaching different ribbons be unique with 3 different integers: width, length and offset

Thomas answered for me :)

> There are really many ways to put individual GNRs together. The basic units created here could be aligned according to their center axis, along their edges, and many other ways. I tend to think these more complicated structures belong to some separate method, or? Maybe some higher-level growth method?

Of course one can determine the exact parameters that it needs to build a ribbon. But at some point, it stops making sense and doing it interactively can be faster. In fact, what I did is tracking the commands issued by the user to build some kind of "fingerprint". You can also provide the fingerprint instead of triggering the interactive mode, so after your first build, you may be able to automatically generate ribbons from that fingerprint (tiling a region a bit more, shifting up or down, etc...)

This is an example of a ribbon that, sure, you can program its structure, but it's just painful (note that there are single chains of a given width at some point):
Screenshot from 2020-04-02 23-28-22

sisl/geom/nanoribbon.py
zerothi: Agreed, removed saturation stuff from this PR, could you in the same go update the argument order? When changed it matches better the other geometry creation tools.

sisl/geom/nanoribbon.py
zerothi: Hmm, I think my snippet below actually takes care of all cases? I'll give you a more detailed example once done.

sisl/geom/nanoribbon.py
tfrederiksen: Fixed

sisl/geom/nanoribbon.py
tfrederiksen: OK, I will take a closer look

zerothi: No need to, unless you want, I have a working solution for armchair, now generalizing for zig-zag :)

sisl/geom/nanoribbon.py
tfrederiksen: Well... only partially. Do you want to enforce the user to specify `bond` and `atom` parameters in addition to `width`?

zerothi: Yeah, I probably want that. This means we have a general nanoribbon, then we can add

- `graphene_nanoribbon` (and the shorter `agnr`/`zgnr`)

In this way it also highlights that the nanoribbon may be a hBN for instance.

Also for `width` I think this would be fair to force users as input.

tfrederiksen: `nanoribbon` fixed

tfrederiksen: Added also the more specific geometries `graphene_nanoribbon`, `agnr` and `zgnr`.

sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,144 @@
...
+            ribbon = ribbon.tile(n + 1, 0)
+            ribbon = ribbon.remove(-1).remove(-1)
+        else:
+            ribbon = ribbon.tile(n, 0)
+        # Invert y-coordinates
+        ribbon.xyz[:, 1] *= -1
+        # Set lattice vectors strictly orthogonal
+        ribbon.cell[:] = np.diag([ribbon.cell[1, 0], -ribbon.cell[0, 1], ribbon.cell[2, 2]])
+        # Sort along x-axis
+        ribbon = ribbon.sort()
```
tfrederiksen, line 63: After merge with the new `sort` method (#197), I think this line should be set explicitly to `sort(axis=0)'

zerothi: Agreed, but why not `axis=(0, 1)` ?

zerothi: Ps, it has been merged, (this is the time for you to play with `git rebase` ;))

First do `git push`
Then you do
```bash
git checkout master
git pull
git checkout nanoribbon # if this is your local branch name
git rebase master
```
Then you have the latest `sort` stuff there ;)

tfrederiksen: Thanks, I managed to fix it :)

sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,147 @@
...
+
+    See Also
+    --------
+    honeycomb : honeycomb lattices
+    graphene : graphene geometry
+    graphene_nanoribbon : graphene nanoribbon
+    agnr : armchair graphene nanoribbon
+    zgnr : zigzag graphene nanoribbon
+    """
+    if not isinstance(bond, float):
```
zerothi, line 31: Users *could* use integers, generally I wouldn't recommend doing any checks on this, it will be caught later if needed. If you really want to check, a better check would be something like this:
```python
from numbers import Real

if isinstance(bond, Real):
    ...
```
the `Real` matches both integers and floats (but not complex values) :)

sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,147 @@
...
+        raise ValueError("nanoribbon: bond needs to be a float!")
+    if not isinstance(atom, (Atom, list, tuple)):
+        raise ValueError("nanoribbon: atom needs to be an instance of Atom (or list of Atoms)!")
+    if not isinstance(width, int):
```
zerothi, line 35: I will merge, but for your information, this is the correct way to do it:
```python
from numbers import Integral

... isinstance(width, Integral)
```
The problem is that `isinstance(np.int32(1), int) is False`, so it will fail if users pass an element from a numpy array.

sisl/geom/nanoribbon.py
```diff
@@ -0,0 +1,147 @@
...
+        # Sort along x-axis
+        ribbon = ribbon.sort(axis=(0, 1))
+    else:
+        raise ValueError("nanoribbon: kind must be armchair or zigzag")
+
+    # Separate ribbons along y-axis
+    ribbon.cell[1, 1] += 10.
+
+    # Movie inside unit cell
+    ribbon = ribbon.move([-np.min(ribbon.xyz[:, 0]), -np.min(ribbon.xyz[:, 1]), 0])
```
zerothi, line 72: I will shift everything so that there are 10 Ang on both sides (better suited for DFT calculations) and place the geometry in the middle.

tfrederiksen:
Maybe we could think of another function `composite_nanoribbon` to create more advanced structures (joining basic blocks), as you proposed?

pfebrer96:
Hey, sorry for commenting now. I'm using now the `nanoribbon` function and it seemed wierd to me that `bond` and `atom` were positional arguments. Is it a wild guess to use 1.42 and "C" as defaults?

I had to do
```
sisl.geom.nanoribbon(1.42, "C", 9)
```
when I expected that this would work:
```
sisl.geom.nanoribbon(9)
```
Actually, I wasn't sure if I needed to pass the kind of ribbon I wanted, but I assumed that the bond and atom would have defaults.

Is it just me? :)

pfebrer96:
Nevermind, I just noticed that I should have used `sisl.graphene_nanorribon`, sorry haha

PR #205, opened 2020-04-04
Sisl visualization and graphical interface

pfebrer96:
What is this
-----
This pull request is to get visibility, so that people knows that this is under developement and I can (hopefully) get some feedback about it. Many syntax things still need to be changed, but the concept is already quite complete, so I'd like to get feedback on it to improve weak points.

For those who don't know, this project is an attempt to build a plotting framework and a graphical interface that can really display the great potential of sisl for processing scientific data.

How can I test it
----
It should be straightforward for sisl users to test it, I put everything needed in my branch (although I know that in production I will have to store it elsewhere) so that it is as simple as:

```
git clone -b GUI https://github.com/pfebrer96/sisl.git
cd sisl
pip install -e .[visualization]
```

for anyone that would like to test it.

Where do I start
------
Under `sisl/viz/Notebooks/tutorials` there are 3 tutorial notebooks that I hope can be quite self-explanatory on introducing people to the framework. Actually I would like feedback on them too :).

I would appreciate it if (and I recommend) you to go trough the `Demo` and the `GUI with Python Demo` in that order. I think that would be enough to get you an idea of the concept.

The notebooks are also here: https://github.com/pfebrer96/sisl.viz-tutorials. But for testing convenience I will keep them inside the package too for now.

**Thanks in advance to any person that takes the time to check it out!**

Cheers!


setup.py
```diff
@@ -52,6 +52,18 @@
     'scipy>=0.18',
     'netCDF4',
     'pyparsing>=1.5.7',
+    # Everyting below is just for convenience while developing the GUI/viz module
```
zerothi, line 55: I would immediately put this in a `extras_require = ['gui', [list...]]` (see how it is formatted currently).

In this way you don't need to change anything.

sisl/io/sile.py
```diff
@@ -306,10 +308,12 @@ def get_siles(attrs=None):
         return list(__siles)
 
     siles = []
-    for sile in __siles:
+    objs = __sile_rules if rules else __siles
```
zerothi, line 311: I don't like these one liners, ;)
You seem to use them a lot but I think it is better to do it explicitly.

Also, could you create a new method, `get_sile_rules`, it seems more appropriate?

sisl/viz/GUI/README.md
```diff
@@ -0,0 +1,26 @@
+This is the graphical interface.
```
zerothi, line 1: This readme file should be moved to the documentation `docs/visualization`, add an entry there.

sisl/viz/GUI/__main__.py
```diff
@@ -0,0 +1,3 @@
+from . import launch
+
+launch(inconsole=True)
```
zerothi, line 3: newline

sisl/viz/GUI/api.py
```diff
@@ -0,0 +1,336 @@
...
+
+from sisl.viz import BlankSession
+from sisl.viz.plotutils import load
+
+app = Flask("SISL GUI API")
+
+class CustomJSONEncoder(JSONEncoder):
+
+	def default(self, obj):
+
```
zerothi, line 20: indentation, sisl uses 4 spaces, you have 8?

sisl/viz/GUI/build/asset-manifest.json
```diff
@@ -0,0 +1,25 @@
+{
```
zerothi, line 1: Why do you have a `build` directory? This should be part of the installation process, no? Not the actual sources.

sisl/viz/README.md
```diff
@@ -0,0 +1,31 @@
+
```
zerothi, line 1: Again, this should be moved to the docs/visualization documentation.

sisl/viz/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+Classes that are used for visualizing simulations results
+
+Range routines
+==============
+
+Miscellaneous routines
+======================
+
+"""
+
```
zerothi, line 18: What happens if the users don't have the modules?
We need to fallback to have the installation without the gui stuff. :)

sisl/viz/inputFields/array.py
```diff
@@ -0,0 +1,17 @@
...
+        "width": "s100% l50%"
+    }
+
+class Array1dInput(ArrayInput):
+
+    _type = 'vector'
+
+class Array2dInput(ArrayInput):
+
+    _type = "matrix"
```
zerothi, line 17: newline

sisl/viz/inputFields/color.py
```diff
@@ -0,0 +1,9 @@
+from ..inputField import InputField
+
+class ColorPicker(InputField):
+    
+    _type = 'color'
+    
+    _default = {
+        "width": "s50% m30% l15%"
+    }
```
zerothi, line 9: newline (perhaps)

sisl/io/sile.py
pfebrer96: Hahahah yes sorry. It is because I'm writing pyhton and javascript at the same time and in javascript is quite common.

sisl/viz/GUI/build/asset-manifest.json
pfebrer96: This is one of the things that I did not know what to do. These are not binaries. It's just minified javascript and css for the browser to read it. So, they are equal for all users and they only need to be changed when there is an update interface.

They are generated from my code in order to reduce the size of the files to the minimum. To generate the build, you need [nodejs](https://nodejs.org/es/) and all the dependencies of my project. With all this, I believe it does not make sense to make the users generate the "build".

Where do you think I should keep it?

pfebrer96: I mean it works perfectly if it is separated from the sisl package. In fact this is what I was doing before bundling it all here so that people could test easily. It just adds a bit more complexity for the user to understand what to do to install it, but it can be done.

pfebrer96:
Thanks for the quick reply Nick and for checking it out! I will fix everything that you proposed.

One specific question that I have though: I don't know if you've noticed, but the plotability of each Sile is defined at the SileClass definition right now (see PDOS, bands or Grid siles). Would you rather move it to the plots instead? So that plots know what siles they can plot instead of siles knowing which plots they use. Somehow it felt like it was an unnecessary intrusion to the package, knowing that all the rest is contained in the viz folder, but at the same time it felt right and seemed easier to set it explicitly at Sile definition.

Let me know what you think when you can, thanks!

zerothi:
> Thanks for the quick reply Nick and for checking it out! I will fix everything that you proposed.
> 
> One specific question that I have though: I don't know if you've noticed, but the plotability of each Sile is defined at the SileClass definition right now (see PDOS, bands or Grid siles). Would you rather move it to the plots instead? So that plots know what siles they can plot instead of siles knowing which plots they use. Somehow it felt like it was an unnecessary intrusion to the package, knowing that all the rest is contained in the viz folder, but at the same time it felt right and seemed easier to set it explicitly at Sile definition.
> 
> Let me know what you think when you can, thanks!

Yeah, I did notice, but I guess I have to get a better feel for what you are doing and how it is discovered via some other means.
Probably the more clean way would be to _patch_ the classes when importing the `viz` module, since then we know for sure the `viz` is existing. But for now this is ok. Patching would also make it easier to modularize, don't you think?

sisl/viz/inputFields/color.py
zerothi: @pfebrer96 also, no camel-case in modules. If the sub-module is meant to be hidden, then prefix with `_input_fields`

pfebrer96:
Yes, so having a file where all the patches are done? What's the best way to do it? Just set the attribute directly? Like

```
import bandsSileSiesta

bandsSileSiesta._plot = (...)
```
or there is a better way?

zerothi:
I don't know what is more cleaner, I just want the flexibility and easy customizability. :)

setup.py
pfebrer96: Done

pfebrer96: Well actually, I named it visualization. If you prefer gui or any other name I can change it

sisl/viz/GUI/README.md
pfebrer96: I'm trying to add things to docs, but I have never used this rst file format. Is there a way I can quickly check how the docs look?

zerothi: Check the other rst documents, I'll work them through (when I get the time) :)

sisl/viz/GUI/build/asset-manifest.json
zerothi: Hmm... I need to think about this, I don't like shipping this in the sources...

pfebrer96: Maybe when the `GUI.launch()` method is used it could try to check if there are updates for the graphical interface and download all the code to store it in the same place that it is right now. 

I don't know how this can be done exactly but I bet it's possible. This would also keep the UI always updated even when the sisl package is not. But maybe at some point there are incompatibilities that make this approach not appropiate.

sisl/viz/__init__.py
pfebrer96: I don't know how to do this or where this sould be done :)

sisl/viz/GUI/build/asset-manifest.json
pfebrer96: I have something that might be good. The user would be able to access the GUI in two different ways:
- Use the interface **online**. The interface is **hosted somewhere**. Right now it's here just as an example: https://sisl-graphical-interface.web.app
- Use the interface **offline**. The interface is bundled with [electron](https://www.electronjs.org/) into a **native app for each OS**. Then people can download it from somewhere. For example: [App Image for Linux](https://www.dropbox.com/s/1l3gawwky9nx94v/ui-0.1.0.AppImage?dl=0)

In both cases the interface can interact either with the API that is running in localhost or with an API that is running anywhere else in the net if it's accessible. (The api can be run just by doing `python path/to/sisl/viz/GUI/api.py`)

In this way, we won't have the GUI files inside sisl but still it can be very easy for people to use it. What do you think? 

You wouldn't even need to have sisl installed to use the interface itself. Imagine the case that you are an experimental and some collaborator wants to share something with you live.  You would just need to go to https://sisl-graphical-interface.web.app, connect to their running API and then you could not only see the plots as they are building them, but also modify them using "their" sisl without any software requirement (not sisl, not even python).

Of course this would need a bit of security, but using API keys to verify the identity would not be that difficult.

zerothi:
Did you do a rebase? If not could you please redo your merges?

zerothi:
I. E. Don't merge, use rebase, in this way it becomes much easier to track what you did vs what I did.

pfebrer96:
Yes, I've been merging for a while now because for some reason it was impossible for me to rebase. Each time there were hundreds of conflicts that I didn't know where they came from. I think it has to do with the fact that I started developing on the master branch but then moved everything to a different branch. Probably I moved it wrong.

I think that since almost all my changes are in the viz folder I will just fork your repo again, start a branch and put the viz folder there. I will lose the history, but I don't think it matters.

zerothi:
Agreed, good idea.

pfebrer96:
Hi Nick!

At today's GM, I showed everyone how easily can you share your analysis with other people that doesn't even have python installed. You just have to run the GUI api in your computer and tell other people to go to https://sisl-siesta.xyz/ (got this domain for 1$/year haha) and connect to your address. And it worked perfectly! They were able to add/remove plots and edit the existing ones.

Alberto then suggested if people without sisl could just go to the website and drop their files to see the plots, which I thought was a very cool idea. I just implemented the drag and drop functionality and now if I'm sharing a session anyone can drop a file and it will automatically be converted to a plot. But of course, this is not exactly what he was asking for. For that, there should be a server with sisl installed available to process files 24/7. 

Then he suggested that the materials cloud people offer "free servers" to build applications that run in their website: https://www.materialscloud.org/work/tools/options. Which I think is pretty cool. But there's a problem, they limit your application to use one of these two options: a framework they developed or Dash. You probably remember that I started with Dash but then ran out from it because of the limitations it imposed, so I really don't want to go back to it for the whole graphical interface :). However, I think it's not a bad idea at all to have a very simple version of the GUI as a Dash app posted in the materials cloud website so that anyone (without any requirement and at any time) can profit from it without us having to spend money on a server. What do you think? It would also give sisl a lot of visibility.

Last but not least: regarding the drag and drop functionality, is there a way to initialize a sile with the bytes of the file directly? I am asking this because the workflow right now looks like:

    GUI      |                                              PYTHON
Drop file ----> Recieve bytes and file name ---> Save to disk ----> get_sile(file_path)

and of course it would be much better to be able to do

    GUI      |                                              PYTHON
Drop file ----> Recieve bytes and file name ---> get_sile(file_name, bytes)

(not exactly this, but somehow pass the bytes so that it doesn't need to read from disk)

Cheers!

zerothi:
Cool, great job Pol!

> At today's GM, I showed everyone how easily can you share your analysis with other people that doesn't even have python installed. You just have to run the GUI api in your computer and tell other people to go to https://sisl-siesta.xyz/ (got this domain for 1$/year haha) and connect to your address. And it worked perfectly! They were able to add/remove plots and edit the existing ones.

Cool!

> Alberto then suggested if people without sisl could just go to the website and drop their files to see the plots, which I thought was a very cool idea. I just implemented the drag and drop functionality and now if I'm sharing a session anyone can drop a file and it will automatically be converted to a plot. But of course, this is not exactly what he was asking for. For that, there should be a server with sisl installed available to process files 24/7.

This would indeed be very nice. :)

> Then he suggested that the materials cloud people offer "free servers" to build applications that run in their website: https://www.materialscloud.org/work/tools/options. Which I think is pretty cool. But there's a problem, they limit your application to use one of these two options: a framework they developed or Dash. You probably remember that I started with Dash but then ran out from it because of the limitations it imposed, so I really don't want to go back to it for the whole graphical interface :). However, I think it's not a bad idea at all to have a very simple version of the GUI as a Dash app posted in the materials cloud website so that anyone (without any requirement and at any time) can profit from it without us having to spend money on a server. What do you think? It would also give sisl a lot of visibility.

Agreed, again, I really have no experience in GUI development and have always tried to stay clear of it ;)

> Last but not least: regarding the drag and drop functionality, is there a way to initialize a sile with the bytes of the file directly? I am asking this because the workflow right now looks like:
> 
> ....GUI......|...............................................PYTHON.....................................................
> Drop file ----> Recieve bytes and file name ---> Save to disk ----> get_sile(file_path)
> 
> and of course it would be much better to be able to do
> 
> ....GUI.....|.........................................PYTHON............................................
> Drop file ----> Recieve bytes and file name ---> get_sile(file_name, bytes)
> 
> (not exactly this, but somehow pass the bytes so that it doesn't need to read from disk)

I guess this would require some pre-load function in the siles, this is not currently implemented. Basically you want users to be able to have everything in RAM for post-processing on the remote server, no?

Hmm. This would be nice and would also be somewhat problematic in some cases. For instance files which are GB of size. Probably users won't push this to the remote, but some might, so you need some way of determining the size of the data and have a hard limit, wouldn't you?

In principle the above bytes could be emulated by pickling the data and storing it in a _sile_ like structure which exposes the same functionality. Hmm...

pfebrer96:
> I guess this would require some pre-load function in the siles, this is not currently implemented. Basically you want users to be able to have everything in RAM for post-processing on the remote server, no?

Yes, because it seems like a waste of time to me that the file is stored in disk and then read from disk, when the information it is already in the python runtime.

> Hmm. This would be nice and would also be somewhat problematic in some cases. For instance files which are GB of size. Probably users won't push this to the remote, but some might, so you need some way of determining the size of the data and have a hard limit, wouldn't you?

When the file is uploaded in the frontend I get some information for free and file size is amongs it. So that is really a very easy problem to solve. Right now I'm just sending to the "server" the bytes and the file name but all I would need to do is send the file size. Then the server should only need to compare that to a user-provided threshold. If not provided, the threshold could be determined based on the total RAM of the computer.

Anyway, if it's so big that the RAM can not handle it, I won't even be able to receive the file in my python application, right? I would need to find a way to directly store it to disk from the front end. And in that case, could sisl handle such a file? It's difficult for me to understand because I don't know much about how memory works :)

> In principle the above bytes could be emulated by pickling the data and storing it in a sile like structure which exposes the same functionality. Hmm...

I don't know exactly what do you mean here, could you give an example?

Isn't this:

```python
with open("file_name.xyz", "rb") as f:
     f.read()
```

the same as this:

```python
import io
with io.BytesIO(b'all the bytes of the file') as f:
     f.read()
```

So you only need to substitute `open("file_name.xyz", "rb")` by `io.BytesIO(b'all the bytes of the file')` in case a bytes object is provided? I may be wrong, because I don't know much about it.

pfebrer96:
Hey Nick! When (if) you have time, could you take a moment to help me with the `splot` command for the command line? I know how to use an argument parser, but I don't know exactly how do you want it to work (By looking at your other implementations I can't seem to fully understand the concept)

The point is that right now you can do:

```
splot <plotable file> [presets]
splot "File.bands" "Dark theme"
```
where presets are previously saved configuration for plots. But I want to be able to do something like:

```
splot ...all the files --presets ...all the presets --setting1 <val for setting 1> --setting2 <val for setting2>  --save <path>, etc...
```

I know this is not very difficult to achieve, but I'm not sure how to do it exactly using the helpers that you already have implemented for argument parsing.

zerothi:
> Hey Nick! When (if) you have time, could you take a moment to help me with the `splot` command for the command line? I know how to use an argument parser, but I don't know exactly how do you want it to work (By looking at your other implementations I can't seem to fully understand the concept)
> 
> The point is that right now you can do:
> 
> ```
> splot <plotable file> [presets]
> splot "File.bands" "Dark theme"
> ```
> 
> where presets are previously saved configuration for plots. But I want to be able to do something like:
> 
> ```
> splot ...all the files --presets ...all the presets --setting1 <val for setting 1> --setting2 <val for setting2>  --save <path>, etc...
> ```
> 
> I know this is not very difficult to achieve, but I'm not sure how to do it exactly using the helpers that you already have implemented for argument parsing.

In my command line utilities I enabled actions to be performed while adding more arguments. In this way one can do multiple things in one go since you constantly know where you are in the parsing tree. I don't know if this is appropriate for your case.  
I.e. is it necessary that the commandline knows which flag is entered at which location? If not, then just do a regular CLI. If you want to know where you are in the CLI and do actions based on this, then look in `Geometry.ArgumentParser` where the sub-classed `argparse.Action` does things on the fly.
1.

pfebrer96:
I thought I didn't need it, but maybe it is interesting to call plot methods from the CLI as you do with `sgeom`. I will leave the CLI as it is right now, because I think it is a great starting point. But maybe in the future I will take a look at it and try to emulate the behavior of `sgeom` for `splot`.

jonaslb:
Hey, just skimming your comments here. Regarding this:
> (not exactly this, but somehow pass the bytes so that it doesn't need to read from disk)

I don't know if you solved it already, but I have previously needed to get a geometry from an fdf string. I tricked the fdfSile by setting the `sile.fh` to a `StringIO` object. Just be aware that the fdfSile likes to close and reopen the file sometimes, so it needs to be either well-formed or you have to override this behaviour as well. Actually here's a hacky snippet (my old solution was even more hacky, perhaps something similar to this could go in sisl, Nick?).

```python
import sisl as si
from io import StringIO

def io_sile(sile_cls):
    class IOSile(sile_cls):
        def __init__(self, filename, mode="r", comment=None, ioobj=None, **kwargs):
            super().__init__(filename, mode=mode, comment=comment, **kwargs)
            self.fh = self.set_io(ioobj)
            
        def set_io(self, io=None):
            if io is None:
                io = StringIO()
            io._close = io.close
            io.close = lambda: None
            self._io = io
            return io
        
        def _open(self):
            if self.fh is not self._io:
                # If %include in fdf
                super()._open()
            elif "w" in self._mode:
                self.fh = self.set_io()
            else:
                self.fh.seek(0)
    IOSile.__name__ = "IO" + sile_cls.__name__
    return IOSile
```

Example
```python
my_io = StringIO(Path("RUN.fdf").read_text())
fdf = io_sile(si.io.fdfSileSiesta)("nonexistent.fdf", ioobj=my_io)
g = fdf.read_geometry()  # will work even if it's in a separate file here (probably, you don't want this in a webservice!)
```

pfebrer96:
Thanks for the help! I've tried to use your code, but with a `BytesIO` instead of a `StringIO`. However, on the `set_io` method, when you are trying to set the `_close` attribute, I get the following error:

```
TypeError.can't set attributes of built-in/extension type '_io.BytesIO'
```

I don't know if I'm doing something wrong or `StringIO` is different from `BytesIO` in the sense that allows setting its attributes.

jonaslb:
Yes, StringIO works, and I don't know of eg fdfSile could work if the file object was bytes instead. But couldn't you just encode/decode (can't remember which it is) your bytes if it's a string anyway? Of course this won't help if it's binary content, but then maybe you won't need to override the close method in that case. If you find you do need to override it, then subclassing and shadowing should work I guess.

zerothi:
Good idea! Probably something similar could be done for the byte objects. Or, as Jonas does it, subclass it and override the method to return a memory stored version. Hmm...

pfebrer96:
> but then maybe you won't need to override the close method in that case.

Do you mean like this?

```python
def io_sile(sile_cls):
    class IOSile(sile_cls):
        def __init__(self, filename, mode="r", comment=None, ioobj=None, **kwargs):
            super().__init__(filename, mode=mode, comment=comment, **kwargs)
            self.fh = self.set_io(ioobj)
            
        def set_io(self, io=None):
            if io is None:
                io = StringIO()
            # io._close = io.close
            # io.close = lambda: None
            self._io = io
            return io
        
        def _open(self):
            if self.fh is not self._io:
                # If %include in fdf
                super()._open()
            elif "w" in self._mode:
                self.fh = self.set_io()
            else:
                self.fh.seek(0)
    IOSile.__name__ = "IO" + sile_cls.__name__
    return IOSile
```

Now I get this error:

```
TypeError.descriptor 'readline' of '_io.BytesIO' object needs an argument
```

I don't know how the implementation of file handlers is done, so I'm not sure what are the needed patches here.

pfebrer96:
Ah, ok. Now I see that I would need to overwrite the readline method of `Sile` but I'm not sure if there are other places where `readline` is used.

I guess I need to pass the encoding method so that it can be decoded into a string. In this case is the bands file, so it makes sense. My question is, what if it's a binary file? The corresponding sile classes never use readline and data is extracted in some other way? Because it would be better if I could just overwrite the `readline` method so that it is decoded if required instead of converting it to a string myself. If I do it myself, I would need to somehow check if it makes sense to decode it into a string.

pfebrer96:
By the way, for the RHO file I get this error:

```
SileError.IOrhoSileSiesta(SrTiO3.RHO, base=/home/pfebrer).read_grid_size could not read grid sizes. (ierr=2)
```

(The bytes are correct, because if I store the file on disk and read it in the normal way it works fine)

zerothi:
The problem is that the binary files are reading through a fortran interface. :)
So you could do something like this (and pre-read everything once)
```python
class InstanceCache:
    """ Wraps an instance to cache *all* results based on `functools.lru_cache`

    Parameters
    ----------
    obj : object
        the object to get cached results
    lru_size : int or dict, optional
        initial size of the lru cache. For integers this
        is the default size of the cache, for a dictionary
        it should return the ``maxsize`` argument for `functools.lru_cache`
    no_cache : searchable (list or dict)
        a list-like (or dictionary) for searching for
        methods that don't require caches (e.g. small methods)
    """
    def __init__(self, obj, lru_size=1, no_cache=None):
        self.__obj = obj

        # Handle user input for lru_size
        if isinstance(lru_size, defaultdict):
            # fine, user did everything good
            self.__lru_size = lru_size
        elif isinstance(lru_size, dict):
            default = lru_size.pop("default", 1)
            self.__lru_size = ChainMap(lru_size, defaultdict(lambda: default))
        else:
            self.__lru_size = defaultdict(lambda: lru_size)

        if no_cache is None:
            self.__no_cache = []
        else:
            self.__no_cache = no_cache

    def __getattr__(self, name):
        attr = getattr(self.__obj, name)
        # Check if the attribute has the cached functionality
        try:
            attr.cache_info()
        except AttributeError:
            # Fix it and set it to this one
            if name in self.__no_cache:
                break

            # We have to make it cacheable
            maxsize = self.__lru_size[name]
            if maxsize == 0:
                # break to not use an lru_cache (maxsize == 0)
                break
            attr = wraps(attr)(lru_cache(maxsize)(attr))

        # offload the attribute to this class (to minimize overhead)
        object.__setattr__(self, name, attr)
        return attr
```
This (untested) should in theory cache every call once they are made. So you have to call a method before it is cached.

Issue #148, opened 2019-11-08
Implement __repr__ for common classes


jonaslb:
When using python interactively, I often type the object name to get the string representation rather than use `print()`. The difference is that the first uses `obj.__repr__` while `print` uses `obj.__str__`. When printing a list, the items in the list are also represented by the `__repr__` rather than `__str__`. 

`__repr__` is currently unimplemented for most sisl objects. So when typing eg. `list(H.atom)` you get the unhelpful output `[<sisl.atom.Atom at 0x7f73efed5b00>, ...]`.

Currently `__str__` on common sisl objects give a whole 'tree' of information, which is pretty nice, but not always what you want. I propose that `__repr__` gives the non-recursive version of the same information. So while eg. `str(atom)` would list all the atomic orbitals of that atom, `repr(atom)` could just tell you how many there are. 

Eg. desired output of `list(H.atom)` could be `[<sisl.Atom: C, Z: 6, mass(au): 12.01000, maxR: 2.57730, no: 13>, ...]`.

([link to relevant python docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__))

zerothi:
That is a good idea!

zerothi:
Since this is mostly helpful for `lists` I would suspect we could do something like:

```python
def __repr__(self):
        """ repr of object """
        import re
        return re.sub(r"\n\s*", " ", str(self))
```
to have everything 1 liners.

What do you think about this?

jonaslb:
It would still recursively call string on subcomponents, eg. hamiltonian calls geometry calls atoms calls orbitals. For printing eg. a list this is definitely too much information for each item. I think it should remain a kind of short summary without going into detail, so repr would just be a short format string in each class really.

I will just throw together a quick PR for this to show what I mean.

zerothi:
I'll close this since it is accepted and partially done, we will add more incrementally, thanks Jonas!

PR #210, opened 2020-04-16
Added __repr__ to some common classes

jonaslb:
Although #148 is still open to some discussion, implementing a few `__repr__` functions seem like a small thing, and specifying what one wants almost amounts to writing the PR itself anyway. So here it is.

I hope the chosen functions are not too uncontroversial, I tried to only put the "most key" summarizing information in, ie. not include the full tree of information when it is stored in child objects.

Otherwise I can add whatever you think is missing.

The `SuperCell.__repr__` is where I add the most 'new' stuff. I thought that having cell lengths and angles were better summarizing information than the full vector coordinates (which can be obtained with `sc.cell` anyway).

### DIFF
--- a/sisl/atom.py
+++ b/sisl/atom.py
@@ -1175,6 +1175,9 @@ class Atom(object):
         orbs = ',\n '.join([str(o) for o in self.orbital])
         return self.__class__.__name__ + '{{{0}, Z: {1:d}, mass(au): {2:.5f}, maxR: {3:.5f},\n {4}\n}}'.format(self.tag, self.Z, self.mass, self.maxR(), orbs)
 
+    def __repr__(self):
+        return f"<sisl.{self.__class__.__name__} {self.tag}, Z={self.Z}, M={self.mass}, maxR={self.maxR()}, no={len(self.orbital)}>"
+
     def __len__(self):
         """ Return number of orbitals in this atom """
         return self.no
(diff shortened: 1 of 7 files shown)


sisl/atom.py
```diff
@@ -1644,6 +1647,9 @@ def __str__(self):
             s += ' {1}: {0},\n'.format(len(idx), str(a).replace('\n', '\n '))
         return s + '}'
 
+    def __repr__(self):
+        return f"<sisl.{self.__class__.__name__} nspecies={len(self._atom)}, na={len(self)}, no={self.no}>"
```
zerothi, line 1651: Instead of doing writing `<sisl.{self.__...` I think we should write: `<{self.__module__}.{self.__class__.__name}` since in that way we can alter the module level and have everything separated and clear :)

I had planned on cleaning up module classes like this for some time. This seems like a good start for its use.

sisl/supercell.py
```diff
@@ -925,6 +925,12 @@ def __str__(self):
         s = ',\n '.join(['ABC'[i] + '=[{:.3f}, {:.3f}, {:.3f}]'.format(*self.cell[i]) for i in (0, 1, 2)])
         return self.__class__.__name__ + ('{{nsc: [{:} {:} {:}],\n ' + s + ',\n}}').format(*self.nsc)
 
+    def __repr__(self):
+        r = lambda x: round(x, 3)  # rounding for display (shorter than fmt-string when numbers are already round)
+        a, b, c = map(r, np.linalg.norm(self.cell, axis=1))
+        alpha, beta, gamma = (r(self.angle(i, j)) for i, j in ((1, 2), (0, 2), (0, 1)))
```
zerothi, line 931: We have `SuperCell.parameters` for this ;)

I'll merge and change.

zerothi: @jonaslb also, you should try and avoid `np.linalg.norm`, I found that it is much slower than doing it manually `(xyz ** 2).sum(0) ** 0.5`, probably since there is tons of redundant code in `linalg.norm` (it has other good uses).

PR #211, opened 2020-04-16
Cleaned setup.py and removed shipped sources

zerothi:
This is now way more lenient for cloning since all *.c code generated by Cython has been removed.

The next release should be checked for correctly shipped sources etc.

Currently, the sources are generated if required, and I hope the build will work as expected.

This will force developers to have Cython installed (but end users does not require it).


zerothi:
@tfrederiksen @jonaslb @pfebrer96 this will soon go in. It fixes some setup stuff I had wanted to clean up for a loong time.

What is important for you is that now the development version requires Cython installed to build the sources. You probably already have it, otherwise it is easily installed.

pfebrer96:
Thanks for the notice Nick! I don't have cython, I'll have to install it then. 

Just one question: If I ask someone to test the visualization module, is there an easy workaround that can be automated or will that person need to have also cython installed?

zerothi:
Sadly, there is no easy work-around, but using `pip3 install [--user] cython` should suffice? Just like any other dependency?

pfebrer96:
Ok, sorry I have no idea about how cython works :). Seems easy then. So it's just:
```
pip install cython
pip install -e <directory where sisl is>
```
?

zerothi:
Yes, that should be it, note that `Cython` should be in uppercase (I just corrected it)

PR #214, opened 2020-04-19
Supervisable objects

pfebrer96:
This is the pull request to test different versions of https://github.com/zerothi/sisl/issues/213.

There are three versions:
- brillouinzonesuperv: *The original one*
- brillouinzone_superv_faster: *A faster way of supervising by using normal functions (and partial) instead of methods*
- brillouinzone_wraps: *Using `functools.wraps`. Surprisingly it takes twice as much as using `SupervisedMethod`*

I only implemented `asaverage()` in each case. Only changed that method and made the parent supervisable on `set_parent`.

To quickly test all versions in a notebook:

```python
import sisl
from sisl.physics.brillouinzone import BandStructure
from sisl.physics.brillouinzonesuperv import BandStructure as BandStructureSuperv
from sisl.physics.brillouinzone_superv_faster import BandStructure as BandStructureSuperv_faster
from sisl.physics.brillouinzone_wraps import BandStructure as BandStructureWraps

H = sisl.get_sile("../Master/Simulations/NPG/relaxedStructure/NPG.fdf").read_hamiltonian()
geom = sisl.Geometry([[0,0,0]])
H = sisl.Hamiltonian(geom)

bz = BandStructure(H, [[0,0,0], [0,0,1]], 2)
bz_spv = BandStructureSuperv(H, [[0,0,0], [0,0,1]], 2)
bz_spv_faster = BandStructureSuperv_faster(H, [[0,0,0], [0,0,1]], 2)
bz_wraps = BandStructureWraps(H, [[0,0,0], [0,0,1]], 2)

# Check that all give the same result
actual_result = bz.asaverage().eigh()
print(( actual_result == bz_spv.asaverage().eigh()).all())
print(( actual_result == bz_spv_faster.asaverage().eigh()).all())
print(( actual_result == bz_wraps.asaverage().eigh()).all())

methods = list(zip([bz_spv, bz_spv_faster, bz_wraps, bz],("Supervised method", "Faster supervision", "Wraps", "Actual")))
def test(test_obj, method_name):
    print("\n")
    print(method_name)
    print("-----------")
    print("\nbz.asaverage()")
    %timeit test_obj.asaverage()
    print("\nbz.eigh")
    %timeit test_obj.eigh
    call = test_obj.eigh
    print("\ncall")
    %timeit call()

# Then run a for loop to test all methods or do it however you want.
# I did it in separate cells of the notebook.

test(*methods[0])
test(*methods[1])
test(*methods[2])
test(*methods[3])
```

Issue #213, opened 2020-04-19
Supervisable objects


pfebrer96:
This might sound stupid at first. But I think it would be a very nice thing to have as a tool to be used.

Why?
----
I didn't really understand how Nick was able to make things like

```python 
BrillouinZone.asaverage().method(...)
```

 work. Today I needed a similar thing for the visualization module and I took the time to read through all the code and I finally understood it. I really enjoyed understanding how does he get the desired functionality :). 

However, there are two things I'd like to discuss here:

- **It would be cool to have this functionality in general**, so that we (in this case, I) don't need to build it each time we need it.
- It is confusing that when you try to **retrieve an attribute** from the brillouinzone object **you get the object itself** instead of the attribute. For example:
```python
import sisl

geom = sisl.Geometry([[0,0,0]])
H = sisl.Hamiltonian(geom)
bz = sisl.BandStructure(H, [[0,0,0], [0,0,1]], 2)
print(bz.H)
```
prints the band structure, not the Hamiltonian. This really pissed me off because I didn't understand anything. Now I understand why it happens, but still I believe it can be confusing.

**Note:** I know that `bz.H` is not a target of this functionality, so this specific problem could be solved by checking that `bz.H` is not callable in `BrillouinZone.__getattr__()`, but still I would like to get the actual method when I ask for `bz.eigenstate`, for example.

Therefore
-----
I'd like to propose the following:

Instead of returning the object (brillouin zone), when a method is requested, return a "`SupervisedMethod`", which is basically an object that **looks exactly like the method, but runs anything else**.

```python
class SupervisedMethod:
    
    def __init__(self, method, really_executing):
        
        self.method = method
        self.really_executing = really_executing
        
    def __call__(self):

        return self.really_executing()
    
    # Everything below is just to make it look like we "really have the method".
    def __getattribute__(self, attr):
        
        real_getattr = object.__getattribute__
        
        if attr not in ["method", "really_executing", "__call__", "__init__"]:
            return real_getattr(self.method, attr)
        else:
            return real_getattr(self, attr)
    
    def __repr__(self):
        return self.method.__repr__()
```

**But what is this `self.really_executing` thing?**. It could be anything, but for our use case is what the class decides that needs to be done. This is the base `Supervisable` class (automatically supervises all method calls):

```python
class Supervisable:
    
    def __init__(self, obj=None):
        # It will not have any effect when the object is supervisable by class inheritance,
        # But it will allow us to make arbitrary objects supervisable
[...]

Issue #182, opened 2020-03-27
BrillouinZone.asdataarray


zerothi:
For example, let's say I got the bands values from BandStructure.eigh() but I want them as a dataarray. I would have to rewrite all the code, assuming that I know how to build a DataArray properly, or copy/paste it in case I knew where the code that I need is located (unlikely to happen with users).

_Originally posted by @pfebrer96 in https://github.com/zerothi/sisl/pull/179#issuecomment-605144003_

I'll see to it and create the necessary things. Thanks

pfebrer96:
I'm using this and works great :100: . Nice work!

PR #215, opened 2020-04-20
ENH: added first Dispatcher method

zerothi:
An extension of ideas proposed by Pol.

This continues work on #213 and #214.

This is different than the proposed idea for
a couple of reasons.

1) Both Pol and Jonas mentioned that using `asaverage`
   was not intuitive since the methods changed the object
   it self AND returned it-self.

   This was a bad decision from me in the first place.

2) I wanted to have some functionality that wasn't there
   yet:
   a) make it clear in the code _when_ a user does dispatching
   b) allow users to customize how dispatching works, e.g. #182
   c) make returned dispatch methods different than the object it self
      I.e. `bz.dispatch.average != bz.dispatch.sum`

This is all accomplished in this merge request.

It is relatively complicated given what it does.
In `sisl/_dispatch.py` the basic implementation exists.

There are 3 classes since we need a class instance (basically
a singleton) that holds the registered functions and allows for
registrering more.

When `bz.dispatch` is called it returns an `ObjectDispatcher` with
the current allowed methods and the object (bz). At this point it is
pretty much like #214, but by allowing specific keys to retrieve the
dispatch method (or actually class).


zerothi:
@pfebrer96 @jonaslb could you please have a look at this.
The concept is a bit different since I wanted clarity.
However, it may easily be emulated in `asaverage` etc.

Please see this code for a use example (and how to extend it without touching sisl):
```python
import numpy as np
import sisl as si
from functools import wraps
import time

# goto example
gr = si.geom.graphene()
H = si.Hamiltonian(gr)

# Now try dispatch and asaverage
bz = si.MonkhorstPack(H, [21, 21, 1])
# install asaverage
bz.asaverage()

def timeit(func, *args, **kwargs):
    t0 = time.time()
    v = func(*args, **kwargs)
    print('  timeit = ', time.time() - t0)
    return v

print("eigh")
old_eigh = timeit(bz.eigh)
new_eigh = timeit(bz.dispatch.average.eigh)

assert np.allclose(old_eigh, new_eigh)

print()

E = np.linspace(-4, 4, 1000)
dist = si.physics.get_distribution('gaussian', smearing=0.05)
def wrap(es, parent, k, weight):
    # some heavy lifting
    DOS = es.DOS(E, distribution=dist)
    PDOS = es.PDOS(E, distribution=dist)
    vel = es.velocity() * es.occupation().reshape(-1, 1)
    return si.oplist([DOS, PDOS, vel])

print("wrap")
old_wrap = timeit(bz.eigenstate, wrap=wrap)
new_wrap = timeit(bz.dispatch.average.eigenstate, wrap=wrap)

for a, b in zip(old_wrap, new_wrap):
    assert np.allclose(a, b)

# Create your own dispatcher
from sisl.physics.brillouinzone import BrillouinZoneDispatcher
class MyDispatcher(BrillouinZoneDispatcher):
    def dispatch(self, method):
        @wraps(method)
        def func(*args, **kwargs):
            print("calling from my dispatcher")
            return method(*args, **kwargs)

        return func

# Register dispatch method to **ALL** BrillouinZone objects
# This works since dispatch is a class instance
si.BrillouinZone.dispatch.register((wrap, MyDispatcher), my=MyDispatcher)

# now use my dispatcher.
# Note that you don't need to do anything fancy with
# your object to allow the dispatcher method.
# Once it is registered with the class, it works for
# *ALL* instances of BrillouinZone
bz.dispatch.my.eigh()
# You can also do fancy indexing
bz.dispatch["my"].eigh()
# this uses a function hash for indexing in the underlying method
# probably useful for automatic handling based on numpy functions
# or something.
bz.dispatch[wrap].eigh()
```

pfebrer96:
That's amazing!
I have some questions though (Keep in mind that I still didn't understand fully the whole implementation yet :)):
- Is there a way to use a dispatcher on an object whose class does not have a ClassDispatcher? Or are all classes in sisl automatically going to have a class dispatcher?
- Would it be possible to have reusable dispatchers? For example, the `AverageDispatcher` is probably useful in many other cases. The only difference will be what is the attribute we are averaging over. Since you register dispatchers as classes (not instances), the way to do it would be a class builder, right? For example:
```python

def build_average_dispatcher(iter_over, ...):
     class YourFreshDispatcher:
             ....
     return YourFreshDispatcher
```
Or would it be a better way?
- Could you give me an example of what you mean here? I don't understand it :/
> this uses a function hash for indexing in the underlying method
> probably useful for automatic handling based on numpy functions
> or something.

zerothi:
> * Is there a way to use a dispatcher on an object whose class does not have a `ClassDispatcher`? Or are all classes in sisl automatically going to have a class dispatcher?

Currently it isn't globally available. But a single line per class should suffice (and since this also works by inheritance it should be very little changes).

I.e. to do all `Sile` simply do this in `BaseSile` etc.

> * Would it be possible to have reusable dispatchers? For example, the `AverageDispatcher` is probably useful in many other cases. The only difference will be what is the attribute we are averaging over. Since you register dispatchers as classes (not instances), the way to do it would be a class builder, right? For example:
> 
> ```python
> def build_average_dispatcher(iter_over, ...):
>      class YourFreshDispatcher:
>              ....
>      return YourFreshDispatcher
> ```
> 
> Or would there be a better way?

Yeah, I also thought about this, however, this is not so easy. In `BrillouinZone` objects we allow more arguments for the methods. I have been thinking about moving this feature like this:

```python
bz_avg_eta = bz.dispatch.average(eta=True)
bz_avg_eta.eigh()
```
which then moves the keywords one level up. This is somewhat problematic since the `wrap` methods is very dependent on _what_ you call and then `bz.dispatch.average(wrap=wrap)` would probably only for for a single subsequent method. :(  
This means that a generic `build_average_dispatcher` that can also build classes that adds arguments is rather difficult(?), no?

> 
> * Could you give me an example of what you mean here? I don't understand it :/
> 
> > this uses a function hash for indexing in the underlying method
> > probably useful for automatic handling based on numpy functions
> > or something.

The thing is that _any-thing_ may be used as a key for the dispatch function. In the posted example `wrap` is a function. And you can look-up in the dictionary using a function (it is hashable).

zerothi:
Note I will change a few things on how to register functions.

jonaslb:
It looks pretty cool! `__get__` was new to me. Not something that gets used a lot :)

The new 'usage interface' (eg. `bz.dispatch.average.eigh`) is certainly easier than writing the for-loop by oneself, which is of course the competition here. And it's also fairly easy to guess what happens from reading it. Although perhaps `dispatch` isn't exactly the best word to use (I thought it meant to send off to a worker node or maybe some kind of parallelism). Other ideas could be `calculate`, `derive`, `quantity`, `values`, `loop`.

> It is relatively complicated given what it does.

This is my primary worry. I don't think more complexity should be introduced without good usecases for that complexity. Sisl is a 'deep' project to get into already.

zerothi:
> The new 'usage interface' (eg. `bz.dispatch.average.eigh`) is certainly easier than writing the for-loop by oneself, which is of course the competition here. And it's also fairly easy to guess what happens from reading it.

Yeah, this was my main motivation, clarity.

> Although perhaps `dispatch` isn't exactly the best word to use (I thought it meant to send off to a worker node or maybe some kind of parallelism). Other ideas could be `calculate`, `derive`, `quantity`, `values`, `loop`.

This is difficult, consider this:
```python
 class WriteDispatcher(BrillouinZoneParentDispatch):
    def dispatch(self, method):
        out = f"{self._obj.parent.__class__.__name__}_{method.__name__}.dat"
        f = si.io.tableSile(out, "w")
        def func(*args, **kwargs):
            v = method(*args, **kwargs)
            f.write_data(v)

        return func

si.BrillouinZone.dispatch.register("write", WriteDispatcher)
# will overwrite file on every call.
bz.dispatch.write.eigh(k=[0.1, .1, 0])
```
this simply writes the values retrieved from `eigh` to a given file. And it will be re-written on every invocation of `eigh` (regardless of `k`).  This dispatch method is not related to `calculate`, `derive`, `quantity`, `values`, `loop`, and hence I think `dispatch` is a more proper word. Another thing to consider is that the dispatch instance should _not_ have a name that may be used in classes, i.e. `calculate` could be used in some classes and hence would force the implementor to rename the `calculate` for this class only.

> > It is relatively complicated given what it does.
> 
> This is my primary worry. I don't think more complexity should be introduced without good usecases for that complexity. Sisl is a 'deep' project to get into already.

Hmm, I am not too concerned about the `sisl/_dispatcher.py` implementation, it is basically only 30 lines of code that does the actual lifting so it is relatively fast to understand. On the other hand it is important that the dispatch classes does not get too complicated for users to extend. This was my main concern. And also the usage, `bz.dispatch.average.eigh`?

How do you feel about this?

Also, I think this new method is more flexible given that you could do (per your request ;)):
```python
bz_avg = bz.dispatch.average
bz_yield = bz.dispatch.yield
```
and do something different for both.

pfebrer96:
> Note I will change a few things on how to register functions.
Whyyy? :( I liked the kwargs way of doing it. It was simple and straightforward.

By the way, I really think that for it to be understandable to general users, maybe 'Dispatch' classes should end with 'er': Averager, Logger, Timer. But maybe it's that I still didn't get used to it.

> Currently it isn't globally available. But a single line per class should suffice (and since this also works by inheritance it should be very little changes).

I don't know if you have something against it, but as you implemented it you can already use it for arbitrary objects:

```python
import numpy as np
import sisl as si
from sisl.physics.brillouinzone import AverageDispatch

# goto example
gr = si.geom.graphene()
H = si.Hamiltonian(gr)
bz = si.MonkhorstPack(H, [21, 21, 1])

AverageDispatch(bz).eigh()
```

which I find a logic way users would expect it to work. Additionally, what do you think of `AbstractDispatch` having a `bind_to` method that would actually plug the ClassDispatcher for you. Like:

```python
# Suposing bz didn't already have a ClassDispatcher
AverageDispatch.bind_to(bz, "average")
bz.dispatch.average.eigh()
```
Another thing: Is it intentional that the `ObjectDispatcher` does not allow registering new dispatches? I think users will expect this to work:

```python
bz.dispatch.register(...)
```

Maybe it could also register it to the class using directly `ClassDispatcher.register()`. Most users, if they receive an object, they won't know that they can access their class with `obj.__class__` and they will have to import the class manually. Although maybe for this cases the dynamical approach `av_obj = AverageDispatch(obj)` is a better fit, so we don't need to register dispatches.

> which then moves the keywords one level up. This is somewhat problematic since the wrap methods is very dependent on what you call and then bz.dispatch.average(wrap=wrap) would probably only for for a single subsequent method. :(

I don't understand why the `wrap` kwarg is not implemented directly in the Hamiltonian's methods. Or if it's general enough, I guess we can live with the AverageDispatch stealing it from ANY method call, can't we? The same applies for `eta`. Why can't it be a kwarg that the AverageDispatch is expected to steal from the method? This could potentially be true for any Dispatch (but I have not thought about it).

pfebrer96:
Actually, there could be a base Dispatch called `Iterator` or `IterateDispatch` which would help for all the dispatches needed in the BrillouinZone, for example. And I'm sure it can be useful for other cases.

zerothi:
> > Note I will change a few things on how to register functions.
> > Whyyy? :( I liked the kwargs way of doing it. It was simple and straightforward.
> 
> By the way, I really think that for it to be understandable to general users, maybe 'Dispatch' classes should end with 'er': Averager, Logger, Timer. But maybe it's that I still didn't get used to it.

Yeah, this is always my worry. For instance you could think that `EigenstateElectron` also gets a dispatcher for `Averager` which averages over occupations or what-ever, then you suddenly have an `Averager` class for two different classes.  
I don't know if they should be named according to their use-scope, e.g. `AveragerBrillouinZone` or something differently, but with the current naming scheme it is ambiguous... :(

> 
> > Currently it isn't globally available. But a single line per class should suffice (and since this also works by inheritance it should be very little changes).
> 
> I don't know if you have something against it, but as you implemented it you can already use it for arbitrary objects:
> 
> ```python
> import numpy as np
> import sisl as si
> from sisl.physics.brillouinzone import AverageDispatch
> 
> # goto example
> gr = si.geom.graphene()
> H = si.Hamiltonian(gr)
> H.construct([(0.1, 1.44), (0, -2.7)])
> bz = si.MonkhorstPack(H, [21, 21, 1])
> 
> AverageDispatch(bz).eigh()
> ```
> 
> which I find a logic way users would expect it to work.

This is fine. It is exactly how users can dispatch easily.

> Additionally, what do you think of `AbstractDispatch` having a `bind_to` method that would actually plug the ClassDispatcher for you?Like:
> 
> ```python
> # Suposing bz didn't already have a ClassDispatcher
> AverageDispatch.bind_to(bz, "average")
> bz.dispatch.average.eigh()
> ```

Hmm, I guess this could be done doing something like:
```python
def bind_to(self, obj, key):
    if not hasattr(obj.__class__, "dispatch"):
        seattr(obj.__class__, "dispatch", ClassDispatcher())
    obj.__class__.dispatch.register(key, self)
```
I don't know if there are any side-effects...
    
> Another thing: Is it intentional that the `ObjectDispatcher` does not allow registering new dispatches? I think users will expect this to work:
> 
> ```python
> bz.dispatch.register(...)
> ```

This is actually by design.
The problem is that `__get__` only works for class instances (`__get__` won't kick in on instance instances). You can see what I do in `__get__` in the `ClassDispatcher`.

This means the following weird situation will happen if you want to register a dispatch class for *all* instances of `BrillouinZone` then you _must_ do it for the class instance.

If we were to allow this:
```python
bz.dispatch.register(...)
```
that register would be local to the instance of the `bz.dispatch` (in the above example you would actually loose that registration since `bz.dispatch` returns a new instance on every invocation).  
For users to get the above to work you would have to force users to do this:
```python
dispatch = bz.dispatch
dispatch.register(...)
dispatch[...].(...)
```
which is not intuitive. I.e. I think users would suspect bz`.dispatch.register(...)` to store the registered class (but it does not).

So I decided not to allow `register` on `ObjectDispatcher`. In principle `ObjectDispatcher` could be two classes, one allowing `register` (to be used by users) and one which does not (to be used in `ClassDispatcher`).  
What do you think?

> Maybe it could also register it to the class using directly `ClassDispatcher.register()`. Most users, if they receive an object, they won't know that the class can be accessed with `obj.__class__` and they will have to import the class manually. Although maybe for this cases the dynamical approach `av_obj = AverageDispatch(obj)` is a better fit, so we don't need to register dispatches.

Hmm, I didn't quite get this... :)

> > which then moves the keywords one level up. This is somewhat problematic since the wrap methods is very dependent on what you call and then bz.dispatch.average(wrap=wrap) would probably only for for a single subsequent method. :(
> 
> I don't understand why the `wrap` kwarg is not implemented directly in the Hamiltonian's methods. 

I didn't do this because then _all_ methods should have this optional keyword which I think is annoying. And also, if you decorate with too nested levels for something that gets called repeatedly it can become quite heavy. Secondly, `eigh(..., wrap=wrap)` would be equivalent to `wrap(eigh(...))` so I don't really see the point in adding this complexity per method.

> Or if it's general enough, I guess we can live with the AverageDispatch stealing it from ANY method call, can't we? The same applies for `eta`. Why can't it be a kwarg that the `AverageDispatch` is expected to steal from the method? This could potentially be true for any Dispatch (but I have not thought about it).

But this is what it is doing currently, they do _steal_ the argument. :)
Another problem with this is that the keywords you _add_ shouldn't be used in the decorated function (for `eta` this is a problem since some physical objects have an `eta` argument related to a complex energy!!!).

zerothi:
> Actually, there could be a base Dispatch called `Iterator` or `IterateDispatch` which would help for all the dispatches needed in the BrillouinZone, for example. And I'm sure it can be useful for other cases.

You can see the latest commit on how to use *nested* things. But how would you do this generally? You would need which keyword argument the iterator is used, and possibly some `lambda` function in case it needs to be wrapped, I foresee some problems ;)

pfebrer96:
My point with the `ObjectDispatcher.register` was to register in the `ClassDispatcher`. But if that's not possible I agree that:

> dispatch = bz.dispatch
> dispatch.register(...)
> dispatch[...].(...)

is not useful.

> But this is what it is doing currently, they do steal the argument. :)
> Another problem with this is that the keywords you add shouldn't be used in the decorated function (for eta this is a problem since some physical objects have an eta argument related to a complex energy!!!).

Then I don't understand why a general `average_dispatch_builder` that generates `<Whatever>AverageDispatch` classes that by default steal `eta` and `wrap` can be used. Could you clarify this. Like, why not:

```python
def build_average_dispatch(iter_over, ...):
     class YourFreshDispatch:
             ....
     return YourFreshDispatch

KAverager = build_average_dispatch(iter_over={"k":"k"}) 
# Meaning that the k keyword argument will be given the value of self.k
# It could also be implemented like
KAverager = build_average_dispatch(iter_over={"k": lambda obj: obj.k})
# Or you could do this of course with class inheritance instead of a class builder

# Then you could build other averagers
AtomAverager = build.average_dispatcher(iter_over={"atoms": lambda obj: obj}) # For geometries

```

zerothi:
> My point with the `ObjectDispatcher.register` was to register in the `ClassDispatcher`.

Ah, ok. Well, this could be done using something like this:
```python
class ObjectDispatcher
def register(self, key, cls):
    if hasattr(self._obj.__class__, "dispatch"):
        self._obj.__class__.dispatch.register(key, cls)
    self._dispatchs[key] = cls
```

> But if that's not possible I agree that:
> 
> > dispatch = bz.dispatch
> > dispatch.register(...)
> > dispatch[...].(...)
> 
> is not useful.
> 
> > But this is what it is doing currently, they do steal the argument. :)
> > Another problem with this is that the keywords you add shouldn't be used in the decorated function (for eta this is a problem since some physical objects have an eta argument related to a complex energy!!!).
> 
> Then I don't understand why a general `average_dispatch_builder` that generates `<Whatever>AverageDispatch` classes that by default steal `eta` and `wrap` can be used. Could you clarify this. Like, why not:
> 
> ```python
> def build_average_dispatch(iter_over, ...):
>      class YourFreshDispatch:
>              ....
>      return YourFreshDispatch
> 
> KAverager = build_average_dispatch(iter_over={"k":"k"}) 
> # Meaning that the k keyword argument will be given the value of self.k
> # It could also be implemented like
> KAverager = build_average_dispatch(iter_over={"k": lambda obj: obj.k})
> # Or you could do this of course with class inheritance instead of a class builder
> 
> # Then you could build other averagers
> AtomAverager = build.average_dispatcher(iter_over={"atoms": lambda obj: obj}) # For geometries
> ```

In the above example you still need to define the _weights_ used for the iterations.  
Hmm, maybe it ain't so difficult. What if you want to customize the `dispatch` method afterwards?

pfebrer96:
> Hmm, maybe it ain't so difficult.

Don't act like passing the weights is difficult after you've build a dispatch framework using metaclasses... hahahahaha

> What if you want to customize the dispatch method afterwards?

Could you give an example of this and how would you do it with the actual implementation?

zerothi:
> > Hmm, maybe it ain't so difficult.
> 
> Don't act like passing the weights is difficult after you've build a dispatch framework using metaclasses... hahahahaha

:+1: 

> > What if you want to customize the dispatch method afterwards?
> 
> Could you give an example of this and how would you do it with the actual implementation?
Well, for instance you would want an averager that returns the data _as is_, i.e. the current iterator. Another averager would be to return an `xarray.DataArray` (this contrived example is not too difficult since you could do the DataArray conversion in the end. But say you want to do something complicated (without `wrap`) to the output of each iteration?

zerothi:
> Ah, ok. Well, this could be done using something like this:
> 
> ```python
> class ObjectDispatcher
> def register(self, key, cls):
>     if hasattr(self._obj.__class__, "dispatch"):
>         self._obj.__class__.dispatch.register(key, cls)
>     self._dispatchs[key] = cls
> ```

now added (and works.

It however has the side-effect:
```python
dispatch = bz.dispatch
bz2.dispatch.register("hello", cls)
```
Now doing `dispatch.hello` would raise a `KeyError` (since it was created prior to the registration).
This could easily be dealt with by having the `ClassDispatcher` contain a `weakref` to all instantiated `ObjectDispatcher` instances and patch those as well. I think this is above the common usage to have any bad impact.

pfebrer96:
> Another averager would be to return an xarray.DataArray (this contrived example is not too difficult since you could do the DataArray conversion in the end.

This brings me to the idea that some Dispatchs might be compatible if you build them in the right way. But that is of course a more complicated thing.

> But say you want to do something complicated (without wrap) to the output of each iteration?

Well in that case

```python
build_average_dispatch(iter_over={"k": lambda obj: obj.k}, wrap=yourWrapThatDoesComplicatedThings)
```
and then the averager would have the user provided `wrap` (inner wrapping level) and the builder provided wrap (outer wrapping level). It is of course complex, but your intent is already complex isn't it? Still, even being complex, it is unambiguous and this would only be inside the average_dispatch_builder. 

The builder's code would look clean by defining the function to be run just once:

```python
actually_run = method
if has_user_wrap:
     actually_run = user_wrap(method)
if has_builder_wrap:
    actually_run = builder_wrap(wrap)
```
with the corresponding argument handling that I'm not sure how it goes now
I mean, if you need something more complex

zerothi:
> > But say you want to do something complicated (without wrap) to the output of each iteration?
> 
> Well in that case
> 
> ```python
> build_average_dispatch(iter_over={"k": lambda obj: obj.k}, wrap=yourWrapThatDoesComplicatedThings)
> ```
> 
> and then the averager would have the user provided `wrap` (inner wrapping level) and the builder provided wrap (outer wrapping level). It is of course complex, but your intent is already complex isn't it? Still, even being complex, it is unambiguous and this would only be inside the average_dispatch_builder.

I probably need to see this in context, could you alter two of the `Dispatch` methods in `sisl/physics/brillouinzone.py` to use the Iterator? ;)
Say the `average` and the `yield`?

pfebrer96:
Ok, I will do it tonight if you don't mind. I have to prepare a presentation now :). 

But maybe but that time you will have already done it haha.

zerothi:
> Ok, I will do it tonight if you don't mind. I have to prepare a presentation now :).

Ok.

> But maybe but that time you will have already done it haha.

No, I also really need to do something else :)

zerothi:
@jonaslb @pfebrer96 
Thanks for your comments, Jonas, could you please give your updated opinion? :)

If this is to go in we need to decide whether the *older* methods should go away. Either they need to be retained _as is_ since one cannot do a drop-in replacement of the `dispatch` method.

I am partly pro on removing the old behaviour. This is more clear and users can do more with it.
What do you think?

jonaslb:
Looks good 👍 I think I won't be implementing my own dispatchers so I left it up to you guys. Like you said, I got what I wanted since they are now separate ;-)

Regarding the old ones, maybe just silently "redirect" them but put/make a deprecation warning. The warning message could contain the approximate time that it'll stop working and how to use the new stuff. Eg. in half a year or some such they could be deleted. Presumably the functions are widely used so I think immediately removing them is not good.

zerothi:
Thanks Jonas, yeah, good suggestion with deprecation notice. I'll do that. :)

pfebrer96:
> I probably need to see this in context, could you alter two of the Dispatch methods in sisl/physics/brillouinzone.py to use the Iterator? ;)
> Say the average and the yield?

I got an approach that seems clean and functional to me. Let's see what you think:

```python
# What is expected from the user is not actually a wrapper, but a
# function that takes the result of the method and returns a modified
# result (at least this is what the brillouin zone does)
# To make things more manageable, we are going to build a wrapper
# that returns the user_provided wrap. At the same time, it also allows 
# controlling which kwargs are seen by the user through user_kwargs 
def user_wrapper(func, wrap):
    if wrap is None:
        # We still need to catch the user_kwargs if there are any
        # So that they are not passed to the method
        def user_wrapped(*args, user_kwargs={}, **kwargs):
            return func(*args, **kwargs)
    else:
        def user_wrapped(*args, user_kwargs={}, **kwargs):
            return wrap(func(*args, **kwargs), **user_kwargs)
    return user_wrapped

class Iterator(AbstractDispatch):
    
    # The iterable can be passed in different ways, so that the user
    # can choose the one that's more convenient for them
    def _get_iterable(self):
        '''
        Normalizes _iter_over to a thing that we can actually iterate over.
        '''
        it_over = getattr(self, "_iter_over")

        if isinstance(it_over, str):
            it_over = getattr(self._obj, it_over)
        elif callable(it_over):
            it_over = it_over(self._obj)
        elif isinstance(it_over, tuple):
            it_over = zip(*[getattr(self._obj, item) for item in it_over])
        # If it's a dict, something has to be made to seemlessly allow kwargs

        return it_over

    def _set_iterable(self, iter_over):
        setattr(self, "_iter_over", iter_over)

    iterable = property(fget=_get_iterable, fset=_set_iterable)
    
    # This defines the function that we are going to actually iterate over
    # It takes into account if wrapping must be done using the "class-specific" wrapper
    # and/or the user provided wrap
    # The code could be inside dispatch, but this allows more modularity and will be handy for
    # the definition of a yieldIterator, for example (see below).
    def _get_actual_function(self, wrap):

        # Note that user_wrapper actually takes care of the case when wrap is None
        actually_run = user_wrapper(method, wrap)
        # Use the build-provided wrapper
        iter_wrapper = getattr(self, "_iter_wrapper", None)
        if iter_wrapper is not None:
            actually_run = iter_wrapper(actually_run)

        return actually_run


    def dispatch(self, method):

        # Let's catch some keywords that will be used by the dispatch
        @wraps(method)
        def func(*args, wrap=None, **kwargs):

            actually_run = self._get_actual_function(wrap)
            for i, item in enumerate(self.iterable):
                actually_run(item, *args, **kwargs)
                #eta.update() 
                # I don't know how do you actually prefer to handle eta so
                # I leave this to you

        return func

# As I promised, here's the yield version with very little additional code
class YieldIterator(Iterator):
    
    def dispatch(self, method):

        @wraps(method)
        def func(*args, wrap=None, **kwargs):

            actually_run = self._get_actual_function(wrap)
            for i, item in enumerate(self.iterable):
                yield actually_run(item, *args, **kwargs)

        return func

# At this point you have probably noticed that it's very easy to build a class that extends
# Iterator however you desire. But still, this dispatch builder is very convenient (I think)
def build_iter_dispatch(iter_over, wrapper=None, yield_iters=False):
    
    # I know you don't like it, you can change it haha
    BaseIterator = YieldIterator if yield_iters else Iterator
    
    class CustomIterator(BaseIterator): 
        
        _iter_over = iter_over  
        _iter_wrapper = staticmethod(wrapper)
    
    return CustomIterator
```

Now to the fun part. How to use it:

A first test to see how things work:
```python
# I defined this dumb class to understand how it should work:
class DumbObject:
    
    def __init__(self):
        
        self.F = [1,2,3,4,5,10]
        self.weights = [0.8, 0.9, 0.4, 0.1, 0.2, 0.3]
        
    def useful_stuff(self, F=1):
        return 2*F

# Now, we are going to define a wrapper to see how the execution actually goes
def build_wrapper(func):
    
    def new_func(*args, **kwargs):
        print("\nACTING BEFORE EXECUTION")
        # We are going to define some kwargs to pass them to the user
        ret = func(*args, **kwargs, user_kwargs={"F": 4, "C": 5})
        print("PROCESSING RESULT\n")
        return ret
    
    return new_func

# Build the test dispatch
TestIterator = build_iter_dispatch("F", wrapper=build_wrapper)

# Now the user can modify each return with wrap, as it could before
def multiply_by3(method_result, **kwargs):
    print(f"I am the user and I am multiplying the method result ({method_result}) by 3.")
    print("Also received some kwargs: {kwargs}")
    return method_result*3

# And then just use the dispatch
MyIterator(obj).useful_stuff(wrap=multiply_by3)

# Note that you can change the iterable of a given dispatch very easily
dispatch = MyIterator(obj)
print(dispatch.iterable) # [1,2,3,4,5,10]
dispatch.iterable = "weights"
print(dispatch.iterable) # [0.8, 0.9, 0.4, 0.1, 0.2, 0.3]
```

And finally a demo of average and yield:
```python
# This is just an example of how the average values can be obtained, to demonstrate
# how one can use the build tool ,but we would need a way to store the values.
# However it would be very easy to build an Averager by building the class ourselves
# inheriting from Iterator
def average_wrapper(func):
    
    def new_func(item_and_weight, *args, **kwargs):
        item, weight = item_and_weight
        result = func(item, *args, **kwargs, user_kwargs={"w": weight})
        print(result*weight)
        return result*weight
    
    return new_func

# We are going to get a tuple of (F, weight) in each iteration. Average wrapper will handle it.
AverageDispatch = build_iter_dispatch(("F", "weights"), wrapper=average_wrapper)
# And just to check that it works
AverageDispatch(obj).useful_stuff()

# Finally the yield dispatch
YieldDispatch = build_iter_dispatch("F", yield_iters=True)
YieldDispatch(obj).useful_stuff() # is a generator
```

So basically, the `Iterator` has all the pieces you need. You can get a custom one for simple cases using `build_iter_dispatch` or you can just extend it to do whatever you want with very little code (as I did with the YieldIterator).

Let me know what you think!

zerothi:
@pfebrer96 @jonaslb 
I considered the name again (`dispatch`) and I agree it ain't. Pretty, so far I couldn't come up with a better one. But now I remembered that Pandas does exactly the same thing for multiple data.

E.g. what do you feel about:
```python
bz.aggregate.average
# with the synonym:
bz.agg.average
```
?
This will be familiar to Pandas users and it will be clearer that this will aggregate over all values in the BrillouinZone?

@pfebrer96 I'll comment on your stuff! :)

zerothi:
> > I probably need to see this in context, could you alter two of the Dispatch methods in sisl/physics/brillouinzone.py to use the Iterator? ;)
> > Say the average and the yield?
> 
> I got an approach that seems clean and functional to me. Let's see what you think:
> 
> ```python
> # What is expected from the user is not actually a wrapper, but a
> # function that takes the result of the method and returns a modified
> # result (at least this is what the brillouin zone does)
> # To make things more manageable, we are going to build a wrapper
> # that returns the user_provided wrap. At the same time, it also allows 
> # controlling which kwargs are seen by the user through user_kwargs 
> def user_wrapper(func, wrap):
>     if wrap is None:
>         # We still need to catch the user_kwargs if there are any
>         # So that they are not passed to the method
>         def user_wrapped(*args, user_kwargs={}, **kwargs):
>             return func(*args, **kwargs)
>     else:
>         def user_wrapped(*args, user_kwargs={}, **kwargs):
>             return wrap(func(*args, **kwargs), **user_kwargs)
>     return user_wrapped
> ```

I think the `user_kwargs` is too strict. I.e. it seems this is equivalent to a `partial`, i.e. in my BZ objects I pass loop variables and some parent information. :(

> ```python
> class Iterator(AbstractDispatch):
>     
>     # The iterable can be passed in different ways, so that the user
>     # can choose the one that's more convenient for them
>     def _get_iterable(self):
>         '''
>         Normalizes _iter_over to a thing that we can actually iterate over.
>         '''
>         it_over = getattr(self, "_iter_over")
> 
>         if isinstance(it_over, str):
>             it_over = getattr(self._obj, it_over)
>         elif callable(it_over):
>             it_over = it_over(self._obj)
>         elif isinstance(it_over, tuple):
>             it_over = zip(*[getattr(self._obj, item) for item in it_over])
>         # If it's a dict, something has to be made to seemlessly allow kwargs
> 
>         return it_over
> 
>     def _set_iterable(self, iter_over):
>         setattr(self, "_iter_over", iter_over)
> 
>     iterable = property(fget=_get_iterable, fset=_set_iterable)
>     
>     # This defines the function that we are going to actually iterate over
>     # It takes into account if wrapping must be done using the "class-specific" wrapper
>     # and/or the user provided wrap
>     # The code could be inside dispatch, but this allows more modularity and will be handy for
>     # the definition of a yieldIterator, for example (see below).
>     def _get_actual_function(self, wrap):
> 
>         # Note that user_wrapper actually takes care of the case when wrap is None
>         actually_run = user_wrapper(method, wrap)
>         # Use the build-provided wrapper
>         iter_wrapper = getattr(self, "_iter_wrapper", None)
>         if iter_wrapper is not None:
>             actually_run = iter_wrapper(actually_run)
> 
>         return actually_run
> 
> 
>     def dispatch(self, method):
> 
>         # Let's catch some keywords that will be used by the dispatch
>         @wraps(method)
>         def func(*args, wrap=None, **kwargs):
> 
>             actually_run = self._get_actual_function(wrap)
>             for i, item in enumerate(self.iterable):
>                 actually_run(item, *args, **kwargs)
> ```

in BZ I also use `item` as a non-first argument, so the `Iterator` class _must_ know which keyword arg to pass to...

> ```python
>                 #eta.update() 
>                 # I don't know how do you actually prefer to handle eta so
>                 # I leave this to you
> 
>         return func
> 
> # As I promised, here's the yield version with very little additional code
> class YieldIterator(Iterator):
>     
>     def dispatch(self, method):
> 
>         @wraps(method)
>         def func(*args, wrap=None, **kwargs):
> 
>             actually_run = self._get_actual_function(wrap)
>             for i, item in enumerate(self.iterable):
>                 yield actually_run(item, *args, **kwargs)
> 
>         return func
> 
> # At this point you have probably noticed that it's very easy to build a class that extends
> # Iterator however you desire. But still, this dispatch builder is very convenient (I think)
> def build_iter_dispatch(iter_over, wrapper=None, yield_iters=False):
>     
>     # I know you don't like it, you can change it haha
>     BaseIterator = YieldIterator if yield_iters else Iterator
>     
>     class CustomIterator(BaseIterator): 
>         
>         _iter_over = iter_over  
>         _iter_wrapper = staticmethod(wrapper)
>     
>     return CustomIterator
> ```
> 
> Now to the fun part. How to use it:
> 
> A first test to see how things work:
> 
> ```python
> # I defined this dumb class to understand how it should work:
> class DumbObject:
>     
>     def __init__(self):
>         
>         self.F = [1,2,3,4,5,10]
>         self.weights = [0.8, 0.9, 0.4, 0.1, 0.2, 0.3]
>         
>     def useful_stuff(self, F=1):
>         return 2*F
> 
> # Now, we are going to define a wrapper to see how the execution actually goes
> def build_wrapper(func):
>     
>     def new_func(*args, **kwargs):
>         print("\nACTING BEFORE EXECUTION")
>         # We are going to define some kwargs to pass them to the user
>         ret = func(*args, **kwargs, user_kwargs={"F": 4, "C": 5})
>         print("PROCESSING RESULT\n")
>         return ret
>     
>     return new_func
> 
> # Build the test dispatch
> TestIterator = build_iter_dispatch("F", wrapper=build_wrapper)
> 
> # Now the user can modify each return with wrap, as it could before
> def multiply_by3(method_result, **kwargs):
>     print(f"I am the user and I am multiplying the method result ({method_result}) by 3.")
>     print("Also received some kwargs: {kwargs}")
>     return method_result*3
> 
> # And then just use the dispatch
> obj = DumbObject()
> TestIterator(obj).useful_stuff(wrap=multiply_by3)
> 
> # Note that you can change the iterable of a given dispatch very easily
> dispatch = TestIterator(obj)
> print(dispatch.iterable) # [1,2,3,4,5,10]
> dispatch.iterable = "weights"
> print(dispatch.iterable) # [0.8, 0.9, 0.4, 0.1, 0.2, 0.3]
> ```
> 
> And finally a demo of average and yield:
> 
> ```python
> # This is just an example of how the average values can be obtained, to demonstrate
> # how one can use the build tool ,but we would need a way to store the values.
> # However it would be very easy to build an Averager by building the class ourselves
> # inheriting from Iterator
> def average_wrapper(func):
>     
>     def new_func(item_and_weight, *args, **kwargs):
>         item, weight = item_and_weight
>         result = func(item, *args, **kwargs, user_kwargs={"w": weight})
>         print(result*weight)
>         return result*weight
>     
>     return new_func
> 
> # We are going to get a tuple of (F, weight) in each iteration. Average wrapper will handle it.
> AverageDispatch = build_iter_dispatch(("F", "weights"), wrapper=average_wrapper)
> # And just to check that it works
> AverageDispatch(obj).useful_stuff()
> 
> # Finally the yield dispatch
> YieldDispatch = build_iter_dispatch("F", yield_iters=True)
> YieldDispatch(obj).useful_stuff() # is a generator
> ```
> 
> So basically, the `Iterator` has all the pieces you need. You can get a custom one for simple cases using `build_iter_dispatch` or you can just extend it to do whatever you want with very little code (as I did with the `YieldIterator`).
> 
> Let me know what you think!

It seems like a lot of boiler plate code to not have stuff repeated (i.e. my `AverageDispatch` is 20-30 lines), hehe. :)
I am still worried about corner cases or how easy it is to extend for additional functionality.

I mean _any_ iterator with different body could be implemented like this:

```python
class Iterator(BrillouinZoneParentDispatch):
    def dispatch(self, method):
        @wraps(method)
        def func(*args, extra_kw1=None, extra_kw2=False, **kwargs):
            obj = ...
            for i, (k, w) in enumerate(zip(obj.k, obj.weight)):
                ... body ...
```
Done.

Hmm. I am not convinced ;)
I.e. if you want to add the above functionality of passing stuff to the `wrap` function based on the iterator, then you have to add lots of code (which doesn't take much in each of them).

pfebrer96:
Well, yeah actually the `dispatch` method is the less useful part of the `Iterator` for library developement. In this regard, my main idea was that the `get_actual_function()` part would make the definition of new Dispatchs cleaner,  less prone to bugs and easier to mantain. Because now this part in your `Dispatch` implementations:

```python
if wrap:
    v = _asoplist(wrap(method(*args, k=k[0], **kwargs), parent=parent, k=k[0], weight=w[0])) * w[0]
    eta.update()
    for i in range(1, len(k)):
        v += _asoplist(wrap(method(*args, k=k[i], **kwargs), parent=parent, k=k[i], weight=w[i])) * w[i]
        eta.update()
else:
    v = _asoplist(method(*args, k=k[0], **kwargs)) * w[0]
    eta.update()
    for i in range(1, len(k)):
         v += _asoplist(method(*args, k=k[i], **kwargs)) * w[i]
         eta.update()
```

would become:
```python
func = self._get_actual_function(method, wrap)
v = _asoplist(func(*args, k=k[0],  **kwargs, user_kwargs={"parent": parent, "k": k[0], "weight": w[0]})) * w[0]
eta.update()
for i in range(1, len(k)):
     v += _asoplist(func(*args, k=k[i],  **kwargs, user_kwargs={"parent": parent, "k": k[i], "weight": w[i]})) * w[i]
     eta.update()
```

for all the `Dispatch` implementations. And I think the `_get_actual_function` method would not even be a specific thing of Iterators, it could go even upper in the inheritance tree. My main point is not that it will save lines of code (which probably will). It is that it will make the code more maintainable, easier to read and therefore also more approachable for future developers.

By the way, I don't understand which limitations are you pointing here, could you clarify please? :)

> I think the user_kwargs is too strict. I.e. it seems this is equivalent to a partial, i.e. in my BZ objects I pass loop variables and some parent information. :(

Having said that, the main point of the implementation is to **give power to the users if they need it**. My opinion is that most users won't take the time to understand what is going on underneath the class, so they won't know what it is available for them to build their classes. Some users might not even be capable of understanding the workflow in a reasonable amount of time. Therefore having things like:

```python
MyIterator = build_iter_dispatch("whatever")
obj = MyIterator(obj)
```

can come in handy for small classes they develop, to automatically iterate all methods over the "whatever" attribute. I don't think they will ever think about using this (at least at the beginning) if they don't have an easy way of doing it. And by easy I mean not having to know that `MyIterator` needs to be a class that inherits from `sisl._dispatcher.AbstractDispatch` with a `dispatch` method that takes a method and returns a function.

Also if they are using an average dispatch that they build they might want to have the same functionality but across a different attribute, for whatever reason.

```python
dispatch = bz.dispatch.average
dispatch.iterable = "cell" # This doesn't make sense I know
# or
dispatch.iterable = lambda bz: bz.k[:20]
# or
dispatch.iterable = lambda bz: bz.k[ bz.k[:,0 ] != 0 ]
```

> in BZ I also use item as a non-first argument, so the Iterator class must know which keyword arg to pass to...

This is why I stated that I wanted to parse dictionaries as possible `iter_over` values in `_get_iterable`. So that you can go like:

```python
dispatch.iterable = {"k":"k", "whatever": lambda obj.k: [ k[0]*23 for k in obj.k] }
```
And then you would be specifying that `obj.k` goes to "k" and "whatever" receives the value that is returned from that lambda.

Well, in summary:

Things that I believe could be helpful for sisl development.
- `_get_actual_function`

User-oriented features:
- The rest

Because yes I see your point that it's maybe not useful at all for you, but it might be useful for people with less knowledge of sisl/python.

pfebrer96:
And by the way I don't know if I have made myself clear, but this:

```python
class Iterator(BrillouinZoneParentDispatch):
    def dispatch(self, method):
        @wraps(method)
        def func(*args, extra_kw1=None, extra_kw2=False, **kwargs):
            obj = ...
            for i, (k, w) in enumerate(zip(obj.k, obj.weight)):
                ... body ...
```
is how I actually expect Iterators that are inside sisl to be defined, unless they are dead simple and then you could use the builder. Actually it would be exactly:

```python
class BzAverager(Iterator):
    _iter_over = ("k", "weight")
    def dispatch(self, method):
        @wraps(method)
        def func(*args, extra_kw1=None, extra_kw2=False, **kwargs):
            obj = ...
            for i, (k, w) in enumerate(self.iterable):
                ... body ...
```

zerothi:
> ```python
> if wrap:
>     v = _asoplist(wrap(method(*args, k=k[0], **kwargs), parent=parent, k=k[0], weight=w[0])) * w[0]
>     eta.update()
>     for i in range(1, len(k)):
>         v += _asoplist(wrap(method(*args, k=k[i], **kwargs), parent=parent, k=k[i], weight=w[i])) * w[i]
>         eta.update()
> else:
>     v = _asoplist(method(*args, k=k[0], **kwargs)) * w[0]
>     eta.update()
>     for i in range(1, len(k)):
>          v += _asoplist(method(*args, k=k[i], **kwargs)) * w[i]
>          eta.update()
> ```
> 
> would become:
> 
> ```python
> func = self._get_actual_function(method, wrap)
> v = _asoplist(func(*args, k=k[0],  **kwargs, user_kwargs={"parent": parent, "k": k[0], "weight": w[0]})) * w[0]
> eta.update()
> for i in range(1, len(k)):
>      v += _asoplist(func(*args, k=k[i],  **kwargs, user_kwargs={"parent": parent, "k": k[i], "weight": w[i]})) * w[i]
>      eta.update()
> ```

I have just updated the currently implemented `Dispatch` to do this. I initially did not do this in Py 2 since I had some serious performance issues when using `wrap` vs. not using `wrap`.

> 
> for all the `Dispatch` implementations. And I think the `_get_actual_function` method would not even be a specific thing of Iterators, it could go even upper in the inheritance tree. My main point is not that it will save lines of code (which probably will). It is that it will make the code more maintainable, easier to read and therefore also more approachable for future developers.
> 
> By the way, I don't understand which limitations are you pointing here, could you clarify please? :)

If we have to change the `Iterator` class after creating it from your factory to pass the correct keywords down through `user_kwargs` then it seems like you won't win a lot? Maybe I don't understand, but it seems to me you _have_ to subclass to be able to do:

> ```python
> v = _asoplist(func(*args, k=k[0],  **kwargs, user_kwargs={"parent": parent, "k": k[0], "weight": w[0]})) * w[0]
> ```


> Having said that, the main point of the implementation is to **give power to the users if they need it**. My opinion is that most users won't take the time to understand what is going on underneath the class, so they won't know what it is available for them to build their classes. Some users might not even be capable of understanding the workflow in a reasonable amount of time. Therefore having things like:
> 
> ```python
> MyIterator = build_iter_dispatch("whatever")
> obj = MyIterator(obj)
> ```
> 
> can come in handy for small classes they develop, to automatically iterate all methods over the "whatever" attribute (with the `wrap` kwarg available). I don't think they will ever think about using this (at least at the beginning) if they don't have an easy way of doing it. And by easy I mean not having to know that `MyIterator` needs to be a class that inherits from `sisl._dispatcher.AbstractDispatch` with a `dispatch` method that takes a method and returns a function.
> 
> Also if they are using an average dispatch that you build they might want to have the same functionality but across a different attribute, for whatever reason.
> 
> ```python
> dispatch = bz.dispatch.average
> dispatch.iterable = "cell" # This doesn't make sense I know
> # or
> dispatch.iterable = lambda bz: bz.k[:20]
> # or
> dispatch.iterable = lambda bz: bz.k[ bz.k[:,0 ] != 0 ]
> ```
> 
> > in BZ I also use item as a non-first argument, so the Iterator class must know which keyword arg to pass to...
> 
> This is why I stated that I wanted to parse dictionaries as possible `iter_over` values in `_get_iterable`. So that you can go like:
> 
> ```python
> dispatch.iterable = {"k":"k", "whatever": lambda obj.k: [ k[0]*23 for k in obj.k] }
> ```
> 
> And then you would be specifying that `obj.k` goes to "k" and "whatever" receives the value that is returned from that lambda.
> 
> Well, in summary:
> 
> Things that I believe could be helpful for sisl development:
> 
> * `_get_actual_function`

The `_get_actual_function` is nothing more than my `_parse_kwargs` without some stuff.

> 
> User-oriented features:
> 
> * The rest
> 
> Because yes I see your point that it's maybe not useful at all for you, but it might be useful for people with less knowledge of sisl/python.

You are probably right. But in my experience people will not do this, already what you try to accomplish is very difficult to know what is _actually_ happening behind the scenes. If users were to adopt this the documentation should be really clear :)

zerothi:
> And by the way I don't know if I have made myself clear, but this:
> 
> > ```python
> > class Iterator(BrillouinZoneParentDispatch):
> >     def dispatch(self, method):
> >         @wraps(method)
> >         def func(*args, extra_kw1=None, extra_kw2=False, **kwargs):
> >             obj = ...
> >             for i, (k, w) in enumerate(zip(obj.k, obj.weight)):
> >                 ... body ...
> > ```
> 
> is how I actually expect Iterators that are inside sisl to be defined, unless they are dead simple and then you could use the builder without passing a wrapper. Actually it would be exactly:
> 
> ```python
> class BzAverager(Iterator, BrillouinZoneParentDispatch):
>     _iter_over = ("k", "weight")
>     def dispatch(self, method):
>         @wraps(method)
>         def func(*args, extra_kw1=None, extra_kw2=False, **kwargs):
>             obj = ...
>             for i, (k, w) in enumerate(self.iterable):
>                 ... body ...
> ```
> ...
> Which again would give all the power to the user to modify at will your dispatch by changing the iterable, since there is nothing specific about it (with no extra lines of code!)

I see, but then again, you are also not _removing_ any lines of code (actually you add 1 line of code :wink:). Sometimes being explicit is better. You are _hiding_ `enumerate(zip(obj.k, obj.weight))` in `enumerate(self.iterable)` through a line at the class definition level `_iter_over = ("k", "weight")`. So to figure out what `k` and `w` are in the loop you have to first figure out what `self.iterable` is doing, _then_ check the `_iter_over`.  
I don't think it adds clarity at all to the build in methods. I only see an advantage for users wanting to build their own simple iterator (without doing fancy stuff) ;)
But perhaps this is enough.

@pfebrer96 what about the name change, https://github.com/zerothi/sisl/pull/215#issuecomment-617998555

jonaslb:
Regarding the name since Nick was reconsidering: I like aggregate or agg! It is definitely pretty close to what the bz does. It might be a bit of a stretch to call getting the full iterator/list/array etc an aggregation though (not a huge problem imo). Otherwise `wrap_parent`, `parametrize` or `collect` would also be reasonable names, I guess?

By the way, all the aggregates (dispatches) are named `<return_type>Dispatch` except the one that returns an iterator (called 'yield' instead). Not a huge issue, but I thought it stuck out a bit.

I also noticed when looking at @pfebrer96 's approach, the way `Iterator` is used irks me a bit... Because it's a general term with [particular meaning](https://docs.python.org/3/glossary.html#term-iterator) (having a `__next__` method), which the proposed class doesn't live up to. Other than that its nice to see someone giving thought to sisl improvements :)

zerothi:
> Regarding the name since Nick was reconsidering: I like aggregate or agg! It is definitely pretty close to what the bz does. It might be a bit of a stretch to call getting the full iterator/list/array etc an aggregation though (not a huge problem imo). Otherwise `wrap_parent`, `parametrize` or `collect` would also be reasonable names, I guess?

Thanks! Yeah, I guess `parametrize` would also work. `param` is a good name as well.
Checking `pandas` doc probably `apply` would also be a good candidate?
- `apply`
- `param`/`parametrize`
- `agg`/`aggregate`
?
 
> By the way, all the aggregates (dispatches) are named `<return_type>Dispatch` except the one that returns an iterator (called 'yield' instead). Not a huge issue, but I thought it stuck out a bit.

Agreed. If changed to `aggregate` I would rename the `*Dispatch` to `*Aggregate` for clarity. It depends on what we choose, however, I don't see the problem with the yield? Do you mean `YieldDispatch` should be named `IteratorDispatch?
Good idea. The yield was a bad name. Probably `iter` would be more appropriate since `iter(...)` is a build-in.

zerothi:
@jonaslb do you prefer `Iterator*` or `Iter*`? Probably the first since that is the nomenclature, and the class does not per see reflect data-type... hmm...

jonaslb:
YYeah its a hard choice between the names, I do actually think `parametrize` is the better one for description, but it is of course a bit long -- and I don't actually think `param` is a good short version because it is also the short of "parameter". So maybe `apply` is better simply because it is shorter. 

I agree with the use of `iter`/`IterDispatch`. Regarding this one, I got an idea: Could the ClassDispatch/ObjectDispatch (possibly renamed) have a "default" dispatcher? Eg it could be the iter version. Then usage for this case would be `for result in bz.apply.eigh():` rather than `for result in bz.apply.iter.eigh():`. 

Actually, and I know this would be a larger change to the current proposed implementation, if we go with `apply`, maybe it'd be cool if we could do `bz.apply.DOS.average()` rather than `bz.apply.average.DOS()`?

pfebrer96:
> Edit: Custom wrappers could be called via bz.apply(myfunc).average()

I like this one :) but what would you expect as a user if you do: `bz.apply(myfunc)`. Because if there is a default dispatch then it should do something, but is it possible to make it do something and at the same time have the funcionality that you were proposing?

Also then you could not store the dispatch so that it runs in all methods without you doing anyhting. I mean:
```python
bz_avg  = bz.apply.average

bz_avg.eigh()
bz_avg.DOS()
bz_avg.PDOS()
```

> for result in bz.apply.eigh():

I also like this. I thought about it and then didn't like it, but seeing this use case I think it can be useful. It is like passing `wrap` but maybe more convenient for some users. 

Regarding the names, `apply` seems like a general one that would fit other cases that do not involve iterating/aggregating. So, if you'd like to make it a consistent word through different classes I think that one is good. Otherwise, `aggregate` sounds also good for this use case.

> If we have to change the Iterator class after creating it from your factory to pass the correct keywords down through user_kwargs then it seems like you won't win a lot? Maybe I don't understand, but it seems to me you have to subclass to be able to do:

Well that's why the builder accepts a wrapper.

```python
def wrapper(method)
   def wrapped(k, *args, **kwargs):
        return method(*args, k=k, user_kwargs={...}, **kwargs)
   return wrapped 

build_iter_dispatch("k", wrapper=wrapper)
```

Or do you mean that you can't have the parent? That would be easy to solve if you pass the object to the wrapper:

```python
def wrapper(obj, method)
   def wrapped(k, *args, **kwargs):
        return method(*args, k=k, user_kwargs={"parent": obj.parent, ...}, **kwargs)
   return wrapped 

build_iter_dispatch("k", wrapper=wrapper)
```

But yeah if you need a wrapper that does complicated things then it is better to subclass if you know how to do it definitely.

> The _get_actual_function is nothing more than my _parse_kwargs without some stuff.

Ok, but then you'll have to write the same when you do dispatching in another class, don't you? I feel like the general framework could benefit more from the things implemented in `BrillouinZone`, but now it is all unnecessarily specific to the brillouin zone. It may be that after making it general, there will not be other use cases. But still at the same time you are opening it up to the users, and therefore many different possible cases may appear. I'm not even saying that the user will necessarily find out about it. Maybe they will ask you how to do something and Dispatch will come to your mind. I believe they would have a better image of sisl if you tell them: It can be done like this:

```python
MyIterator = build_iter_dispatch("whatever")
obj = MyIterator(obj)
```

that if you tell them: It can be done like this:

```python
class MyIterator(AbstractDispatch):
    def dispatch(self, method):
        @wraps(method)
        def func(*args, wrap=None, **kwargs):
            obj = ...
            for i, x in enumerate(obj.whatever):
                ... body ...

obj = MyIterator(obj)
```

Otherwise, you could rename the `AverageDispatch` to `BzKAverageDispatch`, since it is not going to be reusable for any other case.

I won't comment more on this because I already understood that you prefer not to have it, sorry for my persistance :)

pfebrer96:
By the way, just out of curiosity: What's the difference between `oplist` and a numpy array? Is it faster?

zerothi:
> YYeah its a hard choice between the names, I do actually think `parametrize` is the better one for description, but it is of course a bit long -- and I don't actually think `param` is a good short version because it is also the short of "parameter". So maybe `apply` is better simply because it is shorter.

Ok. I just don't like `parametrize` I need to stop to think how it is spelled ;)
I'll change to `apply` :+1: 

> I agree with the use of `iter`/`IterDispatch`. Regarding this one, I got an idea: Could the ClassDispatch/ObjectDispatch (possibly renamed) have a "default" dispatcher? Eg it could be the iter version. Then usage for this case would be `for result in bz.apply.eigh():` rather than `for result in bz.apply.iter.eigh():`.

Good idea. This is now added (the default is `apply.iter`). When registering, one can specify whether it is `default`.

> Actually, and I know this would be a larger change to the current proposed implementation, if we go with `apply`, maybe it'd be cool if we could do `bz.apply.DOS.average()` rather than `bz.apply.average.DOS()`?
> Edit: Custom wrappers could be called via `bz.apply(myfunc).average()`

One _can_ do custom wrappers. Simply do:
```python
bz.apply.iter(method)(...)
```

I see the additional usage of the other things, but lets keep it as is for now. If we can adopt it to do what you suggested, then we may do that. Still with the current implementation one can do:
```python
bz_arr = bz.apply.array
bz_avg = bz.apply.average
bz_arr.DOS(...)
bz_avg.DOS(...)
```
I think the abstraction of doing it _on_ the methods is just not worth it?

zerothi:
> By the way, just out of curiosity: What's the difference between `oplist` and a numpy array? Is it faster?

`oplist` allows one to have a list of different numpy arrays and do mathematical operations:
```python
z, o = np.zeros, np.ones
ol = oplist([z(4), z(6), z(10, 10)])
ol += 2 # valid statement
o_ol = oplist([o(4), o(6), o(10, 10)])
op = ol + o_ol # valid statement
```
this is not so easy in numpy because you can't form a single array due to unmatching dimensions. You would need `xarray.DataSet` or something similar, I guess.

zerothi:
> I won't comment more on this because I already understood that you prefer not to have it, sorry for my persistance :)

No, thanks for your persistence :+1:. It is good with this discussion.

Once this is in, I would be happy to approve a PR adding this functionality *IFF* it is very properly documented ;)

zerothi:
@pfebrer96 in principle the `oplist` can do any mathematical thing just like `map(func, zip(list1, list2))` would do it, this is just a convenience class to make it a bit simpler. In my `oplist` you'll also have the in-place operations which cuts memory requirements by 1/3.

zerothi:
@pfebrer96 @jonaslb if you give your go on the current state, I'll merge and start play with doc! :)

sisl/physics/brillouinzone.py
```diff
@@ -110,6 +120,223 @@
...
+
+        return func
+
+
+@set_module("sisl.physics")
+class DataArrayApply(ArrayApply):
+    __str__ = _apply_str("return xarray.DataArray")
+
+    def dispatch(self, method):
+        """ Dispatch the method by summing """
```
tfrederiksen, line 312: Update this docstring?

zerothi: thanks!

zerothi: @tfrederiksen what is your opinion on this ;) :wink:

tfrederiksen:
Very nice work! I enjoyed this thread and learned new things. For me `apply` is a good choice :)

jonaslb:
> I think the abstraction of doing it on the methods is just not worth it?
It's up to you, I'm not currently writing the code ;) The advantage as I see it is not about functionality, I'm sure you can do what you want either way -- instead it was just about fitting the usage pattern to my mental model of this (which is subjective of course). 

The order of operations at least in my head is 1. one applies a method to all k-points. The most direct/primitive way is a generator expression `(H.method(k=k) for k in bz.k)`. 
Then 2. would be to either (a) collect the results into a list or (b) an aggregation like sum or average (in fact the builtin `sum` could even be used on this example generator expression!). 

So you see the only thing about my idea was that the ordering is a little bit the other way around compared to this, when you write `bz.apply.sum.method()` (ie. you want to apply, but then you specify step 2, and then only last do you specify what to apply in step 1).

zerothi:
> > I think the abstraction of doing it on the methods is just not worth it?
> 
> It's up to you, I'm not currently writing the code ;) The advantage as I see it is not about functionality, I'm sure you can do what you want either way -- instead it was just about fitting the usage pattern to my mental model of this (which is subjective of course).
> 
> The order of operations at least in my head is 1. one applies a method to all k-points. The most direct/primitive way is a generator expression `(H.method(k=k) for k in bz.k)`.
> Then 2. would be to either (a) collect the results into a list or (b) an aggregation like sum or average (in fact the builtin `sum` could even be used on this example generator expression!).
> 
> So you see the only thing about my idea was that the ordering is a little bit the other way around compared to this, when you write `bz.apply.sum.method()` (ie. you want to apply, but then you specify step 2, and then only last do you specify what to apply in step 1).

I see. And I agree. This could first go in when the _older_ methods are deprecated since it requires one to alter `bz.__getattr__` which is currently used for the old back-end.
Once that is done we could do something like this:

```python
# Basically another one of the ObjectDispatcher
class MethodDispatcher:
    __slots__ = ("_obj", "_method", "_dispatchs", "_default")
...
```
Then in `BZ.__getattr__` return a MethodDispatcher. It will be difficult to do in `ObjectDispatcher` since there is no way of knowing whether the user want to do the *default* action `bz.apply.iter.DOS` or `bz.apply.DOS.average`.  
Hmm, note for self at a later point :)

I guess you would still like to retain the behaviour of storing the `bz_avg = bz.apply.average` in one variable, no?

pfebrer96:
Wow, the MethodDispatcher is a very nice way of having both :)

> The order of operations at least in my head is 1. one applies a method to all k-points. The most direct/primitive way is a generator expression (H.method(k=k) for k in bz.k).
> Then 2. would be to either (a) collect the results into a list or (b) an aggregation like sum or average (in fact the builtin sum could even be used on this example generator expression!).

But this order assumes that the arguments are passed to the generator without any previous modification, and it might not be the case, isn't it? In a more general way the dispatch will be a wrapper that can not only act on results given by the generator but also create the generator itself.

pfebrer96:
How does the default work Nick? 

I guess this is valid:
```python
foreig in bz.apply.eigh():
```
Is this valid:
```python
for eig in bz.apply(eigh):
```
?

zerothi:
> How does the default work Nick?

It works by this:
```python
class ObjectDispatcher:
    def __getattr__(self, key):
        """ Retrieve dispatched method by name """
        if key in self._dispatchs:
            return self._dispatchs[key](self._obj)
        return getattr(self._dispatchs[self._default](self._obj), key)
```
So if the method name you wish to call is the same as a registered key, then it won't work.

> I guess this is valid:
> 
> ```python
> for eig in bz.apply.eigh():
> ```

Yes, because `eigh` is not registered. If it was, you'd be having an `ArgumentError` since `method` wasn't passed to `__call__` in `AbstractDispatch`.

> 
> Is this valid:
> 
> ```python
> for eig in bz.apply(eigh):
> ```

No, this isn't valid. But it could be with `MethodDispatcher`, yes, good catch. :)

This is just a little annoying since it isn't clear that `eigh` comes from the `bz.parent`.

zerothi:
I'll try and implement the `MethodDispatcher` later ;)

pfebrer96:
Hahahahah you could not resist.

> Hmm, note for self at a later point :)

A later point means a later point in the day.

jonaslb:
I missed @pfebrer96 's earlier comments, sorry about that.

> what would you expect as a user if you do: bz.apply(myfunc)

I thought of it as a replacement for `bz.apply.iter.eigenstate(wrap=myfunc)`, but I kind of forgot that you had to put "eigenstate" in there to get this functionality (oops). 
On second thought this might be less cool if it just ends up working similar to `bz.call` today.

> (pfebrer): Also then you could not store the dispatch so that it runs in all methods without you doing anyhting.
> (Nick): I guess you would still like to retain the behaviour of storing the bz_avg = bz.apply.average in one variable, no?

I'd say this is less important... My previous comment, back over on the issue, was (intended as), that since this *was* possible, then the different things should at least not be the same object. Besides, if one wants to run many methods over k, then it's probably better to make a wrapper function anyway, to avoid diagonalizing the Hamiltonian many times over.

> But this order assumes that the arguments are passed to the generator without any previous modification, and it might not be the case, isn't it? In a more general way the dispatch will be a wrapper that can not only act on results given by the generator but also create the generator itself.

Yeah, I realize you are seeking a very generalized system. But this was a "most direct/primitive" way. I wouldn't put that generator expression example in a library, it was just intended to show what a user might write themselves for any concrete need :)

zerothi:
@jonaslb now it should be there ;)
It ain't as pretty since it requires some fine-tuning.
```python
bz.apply.eigh.average == bz.apply.average.eigh
bz.apply.eigh() == default dispatch
```
So one can do many things ;)

Note I plan to deprecate `call` since it can be done using `bz.apply.average(method)`

pfebrer96:
One last thing, does it make sense to check if the attribute we are getting is actually a `callable` before wrapping it? 

Just right now, I'm implementing a dispatch to keep plot sessions in sync with the graphical interface, but then you can't get the attributes from the session. 

```python
session = SessionGUIUpdater(session)

a = session.attribute + 2 # I can't do that because the attribute it is converted to a function

```

I can do it in my own implementations, but is there any case in which attributes will be converted to a function? I'm asking because otherwise you have to implement in all the dispatchs the callable check. Otherwise the user will not be able to access attributes. (Unless they know that they can do `session._obj.attribute` of course, but still is not optimal)

zerothi:
> One last thing, does it make sense to check if the attribute we are getting is actually a `callable` before wrapping it?
> 
> Just right now, I'm implementing a dispatch to keep plot sessions in sync with the graphical interface, but then you can't get the attributes from the session.
> 
> ```python
> session = SessionGUIUpdater(session)
> 
> a = session.attribute + 2
> # I can't do that because the attribute is converted to a function
> ```

What is `SessionGUIUpdater` inheriting from, is it an `ObjectDispatcher`?
Because then it should be relatively simple:
```python
attr = self._obj_getattr(self._obj, key)
if callable(attr):
   return MethodDispatcher(...)
return attr
```
?

pfebrer96:
It inherits from `AbstractDispatch`. Isn't this a good way of doing it?

Anyway, yes it is very simple. I was just asking if this might be needed in all cases.

zerothi:
Could you try now?

pfebrer96:
:+1:

Issue #196, opened 2020-04-01
Create an attribute class which can accept events


zerothi:
We need a better way of handling `names` and other attributes.

Having a common class to control attributes and changes to parent structures may be the way forward?

PR #218, opened 2020-05-07
ENH: Category for manipulating categories using boolean operators

zerothi:
This will allow users to generate boolean expressions and select, say, atoms based on simple arithmetic formulas.

This is better described with examples.
```python
import numpy as np
import sisl as si
from sisl.geom.category import *

# Generate some boolean expressions
B = AtomZ(5) # atoms with Z == 5
B2 = AtomNeighbours(2, neigh_cat=B) # atoms with 2 neighbours, both being Z == 5
N = AtomZ(7) # atoms with Z == 7
N2 = AtomNeighbours(2, neigh_cat=N) # atoms with 2 neighbours, both being Z == 7
B3 = AtomNeighbours(3, neigh_cat=B) # ...
N3 = AtomNeighbours(3, neigh_cat=N) # ...

n2 = AtomNeighbours(2) # atoms with 2 neighbours (regardless of neighbour category)
n3 = AtomNeighbours(3) # ...
# n3 == This category _has_ to have 3 neighbours (no specific type)
# AtomNe...() == at least one of the neighbours has to have only 2 neighbours
n3n2 = n3 & AtomNeighbours(1, 3, neigh_cat=n2)
# AtomNe...() == all 3 neighbours must have 3 neighbours with at least one those having only 2 neighbours
# yeah, wrap your head around that one!!!
# It isn't fully correct, I still need to clear a few things
n3n3n2 = AtomNeighbours(3, neigh_cat=n3n2)

# Say if you want one to find only Carbon atoms with 2 neighbours, do this
C_n2 = n2 & AtomZ(6)
# If you want to find atoms with 2 neighbours and they *must* not be Carbon atoms, do this
notC_n2 = n2 ^ AtomZ(6)

# Ok, so now we have defined a few categories
# Lets play...

# Generate HBN
hBN = si.geom.honeycomb(1.42, atom=[si.Atom(5, R=1.43), si.Atom(7, R=1.43)]) * (10, 11, 1)
# remove 10 random atoms
idx = np.random.randint(0, len(hBN), size=10)
hBN = hBN.remove(idx)

# return a list of either NullCategory or `n2` categories
cat = n2.categorize(hBN)
# retrieve all indices where the category is not null
idx_n2 = np.array(cat != NullCategory()).nonzero()[0]
```

This was inspired by #202 to make it more general. However, this may also be used to get indices of atoms for neighbouring sites to holes etc. This would allow easier extraction of DOS for the given atoms.

@tfrederiksen @jonaslb @pfebrer96 I would like your thoughts on this line of progress. And also if you have some ideas! ;)

The relevant files are in `sisl/category/base.py` (will probably be moved to `sisl/category.py`)
and `sisl/geom/category/*.py`.

### DIFF
--- a/sisl/_category.py
+++ b/sisl/_category.py
@@ -3,7 +3,7 @@ from abc import ABCMeta, abstractmethod
 from functools import lru_cache
 import numpy as np
 
-from sisl._internal import set_module, singledispatchmethod
+from ._internal import set_module, singledispatchmethod
 
 __all__ = ["Category", "CompositeCategory", "NullCategory"]
 __all__ += ["AndCategory", "OrCategory", "XOrCategory"]
--- a/sisl/category/__init__.py
+++ b/sisl/category/__init__.py
@@ -1,0 +1,3 @@
+r"""Classification classes """
+
+from .base import *
--- a/sisl/category/__init__.py
+++ b/sisl/category/__init__.py
@@ -1,3 +1,0 @@
-r"""Classification classes """
-
-from .base import *
--- a/sisl/category/base.py
+++ b/sisl/category/base.py
@@ -82,6 +82,55 @@ class Category(metaclass=ABCMeta):
         r""" Override the name of the categorization """
         self._name = name
 
+    @classmethod
+    def kw(cls, **kwargs):
+        """ Create categories based on keywords
+
+        This will search through the inherited classes and
+        return and & category object for all keywords.
+
+        Since this is a class method one should use this
+        on the base category class in the given section
+        of the code.
+        """
+
+        subcls = set()
+        work = [cls]
+        while work:
+            parent = work.pop()
+            for child in parent.__subclasses__():
+                if child not in subcls:
+                    subcls.add(child)
+                    work.append(child)
+
+        del work, parent, child
+
+        # create dictionary look-up
+        subcls = {cl.__name__.lower(): cl for cl in subcls}
+
+        def get_cat(cl, args):
+            if isinstance(args, dict):
+                return cl(**args)
+            return cl(args)
+
+        # Now search keywords and create category
+        cat = None
+        for key, args in kwargs.items():
+            lkey = key.lower()
+            found = ''
+            for name, cl in subcls.items():
+                if name.endswith(lkey):
+                    if found:
+                        raise ValueError(f"{cls.__name__}.kw got a non-unique argument for category name:\n"
+                                         f"    Searching for {name} and found matches {found} and {name}.")
+                    found = name
+                    if cat is None:
+                        cat = get_cat(cl, args)
+                    else:
+                        cat = cat & get_cat(cl, args)
+
+        return cat
+
     @abstractmethod
     def categorize(self, *args, **kwargs):
         r""" Do categorization """
@@ -134,8 +183,8 @@ class Category(metaclass=ABCMeta):
     def __and__(self, other):
         return AndCategory(self, other)
 
-    #def __or__(self, other):
-    #    return OrCategory(self, other)
+    def __or__(self, other):
+        return OrCategory(self, other)
 
     def __xor__(self, other):
         return XOrCategory(self, other)
@@ -285,9 +334,7 @@ class OrCategory(CompositeCategory):
         def cmp(a, b):
             if isinstance(a, NullCategory):
                 return b
-            elif isinstance(b, NullCategory):
-                return a
-            return self
+            return a
 
         if isinstance(catA, list):
             return list(map(cmp, catA, catB))
--- a/sisl/category/base.py
+++ b/sisl/category/base.py
@@ -140,6 +140,9 @@ class Category(metaclass=ABCMeta):
     def __xor__(self, other):
         return XOrCategory(self, other)
 
+    def __invert__(self):
+        return NotCategory(self)
+
 
 @set_module("sisl.category")
 class NullCategory(Category):
@@ -165,6 +168,46 @@ class NullCategory(Category):
     @property
     def name(self):
         return "∅"
+
+
+@set_module("sisl.category")
+class NotCategory(Category):
+    """ A class returning the *opposite* of this class (NullCategory) if it is categorized as such """
+    __slots__ = ("_cat",)
+
+    def __init__(self, cat):
+        super().__init__()
+        if isinstance(cat, CompositeCategory):
+            self.set_name(f"~({cat})")
+        else:
+            self.set_name(f"~{cat}")
+        self._cat = cat
+
+    def categorize(self, *args, **kwargs):
+        r""" Base method for queriyng whether an object is a certain category """
+        cat = self._cat.categorize(*args, **kwargs)
+
+        def check(cat):
+            if isinstance(cat, NullCategory):
+                return self
+            return NullCategory()
+
+        if isinstance(cat, list):
+            return list(map(check, cat))
+        return check(cat)
+
+    @singledispatchmethod
+    def __eq__(self, other):
+        if isinstance(other, NotCategory):
+            return self._cat == other._cat
+        return False
+
+    @__eq__.register(list)
+    @__eq__.register(tuple)
+    @__eq__.register(np.ndarray)
+    def _(self, other):
+        # this will call the list approach
+        return super().__eq__(other)
 
 
 @set_module("sisl.category")
--- a/sisl/category/base.py
+++ b/sisl/category/base.py
@@ -190,6 +190,8 @@ class Category(metaclass=ABCMeta):
         return XOrCategory(self, other)
 
     def __invert__(self):
+        if isinstance(self, NotCategory):
+            return self._cat
         return NotCategory(self)
 
 
--- a/sisl/category/base.py
+++ b/sisl/category/base.py
@@ -1,0 +1,323 @@
+from collections import defaultdict, ChainMap
+from abc import ABCMeta, abstractmethod
+from functools import lru_cache
+import numpy as np
+
+from sisl._internal import set_module, singledispatchmethod
+
+__all__ = ["Category", "CompositeCategory", "NullCategory"]
+__all__ += ["AndCategory", "OrCategory", "XOrCategory"]
+__all__ += ["InstanceCache"]
+
+
+class InstanceCache:
+    """ Wraps an instance to cache *all* results based on `functools.lru_cache`
+
+    Parameters
+    ----------
+    obj : object
+        the object to get cached results
+    lru_size : int or dict, optional
+        initial size of the lru cache. For integers this
+        is the default size of the cache, for a dictionary
+        it should return the ``maxsize`` argument for `functools.lru_cache`
+    no_cache : searchable (list or dict)
+        a list-like (or dictionary) for searching for
+        methods that don't require caches (e.g. small methods)
+    """
+
+    def __init__(self, obj, lru_size=1, no_cache=None):
+        self.__obj = obj
+
+        # Handle user input for lru_size
+        if isinstance(lru_size, defaultdict):
+            # fine, user did everything good
+            self.__lru_size = lru_size
+        elif isinstance(lru_size, dict):
+            default = lru_size.pop("default", 1)
+            self.__lru_size = ChainMap(lru_size, defaultdict(lambda: default))
+        else:
+            self.__lru_size = defaultdict(lambda: lru_size)
+
+        if no_cache is None:
+            self.__no_cache = []
+        else:
+            self.__no_cache = no_cache
+
+    def __getattr__(self, name):
+        attr = getattr(self.__obj, name)
+        # Check if the attribute has the cached functionality
+        try:
+            attr.cache_info()
+        except AttributeError:
+            # Fix it and set it to this one
+            if name in self.__no_cache:
+                # We have to make it cacheable
+                maxsize = self.__lru_size[name]
+                if maxsize != 0:
+                    attr = wraps(attr)(lru_cache(maxsize)(attr))
+
+        # offload the attribute to this class (to minimize overhead)
+        object.__setattr__(self, name, attr)
+        return attr
+
+
+@set_module("sisl.category")
+class Category(metaclass=ABCMeta):
+    r""" A category """
+    __slots__ = ("_name", "_wrapper")
+
+    def __init__(self, name=None):
+        if name is None:
+            self._name = self.__class__.__name__
+        else:
+            self._name = name
+
+    @property
+    def name(self):
+        r""" Name of category """
+        return self._name
+
+    def set_name(self, name):
+        r""" Override the name of the categorization """
+        self._name = name
+
+    @abstractmethod
+    def categorize(self, *args, **kwargs):
+        r""" Do categorization """
+        pass
+
+    def __str__(self):
+        r""" String representation of the class (non-distinguishable between equivalent classifiers) """
+        return self.name
+
+    def __repr__(self):
+        r""" String representation of the class (non-distinguishable between equivalent classifiers) """
+        return self.name
+
+    @singledispatchmethod
+    def __eq__(self, other):
+        """ Comparison of two categories, they are compared by class-type """
+        # This is not totally safe since composites *could* be generated
+        # in different sequences and result in the same boolean expression.
+        # This we do not check and thus are not fool proof...
+        # The exact action also depends on whether we are dealing with
+        # an And/Or/XOr operation....
+        # I.e. something like
+        # (A & B & C) != (A & C & B)
+        # (A ^ B ^ C) != (C ^ A ^ B)
+        if isinstance(self, CompositeCategory):
+            if isinstance(other, CompositeCategory):
+                return (self.__class__ is other.__class__ and
+                        (self.A == other.A and self.B == other.B or
+                         self.A == other.B and self.B == other.A))
+            # if neither is a compositecategory, then they cannot
+            # be the same category
+            return False
+        elif self.__class__ != other.__class__:
+            return False
+        return self == other
+
+    @__eq__.register(list)
+    @__eq__.register(tuple)
+    @__eq__.register(np.ndarray)
+    def _(self, other):
+        return [self.__eq__(o) for o in other]
+
+    def __ne__(self, other):
+        eq = self == other
+        if isinstance(eq, list):
+            return [not e for e in eq]
+        return eq
+
+    # Implement logical operators to enable composition of sets
+    def __and__(self, other):
+        return AndCategory(self, other)
+
+    #def __or__(self, other):
+    #    return OrCategory(self, other)
+
+    def __xor__(self, other):
+        return XOrCategory(self, other)
+
+
+@set_module("sisl.category")
+class NullCategory(Category):
+    r""" Special Null class which always represents a classification not being *anything* """
+    __slots__ = tuple()
+
+    def __init__(self):
+        pass
+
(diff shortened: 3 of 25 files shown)


pfebrer96:
Seems like could be useful to save the user some for loops!

I still have to read the code more to understand some things, but: don't you think it could be useful that `Category` had also a `filter` method so that instead of:

```python
# return a list of either NullCategory or `n2` categories
cat = n2.categorize(hBN)
# retrieve all indices where the category is not null
idx_n2 = np.array(cat != NullCategory()).nonzero()[0]
```
you could do

```python
idx_2 = n2.filter(hBN)
```

Maybe the `filter` (or `sub`) method could actually return the filtered geometry and something like `extract_from`, `indices` or `ìn` could return the indices. I.e.:

```python
filtered_geom = n2.sub(hBN)
#
idx_2 = n2.in(hBN)
```

(I'm proposing `in` meaning "indices IN hBN that fulfill the condition", not as a shortcut for "indices")

And maybe using it on the reverse order
```python

# For receiving geometries
filtered_geom = hBN.sub(n2) # _sanitize_atoms accepts this categories?
# or
filtered_geom = hBN.sub(!n2)
filtered_geom = hBN.without(n2) 

idx_2 = hBN.where(n2)
# or
idx_opposite = hBN.where(!n2)
```

pfebrer96:
By the way, this kind of functionality is very similar to what a pandas dataframe would do isn't it? 

This is a wild idea, I know, but maybe it could be useful to generate dataframes containing some attributes of the geometry:

```python
df = geom.to_df(cols=["Z", "neighs", "angles", "force"])
```

`Geometry` would know how to build these dataframes because it knows what the attributes mean. Then the user would have all the power of dataframes to do not only complex filters but also `groupby`, `describe`, etc...

zerothi:
> ```python
> idx_2 = n2.filter(hBN)
> ```

Something like this would be an ok idea. I am not too sure about the method name... Filter don't quite cut it ... hmm... Probably your `in` would be good.
 
> Maybe the `filter` (or `sub`) method could actually return the filtered geometry and something like `extract_from`, `indices` or `ìn` could return the indices. I.e.:
> 
> ```python
> filtered_geom = n2.sub(hBN)
> #
> idx_2 = n2.in(hBN)
> ```

I think the category shouldn't do too many things that it isn't intended to do. It should do categorization, and that's it. I.e. `geom.sub(n2)` is just as clear (if not clearer)!

> (I'm proposing `in` meaning "indices IN hBN that fulfill the condition", not as a shortcut for "indices")
> 
> And maybe using it on the reverse order
> 
> ```python
> # For receiving geometries
> filtered_geom = hBN.sub(n2) # _sanitize_atoms accepts this categories?
> ```

Yes, this was actually my intent ;)
> ```python
> # or
> filtered_geom = hBN.sub(!n2)
> filtered_geom = hBN.without(n2) 
> 
> idx_2 = hBN.where(n2)
> # or
> idx_opposite = hBN.where(!n2)
> ```

`hBN.sub(...)` is the reverse of `hBN.remove(!...)`, so that should do. I would really try and limit the number of added functions for the `Geometry` class. It is already quite bloated... :(

I still need to figure out how to do `!n2`. I probably need to create a new Category class that emulates this...


Thanks for the feedback!!

zerothi:
> By the way, this kind of functionality is very similar to what a pandas dataframe would do isn't it?
> 
> This is a wild idea, I know, but maybe it could be useful to generate dataframes containing some attributes of the geometry:
> 
> ```python
> df = geom.to_df(cols=["Z", "neighs", "angles", "force"])
> ```
> 
> `Geometry` would know how to build these dataframes because it knows what the attributes mean. Then the user would have all the power of dataframes to do not only complex filters but also `groupby`, `describe`, etc...

Yeah, this could perhaps be a good idea. :)

pfebrer96:
I guess each dataframe column should be build by the corresponding attribute (https://github.com/zerothi/sisl/issues/196), not `Geometry` itself.

(there are lots of references to that issue haha)

zerothi:
> I still need to figure out how to do `!n2`. I probably need to create a new Category class that emulates this...

So I got it ;)

Now one can do:
```python
~n2
```
which is the opposite (`__invert__`) of `n2`. ;)

jonaslb:
- It seems that currently it is possible for atom a to have b as neigbor while b doesn't have a as neighbor (if they have unequal maxR and the separation is between the two maxR). Maybe it'd be best to settle on some consistent definition of 'neighbors' so that one doesn't get this situation accidentally?

- I think having AtomZ (and presumably there should be AtomSymbol and AtomNOrbs and AtomXXX) is too much typing if you need to match several properties. Maybe it'd be better to have a single `AtomCat(Z=None, symbol=None, norb=None, neighbors=None, etc)`. Maybe there could then be a consistent `AtomCat(Z__gt=3)` for 'greater than' filtering. Or `AtomCat(symbol__contains="Ghost")` or similar for string filtering. This would be very similar to using the ORM in Django, where database queries are made that way (Eg a query on people `adult_johns = Person.objects.filter(age__gte=18, gender=Person.GENDER_MALE, firstname__iexact="john")`).

- As pfebrer is making some suggestions on, it needs to be very easy to apply these criteria to a geometry. Already it is not so difficult to do `geom.sub(geom.atom.Z > 5)` (or `geom.a2o` etc), and for simple things like that, it should not be much more complicated to use the categories. Of course the strength is really in the more complicated things like neighbors, but this should be easy as well :) (at least I shouldn't have to type out `NullCategory()`)

- There are a few of commits in the PR which I think are unrelated (eg dynamical matrix reading?)

- The example with the xor operator is explained like "and must not", but these are different: When `a` is false and `b` is true, then `a^b` is true , but `a & ~b` is false.

- `__or__` is commented out?

Overall I think it looks promising :) My advice is just to focus on making application as easy/easier as the example with Z>5, while making sure that powerful expressions are still nearly equally as easy to type out and apply to a geometry.

pfebrer96:
I don't know if this belongs to this discussion, but since I already said something about it, I'm going to comment it here. It's regarding this:

> By the way, this kind of functionality is very similar to what a pandas dataframe would do isn't it?
> 
> This is a wild idea, I know, but maybe it could be useful to generate dataframes containing some attributes of the geometry:
> ```
> df = geom.to_df(cols=["Z", "neighs", "angles", "force"])
>```
> Geometry would know how to build these dataframes because it knows what the attributes mean. Then the user would have all the power of dataframes to do not only complex filters but also groupby, describe, etc...

I just thought that, apart from filtering, this feature in sisl objects could be extremely useful for visualizing. Here's why: For the visualization module I am using plotly. Plotly has a high-level API under `plotly.express` (https://plotly.com/python/plotly-express/). Plotly express implements some plots like scatter, line, histograms, polar, maps... and more. Basically how it works is that, **given a pandas `DataFrame`, you define your plot as columns of this dataframe**. An example of this:

```python
import plotly.express as px

px.scatter3d(df, x="The x column", y="The y column", z="The z column",
         color="Column that defines color", animation_frame="The column that defines the frames", 
         symbol=..., etc  )
```

So, if sisl objects had a `to_df` method as proposed, it would be **trivial** to implement a "sisl.express" in the visualization module that would just parse the object into a dataframe before passing it to any `plotly.express` method. The possibilities would be endless with very simple code:

```python
import sisl
import sisl.viz.express as sx

geom = sisl.Geometry()
sx.scatter(geom, x="z", y="neighs", color="species")
# or
sx.histogram(geom, x="z", color="species", 
     ...other very useful kwargs of plotly express like marginal="violin") 
``` 

And really all that would be happening would be that this line:

```python
sx.*(sisl_obj, x="z", y="neighs", color="species")
```

is converted into this other line:

```python
px.*(sisl_obj.to_df(cols=["z", "neighs", "species"]), x="z", y="neighs", color="species")
```

I don't know, seems pretty exciting to me :)

zerothi:
Thanks Jonas for your insights!

> * It seems that currently it is possible for atom a to have b as neigbor while b doesn't have a as neighbor (if they have unequal maxR and the separation is between the two maxR). Maybe it'd be best to settle on some consistent definition of 'neighbors' so that one doesn't get this situation accidentally?

Yes, this is a problem I had since the inception of sisl.
In one way sisl is doing it *wrong* since one should figure out the _overlap_ of two orbitals. If they overlap they have a matrix element.  
With such a definition one would not have this problem.
Probably the `neighbor` thingy may have an `R` as input. 

I don't know if I should change the `maxR` behaviour such that one has to find overlaps of orbitals? What do you think? It would make everything consistent (and also with Siesta). However, it would break current codes since `maxR` is not used like that currently.

One could have a global switch for this, e.g. `sisl.options.NEIGBOUR = "overlap"|"individual"`? Which defaults to "individual"?

> * I think having AtomZ (and presumably there should be AtomSymbol and AtomNOrbs and AtomXXX) is too much typing if you need to match several properties. Maybe it'd be better to have a single `AtomCat(Z=None, symbol=None, norb=None, neighbors=None, etc)`. Maybe there could then be a consistent `AtomCat(Z__gt=3)` for 'greater than' filtering. Or `AtomCat(symbol__contains="Ghost")` or similar for string filtering. This would be very similar to using the ORM in Django, where database queries are made that way (Eg a query on people `adult_johns = Person.objects.filter(age__gte=18, gender=Person.GENDER_MALE, firstname__iexact="john")`).

I don't think the typing issue is really bad. You build your properties and combine.
```python
prop = AtomCat(Z=3, norb=4)
# vs.
prop = AtomZ(3) & AtomOrbs(4)
```
that's very little difference in typing, also I think the latter is more clear. Generally I also think these categories will be mildly used. So users with experience may easily create wrappers.
What we could do is generate a parent dispatcher:
```python
AtomCat.Z(3).and.NOrbs(4).xor.Z(4)
```

I like your idea about `__*` to change comparisons. Then when doing `AtomZ(3)` one could just do `AtomZ(gt=3)` since it is implicit with `Z`?

> * As pfebrer is making some suggestions on, it needs to be very easy to apply these criteria to a geometry. Already it is not so difficult to do `geom.sub(geom.atom.Z > 5)` (or `geom.a2o` etc), and for simple things like that, it should not be much more complicated to use the categories. Of course the strength is really in the more complicated things like neighbors, but this should be easy as well :) (at least I shouldn't have to type out `NullCategory()`)

Clarity is better than anything else ;)


> * There are a few of commits in the PR which I think are unrelated (eg dynamical matrix reading?)

Yeah, I can see that, I don't get why this is. I'll try and fix it... Thanks for pointing this out!

> * The example with the xor operator is explained like "and must not", but these are different: When `a` is false and `b` is true, then `a^b` is true , but `a & ~b` is false.

Correct, thanks!

> * `__or__` is commented out?

Yes, this is because I am not really sure what to do.
Currently `&` returns the `AndCategory`, `^` returns the category which is not `NullCategory()`, however, what should I do for `|`? I.e. if both are "true", I should return `OrCategory`, but if one of them are true, should I then return the "true" object.
This makes it a bit problematic since the return value of the `categorize` may be used to figure out which `category` it fits in. I.e. the `^` reduces to the true one (or Null), so you always know what to compare with:
```python
n2 = AtomNeighbours(2)
C_n2 = n2 & AtomZ(6)
notC_n2 = n2 ^ AtomZ(6)
...
cats = (C_n2 ^ notC_n2).categorize(hBN)
for cat in cats:
    if cat == C_n2:
        # for sure this is both n2 and Z=6
    elif cat == n2:
        # this will be returned if it has 2 neighbours but Z!=6
   elif cat == AtomZ(6):
       # neighbours not 2 but Z=6
```
this is of course a contrived example.

> Overall I think it looks promising :) My advice is just to focus on making application as easy/easier as the example with Z>5, while making sure that powerful expressions are still nearly equally as easy to type out and apply to a geometry.

Thanks for your comments!

jonaslb:
Maybe there could be two categories, an "overlaps" for the overlap situation, and a "neighborhood" where R then should be specified? I think this would cover the relevant use cases.

About which is clearer, keywords or multiple objects, maybe it is a matter of what one is used to. Perhaps at a later point there can be a `AtomsKwFilter` for my liking ;) I like chaining things, but I don't like that there's some intermediary useless object eg. `category.xor` is possible but not meaningful (although `cat.xor(Z=5)` would make sense).

Regarding the or-category -- you know how the `or` keyword works? Eg. `0 or 1 or 2` is `1` while `2 or 1 or 0` is `2`. In other words it picks the first truthy object and returns it. I realize here we're overloading the binary or operator, so its not exactly the same, but maybe doing the same would be an idea?

pfebrer96:
> I like your idea about __* to change comparisons. Then when doing AtomZ(3) one could just do AtomZ(gt=3) since it is implicit with Z?

On a scale from 1 to 10, how unpythonic would it be that `AtomZ > 3` returned a category? 

So that you could do:
```python
geom.sub(AtomZ > 3)
```

I don't know if that's even possible if `AtomZ` is a class. It probably would need to be an instance. Like so:

```python
class AtomZSingleton:
    
    def __gt__(self, val):
        
        return AtomZCat(gt=val)
    
    def __call__(self, *args, **kwargs):
        
        return AtomZCat(*args, **kwargs)

class AtomZCat:
    
    def __init__(self, **kwargs):
        
        self.gt = kwargs["gt"]
        
AtomZ = AtomZSingleton()
```

and then you could use it in both ways:

```python
geom.sub( AtomZ(gt=3))

geom.sub(AtomZ > 3)
```

I don't know, it looks good to me since it maintains all the previous functionality and adds that extra possibility of using it in an even simpler way if suitable. What do you think?

pfebrer96:
By the way, regarding what Jonas said:

> Perhaps at a later point there can be a AtomsKwFilter for my liking 

I also think that something like this should exist. I always think that having objects is much more convenient and clear for regular usage. However, there are some cases where you need to define things dinamically and your happiness is **infinite** when you discover that somebody has thought about it to make your life easier.

Just as an example, it's obvious that the most convenient and clean way to import modules is:
```python
import sisl
```
but if you need to import a package dinamically based on user input 
```python
importlib.import_module(package)
``` 
is priceless.

zerothi:
> Maybe there could be two categories, an "overlaps" for the overlap situation, and a "neighborhood" where R then should be specified? I think this would cover the relevant use cases.

I was more thinking globally so `geometry.close` also does either neighborhood vs. overlap?

> About which is clearer, keywords or multiple objects, maybe it is a matter of what one is used to. Perhaps at a later point there can be a `AtomsKwFilter` for my liking ;) I like chaining things, but I don't like that there's some intermediary useless object eg. `category.xor` is possible but not meaningful (although `cat.xor(Z=5)` would make sense).

Actually I agree, it was a bad example ;)
 
> Regarding the or-category -- you know how the `or` keyword works? Eg. `0 or 1 or 2` is `1` while `2 or 1 or 0` is `2`. In other words it picks the first truthy object and returns it. I realize here we're overloading the binary or operator, so its not exactly the same, but maybe doing the same would be an idea?

Thanks! I didn't think about it this way. I'll do this :)

zerothi:
> > I like your idea about __* to change comparisons. Then when doing AtomZ(3) one could just do AtomZ(gt=3) since it is implicit with Z?
> 
> On a scale from 1 to 10, how unpythonic would it be that `AtomZ > 3` returned a category?
> 
> So that you could do:
> 
> ```python
> geom.sub(AtomZ > 3)
> ```
> 
> I don't know if that's even possible if `AtomZ` is a class. It probably would need to be an instance. Like so:

I think it would be doable, but you'd have to use metaclasses for this (or your work-around). Hmm. I don't particularly like the complexity it adds, I don't know if metaclasses for this would give unwanted consequences. ;)

> 
> ```python
> class AtomZSingleton:
>     
>     def __gt__(self, val):
>         
>         return AtomZCat(gt=val)
>     
>     def __call__(self, *args, **kwargs):
>         
>         return AtomZCat(*args, **kwargs)
> 
> class AtomZCat:
>     
>     def __init__(self, **kwargs):
>         
>         self.gt = kwargs["gt"]
>         
> AtomZ = AtomZSingleton()
> ```
> 
> and then you could use it in both ways:
> 
> ```python
> geom.sub(AtomZ(gt=3))
> 
> geom.sub(AtomZ > 3)
> ```

Hmm.. Yes, this could work. But it would require many classes, 2 classes per Category?

> I don't know, it looks good to me since it maintains all the previous functionality and adds that extra possibility of using it in an even simpler way if suitable. What do you think?

I don't think we should go there yet, having 1 solution which works good is ideal. Also, your example works good for single elements, but for `kwargs` you would have to do something like:
```python
AtomNeighbours == {"min": 3, ...}
```
which I don't think improves readability... ;)

zerothi:
> About which is clearer, keywords or multiple objects, maybe it is a matter of what one is used to. Perhaps at a later point there can be a `AtomsKwFilter` for my liking ;)

I have now enabled the _kwfilter_ thingy by this:
```python
AtomCategory.kw(Z=5, neighbours={"max": 2, "min": 1})
AtomZ(5) & AtomNeighbours(1, 2)
```
where the two are equivalent.  
I have extended the base category so that keywords are automatically searching the sub-classes of the class calling `kw`. So if one does a sub-class in ones own script, you'll also be able to use that on a keyword basis in the above ;) Cool ;)
Although the names has to be uniquely found (I am now using `lower().endswith`) which probably isn't the best. But I think this is required since `key.lower() in cls.__name__.lower()` could give problems for all sub-categories which have a `Z` in their name... ;)

Perhaps the `kw` method name is a bit non-descriptive, could we improve it a bit?

pfebrer96:
> Hmm.. Yes, this could work. But it would require many classes, 2 classes per Category?

But this could be easily automathized, couldn't it?

With a "singleton builder":

```python
def category_singleton_factory(Cat):
    
    class CatSingleton:

        def __call__(self, *args, **kwargs):

            return Cat(*args, **kwargs)
        
        def __gt__(self, val):
            return Cat(gt=val)
        
        def __lt__(self, val):
            return Cat(lt=val)
        
        def __eq__(self,val):
            return Cat(val)
        
        def __repr__(self):
            return Cat().__repr__()
        
    #CatSingleton.__doc__ += Cat.__doc__ # Add the documentation of the category
    CatSingleton.__name__ = Cat.__name__
    
    return CatSingleton()
```

And then for each class you would just need to:

```python
class AtomZCat:
    
    def __init__(self, **kwargs):
        print(kwargs)
    
AtomZ = category_singleton_factory(AtomZCat)
```

> Also, your example works good for single elements, but for kwargs you would have to do something like:

Yes, of course this is only helpful for simple cases, but I think it already gives good advantages in usability. For your example of a range there could be shortcuts like:

```python
AtomZ == [3,5] #But maybe that shouldn't be a range
AtomZ == range(3,5) # This is really not helpful

# Maybe something like (you don't need the singleton implementation for this)
AtomZ.range(3,5)
```

Or AtomZCat could implement comparison with int so that this could work `3 < AtomZ < 5`. `3 < AtomZ` would return a category, and then `AtomZCat` would take care of the second comparison.

But still, my idea was **just for simple comparisons**, then it starts getting less advantageous relative to the current approach.

zerothi:
> But still, my idea was **just for simple comparisons**, then it starts getting less advantageous relative to the current approach.

I must admit I don't particularly like this. So I will say this is a no-go. `AtomZ(range(1, 3))` is now possible.

I'll have to think about the other things. There are many things one can do. But it should also be maintainable. ;)

tfrederiksen:
Sorry, I didn't find the time yet to explore this development, but I am sure it is great! I will revisit #202 to make use of this functionality.

zerothi:
@tfrederiksen no worries ;)

If you don't mind, I would like some feedback on this first. Then we adjust categories, and finally we can do the bond-completion. :) Then I think it could be rather cool ;)

Issue #225, opened 2020-05-29
Unfinalized sparsegeometry and math operations


zerothi:
**Describe the bug**

It seems that math operations on non-finalized sparse geometries produces wrong results.
I need to look into this!

zerothi:
I checked TB_07 with finalized matrices or not, and all gets the same result.

PR #226, opened 2020-06-03
ufunc on SparseCSR and derived types

zerothi:
Enable numpy.ufunc i.e. `np.exp(Hamiltonian)` should now work.


zerothi:
This should fix #225

PR #227, opened 2020-06-04
Fixing atoms keywords, all around

### LINKED ISSUE #206: Keyword arguments for atomic indices are not consistent
**Describe the issue**

We need to create consistency throughout sisl (`atom`, `idx` and others are currently used)

We should unify

I dislike `atom` since it may indicate `Atom` for users, I dislike `idx/index` since it does not clarify index.

Possibly candidates:
- `aidx`
- `atom_index`
- `a_index`
- `aindex`
- `idx_atom`
- `i_atom`
- `iatom`

The same applies to orbitals.

Generally I use `ia` for single atomic indices (and `io` for orbitals).

zerothi:
This PR tries to fix #206 by changing all relevant entries of `atom` to `atoms` and `orbital` to `orbitals`.

This may break backwards compatibility to a rather great extend.

I have retained the `atom` keyword in a few places where a single index is _required_. For instance `Geometry.insert`

@sofiasanz @tfrederiksen @pfebrer96 @jonaslb I suspect this may break some of your work.

Could you please try and install this locally, and run your scripts to see if it works, and if it breaks too much? ;)

Thanks!!!

Once this is in I will plan a soon release to get it out there... Then we may iterate a bit on the left-overs (I assume there are some things I forgot!)

### DIFF
--- a/docs/tutorials/tutorial_06_square.py
+++ b/docs/tutorials/tutorial_06_square.py
@@ -14,7 +14,7 @@ print(H)
 # Specify matrix elements
 for ias, idxs in square.iter_block():
     for ia in ias:
-        idx_a = square.close(ia, R=[0.1, 1.1], idx=idxs)
+        idx_a = square.close(ia, R=[0.1, 1.1], atoms=idxs)
         H[ia, idx_a[0]] = -4.
         H[ia, idx_a[1]] = 1.
 
--- a/sisl/atom.py
+++ b/sisl/atom.py
@@ -1277,69 +1277,69 @@ class Atoms:
     # Using the slots should make this class slightly faster.
     __slots__ = ['_atom', '_specie', '_firsto']
 
-    def __init__(self, atom=None, na=None):
+    def __init__(self, atoms=None, na=None):
 
         # Default value of the atom object
-        if atom is None:
-            atom = Atom('H')
+        if atoms is None:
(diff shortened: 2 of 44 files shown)


jonaslb:
Thanks for the heads up, fortunately it seems that `geometry.atom` still works but raises a warning. If it didn't work, that would be a bit of trouble.

I don't think I have been using the keyword in places where it was a positional parameter, so those renames won't affect me. And I think this is the case for most of the renames, right?

So overall, even though I don't have the full overview just now, I don't think I will get so much trouble with this.

zerothi:
> Thanks for the heads up, fortunately it seems that `geometry.atom` still works but raises a warning. If it didn't work, that would be a bit of trouble.
> 
> I don't think I have been using the keyword in places where it was a positional parameter, so those renames won't affect me. And I think this is the case for most of the renames, right?
> 
> So overall, even though I don't have the full overview just now, I don't think I will get so much trouble with this.

Great, thanks! Agreed, they are mostly positional, however in the tbtrans interface, there are some *keyword* only accesses which are changed... :(

pfebrer96:
All good for me! I never use the keyword :)

zerothi:
Great @tfrederiksen @sofiasanz I just tried Hubbard, and all looks good, there are some minor changes required here and there, but I think it should be fine :)

This will go in then!

PR #229, opened 2020-06-10
Fix sanity check in sparse_geometry.append

jonaslb:
I was running into some issues with `spgeom.append` and thought that the sanity check was to blame. I'm pretty sure that I found an error there: When deciding which orbitals to exclude from `spgeom.edges`, the edges which are excluded are in fact the edges that should be *included*. 
I've changed that in this PR.

I later found out that my issues were entirely due to my own errors, but I only realized because I also enhanced the error message.

### DIFF
--- a/sisl/sparse_geometry.py
+++ b/sisl/sparse_geometry.py
@@ -1,6 +1,8 @@
 from numbers import Integral
 import warnings
 import functools as ftool
+import itertools
+import operator
 import numpy as np
 from numpy.lib.mixins import NDArrayOperatorsMixin
 from numpy import (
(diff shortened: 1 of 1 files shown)


zerothi:
> I was running into some issues with `spgeom.append` and thought that the sanity check was to blame. I'm pretty sure that I found an error there: When deciding which orbitals to exclude from `spgeom.edges`, the edges which are excluded are in fact the edges that should be _included_.
> I've changed that in this PR.
> 
> I later found out that my issues were entirely due to my own errors, but I only realized because I also enhanced the error message.

Hmm, I don't get this, is there an error, or did this only improve the message? ;)
Your first paragraph suggests the former, while the last suggests the latter...

zerothi:
Ok, it seems it does fix something, the exclude array was wrong.

PR #230, opened 2020-06-12
Method to smooth grid and (much faster interpolation?)

pfebrer96:
For the sisl visualization module I wanted to implement methods to smoothen a grid. This is useful for cases where you want to see a trend that is at a bigger scale than interatomic distances. I believe there's a siesta utility called `macroave` that does precisely this, but I thought that it would be very easy to implement in python and can be very useful in some cases for sisl grids. This is why I propose this pull request to implement it in the `Grid` object itself.

To do the smoothing, I use the gaussian and uniform filters in [scipy.ndimage module](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html#scipy.ndimage.gaussian_filter). In the same module, I found a function called `zoom`, which interpolates an N-dimensional image to a finer/more spaced grid. I tested it, and it seems to be much faster than the current implementation (4-8 times faster), so I also propose it in this PR. I've kept the `old_interp` for comparison. I believe it does exactly the same (old version does linear interpolation and the new version does spline interpolation of order 1), but maybe I'm missing something and the speed up is just because it's a cheaper (and worse) method.

I was so amazed by all the different possibilities that the `scipy.ndimage` provides that I exposed a public method (`apply`), that let's you apply any function in that module and get a new grid. The name should probably be more specific, but I think the concept is useful and opens many possibilities. Maybe apply here could be something similar to the `PlotHandler`, which has different "engines" (numpy, scipy, xarray, itk, scikit-image...).

Demo
-------
### Interpolation benchmarks
Running this code in my laptop:
```python
import time

grid = sisl.Grid(0.1, sc=10)
grid.grid = np.random.rand(*grid.shape)

# New interpolation implementation
start = time.time()
gr = grid.interp((400,400,400))
new_time = time.time() - start

# Old interpolation implementation
start = time.time()
gr_old = grid.old_interp((400,400,400))
old_time = time.time() - start
print(f'Old interp: {old_time:.2f} s\n New interp: {new_time:.2f} s\n Ratio: {(old_time/new_time):.2f}')
```

Yields:
Old interp: 20.24 s
New interp: 2.61 s
Ratio: 7.75

By the way, the `scipy.ndimage.zoom` accepts zooming ratios (not shape): I've kept `shape` as the input for backwards compatibility, but I think in practice it could be more useful to accept zooming ratios, or maybe grid fineness in Ang. Up to you.

### Smooth functions
The way you can get a smoother grid is (explicitly passing the defaults):
```python
smooth_grid = grid.smooth(method='gaussian', r=0.6)
```

I tested it in a RHO file that I have:
- Visualizing just one axis:
image
- Visualizing two axes:
image
image
image

### Just for fun

```python
grid.smooth(r=0.6).apply('percentile_filter', 20, 10).apply('laplace')
```

image

```python
gr = averaged.smooth(r=0.6).apply('sobel', axis=0).smooth('uniform')
```

image

### And measurements

```python
grid.apply('center_of_mass')
# Label tries to find different isolated objects in the N-dimensional image
# according to a rule of connectivity (here the default).
grid.apply('label') 
```

Just wanted to show that this has a lot of potential for processing grids (and seems to be very fast!), I'm not saying that this examples are meaningful.

Let me know if you think it's worth it. Cheers!

### DIFF
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -139,7 +139,30 @@ class Grid(SuperCellChild):
         """
         self.grid.fill(val)
 
-    def interp(self, shape, method='linear', **kwargs):
+    def interp(self, shape, order=1, **kwargs):
+        """ 
+        Interpolate grid values to a new grid.
+
+        It uses the `zoom` method in `scipy.ndimage`, which creates a finer or
+        # more spaced grid using spline interpolation
+
+        Parameters
+        ----------
+        shape : int, array_like
+            the new shape of the grid
+        order : int 0-5, optional
+            the order of the spline interpolation. 
+            1 means linear, 2 quadratic, etc...
+        **kwargs : dict
+            optional arguments passed to the interpolation algorithm
+            The interpolation routine is `scipy.ndimage.zoom`
+        """
+        # Calculate the zoom_factors
+        zoom_factors = np.array(shape) / self.shape
+
+        return self.apply('zoom', zoom_factors, mode='wrap', order=order, **kwargs)
+
+    def old_interp(self, shape, method='linear', **kwargs):
         """ Interpolate grid values to a new grid
 
         Parameters
@@ -195,6 +218,72 @@ class Grid(SuperCellChild):
 
         return grid
 
+    def smooth(self, method='gaussian', r=None, mode='wrap', **kwargs):
+        '''
+        Make a smoother grid by applying a filter.
+
+        Parameters
+        -----------
+        method: {'gaussian', 'uniform'}
+            the type of filter to apply to smoothen the grid.
+        r: float or array-like of float, optional
+            the radius of the filter in Angstrom for each axis.
+            If the method is 'gaussian', this is the standard deviation!
+            
+            If a single float is provided, then the same distance will be used for all axes.
+            
+            The defaults are 0.6 Ang for a gaussian filter and 1.5 Ang for a uniform filter.
+        '''
+        
+        if r is None:
+            r = 0.6 if method == 'gaussian' else 1.5
+        # Normalize the radius input to a list of radius
+        if isinstance(r, (int, float)):
+            r = [r]*3
+        
+        # Calculate the size of the kernel in pixels (in case the
+        # gaussian filter is used, this is the standard deviation)
+        size = (r/np.linalg.norm(self.dcell, axis=1)).astype(int)
+        
+        # Update the kwargs accordingly
+        if method == 'gaussian':
+            kwargs['sigma'] = size  
+        elif method == 'uniform':
+            kwargs['size'] = size
+        
+        # Apply the method
+        return self.apply(f'{method}_filter', mode=mode, **kwargs)
+
+        
+    def apply(self, method, *args, **kwargs):
+        '''
+        Applies a function of `scipy.ndimage` to the grid and returns a new grid.
+
+        There are many interesting functions in the `scipy.ndimage` submodule. You can
+        check for yourself: https://docs.scipy.org/doc/scipy/reference/ndimage.html
+
+        Parameters
+        -----------
+        method: str or function
+            if it's a string, it should be the name of a function in `scipy.ndimage`.
+
+            if it's a function, it can really be anything. It will get the numpy grid as
+            the first argument.
+        **args and **kwargs:
+            arguments that go directly to the function call
+        '''
+
+        if isinstance(method, str):
+            import scipy.ndimage
+            method = getattr(scipy.ndimage, method)
+
+        grid = self.copy()
+        del grid.grid
+
+        grid.grid = method(self.grid, *args, **kwargs)
+
+        return grid
+            
     @property
     def size(self):
         """ Total number of elements in the grid """
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -224,7 +224,7 @@ class Grid(SuperCellChild):
 
         Parameters
         -----------
-        method: {'gaussian', 'uniform'}
+        method: {'gaussian', 'uniform'}, optional
             the type of filter to apply to smoothen the grid.
         r: float or array-like of float, optional
             the radius of the filter in Angstrom for each axis.
@@ -233,6 +233,11 @@ class Grid(SuperCellChild):
             If a single float is provided, then the same distance will be used for all axes.
             
             The defaults are 0.6 Ang for a gaussian filter and 1.5 Ang for a uniform filter.
+        mode: {'wrap', 'mirror', 'constant', 'reflect', 'nearest'}, optional
+            determines how to compute the borders of the grid.
+            The default is wrap, which accounts for periodic conditions. 
+            
+            See https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html#scipy.ndimage.gaussian_filter
         '''
         
         if r is None:
@@ -253,8 +258,7 @@ class Grid(SuperCellChild):
         
         # Apply the method
         return self.apply(f'{method}_filter', mode=mode, **kwargs)
-
-        
+  
     def apply(self, method, *args, **kwargs):
         '''
         Applies a function of `scipy.ndimage` to the grid and returns a new grid.
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -17,7 +17,7 @@ from .shape import Shape
 from .utils import default_ArgumentParser, default_namespace
 from .utils import cmd, strseq, direction, str_spec
 from .utils import array_arange
-from .utils import call_func
+from .utils import import_attr
 from .utils.mathematics import fnorm
 
 from .supercell import SuperCellChild
@@ -239,7 +239,10 @@ class Grid(SuperCellChild):
             arguments that go directly to the function call
         """
 
-        result = call_func(function_, self.grid, *args, **kwargs)
+        if isinstance(function_, str):
+            function_ = import_attr(function_)
+
+        result = function_(self.grid, *args, **kwargs)
 
         # Maybe the result is not a grid, because there are methods that actually
         # do measurements of the grid
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -281,10 +281,16 @@ class Grid(SuperCellChild):
             import scipy.ndimage
             method = getattr(scipy.ndimage, method)
 
-        grid = self.copy()
-        del grid.grid
+        result = method(self.grid, *args, **kwargs)
 
-        grid.grid = method(self.grid, *args, **kwargs)
+        # Maybe the result is not a grid, because there are methods that actually
+        # do measurements of the grid (e.g. "center_of_mass")
+        if not isinstance(result, np.ndarray) or result.ndim != 3:
+            return result
+
+        # If the result is a grid, we will generate a copy of this one with the new grid values
+        grid = self.copy()
+        grid.grid = result        
 
         return grid
             
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -17,6 +17,7 @@ from .shape import Shape
 from .utils import default_ArgumentParser, default_namespace
 from .utils import cmd, strseq, direction, str_spec
 from .utils import array_arange
+from .utils import call_func
 from .utils.mathematics import fnorm
 
 from .supercell import SuperCellChild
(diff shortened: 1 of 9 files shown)


sisl/grid.py
```diff
@@ -139,7 +139,30 @@ def fill(self, val):
         """
         self.grid.fill(val)
 
-    def interp(self, shape, method='linear', **kwargs):
+    def interp(self, shape, order=1, **kwargs):
```
zerothi, line 142: Add `mode` to args, say you have a grid where you don't care about periodicity, you may want to do two interps with 
one wrap, another constant.

sisl/grid.py
```diff
@@ -139,7 +139,30 @@ def fill(self, val):
...
+        **kwargs : dict
+            optional arguments passed to the interpolation algorithm
+            The interpolation routine is `scipy.ndimage.zoom`
+        """
+        # Calculate the zoom_factors
+        zoom_factors = np.array(shape) / self.shape
+
+        return self.apply('zoom', zoom_factors, mode='wrap', order=order, **kwargs)
+
+    def old_interp(self, shape, method='linear', **kwargs):
```
zerothi, line 165: Just delete the old one, to ensure we don't break old code we could put in a check `kwargs.get("method", None)` and check for "linear". Then set corresponding values.

sisl/grid.py
```diff
@@ -195,6 +218,76 @@ def interp(self, shape, method='linear', **kwargs):
 
         return grid
 
+    def smooth(self, method='gaussian', r=None, mode='wrap', **kwargs):
+        '''
```
zerothi, line 222: Use `"""` instead of `'''`

sisl/grid.py
```diff
@@ -195,6 +218,76 @@ def interp(self, shape, method='linear', **kwargs):
...
+            The defaults are 0.6 Ang for a gaussian filter and 1.5 Ang for a uniform filter.
+        mode: {'wrap', 'mirror', 'constant', 'reflect', 'nearest'}, optional
+            determines how to compute the borders of the grid.
+            The default is wrap, which accounts for periodic conditions. 
+            
+            See https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html#scipy.ndimage.gaussian_filter
+        '''
+        
+        if r is None:
+            r = 0.6 if method == 'gaussian' else 1.5
```
zerothi, line 244: write this out ;)

sisl/grid.py
```diff
@@ -195,6 +218,76 @@ def interp(self, shape, method='linear', **kwargs):
...
+        # Normalize the radius input to a list of radius
+        if isinstance(r, (int, float)):
+            r = [r]*3
+        
+        # Calculate the size of the kernel in pixels (in case the
+        # gaussian filter is used, this is the standard deviation)
+        size = (r/np.linalg.norm(self.dcell, axis=1)).astype(int)
```
zerothi, line 251: `np.linalg.norm` is ridiculously slow, use `sisl.utils.fnorm` instead.

Also, use `astype(np.int32)` for explicitness. Another idea would be to first round, then astype.
What you have is equivalent to truncation.

sisl/grid.py
```diff
@@ -195,6 +218,76 @@ def interp(self, shape, method='linear', **kwargs):
...
+        
+        # Update the kwargs accordingly
+        if method == 'gaussian':
+            kwargs['sigma'] = size  
+        elif method == 'uniform':
+            kwargs['size'] = size
```
zerothi, line 257: Add
```python
else:
    raise ValueError( method is not in gaussian / uniform
```

sisl/grid.py
```diff
@@ -195,6 +218,76 @@ def interp(self, shape, method='linear', **kwargs):
...
+        
+        # Apply the method
+        return self.apply(f'{method}_filter', mode=mode, **kwargs)
+  
+    def apply(self, method, *args, **kwargs):
+        '''
+        Applies a function of `scipy.ndimage` to the grid and returns a new grid.
```
zerothi, line 264: Correct documentation, you may apply *any* function, but a str converts to `scipy.ndimage.<>`

sisl/grid.py
```diff
@@ -195,6 +218,76 @@ def interp(self, shape, method='linear', **kwargs):
...
+        if method == 'gaussian':
+            kwargs['sigma'] = size  
+        elif method == 'uniform':
+            kwargs['size'] = size
+        
+        # Apply the method
+        return self.apply(f'{method}_filter', mode=mode, **kwargs)
+  
+    def apply(self, method, *args, **kwargs):
+        '''
```
zerothi, line 263: Change `''''` to `"""`

sisl/grid.py
```diff
@@ -195,6 +218,76 @@ def interp(self, shape, method='linear', **kwargs):
...
+            if it's a string, it should be the name of a function in `scipy.ndimage`.
+
+            if it's a function, it can really be anything. It will get the numpy grid as
+            the first argument.
+        **args and **kwargs:
+            arguments that go directly to the function call
+        '''
+
+        if isinstance(method, str):
+            import scipy.ndimage
```
zerothi, line 281: Move this import to the top of the file.

sisl/grid.py
pfebrer96: Do you mean to not have a default for `r` or that I should use more than one line for the if/else? 😅  By the way I just realized that I am calculating the kernel size wrong for the uniform case, if `r` is the radius, the size should be doubled (default radius halved).

pfebrer96:
Glad you liked it! I will fix things on monday, because my laptop ran out of battery and my charger is broken :) I will also add some tests.

zerothi:
Great, thanks! Could you also delete the old interp function and squash commits!

sisl/grid.py
zerothi: Hehe, two lines.  
Or you could do:
``python
if r is None:
    r = {"gaussian": 0.6}.get(method, 1.5)
```
in which case it becomes easy to extend defaults?

File: sisl/grid.py (hunk 7 above)
Comment (zerothi) [reply sisl/grid.py:264]: Coming to think of it, I don't think we should allow `method` as `str`, it is rather obscure that it takes functions from `scipy.ndimage`, if anything it should be a string with the full module name + method such that one could also do:
`method="numpy.interpolation.interp` or whatever?

The use of `importlib` would be good here, it seems you are the expert in this ;)

Comment (zerothi) [reply sisl/grid.py:264]: Or centralize a `str2method` method? in `sisl.utils` or something?

Comment (pfebrer96) [reply sisl/grid.py:264]: > The use of importlib would be good here, it seems you are the expert in this ;)

Yes, sure, I'm a core developer of importlib :smile: 

> Or centralize a str2method method? in sisl.utils or something?

Sounds good to me!

Do you still think it's obscure if we have a keyword `module='scypy.ndimage'`?

Comment (zerothi) [reply sisl/grid.py:264]: hmm... No I think that is bad practice since a function *could* use a keyword "module".
In any case it isn't hard for users to create this full string.

**Comment by pfebrer96** [comment]
Do you think it makes sense to interpret shape as zooming factors if they are floats?

You could also mix shape and zoom: `shape=(300, 2.5, 400)` would mean give me this specific shape in the first and third axis and just 2.5x in the second one.

**Comment by zerothi** [comment]
> Do you think it makes sense to interpret shape as zooming factors if they are floats?
> 
> You could also mix shape and zoom: `shape=(300, 2.5, 400)` would mean give me this specific shape in the first and third axis and just 2.5x in the second one.

No, ;)
Bad idea. If users forget to cast to int's and then pass (300., 200., 400.) ;)
Thats one giant grid.

**Comment by pfebrer96** [comment]
Yeah I was thinking on detecting fake floats also, but then you can not 2x.

Can I add an extra keyword for zoom?

**Comment by zerothi** [comment]
> Yeah I was thinking on detecting fake floats also, but then you can not 2x.
> 
> Can I add an extra keyword for zoom? :)

Hmm, I don't think it is that useful.
```python
grid.interp((grid.shape * np.array([2., 2.5, 3.])).astype(np.int32))
```

In the beginning of the project I was also in favour of allowing many things in one method. However, it proves to be very difficult to actually explain what exactly is happening since there are competing options. And this just leads to frustration for new users (not expert users). So I would try and keep interfaces as simple as possible. :)

sisl/grid.py
pfebrer96: Ok :+1: 

By the way, a function could also accept a keyword argument called "method" (actually, it is more likely than "module"), so should I rename it?

pfebrer96:
Ok, I just thought it is more common that you know how much finer or spaced you want your grid than knowing the actual shape.

zerothi:
> Ok, I just thought it is more common that you know how much finer or spaced you want your grid than knowing the actual shape.

I found the contrary. Typically I want to compare with another system and thus have explicit shapes I want to compare with. Well, in any case there is an easy work-around which I think is good :)

pfebrer96:
:+1:

sisl/grid.py
zerothi: hmm, yeah. Perhaps this is a good idea.

`method_` would be pretty good, what do you think?
Still explicit but hides it from the user since it will most likely be used non-keyword.

pfebrer96: Good? Should I call it `function_`, since it actually doesn't need to be a method?

zerothi: Agreed :)

pfebrer96:
I'm trying to run tests but I get:

ImportError while loading conftest 'sisl/sisl/conftest.py'.
../__init__.py:57: in <module>
    from . import info
E   ImportError: cannot import name 'info'

Do you know why? :sweat_smile:

zerothi:
> I'm trying to run tests but I get:
> 
> ```
> ImportError while loading conftest 'sisl/sisl/conftest.py'.
> ../__init__.py:57: in <module>
>     from . import info
> E   ImportError: cannot import name 'info'
> ```
> 
> Do you know why?

You shouldn't run the tests in the repo.
You should do something like:
```bash
python3 setup.py install --prefix=<>
cd <other directory>
pytest --pyargs sisl
```
It is because `info.py` is autogenerated when needed upon installation. :)

pfebrer96:
Oh I see, the problem was that I forgot the `-e` when installing, so `info.py` was not generated inplace. Or is it wrong to run the tests in the repo after you've installed with `-e`?

sisl/grid.py
pfebrer96: Is there a reason you hardcoded the axis in `fnorm`?

zerothi:
> Oh I see, the problem was that I forgot the `-e` when installing, so `info.py` was not generated inplace. Or is it wrong to run the tests in the repo after you've installed with `-e`?

Ok, as long as it works ;)

sisl/grid.py
zerothi: No ;)
Not really. I have just committed to you can choose, default to last axis.

Please rebase ;)

pfebrer96:
A part from `str2func`, there is a `call_func` which gets the function and call it. It seemed convenient to me, but you can remove it if you want (modify also `Grid.apply`).

pfebrer96:
Ok, I don't know how to do it now, because I changed `str2func` to `str2var` to note that in fact you can load a class, or any variable.

Where should I put the type check then? In `Grid.apply`?

pfebrer96:
> Could you fix the history? You merged in from master a couple of times

By the way I don't know how to do this without being afraid of messing everything up :(

zerothi:
> Ok, I don't know how to do it now, because I changed `str2func` to `str2var` to note that in fact you can load a class, or any variable.

1. Since it just gets an attribute of the module, probably `str2attr` would be more appropriate?
2. Don't do a function check, `str2attr` should _only_ be allowed `str` arguments.
3. There is no need to differentiate between objects and functions, an object may implemente `__call__` so in case the user supplies a non-callable attribute, then an error is automatically raised. 

So in `Grid.apply` I think you only need:
```python
if isinstance(function_, str):
    function_ = str2attr(function_
```

If you wish, you can add
```python
if not callable(function_):
    raise ..
```
in `Grid.apply`, but I don't think you need to :)

zerothi:
> > Could you fix the history? You merged in from master a couple of times
> 
> By the way I don't know how to do this without being afraid of messing everything up :(

Ok, I'll have a go at it once you are done.
Once I did it I will tell you how you could have done it, then you can try on your local branch ;)

pfebrer96:
> Since it just gets an attribute of the module, probably str2attr would be more appropriate

`import_attr` ?

zerothi:
> > Since it just gets an attribute of the module, probably str2attr would be more appropriate
> 
> `import_attr` ?

Good idea!!! :) I like this much better :+1:

pfebrer96:
Everything is done now, you can do your commit removing magic :)

zerothi:
Great thanks! Great you added tests!!

zerothi:
Ok, in this case it wasn't too difficult to clean up.

1. Call `git checkout master ; git pull` (to ensure you have the latest master
2. Call `git checkout smooth-grid`
3. Call `git rebase -i master`
Rebase all commits onto master., you'll get something like this:
```
d 6f6094714 Modified gitignore
pick 6b773e687 enh: grid smooth method and faster interpolation
s 581559864 maint: fixed some docs for Grid.smooth
s 919535e70 maint: Grid.apply checks if the returns are a grid
s b7708274c maint: fixed things in Grid.apply and wrote tests for Grid.smooth
s a728e96ce maint: added tests for utils.call_func
s ef4528af7 maint: removed call_func
```
Here edited to get a clean history. See the command what each does.

zerothi:
Merged!

PR #243, opened 2020-07-13
Fdf's _r_grid_bin was using get_sile_class without importing it

pfebrer96:
There was a bug and `fdf_sile.read_grid("VT")` did not work.

I don't know how to add tests for `fdf_sile.read_grid()`, but it would probably be needed, because this is the second time I've found it fails :)


zerothi:
Yeah, tests for fdf are quite sparse and not really easy... I'll see if I can add some...

Merged! Thanks!

PR #246, opened 2020-07-28
Method to calculate grid isosurfaces

pfebrer96:
When plotting grids with plotly I had the problem that non-orthogonal grids didn't make sense because there was no way to indicate the vectors of the cell.

I submitted an issue in plotly and they proposed a workaround, which is to calculate the isosurfaces first with `skimage.measure.marching_cubes` and then apply a transformation to the vertices to get their real coordinates taking into account the cell vectors.

I thought that maybe this is useful for other cases, not only for the plotly visualization module. This would definitely be useful for representing the grids in blender (because it gives you all that you need to build a 3d mesh), and maybe for someone who wants to do some quantitative analysis with isosurfaces (I'm not sure when would this be useful).

If you think it is worth it to include this in the `Grid` class I will write some tests for it. Otherwise I will just do the isosurface calculation in the visualization module.

Cheers!

### DIFF
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -180,6 +180,68 @@ class Grid(SuperCellChild):
         # Apply the scipy.ndimage.zoom function and return a new grid
         return self.apply(ndimage_zoom, zoom_factors, mode=mode, order=order, **kwargs)
 
+    def isosurface(self, level=None, step_size=1, **kwargs):
+        """Calculates the isosurface for a given value.
+        
+        It uses skimage.measure.marching_cubes, so you need to have scikit-image installed.
+
+        Parameters
+        ----------
+        level: float
+            contour value to search for isosurfaces in the grid. 
+            If not given or None, the average of the min and max of the grid is used.
+        step_size: int
+            step size in voxels. Larger steps yield faster but coarser results. 
+            The result will always be topologically correct though.
+        **kwargs:
+            optional arguments passed directly to `skimage.measure.marching_cubes`
+            for the calculation of isosurfaces.
+        
+        Returns
+        ----------
+        numpy array of shape (V, 3)
+            Verts. Spatial coordinates for V unique mesh vertices.
+
+        numpy array of shape (n_faces, 3)
+            Faces. Define triangular faces via referencing vertex indices from verts. 
+            This algorithm specifically outputs triangles, so each face has exactly three indices.
+
+        numpy array of shape (V, 3)
+            Normals. The normal direction at each vertex, as calculated from the data.
+
+        numpy array of shape (V, 3)
+            Values. Gives a measure for the maximum value of the data in the local region near each vertex. 
+            This can be used by visualization tools to apply a colormap to the mesh.
+
+        See Also
+        --------
+        skimage.measure.marching_cubes : method used for interpolation
+        
+        """
+        try:
+            import skimage.measure
+        except:
+            raise ModuleNotFoundError("You need to have scikit-image installed in order to calculate isosurfaces.")
+
+        # Run the marching cubes algorithm to calculate the vertices and faces
+        # of the requested isosurface.
+        verts, *returns = skimage.measure.marching_cubes(
+            self.grid, level=level, spacing=fnorm(self.dcell), step_size=step_size, **kwargs
+        )
+
+        # Define the transformation matrix to get the real vertices
+        # This is because skimage calculates the vertices as if it was an orthogonal cell
+        T = self.cell/fnorm(self.cell).T
+
+        # Check that the transformation matrix is definite positive
+        if np.linalg.det(T) < 0:
+            T[:, 2] = -T[:, 2]  # change the orientation of the third vector
+
+        # Apply the transformation and get the real coordinates of the vertices
+        verts = (np.dot(T, verts.T)).T
+
+        return (verts, *returns)
+
     def smooth(self, r=0.7, method="gaussian", mode="wrap", **kwargs):
         """
         Make a smoother grid by applying a filter.
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -180,23 +180,23 @@ class Grid(SuperCellChild):
         # Apply the scipy.ndimage.zoom function and return a new grid
         return self.apply(ndimage_zoom, zoom_factors, mode=mode, order=order, **kwargs)
 
-    def isosurface(self, level=None, step_size=1, **kwargs):
+    def isosurface(self, level, step_size=1, **kwargs):
         """Calculates the isosurface for a given value.
-        
-        It uses skimage.measure.marching_cubes, so you need to have scikit-image installed.
+
+        It uses `skimage.measure.marching_cubes`, so you need to have scikit-image installed.
 
         Parameters
         ----------
         level: float
             contour value to search for isosurfaces in the grid. 
             If not given or None, the average of the min and max of the grid is used.
-        step_size: int
+        step_size: int, optional
             step size in voxels. Larger steps yield faster but coarser results. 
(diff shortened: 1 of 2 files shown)


sisl/grid.py
```diff
@@ -180,6 +180,68 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
         return self.apply(ndimage_zoom, zoom_factors, mode=mode, order=order, **kwargs)
 
+    def isosurface(self, level=None, step_size=1, **kwargs):
+        """Calculates the isosurface for a given value.
+        
+        It uses skimage.measure.marching_cubes, so you need to have scikit-image installed.
+
+        Parameters
+        ----------
+        level: float
```
zerothi, line 190: I think `level` should be a required input. The average min/max seems rather arbitrary ;)

sisl/grid.py
```diff
@@ -180,6 +180,68 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+            contour value to search for isosurfaces in the grid. 
+            If not given or None, the average of the min and max of the grid is used.
+        step_size: int
```
zerothi, line 193: I think this is not needed, it will already be handled by `kwargs` and I guess (?) that users will tend to want the same resolution?

sisl/grid.py
```diff
@@ -180,6 +180,68 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+
+        See Also
+        --------
+        skimage.measure.marching_cubes : method used for interpolation
+        
+        """
+        try:
+            import skimage.measure
+        except:
+            raise ModuleNotFoundError("You need to have scikit-image installed in order to calculate isosurfaces.")
```
zerothi, line 224: Change to:

`f"{self.__class__.__name__}.isosurface requires scikit-image to be installed"`

sisl/grid.py
```diff
@@ -180,6 +180,68 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+
+        # Run the marching cubes algorithm to calculate the vertices and faces
+        # of the requested isosurface.
+        verts, *returns = skimage.measure.marching_cubes(
+            self.grid, level=level, spacing=fnorm(self.dcell), step_size=step_size, **kwargs
+        )
+
+        # Define the transformation matrix to get the real vertices
+        # This is because skimage calculates the vertices as if it was an orthogonal cell
+        T = self.cell/fnorm(self.cell).T
```
zerothi, line 234: add a space on both sides of `/`

sisl/grid.py
```diff
@@ -180,6 +180,68 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+
+        # Check that the transformation matrix is definite positive
+        if np.linalg.det(T) < 0:
+            T[:, 2] = -T[:, 2]  # change the orientation of the third vector
+
+        # Apply the transformation and get the real coordinates of the vertices
+        verts = (np.dot(T, verts.T)).T
```
zerothi, line 241: change to `T.dot(verts.T).T`, simpler and cleaner ;)

sisl/grid.py
```diff
@@ -180,6 +180,68 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+
+        return (verts, *returns)
```
zerothi, line 243: remove parenthesis ;)

sisl/grid.py
pfebrer96: Yes, I thought so too, I'm going to change it.

I also thought that maybe it is useful to let the user specify the level as a percentage from min to max (0 -> min, 1 -> max) so  that the same value can be used for different grids and the user does not need to check what are the exact values contained in the grid before using the `isosurface` method. I don't know how to make it easy to accept both the value and the percentage. What do you think?

sisl/grid.py
pfebrer96: I was not sure about this, because I think the typical resolution of a grid in SIESTA results in a very big amount of vertices and faces, which can be hard for the rendering engines to display (depending on the computer of course). Sometimes they don't add any visual information versus a surface that is defined with less resolution. Of course the default needs to be 1, but by putting it here I wanted to emphasize that you can tune the parameter, so that no one gets stuck with a too large number of vertices.

If you think this is not important or that it is enough to point to the docs of `skimage.measure.marching_cubes` I will remove it.

sisl/grid.py
pfebrer96: Apparently it is invalid syntax to write

```python
return verts, *returns
```

I wrote it like this at the beginning but it raised a syntax error.

sisl/grid.py
zerothi: I don't think the percentage would be of great use. For various reasons. Consider a comparison of simulations where one in the set has a much higher maximum. Then you can't compare by percentage.

sisl/grid.py
zerothi: Ok, agreed, keep it.

sisl/grid.py
zerothi: Ok, I was unaware of this, so keep it. :)

sisl/grid.py
pfebrer96: You are right, no percentage argument.

pfebrer96:
Everything is done now and I added some tests. I don't know how to specify that the test should fail if `scikit-image` is not installed. If you could to this, I think there's nothing else to do :)

zerothi:
> Everything is done now and I added some tests. I don't know how to specify that the test should fail if `scikit-image` is not installed. If you could to this, I think there's nothing else to do :)

You simply do this:
```python
def test_<>(...):
    pytest.importorskip("scikit.image", reason="scikit-image not available")
    <body of test>
```

sisl/tests/test_grid.py
```diff
@@ -182,6 +182,31 @@ def test_interp_extrap(self, setup):
...
+        grid.grid = np.tile([1, 2, 3, 4, 5, 4, 3, 2, 1, 0], 100).reshape(10, 10, 10)
+
+        # Calculate the isosurface for the value of 2.5
+        verts, *returns = grid.isosurface(2.5)
+
+        # The third dimension should contain only two coordinates
+        # [1, 2, (HERE) 3, 4, 5, 4, 3, (HERE) 2, 1, 0]
+        assert np.unique(verts[:, 2]).shape == (2,)
+
+        # If the grid is non-orthogonal, there should be 20 unique values
```
zerothi, line 200: Split into two tests! Great you added tests!

```
def test_isosurface_orthogonal(self)

def test_isosurface_non_orthogonal(self)
```

pfebrer96:
Done :+1:

pfebrer96:
Nick, I'm thinking now that instead of returning the vertices, faces, normals, etc... we could return an `Isosurface` object that inherits from `Shape`. In that way we could use the methods that you already have there. E.g.: to get the atoms that are inside the isosurface, for example.

I'm not familiar with the implementation of shapes, is this possible?

zerothi:
> Nick, I'm thinking now that instead of returning the vertices, faces, normals, etc... we could return an `Isosurface` object that inherits from `Shape`. In that way we could use the methods that you already have there. E.g.: to get the atoms that are inside the isosurface.

Good idea, but no.. :(
The `Shape`s are not meant for this. Currently they only encompass volumetric data. If planes and polygon-surfaces are needed later, then it requires a major effort on this part first before doing this.

But I will have it in mind if need be :)

Thanks for the suggestions!

pfebrer96:
Right, I was thinking only about closed isosurfaces, but I didn't notice that isosurfaces do not necessarily create a closed volume.

PR #248, opened 2020-08-03
Blender support

pfebrer:
(Continuation of https://github.com/zerothi/sisl/pull/171. Just built the branch again to allow smooth rebasing as happened with the plotly module.)

*This is the beggining of an era of beautiful slide presentations.*

Here I have implemented the first basic functionalities that will allow an easy interaction between scientific data manipulated by sisl and beautiful 3D image rendering by blender (https://www.blender.org/). The aim of it is to **introduce people into 3D image rendering** and provide a set of automatic tools that will require little to no interaction of the user with blender's GUI. However, it is meant to also **serve people who is already familiar with blender by automating some repetitive tasks** and while still leaving room for the user to display their creativity.

How to use it
------

A function has been implemented under sisl.blender that will take care of everything if you pass it a renderable object *(at the moment, only Geometry is renderable)*. The easiest way to use it is by building a script that looks something like this:
```python
import sisl

geom = sisl.get_sile(<your path to a geometry file here>).read_geometry()

sisl.blender(geom, camera_settings={}, lighting_settings={}, render_settings={})
```
Then, you can run this python script within blender in many ways, one of them being:
```
blender -P path/to/your/script.py
```
from the terminal, which will open blender, run your script and then leave the GUI open. Use the *-b* flag if you want to run blender in background mode.

*Check the docs of the sisl.blender function under _blender.py, and specially the docs on the `camera()`, `lighting()` and `render()` methods of the `BlenderScene` class for more info on how to use it. I believe they are quite explanatory. If not, don't hesitate to ask questions here.*

How to set it up
-----
For this tools to work, you need a python environment that contains both the sisl package and blender modules (bpy, mathutils, addon_utils). There are three main ways (that I know) to set up an environment that satisfies these requirements:
- **Make blender "aware" of your environment (recommended):** This option is the best one as you don't need to modify blender's filesystem and therefore you can do it with the snap installation (from the Ubuntu Store). Follow the instructions here https://blender.stackexchange.com/a/76124, it's very clear. 

    Basically you just need to have a python environment with the same python version as blender (run `import sys; print(sys.version_info)` inside blender, for blender 2.83 is python 3.7.4) and install sisl there. Then add the path of the environment to blender's python. Note that you can make blender automatically add the path on start up (last point of the stack exchange comment)

- **Install sisl in blender's python environment:** Download blender from https://www.blender.org/download/ (*NOT FROM THE UBUNTU SOFTWARE STORE, because it won't let you modify the filesystem and we need to*). Then find the python binary under 2.82/python/bin. You then need to install pip for that python and with that install sisl:
```
<path/to/blender/python> -m ensurepip
<path/to/blender/python> -m pip install sisl
```
If you can't install sisl because of an error that is something like: *Python.h no such file or directory*, it is because you are missing some headers. If blender's python is python3.7, they can be installed in linux like so:
```
sudo apt-get install python3.7-dev
```
However, it seems like the python for blender it is unable to find them (at least for me), so you will probably need to specify their location like so:

```
CFLAGS="-I/usr/include/python3.7" <path/to/blender/python> -m pip install sisl
```

as suggested by Nick.

- **Install blender as a package in your sisl environment:** This should be possible either by building blender yourself as a python package (https://wiki.blender.org/wiki/Building_Blender/Other/BlenderAsPyModule,https://blender.stackexchange.com/questions/117200/how-to-build-blender-as-a-python-module) or using prebuilt versions that can be installed via pip (https://pypi.org/project/bpy/, https://pypi.org/project/blender/). *I have not been able to accomplish any of the two.*

**Let me know your opinion about it, suggest changes in the structure of the code or request further functionality that you think would be necessary.** Thanks!

### DIFF
--- a/sisl/_blender.py
+++ b/sisl/_blender.py
@@ -1,0 +1,501 @@
+from scipy.spatial.transform import Rotation
+import numpy as np
+from math import acos
+import os
+from ._array import arrayd
+
+try:
+    import bpy
+    from mathutils import Matrix, Vector
+
+    try:
+        import addon_utils
+        INSIDE_BLENDER = True
+        BLENDER_ASPACKAGE = False
+        
+    except ModuleNotFoundError:
+        BLENDER_ASPACKAGE = True
+        INSIDE_BLENDER = False
+
+    BLENDER_AVAIL = True
+
+except ModuleNotFoundError as BLENDER_IMPORT_EXCEPTION:
+    BLENDER_AVAIL = False
+    BLENDER_ASPACKAGE = False
+    INSIDE_BLENDER = False
+
+__all__ = ['BlenderScene', 'blender']
+
+class BlenderScene:
+
+    _POVs = {
+        '+z': (0,0,1),
+        '-z': (0,0,-1),
+        '+y': (0,1,0),
+        '-y': (0,-1,0),
+        '+x': (1,0,0),
+        '-x': (-1,0,0),
+    }
+
+    def __init__(self, objs = [], filepath = None):
+
+        #Define some attributes that will be helpful so that bpy does not need to be imported elsewhere
+        self.ops = bpy.ops
+        self.context = bpy.context
+        self.data = bpy.data
+        self.atomic_blender = lambda: addon_utils.enable("io_mesh_atomic")
+
+        self.filepath = filepath or os.path.join(os.path.expanduser("~"), "image")
+
+        #Clean the scene of any objects that were there (e.g. the default cube)
+        self.clean()
+
+        self.objs = []
+        self.lights = []
+
+        #If objects were provided on initialization, add them to the scene
+        for obj in objs:
+            self.add(obj)
+
+    def update_bounding_box(self):
+        '''
+        Updates the bounding box of the model so that lights and cameras can be placed correctly.
+        '''
+        self.bounding_box = np.array([
+            np.min([np.min(obj.bound_box, axis=0 ) for obj in self.data.objects], axis = 0),
+            np.max([np.max(obj.bound_box, axis=0 ) for obj in self.data.objects], axis = 0),
+        ]).T
+
+        return self
+
+    def lim(self, min_max = None, axis = None):
+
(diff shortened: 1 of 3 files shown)


pfebrer96:
Thanks to `Grid.isosurface`, now grid isosurfaces can be represented in blender. It's fast and they look great :)

However, I think we should rethink how to do things in blender. It may be worth it to profit from all the processing and settings that are already available in the plots from the "plotly visualization module". The only difference would be the rendering part. 

For example one could do

```python
H = ...load the hamiltonian
wf = H.plot.wavefunction(i=20)

# Show it with plotly
wf.show()

# Add it to the scene in blender
scene = BlenderScene()
scene.add(wf)
```

The blender part would not need to work exactly like this but it's just to give an example. Then, the `__blender__` method would not be in the classes like `Geometry` and `Grid` but in each plot, giving much more versatility because from the same class you could have multiple representations that are implemented in each plot.

And of course the way that you can do it now:

```python

geometry = ...load the geometry
scene.add(geometry)

```

would load the default plot. I.e. would be the same as doing:

```python
scene.add(geometry.plot())
```

I believe this not only would be very handy for users, as they wouldn't have to learn two different ways of representing things, but also would make it easier to maintain the code, as there would be less and it wouldn't be repeated.

zerothi:
Thanks, I am in the process of trying out your #205 PR, could you please hold on with doing more on this branch (just so we can finish the other one).

pfebrer96:
Yes :+1:. I just wanted to point out this possibility to take it into account when giving the final shape to #221.

zerothi:
@pfebrer could you rebase here? Once this is in it may be ideal to try the different backend approach?

pfebrer:
I think this PR could be totally discarded and we could focus on the
implementation of the dispatcher approach. After that is settled I can try
to incorporate blender.

What do you think?

El dj., 4 de febr. 2021, 12:29, Nick Papior <notifications@github.com> va
escriure:

zerothi:
Ok, sounds like a plan!

PR #250, opened 2020-08-06
Making sure siles are pickleable

pfebrer96:
I have two questions:
- Regarding tests: In the visualization module, I define a `BasePlotTester` class that I then use for each specific class. For example, where I want to test the bands plot, I define `BandsPlotTester(BasePlotTester)`. In this way I make sure that the basics are tested across all specific plots without having to rewrite the code. It makes me feel quite confident that plots work. Is this bad practice? I'm asking because the same could be done for siles, and by just adding a test in the `BaseSileTester` we would make sure that all siles are pickleable, for example.

- Do you want to use this PR to make sure all classes (not only siles) are pickleable so that we don't need to worry anymore?

### DIFF
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -425,6 +425,16 @@ class BaseSile:
             raise AttributeError(f"The filehandle for {self.file} has not been opened yet...")
         return getattr(self.fh, name)
 
+    def __getstate__(self):
+        """This method and __setstate__ avoid errors when pickling siles"""
+
+        return self.__dict__ 
+
+    def __setstate__(self, state):
+        """This method and __getstate__ avoid errors when pickling siles"""
+            
+        self.__dict__ = state
+
     @classmethod
     def _ArgumentParser_args_single(cls):
         """ Default arguments for the Sile """
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -427,12 +427,10 @@ class BaseSile:
 
     def __getstate__(self):
         """This method and __setstate__ avoid errors when pickling siles"""
-
         return self.__dict__ 
 
     def __setstate__(self, state):
-        """This method and __getstate__ avoid errors when pickling siles"""
-            
+        """This method and __getstate__ avoid errors when pickling siles""" 
         self.__dict__ = state
 
     @classmethod
--- a/sisl/io/tests/test_sile.py
+++ b/sisl/io/tests/test_sile.py
@@ -1,29 +1,33 @@
 from pathlib import Path
-import pickle
 import pytest
 import tempfile
 
 import sisl
 from sisl.io.sile import __siles
 
+@pytest.fixture(params=["pickle", "dill"])
+def pickle_module(request):
+    return request.param
 
 @pytest.fixture(params=__siles)
 def sile(request):
     sile_cls = request.param
 
-    return sile_cls("test")
+    return sile_cls("test", _open=False)
 
-def test_pickling(sile):
 
+def test_pickling(sile, pickle_module):
+
+    pickle_module = pytest.importorskip(pickle_module)
     tmp_file = tempfile.NamedTemporaryFile(delete=False)
     error = None
 
     try:
         with open(tmp_file.name, "wb") as fh:
-            pickle.dump(sile, fh)
+            pickle_module.dump(sile, fh)
 
         with open(tmp_file.name, "rb") as fh:
-            loaded_sile = pickle.load(fh)
+            loaded_sile = pickle_module.load(fh)
     except Exception as e:
         error = e
 
--- a/sisl/io/tests/test_sile.py
+++ b/sisl/io/tests/test_sile.py
@@ -1,0 +1,35 @@
+from pathlib import Path
+import pickle
+import pytest
+import tempfile
+
+import sisl
+from sisl.io.sile import __siles
+
+
+@pytest.fixture(params=__siles)
+def sile(request):
+    sile_cls = request.param
+
+    return sile_cls("test")
+
+def test_pickling(sile):
+
+    tmp_file = tempfile.NamedTemporaryFile(delete=False)
+    error = None
+
+    try:
+        with open(tmp_file.name, "wb") as fh:
+            pickle.dump(sile, fh)
+
+        with open(tmp_file.name, "rb") as fh:
+            loaded_sile = pickle.load(fh)
+    except Exception as e:
+        error = e
+
+    Path(tmp_file.name).unlink()
+
+    if error is not None:
+        raise error
+
+    


zerothi:
> * Regarding tests: In the visualization module, I define a `BasePlotTester` class that I then use for each specific class. For example, where I want to test the bands plot, I define `BandsPlotTester(BasePlotTester)`. In this way I make sure that the basics are tested across all specific plots without having to rewrite the code. It makes me feel quite confident that plots work. Is this bad practice? I'm asking because the same could be done for siles, and by just adding a test in the `BaseSileTester` we would make sure that all siles are pickleable, for example.

This is what is named a `fixture` in `pytest`, please look it up ;)
That is the correct way to do it. Potentially using `pytest.mark.parametrize` with different classes., grep for those things in `sisl` and you'll find plenty of things, do not hesitate to ask.

> * Do you want to use this PR to make sure all classes (not only siles) are pickleable so that we don't need to worry anymore?
No I think the majority of the classes already are pickeable, so lets concentrate on one thing at a time. First siles, then if you have problems with other classes, open an issue so we can track it.

sisl/io/sile.py
```diff
@@ -425,6 +425,16 @@ def __getattr__(self, name):
             raise AttributeError(f"The filehandle for {self.file} has not been opened yet...")
         return getattr(self.fh, name)
 
+    def __getstate__(self):
+        """This method and __setstate__ avoid errors when pickling siles"""
+
```
zerothi, line 430: Remove empty line between doc and code

sisl/io/sile.py
```diff
@@ -425,6 +425,16 @@ def __getattr__(self, name):
...
+        return self.__dict__ 
+
+    def __setstate__(self, state):
+        """This method and __getstate__ avoid errors when pickling siles"""
+
```
zerothi, line 435: same here.

pfebrer96:
Thanks for the tips! I'll use fixtures from now on :)

There's a test for all siles pickling with both `pickle` and `dill`. There are two types of siles that don't pass the tests: those from `sisl.io.phtrans` and those from `sisl.io.siesta.binaries`. It seems like it is unable to find the classes at those paths:

```
Can't pickle <class 'sisl.io.phtrans.phtprojncSilePHtrans'>: it's not found as sisl.io.phtrans.phtprojncSilePHtrans
```

Any idea why?

jonaslb:
What happens if the sile has been used before pickling? It might contain an open file handle in that case.
Further siles might accumulate some additional internal state which will be invalid after unpickling even if the filehandle was reopened.

It should also be considered whether this is expected behaviour at all -- file handles are not pickleable, so why should Siles be -- and what happens if you try? The usual thing to do is to pickle the path instead of file, because then at least it is clear to the user that state is reset.

In any case, even if we really want to pickle siles, I think there shouldn't be a general Sile-implementation, because it cannot know how the childs store their internal state. Or at least it should be ensured that an exception is raised if no specific implementation exists.

zerothi:
> What happens if the sile has been used before pickling? It might contain an open file handle in that case.
> Further siles might accumulate some additional internal state which will be invalid after unpickling even if the filehandle was reopened.
> 
> It should also be considered whether this is expected behaviour at all -- file handles are not pickleable, so why should Siles be -- and what happens if you try? The usual thing to do is to pickle the path instead of file, because then at least it is clear to the user that state is reset.
> 
> In any case, even if we really want to pickle siles, I think there shouldn't be a general Sile-implementation, because it cannot know how the childs store their internal state. Or at least it should be ensured that an exception is raised if no specific implementation exists.

Thanks for your comments Jonas, I am not really aware of any of these issues. So nice you bring them to the table. I think Pol is mainly interested in some of the data in the file. I agree that the pickling does not necessarily carry the actual file information and thus the potential use is not so big.
@pfebrer96 which state is it that the plotting mechanisms wants? Is it only used for reading the file name (a simple str), or do you expect it to carry all file information?

pfebrer96:
Thanks for the comments @jonaslb !

In the case of plots all I want basically is to store the path, but I didn't want to write an additional check before pickling to convert all siles to paths and another one after unpickling to do the opposite. So for my case it would be enough if just the path is conserved through the pickling/unpickling process. I just didn't know if the other parameters were meaningful to store. That's why I stored the full `__dict__`.

Should I not store the information in siles?

zerothi:
> Thanks for the comments @jonaslb !
> 
> In the case of plots all I want basically is to store the path, but I didn't want to write an additional check before pickling to convert all siles to paths and another one after unpickling to do the opposite. So for my case it would be enough if just the path is conserved through the pickling/unpickling process. I just didn't know if the other parameters were meaningful to store. That's why I stored the full `__dict__`.
> 
> Should I not store the path information in siles maybe?

Ah, but if this is what you want, why not just do `str(obj)`, done! That string representation should contain the path + other information

pfebrer96:
Well I'm not pickling the sile specifically :sweat_smile:.

I'm pickling the whole plot and I expect everything to be pickled automatically. That's why if `Sile` doesn't have a protocol to be pickled I need to find all siles that are currently within the state of the plot and convert them to paths, and somehow store that information so that I can restore the siles again when the plot is unpickled.

pfebrer96:
So what I'm asking is: Is the solution storing just the `Path` object (never the sile) and then get the siles each time that the plot requires some information from the path?

zerothi:
I think, given @jonaslb's input; that the correct approach would be:

1. Remove pickling from `Siles`
2. Override plotable's pickling method (`__state__`) to store `Path` in the pickling process
3. When restoring from a pickled object, recreate the siles. This however, may be problematic since if files are removed/deleted, exceptions will happen. In that case perhaps fall back to simply `Path` stored in the plotable?

@jonaslb your opinion would be of value here ;)

jonaslb:
Ok, yes I agree that it would be the correct approach to handling Siles because they are stateful and that state cannot be transferred. I don't think your point 3. is really problematic, its better to throw an immediate exception if someone tries to do nonsense, instead of making "unclean" objects that will just throw exceptions at a later point. Actually -- I think siles should throw an error if one tries to instantiate them with a non-existent path in read mode, just like if you try to open a non-existent file (they don't currently, I think).

I think there is an alternative way though -- we could declare that Siles must be stateless (only storing path and mode) and rework the ones that do have some state to be stateless instead (fdfSile comes to mind, might be trouble with the binary siles). In that case I think it would be more reasonable to have pickling because then siles are no longer "file-like".

pfebrer96:
>  I don't think your point 3. is really problematic, its better to throw an immediate exception if someone tries to do nonsense, instead of making "unclean" objects that will just throw exceptions at a later point.

In the case of plots you need to handle that exception, because you can save a plot in your filesystem and then send it to someone. Even if the plot contains a sile pointing to a path that is no longer available, there should be no problem in displaying and modifying the plot as long as the reading of the sile is not triggered.

zerothi:
> Ok, yes I agree that it would be the correct approach to handling Siles because they are stateful and that state cannot be transferred. I don't think your point 3. is really problematic, its better to throw an immediate exception if someone tries to do nonsense, instead of making "unclean" objects that will just throw exceptions at a later point. Actually -- I think siles should throw an error if one tries to instantiate them with a non-existent path in read mode, just like if you try to open a non-existent file (they don't currently, I think).

Good idea, feel free to make a PR for this! ;)
 
> I think there is an alternative way though -- we could declare that Siles must be stateless (only storing path and mode) and rework the ones that do have some state to be stateless instead (fdfSile comes to mind, might be trouble with the binary siles). In that case I think it would be more reasonable to have pickling because then siles are no longer "file-like".

Hmm. I would rather we have consistency on Siles. And currently binary files requires a state to handle where they are in the file.
Since this is the way Python handles it, I think we need a _really_ good reason for not doing the same?

pfebrer96:
By the way, if plots and sessions are something to be saved/loaded, should I create siles for them and keep the pickling functionality in the siles?

zerothi:
> By the way, if plots and sessions are something to be saved/loaded, should I create siles for them and keep the pickling functionality in the siles?

zerothi:
Woops... Hmm good question.
A session is per see not a Sile, however, one could create a specific `SessionSile` which has some special functionality. Hmm.. That needs to be thought through.

pfebrer96:
> A session is per see not a Sile

Why not? Just to be clear, by session I mean my `Session` class, which basically stores plots organized in tabs.

zerothi:
> > A session is per see not a Sile
> 
> Why not? Just to be clear, by session I mean my `Session` class, which basically stores plots organized in tabs.

A `Sile` is by definition a file. A session is in your case a tab, or something that holds a state of something, it can live without any harddisk, the `Sile` can't.

pfebrer96:
Yes, but I mean, there could be a `SessionSile` that is to `Session` the same as `tshsSileSiesta` is to `Hamiltonian`. Or not?
I'm not saying that `Session` is a sile itself,  just that `SessionSile` would be a `Sile` from where sessions can be read or where sessions can be written. Does it make sense? Why is it that special?

zerothi:
> Yes, but I mean, there could be a `SessionSile` that is to `Session` the same as `tshsSileSiesta` is to `Hamiltonian`. Or not?
> I'm not saying that `Session` is a sile itself, just that `SessionSile` would be a `Sile` from where sessions can be read or where sessions can be written. Does it make sense? Why is it that special?

I think we agree, that was what I wrote in https://github.com/zerothi/sisl/pull/250#issuecomment-670498370
I was just not sure exactly how that would work. And whether there are other more easy solutions for this; i.e. sessions contain binary encoded stuff (I guess) and how to store that efficiently is not something I know...

zerothi:
@pfebrer96 and @jonaslb I think we have agreed that this will not go in. We will not have pickling on siles.

Any pickling on sessions or classes which contain sile-objects must have some work-around method to do so.

pfebrer96:
Yes! You can close this.

PR #253, opened 2020-08-12
Geometry._sanitize_atoms did not accept generic categories

pfebrer96:
Basically:

```python
import sisl

geom = sisl.geom.graphene()
geom.sub({"Z": 6, "neighbours": 2})
```

was raising an error because `AtomCategory.kw()` returned an `AndCategory`, which was not registered in `_sanitize_atoms`. My approach has been to create a new class `GenericCategory` for the categories that are not "context-specific" (null, and, not, etc...) so that we don't need to register `Category` (which wouldn't make sense), but maybe it's not the best one.

Also in this same PR I made `Category.kw` able to return categories that don't accept any arguments, so that one can do:

```python
geom.sub({"odd": True})
```

(otherwise impossible). At the same time I enabled:

```python
geom.sub({"even": False})
```

for these cases, which returns a `NotCategory`.


sisl/_category.py
```diff
@@ -111,8 +111,20 @@ def kw(cls, **kwargs):
...
+            
+            try:
+                return cl(args)
+            except TypeError as e:
+                # Here we are trying to catch the fact that some classes
+                # don't accept any arguments
+                if args is True:
+                    return cl()
+                elif args is False:
+                    return ~cl()
```
zerothi, line 123: Hmm, I don't particularly fancy this, thus you can only return `NotCategory` for categories without arguments? I think we should do something like:

```python
if args is None:
   try: return cl(args)
   except: TypeError as e
   return cl()
```
This makes more sense for end-users.

sisl/_category.py
```diff
@@ -196,7 +208,18 @@ def __invert__(self):
 
 
 @set_module("sisl.category")
-class NullCategory(Category):
+class GenericCategory(Category):
```
zerothi, line 211: I think we don't need this, we should simply register `Category` in `_sanitize_atoms` if the user passed categories which does not accept the geom as arguments errors will be raised.
This will allow us to enable `OrbitalCategory` and not need to register those as individuals.

zerothi: Hmm, I see your point, but I think for maintainability this is better, no?

sisl/_category.py
pfebrer96: Yes I also thought this at the beggining, but do you think it makes more sense for a user to do:

 ```python
geom.sub({"odd": None})
```
than

```python
geom.sub({"odd": True})
```
?

After giving it some thought, I came to the conclusion that categories that don't accept arguments are like a switch, so for me it makes sense that you can set them to `True` or `False` and it doesn't bother me that you can not do this with other categories that have a wider range of possibilities. What do you think?

sisl/_category.py
pfebrer96: Yes, maybe it will look a bit ugly if you start to register lots of categories. Still, maybe it's good to keep the distinction even if we're not using it in this case for quick type checking (in sisl or user code)? 

PS: Does it make sense to do type checking in the category itself to raise a meaningful error or is it a waste of time that will decrease performance?

sisl/_category.py
zerothi: Hmm, I think argument-less categories should be indifferent of arguments.

To get true/false perhaps we could do
```python
geom.sub({"~odd":None}) == geom.sub({"even": None})
```
this would then also work for other classes and such characters in class names are not allowed (`~`, `&`, `|` could be implemented). I see your point about `True/False`, but it just feels disconnected from the rest of the `kw` method and purely from this it seems dubious to allow it, in fact it would probably be better to disallow booleans for categories without arguments since it could confuse?  
Also, the possibilities with `Category` allows users to extend to their liking (which is what I highly encourage!!), and as such I think the most simple and consistent interface would be the best approach. A user may wish to create categories which have "proper" defaults, and simply use these if `geom.sub({"user_custom":None})`, that would allow users to change defaults in _one_ place, and not worry about other places of their code.

@jonaslb your two cents would be appreciated here ;)

sisl/_category.py
zerothi: > Yes, maybe it will look a bit ugly if you start to register lots of categories. Still, maybe it's good to keep the distinction even if we're not using it in this case for quick type checking (in sisl or user code)?

Hmm, perhaps the generic is a good idea in any case. Let me play a bit with `__instancecheck__` ;)

> PS: Does it make sense to do type checking in the category itself to raise a meaningful error or is it a waste of time that will decrease performance?
Bad idea, and yeah, for large systems this would be a performance hit...

sisl/_category.py
pfebrer96: Ok, love the idea of being able to use "~" to negate any category! :)

The `|` and `&` could work something like:
```python
geom.sub({"Z|even": (6, None)})
```
(you pass a tuple with an item for each category)

zerothi: hmm.. no not the last! ;)

I think this requires some more thought... Perhaps lets start with `~` :+1: 
Perhaps there is a way to easier build such a dictionary. The nice thing about Py3 is that kwargs is an OrderedDict!!!! :)

pfebrer96: > The nice thing about Py3 is that kwargs is an OrderedDict!!!!

But the dict you pass to `geom.sub` isn't, so can you be sure that order is kept?

sisl/_category.py
zerothi: Sadly, that won't work... hmm...

sisl/_category.py
zerothi: > > The nice thing about Py3 is that kwargs is an OrderedDict!!!!
> 
> But the dict you pass to `geom.sub` isn't, so can you be sure that order is kept?

So true... *shit*...

zerothi: That would require users do `geom.sub(Category.kw(...))`...

pfebrer96: By the way, as the functionality of keyword arguments seems that it will be quite handy, would it make sense that the `AtomCategory` class is itself callable and doing `AtomCategory(Z=6)` is the same as doing `AtomCategory.kw(Z=6)`? 

Basically we just would have to add a `__new__` method to `Category`:

```python
def __new__(cls, *args, **kwargs):
    try:
        return super().__new__(cls)
    except TypeError:
        # The class can not be instantiated because it has abstract methods (it is a parent class)
        return cls.kw(**kwargs)
```

Makes sense to me since it is never going to get instantiated anyway because it has abstract methods. Also, it will be more beginner friendly :) (probably a begginer won't understand why the `kw` method is called `kw`).

pfebrer96: Oh, and another problem, if you use `geom.sub(Category.kw(...))` to keep the order, you can not use characters such as `~`, `|`, etc in the names because that would be a syntax error.

sisl/_category.py
zerothi: Having played a bit around with this. Yes, I think we should have the generic class. So keep it as is!

sisl/_category.py
jonaslb: Dictionaries are ordered since Python 3.6 in implementation and since 3.7 in the language standard, I remember reading about this for both releases. So as such that shouldn't be a problem I suppose.

However I don't think it's worth doing something that would essentially require us to parse python-like syntax in strings, because then I think people should/would just be writing the code instead of learning a new pseudo-language. I also prefer typing `dict(Z=6, neighbors=2)` when using several keys over the curly-braces-and-strings way, and as pfebrer has already pointed out that wouldn't work with operators in the dict keys.

But as for Categories that don't take arguments -- maybe we just shouldn't have them? I kind of consider them comparison operators, so I think there should be something to compare with. The example with "even"/"odd" probably isn't very useful, but even if it was, I think a `IdxModCategory` would be more appropriate (or just using `np.arange(1, geom.na, 2)`...). Or even if we *really really* wanted an `IdxIsEvenCategory`, it could just take a boolean as it's argument, right? I guess this will be effectively approximately the same as is already done here but maybe there comes a category that wants to do custom things with a boolean argument.

zerothi: > However I don't think it's worth doing something that would essentially require us to parse python-like syntax in strings, because then I think people should/would just be writing the code instead of learning a new pseudo-language. I also prefer typing `dict(Z=6, neighbors=2)` when using several keys over the curly-braces-and-strings way, and as pfebrer has already pointed out that wouldn't work with operators in the dict keys.

Agreed, thanks for bringing back the roots. We shouldn't clutter the code with stuff that could already be solved (unless it does it truly elegantly and intuitively).
 
> But as for Categories that don't take arguments -- maybe we just shouldn't have them? I kind of consider them comparison operators, so I think there should be something to compare with. The example with "even"/"odd" probably isn't very useful, but even if it was, I think a `IdxModCategory` would be more appropriate (or just using `np.arange(1, geom.na, 2)`...). Or even if we _really really_ wanted an `IdxIsEvenCategory`, it could just take a boolean as it's argument, right? I guess this will be effectively approximately the same as is already done here but maybe there comes a category that wants to do custom things with a boolean argument.

Hmm, I think it is fine to have them, but you are probably correct that such cases that take no arguments could easily be managed in other ways.

@pfebrer96 :
> By the way, as the functionality of keyword arguments seems that it will be quite handy, would it make sense that the `AtomCategory` class is itself callable and doing `AtomCategory(Z=6)` is the same as doing `AtomCategory.kw(Z=6)`?
> 
> Basically we just would have to add a `__new__` method to `Category`:
> 
> ```python
> def __new__(cls, *args, **kwargs):
>     try:
>         return super().__new__(cls)
>     except TypeError:
>         # The class can not be instantiated because it has abstract methods (it is a parent class)
>         return cls.kw(**kwargs)
> ```

I don't think this would work, at least what I tried didn't make it run. I think also there is no reason for this complexity.
If the `kw` is a bad method name (which I agree it is), we should change it so it is clear. Like `Category.build` or something?

> Makes sense to me since it is never going to get instantiated anyway because it has abstract methods. Also, it will be more beginner friendly :) (probably a begginer won't understand why the `kw` method is called `kw`).

Hmm, actually, they currently don't... Neither `Category` nor `AtomCategory` are abstract classes, and even if they were, I don't think it would change it.
Also, with `__new__` you can't change `*args` and `**kwargs` when they are passed to `__init__`, so basically what you would do is:
```python
type(AtomCategory.kw(Z=5, neighbours=3)).__init__(Z=5, neighbours=3)
```
which obviously fails.

I agree with Jonas, the extra stuff in `.kw` should be removed. even/odd could be realized in other ways (I will add a new issue for this). No need to add `~` checks in the code for keys, if users want special stuff, they should do it specifically.

pfebrer96: I agree with Jonas too! It's probably much better to let python parse it than do it ourselves and have to maintain it, but then I find it even more useful that you can initialize categories directly by calling `Category` (or `AtomCategory`), because this:

```python
from sisl import AtomCategory

geom = ...
cat = AtomCategory(Z=6, neighbours=2) | ~AtomCategory(Z=6)
geom.sub(cat)
```

is probably better than

```python
from sisl import AtomCategory

geom = ...
cat = AtomCategory.build(Z=6, neighbours=2) | ~AtomCategory.build(Z=6)
geom.sub(cat)
```

and:

```python
from sisl.geom.category import AtomZ, AtomNeighbours

geom = ...
cat = (AtomZ(6) & AtomNeighbours(2)) | ~AtomZ(6)
geom.sub(cat)
```

(this one is clear, but the problem is that you need to import each category that you want to use). Between the 1st and the 2nd, I like the name `build` that you proposed, but for me it seems reasonable to expect that you are building a category also by calling `AtomCategory` directly.

zerothi: > I agree with Jonas too! It's probably much better to let python parse it than do it ourselves and have to maintain it, but then I find it even more useful that you can initialize categories directly by calling `Category` (or `AtomCategory`), because this:
> 
> ```python
> from sisl import AtomCategory
> 
> geom = ...
> cat = AtomCategory(Z=6, neighbours=2) | ~AtomCategory(Z=6)
> geom.sub(cat)
> ```
> 
> is probably better than
> 
> ```python
> from sisl import AtomCategory
> 
> geom = ...
> cat = AtomCategory.build(Z=6, neighbours=2) | ~AtomCategory.build(Z=6)
> geom.sub(cat)
> ```
> 
> and:
> 
> ```python
> from sisl.geom.category import AtomZ, AtomNeighbours
> 
> geom = ...
> cat = (AtomZ(6) & AtomNeighbours(2)) | ~AtomZ(6)
> geom.sub(cat)
> ```
> 
> (this one is clear, but the problem is that you need to import each category that you want to use). Between the 1st and the 2nd, I like the name `build` that you proposed, but for me it seems reasonable to expect that you are building a category also by calling `AtomCategory` directly.

I agree that it would be better, but it requires a metaclass that changes the `__call__` method to divert the `*args` and `**kwargs` for correct handling. If you provide a PR for this, I can't see why it wouldn't go in but it requires the metaclass `__call__` does the normal thing if it ain't the `AtomCategory` class calling it.

Again, I think Jonas is correct that this is fine-tuning that isn't necessary at its currents state. If you wish to discuss this further it should go in another issue, it doesn't belong to this PR, and certainly would require a new PR.  
We are really talking about minor details that are convenient but would again be a little bit unuseful since the direct classes allows more clear usage patterns.

Lets focus on this PR, and do a new one to keep focus...

pfebrer96: > the extra stuff in .kw should be removed

The part about calling argument-less categories too? I just realised that you could do that anyway by passing an empty dictionary (e.g. `geom.sub({"even": {}})` :sweat_smile:

zerothi: > I don't understand why you say that `Category` is not an abstract class. It has `ABCmeta` as its metaclass
> 
> https://github.com/zerothi/sisl/blob/69abded3a651df6f36aba716c9a4e7372600d54f/sisl/_category.py#L66
> 
> and it has abstract methods, doesn't it?

The metaclass doesn't expose any abstract methods, so the `TypeError` will never happen.

You would have to do:
```python
def __new__(cls, *args, **kwargs):
    if cls == AtomCategory:
          return cls.kw(*args, **kwargs)
    return super().__new__(cls)
```
I think, feel free to open a new PR if it works! ;)

Actually coming to think of it, this is a good idea since we don't want instances of `AtomCategory/Category`!

zerothi: > > the extra stuff in .kw should be removed
> 
> The part about calling argument-less categories too? I just realised that you could do that anyway by passing an empty dictionary (e.g. `geom.sub({"even": {}})`)

Yeah, do it like that, it is not pretty, but works, so retain old behaviour.

pfebrer96: > The metaclass doesn't expose any abstract methods, so the TypeError will never happen.

Why is `categorize` not exposed? Is it overwritten somewhere else?

> If you provide a PR for this, I can't see why it wouldn't go in but it requires the metaclass __call__ does the normal thing if it ain't the AtomCategory class calling it.

I was just trying this :) So I will probably submit it after this is merged.

zerothi: Ah, yeah, I had totally overlooked this. Please open a minimal PR, I have some suggestions, lets see if you do it like I thought (for generic usage) ;)

pfebrer96: > Why is categorize not exposed? Is it overwritten somewhere else?

Can I know the answer to this? It's bothering me that I don't understand it haha

zerothi: > > Why is categorize not exposed? Is it overwritten somewhere else?
> 
> Can I know the answer to this? It's bothering me that I don't understand it haha

I was wrong, that was my point in https://github.com/zerothi/sisl/pull/253#discussion_r470536195
You were correct, `categorize` was still an `abstractmethod` hence the `TypeError` will be raised.

pfebrer96: Ah ok, because of some change that I did while trying things, `AtomCategorize()` didn't raise an error so that's why I was very confused now and thought that you were actually right, sorry for the misunderstanding :)

pfebrer96:
All good to go! (only the bug fix remains).

By the way, I don't know if you wanted to register the `GenericCategory` differently than I did.

zerothi:
Thanks!

PR #255, opened 2020-08-14
Make parent categories callable and avoid their instantiation

pfebrer96:
Following the discussion in https://github.com/zerothi/sisl/pull/253, this makes `AtomCategory(Z=6)` equivalent to `AtomCategory.kw(Z=6)`.

### DIFF
--- a/sisl/_category.py
+++ b/sisl/_category.py
@@ -61,14 +61,15 @@ class InstanceCache:
         object.__setattr__(self, name, attr)
         return attr
 
+
 class CategoryMeta(ABCMeta):
 
     def __call__(cls, *args, **kwargs):
-
         try:
             return super().__call__(*args, **kwargs)
         except TypeError:
             return cls.kw(**kwargs)
+
 
 @set_module("sisl.category")
 class Category(metaclass=CategoryMeta):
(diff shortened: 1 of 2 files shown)


zerothi:
Let me add some comments that is mainly for my own remembrance

One would expect the following to work:
```python
class Category(...):
    def __new__(cls, *args, **kwargs):
        try:
            return super().__new__(cls)
        except TypeError:
            cat = cls.kw(**kwargs)
            return cat
```
As the standard says [here](https://docs.python.org/3/reference/datamodel.html#object.__new__):

> If __new__() does not return an instance of cls, then the new instance’s __init__() method will not be invoked.

So one would expect `__init__` not to be called in the above.  
The problem with the above is that `isinstance(cat, cls) is True` and thus we will never be able to succeed this (unless it returned a `CompositeCategory` (which then isn't a subclass of `AtomCategory`) because `cat` is an instance of a subclass.

So it all depends on the inheritance order and how they are called.

pfebrer96:
Wow, now I understand why it worked when I did `AtomCategory(Z=6, neighbours=2)` and not when I did `AtomCategory(neighbours=2)`. This will be very useful to know, thanks!

pfebrer96:
Done!

PR #256, opened 2020-08-17
enh: added fractional and coordinate categories

zerothi:

@pfebrer96 @jonaslb @sofiasanz @tfrederiksen please check out this new categories stuff for site categories.

It enables one to search for specific sites via coordinates or via fractional coordinates.

Here are two tests:
```python
def test_geom_category_frac_site():
    hBN_gr = bilayer(1.42, Atom[5, 7], Atom[6]) * (4, 5, 1)
    mid_layer = np.average(hBN_gr.xyz[:, 2])

    A_site = AtomFracSite(graphene())
    bottom = AtomCoordinate(z_lt=mid_layer, z=(-10, mid_layer))
    top = ~bottom

    bottom_A = A_site & bottom
    top_A = A_site & top

    cat = (bottom_A | top_A).categorize(hBN_gr)
    str(cat)
    cnull = 0
    for i, c in enumerate(cat):
        if c == NullCategory():
            cnull += 1
        elif hBN_gr.xyz[i, 2] < mid_layer:
            assert c == bottom_A
        else:
            assert False
    assert cnull == len(hBN_gr) // 4 * 3

def test_geom_category_frac_A_B_site():
    gr = graphene() * (4, 5, 1)

    A_site = AtomFracSite(graphene())
    B_site = AtomFracSite(graphene(), foffset=(-1/3, -1/3, 0))

    cat = (A_site | B_site).categorize(gr)
    for i, c in enumerate(cat):
        if i % 2 == 0:
            assert c == A_site
        else:
            assert c == B_site
```
This allows one to distinguish between different sites.

Please let me know if something could be changed! :)

### DIFF
--- a/sisl/_category.py
+++ b/sisl/_category.py
@@ -10,6 +10,9 @@ __all__ += ["AndCategory", "OrCategory", "XOrCategory"]
 __all__ += ["InstanceCache"]
 
 
+_list_types = (tuple, list, np.ndarray)
+
+
 class InstanceCache:
     """ Wraps an instance to cache *all* results based on `functools.lru_cache`
 
@@ -205,7 +208,7 @@ class Category(metaclass=CategoryMeta):
 
     def __ne__(self, other):
         eq = self == other
-        if isinstance(eq, list):
+        if isinstance(eq, _list_types):
             return [not e for e in eq]
         return not eq
 
@@ -281,12 +284,13 @@ class NotCategory(GenericCategory):
         r""" Base method for queriyng whether an object is a certain category """
         cat = self._cat.categorize(*args, **kwargs)
 
+        _null = NullCategory()
         def check(cat):
             if isinstance(cat, NullCategory):
                 return self
-            return NullCategory()
+            return _null
 
-        if isinstance(cat, list):
+        if isinstance(cat, _list_types):
             return list(map(check, cat))
         return check(cat)
 
@@ -378,7 +382,7 @@ class OrCategory(CompositeCategory):
                 return b
             return a
 
-        if isinstance(catA, list):
+        if isinstance(catA, _list_types):
             return list(map(cmp, catA, catB))
         return cmp(catA, catB)
 
@@ -412,7 +416,7 @@ class AndCategory(CompositeCategory):
                 return b
             return self
 
-        if isinstance(catA, list):
+        if isinstance(catA, _list_types):
             return list(map(cmp, catA, catB))
         return cmp(catA, catB)
 
@@ -448,7 +452,7 @@ class XOrCategory(CompositeCategory):
             # is exclusive, so we return the NullCategory
             return NullCategory()
 
-        if isinstance(catA, list):
+        if isinstance(catA, _list_types):
             return list(map(cmp, catA, catB))
         return cmp(catA, catB)
 
--- a/sisl/_category.py
+++ b/sisl/_category.py
@@ -116,6 +116,32 @@ class Category(metaclass=CategoryMeta):
         self._name = name
 
     @classmethod
+    @abstractmethod
+    def is_class(cls, name):
+        r""" Query whether `name` matches the class name by removing a prefix `kw`
+
+        This is important to ensure that users match the full class name
+        by omitting the prefix returned from this method.
+
+        This is an abstract method to ensure sub-classes of `Category`
+        implements it.
+
+        For instance:
+
+        .. code::
+
+            class MyCategory(Category):
+
+                @classmethod
+                def is_class(cls, name):
+                     # strip "My" and do comparison
+                     return cl.__name__.lower()[2:] == name.lower()
+
+        would enable one to compare against the *base* category scheme.
+        """
+        pass
+
+    @classmethod
     def kw(cls, **kwargs):
         """ Create categories based on keywords
 
@@ -152,7 +178,7 @@ class Category(metaclass=CategoryMeta):
             lkey = key.lower()
             found = ''
             for name, cl in subcls.items():
-                if name.endswith(lkey):
+                if cl.is_class(name):
                     if found:
                         raise ValueError(f"{cls.__name__}.kw got a non-unique argument for category name:\n"
                                          f"    Searching for {name} and found matches {found} and {name}.")
@@ -233,10 +259,14 @@ class GenericCategory(Category):
     """Used to indicate that the category does not act on specific objects
 
     It serves to identify categories such as `NullCategory`, `NotCategory`
-    and composite categories and distinguish them from categories that have
+    and `CompositeCategory` and distinguish them from categories that have
     a specific object in which they act.
     """
-    pass
+    @classmethod
+    def is_class(cls, name):
+        # never allow one to match a generic class
+        # I.e. you can't instantiate a Null/Not/And/Or/XOr category by name
+        return False
 
 
 @set_module("sisl.category")
--- a/sisl/_category.py
+++ b/sisl/_category.py
@@ -178,7 +178,7 @@ class Category(metaclass=CategoryMeta):
             lkey = key.lower()
             found = ''
             for name, cl in subcls.items():
-                if cl.is_class(name):
+                if cl.is_class(key):
                     if found:
                         raise ValueError(f"{cls.__name__}.kw got a non-unique argument for category name:\n"
                                          f"    Searching for {name} and found matches {found} and {name}.")
--- a/sisl/geom/category/__init__.py
+++ b/sisl/geom/category/__init__.py
@@ -1,3 +1,5 @@
 from .base import *
 from ._neighbours import *
 from ._kind import *
+from ._coord import *
+
--- a/sisl/geom/category/__init__.py
+++ b/sisl/geom/category/__init__.py
@@ -2,4 +2,3 @@ from .base import *
 from ._neighbours import *
 from ._kind import *
 from ._coord import *
-
--- a/sisl/geom/category/_coord.py
+++ b/sisl/geom/category/_coord.py
@@ -1,0 +1,231 @@
+from functools import partial, wraps
+import operator
+from numbers import Integral
+
+import numpy as np
+from numpy import dot, fabs, where
+
+from sisl._internal import set_module, singledispatchmethod
+from sisl._help import isiterable
+from sisl.utils.misc import direction
+from sisl.shape import *
+from sisl.supercell import SuperCell, SuperCellChild
+from sisl._supercell import cell_invert
+import sisl._array as _a
+from .base import AtomCategory, NullCategory, _sanitize_loop
+
+
+__all__ = ["AtomFracSite", "AtomCoordinate"]
+
+
+@set_module("sisl.geom")
+class AtomFracSite(AtomCategory):
+    r""" Classify atoms based on fractional sites for a given supercell
+
+    Match atomic coordinates based on the fractional positions.
+
+    Parameters
+    ----------
+    sc : SuperCell, SuperCellChild or argument to SuperCell
+       an object that defines the lattice vectors (will be passed through to `SuperCell`
+       if not an object instance of `SuperCell` or `SuperCellChild`
+    atol : float, optional
+       the absolute tolerance (in Ang) to check whether the site is an integer
+       site.
+    offset : array_like, optional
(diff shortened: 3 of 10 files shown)


pfebrer96:
Hi Nick!

Regarding the coordinates, I think it would be useful for the end users that `Geometry` had `x`, `y` and `z` properties such as:

```python
@property
def x(self):
    return self.xyz[:,0]
```

and same for the fractional coordinates. Because for cases where you want to just filter according to a coordinate it is much simpler an intuitive to do:

```python
geom.sub(geom.x < 5)
```

than

```python
geom.sub({"coordinate": {"x_lt": 5}})
```

Also, I was thinking in ways to combine a boolean array with sisl categories, e.g.:

```python
geom.x < 5 | AtomCategory(neighbours=5)
```

but I don't know how would this work since in this case numpy gets total control of the operation.

zerothi:
Thanks for taking a look!

> Regarding the coordinates, I think it would be useful for the end users that `Geometry` had `x`, `y` and `z` properties such as:
> 
> ```python
> @property
> def x(self):
>     return self.xyz[:,0]
> ```
> 
> and same for the fractional coordinates. Because for cases where you want to just filter according to a coordinate it is much simpler an intuitive to do:
> 
> ```python
> geom.sub(geom.x < 5)
> ```
> 
> than
> 
> ```python
> geom.sub({"coordinate": {"x_lt": 5}})
> ```

Well, `geom.sub(geom.xyz[:, 0] < 5)` I think is just fine (and already functioning)?

And `cat.categorize(geom, (geom.xyz[:, 0] < 5).nonzero()[0])` should already work?

I don't think we should add `.x`, `.fx` etc. to the property list. I am already struggling with too many methods for the `Geometry` class, and I think cluttering it with even more isn't helpful enough. :)

> 
> Also, I was thinking in ways to combine a boolean array with sisl categories, e.g.:
> 
> ```python
> geom.x < 5 | AtomCategory(neighbours=5)
> ```
> 
> but I don't know how would this work since in this case numpy gets total control of the operation.

Yeah, this would require us to implement the mixin's from ufuncs into the `Category` class. this could be added in the future, but for now not necessary. Again, as Jonas mentioned, categories are probably already "expert" use and thus the slight overhead of coding something more explicit isn't really a pain ;)
I agree that the `key` interface for having the specification isn't really optimal, so any idea to improve ease this would be nice...

pfebrer96:
> Well, geom.sub(geom.xyz[:, 0] < 5) I think is just fine (and already functioning)?

Yes, it's just that it took me a while to realize that I could do that when I was beggining to use python/sisl :sweat_smile:, and honestly it bothers me that I need to write the `xyz[:, 0]` part each time, but maybe it's just that I'm too lazy :)

> I agree that the key interface for having the specification isn't really optimal, so any idea to improve ease this would be nice...

Why not only accepting intervals? I.e. instead of `AtomCoordinates(x_lt=5)` one should do `AtomCoordinates(x=(None, 5))`.

Also, have you thought about creating one category for each so that they can be easily accesed from the `atoms` argument? E.g.:

```python
class AtomX(AtomCategory):
    def __new__(cls, interval):
        return AtomCoordinate(x=interval)
```
this allows a user to do:

```python
geom.sub({"x": (2, 3)})
# instead of
geom.sub({"coordinate": {"x": (2,3)}})
```

and so on for fractional and absolute coordinates (these would of course be created dinamically, not individually declared). Furthermore, we could make them all use a metaclass that implements `__lt__` and `__gt__`, so that

```python
AtomX < 3 == AtomX((None, 3))
```

(I liked Jonas' idea about not parsing strings in general and use python syntax as much as possible)

Maybe these categories could also be found as attributes of the `AtomCoordinate` class so that you wouldn't need to import each of them. That is:

```python
AtomCoordinate.fx > 0.4 == AtomFx((0.4, None))
```

Also, it confused me that you somehow kind of implemented the `AndCategory` functionality inside `AtomCoordinate`. Is it necessary? I mean, what's the main difference between letting `AndCategory` handle it and handling it inside of `AtomCoordinate`?

Would there be a significant difference in performance if one does this:
```python
AtomCoordinate.fx > 0.4 & AtomCoordinate.x < 10
# or
AtomCategory(fx=(0.4, None), x=(None, 10))
```

instead of this:

```python
AtomCoordinate(fx_gt=0.4, x_lt=10)
```

That's enough, I hope you don't mind all the proposals, maybe one of them will actually make sense :)

zerothi:
> > Well, geom.sub(geom.xyz[:, 0] < 5) I think is just fine (and already functioning)?
> 
> Yes, it's just that it took me a while to realize that I could do that when I was beggining to use python/sisl , and honestly it bothers me that I need to write the `xyz[:, 0]` part each time, but maybe it's just that I'm too lazy :)

I think you are being too lazy ;)

There are _many_ things one could shorten, but it would blow up the code unnecessarily.

> > I agree that the key interface for having the specification isn't really optimal, so any idea to improve ease this would be nice...
> 
> Why not only accepting intervals? I.e. instead of `AtomCoordinates(x_lt=5)` one should do `AtomCoordinates(x=(None, 5))`.

It already accepts intervals, `x=(-2, 4)` would be an interval :) I have now amended your `None` suggestion, good!
> 
> Also, have you thought about creating one category for each so that they can be easily accesed from the `atoms` argument? E.g.:
> 
> ```python
> class AtomX(AtomCategory):
>     def __new__(cls, interval):
>         return AtomCoordinate(x=interval)
> ```
> 
> this allows a user to do:
> 
> ```python
> geom.sub({"x": (2, 3)})
> # instead of
> geom.sub({"coordinate": {"x": (2,3)}})
> ```
> 
> and so on for fractional and absolute coordinates (these would of course be created dinamically, not individually declared). Furthermore, we could make them all use a metaclass that implements `__lt__` and `__gt__`, so that
> 
> ```python
> AtomX < 3 == AtomX((None, 3))
> ```
> 
> (I liked Jonas' idea about not parsing strings in general and use python syntax as much as possible)

I have now changed the name of the class to `AtomXYZ`, shorter and still with same intent.
I haven't carefully thought through the above, but that would limit the `__eq__` function for the objects. I.e. by using `<` operators to create new ones we would not be able to compare different categories. I don't know if that would be meaningful at all...

However, to me it would be quite strange that `cat == cat` is an instance method, while `AtomX < 3` is a class method? It is rather un-intuitive to me?

> Maybe these categories could also be found as attributes of the `AtomCoordinate` class so that you wouldn't need to import each of them. That is:
> 
> ```python
> AtomCoordinate.fx > 0.4 == AtomFx((0.4, None))
> ```

This however, could be a good idea. :)

> 
> Also, it confused me that you somehow kind of implemented the `AndCategory` functionality inside `AtomCoordinate`. Is it necessary? I mean, what's the main difference between letting `AndCategory` handle it and handling it inside of `AtomCoordinate`?

There is no difference ;)
It is just easier to do ranges:
```python
AtomXYZ(x=(3, 6), y=(-2, 5))
```
for instance.

> Would there be a significant difference in performance if one does this:
> 
> ```python
> AtomCoordinate.fx > 0.4 & AtomCoordinate.x < 10
> # or
> AtomCategory(fx=(0.4, None), x=(None, 10))
> ```
> 
> instead of this:
> 
> ```python
> AtomCoordinate(fx_gt=0.4, x_lt=10)
> ```

Yes, there would be a performance penalty, the internal `and` uses `numpy` while the other one uses a `zip` loop.

> That's enough, I hope you don't mind all the proposals, maybe one of them will actually make sense :)

Thanks!
I changed name to `AtomXYZ` and allowed `None` as bounds in the range construct.

zerothi:
hmm, actually changing the name to `AtomXYZ` was a bad idea... `AtomCategory(Z=` will not know whether to use `AtomZ` or `AtomXYZ`!!!

What to do? Perhaps each class should have a specifier that is queried?

pfebrer96:
> hmm, actually changing the name to AtomXYZ was a bad idea... AtomCategory(Z= will not know whether to use AtomZ or AtomXYZ!!!
> 
> What to do? Perhaps each class should have a specifier that is queried?

Yes, I think that this is a problem of how the classes are retrieved in `Category.kw`. I thought this when I first saw the implementation.

Probably the subclasses should be registered when they are created so that `AtomCategory` is aware of it and it does not need to find all the subclasses each time. 

Otherwise we could use a more restrictive condition to see if the class name meets the requirement. I.e. instead of:

```python
name.endswith(lkey):
```

something like

```python
# Get the identifier of the class, which will be "Atom" for AtomCategory
cls_id = cls.__name__.replace("Category", "")
# And then inside the loop just remove it from the subclass name
name.replace(cls_id, "") == lkey
```

> I have now changed the name of the class to AtomXYZ, shorter and still with same intent.

My intent with the multiple categories was more to remove one level of nesting in the `atoms` argument, not that the name was shorter :) But still is probably a better name.

> I haven't carefully thought through the above, but that would limit the __eq__ function for the objects. I.e. by using < operators to create new ones we would not be able to compare different categories. I don't know if that would be meaningful at all...
> 
> However, to me it would be quite strange that cat == cat is an instance method, while AtomX < 3 is a class method? It is rather un-intuitive to me?

I don't understand, because what I propose is that the metaclass implements `__lt__` and `__gt__` for the class object. These would not be class methods, therefore they won't be available from the instance. E.g.:

```python
class CoordMeta(type):
    def __lt__(self, other):
        return self((None, other))

class AtomX(metaclass=CoordMeta, AtomCategory):
    def __new__(cls, interval):
        return AtomCoordinate(x=interval)

# enables:
AtomX < 5
# But does not do anything when you do
AtomX((3, None)) < 5
# which would raise an error '<' not supported between instances of 'AtomX' and 'int'
```

zerothi:
> > hmm, actually changing the name to AtomXYZ was a bad idea... AtomCategory(Z= will not know whether to use AtomZ or AtomXYZ!!!
> > What to do? Perhaps each class should have a specifier that is queried?
> 
> Yes, I think that this is a problem of how the classes are retrieved in `Category.kw`. I thought this when I first saw the implementation.
> 
> Probably the subclasses should be registered when they are created so that `AtomCategory` is aware of it and it does not need to find all the subclasses each time.

Yeah, this just requires another meta-class scheme to handle this, besides, I think the gains are extremely small given that we'll probably never go beyond 200 classes, and since these are created one time only, it shouldn't really matter (performance wise).

What we probably should do is to have a class-method that needs to be overwritten for `Category`, `AtomCategory` etc. 
I'll amend to this one since it is required.
I'll also disable `Category.kw` such that it should be a subclass.

> 
> Otherwise we could use a more restrictive condition to see if the class name meets the requirement. I.e. instead of:
> 
> ```python
> name.endswith(lkey):
> ```
> 
> something like
> 
> ```python
> # Get the identifier of the class, which will be "Atom" for AtomCategory
> cls_id = cls.__name__.replace("Category", "")
> # And then inside the loop just remove it from the subclass name
> name.replace(cls_id, "") == lkey
> ```
> 
> > I have now changed the name of the class to AtomXYZ, shorter and still with same intent.
> 
> My intent with the multiple categories was more to remove one level of nesting in the `atoms` argument, not that the name was shorter :) But still is probably a better name.
> 
> > I haven't carefully thought through the above, but that would limit the **eq** function for the objects. I.e. by using < operators to create new ones we would not be able to compare different categories. I don't know if that would be meaningful at all...
> > However, to me it would be quite strange that cat == cat is an instance method, while AtomX < 3 is a class method? It is rather un-intuitive to me?
> 
> I don't understand, because what I propose is that the metaclass implements `__lt__` and `__gt__` for the class object. These would not be class methods, therefore they won't be available from the instance. E.g.:
> 
> ```python
> class CoordMeta(type):
>     def __lt__(self, other):
>         return self((None, other))
> 
> class AtomX(metaclass=CoordMeta, AtomCategory):
>     def __new__(cls, interval):
>         return AtomCoordinate(x=interval)
> 
> # enables:
> AtomX < 5
> # But does not do anything when you do
> AtomX((3, None)) < 5
> # which would raise an error '<' not supported between instances of 'AtomX' and 'int'
> ```
> 
> Therefore implementing the comparison of categories can still be done without problem.

Good point, I agree this would be a good idea. Feel free to make a PR for this ;)

zerothi:
@pfebrer96 any final comments before this goes in? :)

pfebrer96:
All good for me! :+1:

zerothi:
Merged!

PR #257, opened 2020-08-18
RealspaceSI: Using unfold=(1,1,1) no longer overwrites nsc on Hsurface

jonaslb:
As the title says, a surface hamiltonian would effectively be destroyed in `RealSpaceSI` when using `unfold=(1,1,1)`, because no copy-operations were performed before setting nsc in a few methods in that case. This is probably a kind of unusual value to use for `unfold`, so hopefully noone was affected. `RealSpaceSE` is not affected because it always uses a tile operation in the semi-infinite direction in its corresponding methods.

- [ ❌ ] Closes #xxxx
 - [ ❌ ] Tests added
 - [ ❌ ] Documentation for functionality
 - [ ✔️ ] User visible changes (including notable bug fixes) are documented in `CHANGELOG`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,9 @@
 - Ensured ghost atoms in Siesta are handled with separate
 	class, AtomGhost, #249
 
+- Using `si.RealspaceSI` with `unfold=(1,1,1)` no longer results in `nsc` on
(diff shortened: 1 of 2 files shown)


zerothi:
Thanks for finding this!

PR #258, opened 2020-08-19
Making coordinate categories more accessible

pfebrer96:
As discussed in https://github.com/zerothi/sisl/pull/256, one category is created for each "coordinate direction". This allows one to:
- Use coordinate categories without needing to nest them:

    ```python
    geom.sub({"x": (3, None)})
    # instead of
    geom.sub({"xyz": {"x": (3, None)}})
    ```
- Create your categories using python syntax:

    ```python
    geom.sub(AtomXYZ.fz < 0.5)
    ```

Let me know how does it look to you!

 - [ ] Tests added


sisl/geom/category/_coord.py
```diff
@@ -232,3 +233,55 @@ def __eq__(self, other):
...
                 return True
         return False
+
+# The following code is just to make the use of coordinate categories more convenient
+# We will create one individual category for each direction. In this way, one can access
+# them directly in the keyword category builder. E.g.: `AtomCategory(x=(1,2))` creates `AtomXYZ(x=(1,2))`
+# without needing to nest "x", as in `AtomCategory(xyz={"x": (1,2)})`.
+
+
+class CoordMeta(CategoryMeta):
```
zerothi, line 243: To conform to the class name, name this `AtomXYZMeta`, I guess this meta is only applicable to this class, so explicitness is good! :+1:

sisl/geom/category/_coord.py
```diff
@@ -232,3 +233,55 @@ def __eq__(self, other):
...
+    If `AtomX` has been built using this metaclass:
+    >>> AtomX < 5 == AtomX((None, 5)) == AtomXYZ(x=(None, 5))
+    >>> AtomX > 5 == AtomX((5, None)) == AtomXYZ(x=(5, None))
+    """
+
+    def __lt__(self, other):
+        return self(None, other)
+
+    def __gt__(self, other):
+        return self(other, None)
```
zerothi, line 262: also `__le__` and `__ge__`, in fact I think this is what is being done internally anyways

sisl/geom/category/_coord.py
```diff
@@ -232,3 +233,55 @@ def __eq__(self, other):
...
+
+# Iterate over all directions that deserve an individual class
+for key in ("x", "y", "z", "f_x", "f_y", "f_z", "a_x", "a_y", "a_z"):
+
+    # The underscores are just kept for the key that is passed to AtomXYZ,
+    # but the name of the category will have no underscore
+    name = key.replace("_", "")
+
+    # Create the class for this direction
+    new_cls = CoordMeta(f"Atom{name}", (AtomCategory, ), {"__new__": _new, "_coord_key": key})
```
zerothi, line 281: for consistency I would prefer if we do `f"Atom{name.upper()}"`

sisl/geom/category/_coord.py
zerothi: Also, would `AtomX in (-2, 3)` be counter-intuitive? Would users assume X _has_ to be `-2` or `3`?

zerothi (changes requested):
Looks good!!!

sisl/geom/category/_coord.py
pfebrer96: Hmm good point. 

> Would users assume X has to be -2 or 3?

I think this would be the logical thing to believe knowing what `in` does in python

Since `AtomX(-2,3)` already achieves the goal, maybe it is unnecessary. But I'll add it if you want. This is the reason I have also not bothered to support ` -2 < AtomX < 3`.

sisl/geom/category/_coord.py
pfebrer96: Also thought about this, but then you would access them like `AtomXYZ.FX`, which looks uglier to me than `AtomXYZ.fx`. What do you prefer, uppercase in the class name and lowercase in the attribute or all uppercase?

sisl/geom/category/_coord.py
zerothi: Well, the problem with `in` is that `2 in [1, 3]` will be false, that was my concern... Ok, lets keep it out for now!

sisl/geom/category/_coord.py
zerothi: Class names uppercase, attributes lowercase.

sisl/geom/category/_coord.py
pfebrer96: Yes, I was agreeing with that here :)

> I think this would be the logical thing to believe knowing what `in` does in python

sisl/geom/category/_coord.py
pfebrer96: I changed the name, but this could be applicable to neighbours for example, right?

pfebrer96: Well, just the concept, probably not exactly this meta.

sisl/geom/category/_coord.py
pfebrer96: > also __le__ and __ge__, in fact I think this is what is being done internally anyways

We should probably differentiate somehow between `gt` and `ge`, no?

zerothi: > Yes, I was agreeing with that here :)
> 
> > I think this would be the logical thing to believe knowing what `in` does in python

Ok, it was just that that is not the python interpretation. But I think we agree that this *could* be useful, lets see if anybody wants it.


> We should probably differentiate somehow between `gt` and `ge`, no?

I agree. It would just require that `<direction>_gt` gets passed correctly and with expanded keywords.

zerothi: I have tried to play with `__contains__` I can't get it to work. I will push to this PR some changes.

zerothi:
Hmm, I think we have to change the class names, this `AtomZ` will clash with the class constructed here...

I don't think there is anyway around... `AtomXYZ[XYZ]` is probably the way to go?

zerothi:
> Hmm, I think we have to change the class names, this `AtomZ` will clash with the class constructed here...
> 
> I don't think there is anyway around... `AtomXYZ[XYZ]` is probably the way to go?

Or `AtomX_` so `x_=<>` works? hmm... Ideas?

sisl/geom/category/_coord.py
pfebrer96: > Ok, it was just that that is not the python interpretation.

I'm getting confused now, the python interpretation is that `2 in [1, 3]` is `False`, right? I was agreeing with this and therefore saying that it is probably not worth it to implement `AtomX in [1, 3]` for ranges, specially since one can already do `AtomX(1,3)`. Or do you mean that you want `AtomX in [1, 3]` to actually match only 1 and 3?

pfebrer96:
> Hmm, I think we have to change the class names, this AtomZ will clash with the class constructed here...
> 
> I don't think there is anyway around... AtomXYZ[XYZ] is probably the way to go?

Right :cold_sweat: 

I think the problem is still the way that categories are registered, because I'm sure everybody would understand that `Z` is for the atomic number and `z` is for the coordinate, so if one could register each class individually with its own name the problem would be solved. We wouldn't need a metaclass for this, just using `__init_subclass__` would be enough. The default name could be always the lowercase version of the class name (e.g. `AtomNeighbours` --> `neighbours`) and then we would just specify the name "Z" for the atomic number Z. The coordinate classes you could call them `AtomXYZz` but still their access name in kw could be "z".

sisl/geom/category/_coord.py
zerothi: > > Ok, it was just that that is not the python interpretation.
> 
> I'm getting confused now, the python interpretation is that `2 in [1, 3]` is `False`, right? I was agreeing with this and therefore saying that it is probably not worth it to implement `AtomX in [1, 3]` for ranges, specially since one can already do `AtomX(1,3)`. Or do you mean that you want `AtomX in [1, 3]` to actually match only 1 and 3?

Ok, then we agree. :+1:

zerothi:
> > Hmm, I think we have to change the class names, this AtomZ will clash with the class constructed here...
> > I don't think there is anyway around... AtomXYZ[XYZ] is probably the way to go?
> 
> Right 
> 
> I think the problem is still the way that categories are registered, because I'm sure everybody would understand that `Z` is for the atomic number and `z` is for the coordinate, so if one could register each class individually with its own name the problem would be solved. We wouldn't need a metaclass for this, just using `__init_subclass__` would be enough. The default name could be always the lowercase version of the class name (e.g. `AtomNeighbours` --> `neighbours`) and then we would just specify the name "Z" for the atomic number Z. The coordinate classes you could call them `AtomXYZz` but still their access name in kw could be "z".

Hmm, I would rather not go this direction. This calls for "corner" cases which are less clear. What about `AtomX_` so it would be `AtomCategory(x_=5)` not pretty, but sufficient? By the way, this is what the `operator` library does for built-ins.

zerothi:
@pfebrer96 please see the latest commits I did

pfebrer96:
> What about AtomX_ so it would be AtomCategory(x_=5) not pretty, but sufficient?

I don't know, I think people won't understand why they have to use `x_` and not `neighbours_` :(. Also I don't think anyone would expect that by doing `geom.sub({"z": ...})`, they are classifying by atomic number. This is the same reason why `geom.atoms.z` is not implemented and you have to do `geom.atoms.Z`. 

I agree that it is a bit more complex though, I don't know-

sisl/geom/category/_coord.py
```diff
@@ -295,15 +285,18 @@ def _new(cls, *interval, **kwargs):
...
             return f"{key}_{k}", v
-        new_kwargs = dict(map(_apply_key, *zip(*kwargs.items())))
+        if len(kwargs) > 0:
+            new_kwargs = dict(map(_apply_key, *zip(*kwargs.items())))
+        else:
+            new_kwargs = {}
 
         # Convert interval to correct interpretation
         if len(interval) > 1:
             raise ValueError(f"{cls.__name__} makes only sense to instantiate with 1 non-keyword argument")
```
pfebrer96, line 295: Actually, I was supporting `AtomX((1, 3))` because that's how you can call it from the kw builder (e.g.: `geom.sub({"x": (1,3)})`). You don't want to support this?

zerothi: I do want to support it, I can see I messed that up, I'll correct.

zerothi: fixed

pfebrer96: My mistake, I read it wrong :)  But I don't understand what do you expect `interval` to be. Because if `len(interval) == 1`, `ìnterval` is probably something like `((-2, 3), )`

zerothi: Yeah, the point is that one should not be able to call `AtomX(4.0, 3.0)` (this would explicitly fail), nor `AtomX(3.)` (this would fail in `AtomXYZ.__init__` due to ambiguity). It cannot be deferred. So only one interval is allowed.

zerothi: I expect `interval` to be a single tuple, `isinstance(interval[0], (list, tuple, np.ndarray))` and nothing more, otherwise I can't know what it means?

pfebrer96: > My mistake, I read it wrong :) But I don't understand what do you expect interval to be. Because if len(interval) == 1, ìnterval is probably something like ((-2, 3), )

Sorry, your comment didn't show up, that's why I commented this.

I used `*interval` so that when you did `AtomX(3, 4)`, `interval` was `(3,4)`, but if you don't want to support it (that's what I understand when you raise the error on `len(interval) > 1`), you can just have `interval=None` as the first argument and that will save us some checks right?

zerothi: No, that is a good idea, see commit, fixed.

zerothi:
I think it is there now. :)

Could you please have a look, `AtomXYZ.x` and `AtomCategory(z=(-1, 1), Z=5)` should now work!

pfebrer96:
Good! :)

One last question, what's the purpose of setting the `__module__`? I thought that you could then import the class from that module, but I can not do:

```python
from sisl.geom import Atomx
```

is it just for aesthethic reasons? (when you print the class or something)

zerothi:
> One last question, what's the purpose of setting the `__module__`? I thought that you could then import the class from that module, but I can not do:
> 
> ```python
> from sisl.geom import Atomx
> ```
> 
> is it just for aesthethic reasons? (when you print the class or something)

It is for aesthetic reasons, when errors are printed the full module path are given, in this case in turns out it will be an abc.Atomx class which doesn't make sense. 

There are still places where this hasn't been fixed, but in should be. :)

pfebrer96:
Great, then should we do

```python
new_cls = set_module("sisl.geom")(new_cls)
```

instead of setting the `__module__` attribute "directly"? Just so that if at some point you extend or change the functionality of `set_module` this is updated too.

zerothi:
Yes, good idea, feel free to fix this on the branch, then I'll merge tomorrow!

zerothi (approved):

PR #263, opened 2020-09-09
Supporting the creation of Binder docker images

pfebrer96:
This PR aims to ease the introduction of sisl to new users, as well as providing an interactive playground in the cloud for people to very easily interact with new features, tutorials, or demos without needing to install/update sisl and/or download the corresponding notebooks.

This is achieved by supporting the creation of [binder](https://jupyter.org/binder) docker images, which serve as an environment to allow people to run interactive sessions in the cloud.  https://mybinder.org/ is an implementation of [`BinderHub`](https://binderhub.readthedocs.io/en/latest/) that provides free cloud computing (I don't know why :sweat_smile:), but of course it has its limitations (memorywise). However, one can deploy their own instance of `BinderHub` to provide service for their users.

To support it, we just needed to specify that `gcc` and `gfortran` had to be installed. You can check that it works by clicking this link: https://mybinder.org/v2/gh/pfebrer96/sisl/binder. Then create a new notebook or open the tutorial ones.

In this same PR, all notebooks that are used in the docs now have a badge to launch them in binder (not working yet because they point to zerothi/sisl/master).

Some things that still need to be thought:

- [x] I believe binder uses `setup.py` and `requirements.txt` to install the required packages (I'm not sure if only `setup.py` is used). This results in the docker image only having the minimum requirements, i.e. not `matplotlib`, `xarray`, `tqdm`, etc... We should probably build the docker image with all the dependencies, right?
- [ ] MyBinder rebuilds the image automatically after a commit is pushed. However, this is only done when someone requests to use it. Since sisl takes a long time to install, this is not optimal because the user might have to wait a while before being able to access the notebooks. Maybe there is a way to turn off auto rebuilds and trigger them manually, but I haven't found it yet.
- [ ] I think there are smart ways to separate the environment (sisl package) from the content (notebooks) so that when a change is made in the notebooks it does not trigger a rebuild, but I have to look into it.


zerothi:
In my latest commit you'll see the fixes done for the binder stuff and special requirements.txt file placement.

I think the image should just point to a specific release, that would only tricker the build at the first invocation, all others should be a lot faster. However, this won't be the "latest" version.

We should probably do a few test releases to figure this out, i.e. play with `docs/conf.py` url's and figure out the best way.

zerothi:
However, I can't use mybinder at its current implementation? DId you succeed prior to my commit?

pfebrer96:
Yes, it worked before.

I also tried to place things in a specific binder folder, but the problem is that then it doesn't read the `setup.py` file, which is in the root folder. This is the main problem. That's why you can import numpy (which is in `requirements.txt`), but not `sisl`.

PS: Is `.binder` a valid name for the folder? I thought it needed to be `binder`.

zerothi:
Yeah, the documentation clearly states that the `.binder` and `binder` directories are both good :)

I prefer `.binder` to hide it from regular users. :)

pfebrer96:
> I think the image should just point to a specific release, that would only tricker the build at the first invocation, all others should be a lot faster. However, this won't be the "latest" version.

It's easy to create the badges to point to an specific release:

"https://mybinder.org/v2/gh/sisl/{{ env.config.release|e }}?filepath={{ docname|e }}"

but the problem is that it doesn't work when env.config.release is "latest" or "stable", which will be most of the time.

zerothi:
> > I think the image should just point to a specific release, that would only tricker the build at the first invocation, all others should be a lot faster. However, this won't be the "latest" version.
> 
> It's easy to create the badges to point to an specific release:
> 
> ```
> "https://mybinder.org/v2/gh/sisl/{{ env.config.release|e }}?filepath={{ docname|e }}"
> ```
> 
> but the problem is that it doesn't work when env.config.release is "latest" or "stable", which will be most of the time.

Exactly, but we should somehow back-track to the latest stable version. That is easy on my local box by using the git command, but may not be so easy in automated setups...

.binder/postBuild
```diff
@@ -1,8 +1,10 @@
...
 
 # Stop at errors
-set -e
+set -ex
 
 echo "### mybinder environment"
 env
 echo "### mybinder environment^^^"
+
+pip install .
```
pfebrer96, line 10: Nice :)

zerothi: hehe...

zerothi: I am currently waiting for binder to finish the installation, agreed it is very slow. I can't see from the environment why it should be slower than any other binder projects...

zerothi: now its working! :)

pfebrer96: I'm watching your build hahah, finished!

By the way, if we do `pip install .` then `requirements.txt` does not need to contain sisl's requirements right? Maybe it's better to remove them and keep there only the optional external ones, just to avoid inconsistencies and remove the need to update multiple files.

zerothi: Yeah, good idea, now changed.

zerothi:
I think I'll just merge this and then we can always later adapt. Thanks for the PR!!!

PR #267, opened 2020-09-18
bug: skewed isosurfaces were not properly calculated

pfebrer96:
Skewed isosurfaces were still not properly calculated. It was easier to do it good than to do it bad (see how simple the code is now). Sorry :)

Visual proof that it works now by overlapping a RHO file with its geometry:

With the old implementation:

newplot - 2020-09-18T152305 452

With the new implementation:

newplot - 2020-09-18T152114 542


zerothi:
Great, could you add a test of a skewed lattice which interpolates up and down and check.

pfebrer96:
I don't understand what do you mean. Is there any relation between the interpolation test and the isosurface?

zerothi:
You are correct!

PR #268, opened 2020-09-21
Allowing easy linear transformations on grids

pfebrer96:
Yet another feature that I need to use in the visualization module but I thought it would be a good addition for the package.

It uses `scipy.ndimage.affine_transform` to **apply a linear transformation to the grid**, therefore changing its supercell. It handles the process to make sure that the new obtained grid fits exactly in the output array (which will be the minimum bounding box) and the new coordinates are correctly set.

This can be useful to **visualize skewed grids** in 1D and 2D for example (in 3D we already have `isosurface`), but also probably to allow easily **overlapping grids with different cells** (to perform mesh integrals, for example?).

Here is a visual example of what it does. For a tiled ([4,3,1]) graphite grid, I used 

```python
grid.transform_axes(cval=np.nan, output_shape=(200,300,100))
``` 

to get from 

newplot - 2020-09-22T012134 350

to

newplot - 2020-09-22T012140 455

One can use `cval` to fill the empty spaces, in this case I used `np.nan`. Also if you set the mode to `"wrap"`, you get the space filled with the periodic images:
newplot - 2020-09-22T012956 660

Although it's important to know that **there is a discontinuity in the borders**. This is caused by a "bug" in scipy that will be solved when https://github.com/scipy/scipy/pull/12767 gets in, which is in the 1.6.0 milestone. It adds a "grid-wrap" mode that is what we need. I tried it already and it works, so we should be good.

newplot - 2020-09-22T013017 591

Similarly, there's a pull request done also for `"grid-constant"` which will solve the gap that appears in constant mode (see plot above).

Some things that should be taken into consideration:
- Should we do something to the grid's geometry when we apply the linear transformation?
- If the cell contains an axis with a negative contribution, the coordinates after the linear transformation will not match the ones from the geometry. To solve this, we would need to shift the grid. Maybe a `Grid.shift` or `Grid.set_origin` method could be useful? One can already shift it with `grid.apply("scipy.ndimage.shift", pixels_shift, mode="grid-wrap")`, but perhaps it is important enough to provide a method for it?
- Scipy minimum version could be pumped to `1.6.0` when the release is done, if there is no problem with that.

Let me know if you think this can be useful! If so, I'll write some tests for it.

### DIFF
--- a/sisl/grid.py
+++ b/sisl/grid.py
@@ -8,7 +8,7 @@ from numpy import floor, dot, add, cos, sin
 from numpy import ogrid, take, asarray
 from scipy.sparse import diags as sp_diags
 from scipy.sparse import SparseEfficiencyWarning
-from scipy.ndimage import zoom as ndimage_zoom
+from scipy.ndimage import zoom as ndimage_zoom, affine_transform
 
 from ._internal import set_module
 from . import _array as _a
@@ -179,6 +179,84 @@ class Grid(SuperCellChild):
 
         # Apply the scipy.ndimage.zoom function and return a new grid
         return self.apply(ndimage_zoom, zoom_factors, mode=mode, order=order, **kwargs)
+    
+    def transform_axes(self, axes=np.eye(3), output_shape=None, mode="constant", order=1, **kwargs):
+        """
+        Applies a linear transformation to the grid to get it relative to arbitrary axes.
+
+        This method can be used, for example to get the values of the grid with respect to
+        the standard basis, so that you can easily visualize it or overlap it with other grids
+        (e.g. to perform integrals).
+
+        Parameters
+        -----------
+        axes: array-like of shape (3,3)
+            these axes represent the directions that you want to use as references for
+            the new grid.
+
+            The length of the axes does not have any effect! They will be rescaled to create
+            the minimum bounding box necessary to accomodate the unit cell.
+        output_shape: array-like of int of shape (3,), optional
+            the shape of the final output. If not provided, the current shape of the grid
+            will be used. 
+            
+            Notice however that if the transformation applies a big shear to the image (grid)
+            you will probably need to have a bigger output_shape.
+        mode: str, optional
+            determines how to handle borders. See scipy docs for more info on the possible values.
+        order : int 0-5, optional
+            the order of the spline interpolation to calculate the values (since we are applying
+            a transformation, we don't actually have values for the new locations and we need to
+            interpolate them)
+            1 means linear, 2 quadratic, etc...
+        **kwargs:
+            the rest of keyword arguments are passed directly to `scipy.ndimage.affine_transform`
+
+        See also
+        ----------
+        scipy.ndimage.affine_transform : method used to apply the linear transformation.
+        """
+        # Take the current shape of the grid if no output shape was provided
+        if output_shape is None:
+            output_shape = self.shape
+
+        # Get the current cell in coordinates of the destination axes
+        inv_axes = np.linalg.inv(axes)
+        projected_cell = self.cell.dot(inv_axes)
+
+        # From that, infere how long will the bounding box of the cell be
+        lengths = abs(projected_cell).sum(axis=0)
+
+        # Create the transformation matrix. Since we want to control the shape
+        # of the output, we can not use self.dcell directly, we need to modify it.
+        scales = output_shape / lengths
+        forward_t = (self.dcell.dot(inv_axes)*scales).T
+
+        # Scipy's affine transform asks for the inverse transformation matrix, to
+        # map from output pixels to input pixels. By taking the inverse of our
+        # transformation matrix, we get exactly that.
+        tr = np.linalg.inv(forward_t)
+
+        # Calculate the offset of the image so that all points of the grid "fall" inside
+        # the output array. 
+        # For this we just calculate the centers of the input and output images
+        center_input = 0.5 * (np.array(self.shape) - 1)
+        center_output = 0.5 * (np.array(output_shape) - 1)
+
+        # And then make sure that the input center that is interpolated from the output
+        # falls in the actual input's center
+        offset = center_input - np.dot(tr, center_output)
+
+        # We pass all the parameters to scipy's affine_transform
+        transformed_image = affine_transform(self.grid, tr, order=1, offset=offset, 
+            output_shape=output_shape, mode=mode, **kwargs)
+
+        # Create a new grid with the new shape and the new axes (notice how the axes
+        # are rescaled from the input axes to fit the actual coordinates of the system)
+        grid = self.__class__(output_shape, sc=axes*lengths.reshape(3,1))
+        grid.grid = transformed_image
+            
+        return grid
 
     def isosurface(self, level, step_size=1, **kwargs):
         """Calculates the isosurface for a given value.


pfebrer96:
Hey Nick, do you think it makes sense for this to go in? Otherwise I'll add the linear transformation code to `GridPlot` :)

sisl/grid.py
```diff
@@ -179,6 +179,84 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
 
         # Apply the scipy.ndimage.zoom function and return a new grid
         return self.apply(ndimage_zoom, zoom_factors, mode=mode, order=order, **kwargs)
+    
+    def transform_axes(self, axes=np.eye(3), output_shape=None, mode="constant", order=1, **kwargs):
```
zerothi, line 183: Instead of having `axes` as input, perhaps the first argument could be either:
- `SuperCell` input
- ndarray, passed to `SuperCell` to generate sc
- `Grid`, to get parent sc and `output_shape`

I.e. something like this:
```python
def ...(self, obj, *args, ...)
    def_shape = self.shape
    if isinstance(obj, Grid):
        sc = obj.sc
        def_shape = obj.shape
    elif not isinstance(obj, SuperCell):
        sc  = obj
    else:
        sc = SuperCell(obj)
    ... do stuff with sc instead of axes
```

While I can see you re-use the `output_shape` keyword from `scipy`, I am not sure this is the best wording... I.e. all other places we use `shape`... Hmm.. really confusing `scipy` choose this wording...

sisl/grid.py
```diff
@@ -179,6 +179,84 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+        """
+        Applies a linear transformation to the grid to get it relative to arbitrary axes.
+
+        This method can be used, for example to get the values of the grid with respect to
+        the standard basis, so that you can easily visualize it or overlap it with other grids
```
zerothi, line 188: "standard basis" -> "cartesian basis"?

sisl/grid.py
```diff
@@ -179,6 +179,84 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+        See also
+        ----------
+        scipy.ndimage.affine_transform : method used to apply the linear transformation.
+        """
+        # Take the current shape of the grid if no output shape was provided
+        if output_shape is None:
+            output_shape = self.shape
+
+        # Get the current cell in coordinates of the destination axes
+        inv_axes = np.linalg.inv(axes)
```
zerothi, line 224: this one is inherent to `SuperCell` so once that is used this is easier :)

sisl/grid.py
```diff
@@ -179,6 +179,84 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+
+        # Scipy's affine transform asks for the inverse transformation matrix, to
+        # map from output pixels to input pixels. By taking the inverse of our
+        # transformation matrix, we get exactly that.
+        tr = np.linalg.inv(forward_t)
+
+        # Calculate the offset of the image so that all points of the grid "fall" inside
+        # the output array. 
+        # For this we just calculate the centers of the input and output images
+        center_input = 0.5 * (np.array(self.shape) - 1)
```
zerothi, line 243: use `_a.arrayd(...)` to cast to floats

sisl/grid.py
```diff
@@ -179,6 +179,84 @@ def interp(self, shape, order=1, mode="wrap", **kwargs):
...
+        # falls in the actual input's center
+        offset = center_input - np.dot(tr, center_output)
+
+        # We pass all the parameters to scipy's affine_transform
+        transformed_image = affine_transform(self.grid, tr, order=1, offset=offset, 
+            output_shape=output_shape, mode=mode, **kwargs)
+
+        # Create a new grid with the new shape and the new axes (notice how the axes
+        # are rescaled from the input axes to fit the actual coordinates of the system)
+        grid = self.__class__(output_shape, sc=axes*lengths.reshape(3,1))
```
zerothi, line 256: here it may make sense to make the `self.geometry` match the transformed shape and linear vectors?
In that way we have atomic coordinates again.

zerothi: Also, just pass `[1, 1, 1]` as the shape, when overriding the `grid` in the next line the correct shape will be used.
This will reduce memory requirements by a half!

zerothi (commented):
Thanks. Looks very interesting.

Some comments:
- could we imagine other transforms? I.e. does it make sense to name the method `transform_axes`? In fact this is nothing but an interpolation to a different supercell, no?
- could `interp`/`apply` also be used as a backend for this method? i.e. by some simple ways of doing it?

zerothi:
I am a bit reluctant to bump the required version of scipy to 1.6 (it is very new!), but we could for some time have it optionally defined/imported.

```python
has_affine_transform = map(int, scipy.version.version.split('.'))[:2] > (1, 6)
```
or something?

zerothi:
Also, I don't really get why your graphite (which I guess is the hexagonal lattice vectors) are shown as a square in your first image? I mean, isn't this simply a wrong plot?

pfebrer96:
> - could we imagine other transforms? I.e. does it make sense to name the method transform_axes? In fact this is nothing but an interpolation to a different supercell, no?

Yes, it is an interpolation to a different supercell.

> - could interp/apply also be used as a backend for this method? i.e. by some simple ways of doing it?

I thought that this could be integrated to `interp`, since, as you said, it is an interpolation to a different supercell and the `output_shape` will determine how much we zoom. I don't understand though why do you say that `interp` could be the backend.

Do you mean like `transform_axes` ---calls----> `interp`? I saw it more like `interp` ----calls----> `_transform_axes`.

> I am a bit reluctant to bump the required version of scipy to 1.6 (it is very new!), but we could for some time have it optionally defined/imported.

Whatever you prefer. I think that this affects all `scipy.ndimage` methods (i.e. it also affects `Grid.interp`), not just the affine transform. It's not a big deal if you just want it for visualization, but it might be bad if you do some computations with the resulting grid, as the borders will be wrong.

> Also, I don't really get why your graphite (which I guess is the hexagonal lattice vectors) are shown as a square in your first image? I mean, isn't this simply a wrong plot?

It's just a picture of what the grid array looks like. So in this case the lattice vectors `a` and `b` are the `x` and `y` axes of the plot. That's why I need to apply linear transforms to display the image correctly in `GridPlot` :)

sisl/grid.py
pfebrer96: Yes, good idea. 

However, do I keep the feature that the vectors in `sc` only indicate the directions and are scaled as needed? In the case of cartesian coordinates it seems much more reasonable, but I don't know about the other cases.

sisl/grid.py
pfebrer96: Yes, probably. I used standard basis since it is how it's called in math and that's all the method does (math). But going back to the physics world it probably makes more sense.

sisl/grid.py
pfebrer96: Although I still need to do `icell = sc.icell` right? Otherwise it will recalculate the inverse each time I do `sc.icell`.

pfebrer96: I noticed that `cell_invert` is 7 times faster than `np.linalg.inv`, but instancing a `SuperCell` adds a lot of overhead:

```python
sinvert = sisl._supercell.cell_invert
npinvert = np.linalg.inv

%timeit npinvert(cell)
%timeit sinvert(cell)
%timeit sisl.SuperCell(cell).icell
```
_7.36 µs ± 66.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
1.06 µs ± 11.1 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
48.4 µs ± 142 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)_

Should I use `cell_invert` directly and don't accept supercells? :thinking:

zerothi: No, if you compare with the scipy call this part is nothing ;)

So prefer readability and usability in this case.

zerothi:
> > Also, I don't really get why your graphite (which I guess is the hexagonal lattice vectors) are shown as a square in your first image? I mean, isn't this simply a wrong plot?
> 
> It's just a picture of what the grid array looks like. So in this case the lattice vectors `a` and `b` are the `x` and `y` axes of the plot. That's why I need to apply linear transforms to display the image correctly in `GridPlot` :)

So what this does is enlarging the matrix and fills with `cval` in the skewed-out regions?
Why not just do axis transforms in the GridPlot method? seems much simpler?

zerothi:
See [here](http://zerothi.github.io/sisl/docs/latest/tutorials/tutorial_siesta_2.html) which seems to be the same thing?

sisl/grid.py
pfebrer96: Right :sweat_smile:

pfebrer96:
> See here which seems to be the same thing?

It is exactly the same. There is no easy way to draw a skewed image in plotly though, you have to provide the image already skewed yourself (at least for now, I plan to propose a PR).

I realised that you can create a skewed contour plot in plotly similarly to how you do it in matplotlib. The problem is that it only shows discrete levels, not the actual data points (same as matplotlib's `contourf`), and it doesn't display the data on hover.

I can definitely add the affine transformation inside `GridPlot`'s code, this was just in case it was useful for any other cases where you may want to overlap grids with different axes. I don't know, maybe this doesn't have any use case apart from visualization, then it doesn't make sense to include it in `Grid` :)

zerothi:
> > See here which seems to be the same thing?
> 
> It is exactly the same. There is no easy way to draw a skewed image in plotly though, you have to provide the image already skewed yourself (at least for now, I plan to propose a PR).

Ok, now I have a better understanding of what it does.

> I realised that you can create a skewed contour plot in plotly similarly to how you do it in matplotlib. The problem is that it only shows discrete levels, not the actual data points (same as matplotlib's `contourf`), and it doesn't display the data on hover.

Ok.

> I can definitely add the affine transformation inside `GridPlot`'s code, this was just in case it was useful for any other cases where you may want to overlap grids with different axes. I don't know, maybe this doesn't have any use case apart from visualization, then it doesn't make sense to include it in `Grid` :)

Ok, I think for now this should then go into `GridPlot`. However, we should consider this method as a way to interpolate *any* grid to *any* other unit-cell. This would be really nice to have, for instance in the following workflow:
1. calculate density for hexagonal cell
2. interpolate density to cartesian lattice and tile it
3. reuse density as input for siesta.

So once we have a better view of things we could put it in, but please keep it visible! It seems like it could be handy!

pfebrer96:
> See here which seems to be the same thing?

I just realized that doing it in this way assumes that there are two cell vectors that are not orthogonal, but **a third one that is orthogonal to the other two**. E.g. a 2D material with the third lattice vector being `[0,0,z]`. Otherwise, the slice that you want to take is not perpendicular to any of the grid axes and this is no longer valid.

So, to make sure it works in the most general case I think there's no better way to do it than first moving the grid to cartesian coordinates and then plotting it. Unless you want to define a `cross_section` method that is general enough to take a cross section in any direction of space (not necessarily the grid axes).

Anything that I might be missing?

pfebrer96:
Update: I added the code to `GridPlot` in https://github.com/pfebrer96/sisl/commit/2ee332249c122ec9a6cf0a559473171b07d81fb9,  (as `GridPlot._transform_grid_cell`) but it can be moved to `Grid` at any moment.

zerothi:
I'll close this for now then.

PR #270, opened 2020-09-22
Fixed reading geometries in .AXSF-files

jonaslb:
Reading axsf-files didn't really work before, as sisl would read the entire file and put all of the steps into one big geometry.

I rewrote the parsing code and changed the call signature for `axsfSile.read_geometry` (but kept the one for `xsfSile.read_geometry`) to help with reading many geometries in large axsfs. It shouldnt matter that I changed it, wrt compat, because it didn't really work before anyway.

 - [x] Tests added

### DIFF
--- a/sisl/io/xsf.py
+++ b/sisl/io/xsf.py
@@ -1,4 +1,5 @@
 import os.path as osp
+from numbers import Integral
 import numpy as np
 
 # Import sile objects
@@ -7,6 +8,7 @@ from .sile import *
 from sisl._internal import set_module
 from sisl import Geometry, AtomUnknown, SuperCell
 from sisl.utils import str_spec
+import sisl._array as _a
 
 
 __all__ = ['xsfSile', 'axsfSile']
@@ -115,27 +117,22 @@ class xsfSile(Sile):
                 self._write(fmt_str.format(geometry.atoms[ia].Z, *geometry.xyz[ia, :]))
 
     @sile_fh_open()
-    def _r_geometry_multiple(self, steps, reuse_geom0=False, load_data=False, squeeze=False):
+    def _r_geometry_multiple(self, steps, ret_data=False, squeeze=False):
         asteps = steps
         steps = dict((step, i) for i, step in enumerate(steps))
 
-        if reuse_geom0:
-            cell = [None]
-            cell_set = [False]
-            xyz_set = [False]
-            atom = [None]
-            xyz = [None]
-        else:
-            cell = [None] * len(steps)
-            cell_set = [False] * len(steps)
-            xyz_set = [False] * len(steps)
-            atom = [None for _ in steps]
-            xyz = [None for _ in steps]
+        # initialize all things
+        cell = [None] * len(steps)
+        cell_set = [False] * len(steps)
+        xyz_set = [False] * len(steps)
+        atom = [None for _ in steps]
+        xyz = [None for _ in steps]
         data = [None for _ in steps]
-        data_set = [not load_data for _ in steps]
+        data_set = [not ret_data for _ in steps]
 
         line = " "
         all_loaded = False
+
         while line != '' and not all_loaded:
             line = self.readline()
 
@@ -144,6 +141,7 @@ class xsfSile(Sile):
             kw = line.split()[0]
             if kw not in ("CONVVEC", "PRIMVEC", "PRIMCOORD"):
                 continue
+
             step = _get_kw_index(line)
             if step != -1 and step not in steps:
                 continue
@@ -152,7 +150,7 @@ class xsfSile(Sile):
                 step = idstep = istep = None
             else:
                 idstep = steps[step]
-                istep = 0 if reuse_geom0 else idstep
+                istep = idstep
 
             if kw == "CONVVEC":
                 if step is None:
@@ -165,7 +163,7 @@ class xsfSile(Sile):
                 else:
                     cell_set[istep] = True
 
-                icell = np.zeros((3, 3), np.float64)
+                icell = _a.zerosd([3, 3])
                 for i in range(3):
                     line = self.readline()
                     icell[i] = line.split()
@@ -180,7 +178,7 @@ class xsfSile(Sile):
                 else:
                     cell_set[istep] = True
 
-                icell = np.zeros((3, 3), np.float64)
+                icell = _a.zerosd([3, 3])
                 for i in range(3):
                     line = self.readline()
                     icell[i] = line.split()
@@ -191,8 +189,8 @@ class xsfSile(Sile):
 
             elif kw == "PRIMCOORD":
                 if step is None:
-                    raise ValueError(
-                        "The file contains an unindexed (or somehow malform) 'PRIMCOORD'"
+                    raise ValueError(f"{self.__class__.__name__}"
+                        " contains an unindexed (or somehow malformed) 'PRIMCOORD'"
                         " section but you've asked for a particular index. This"
                         f" shouldn't happen. line:\n {line}"
                     )
@@ -204,10 +202,9 @@ class xsfSile(Sile):
                 for _ in range(int(line[0])):
                     line = self.readline().split()
                     if not xyz_set[istep]:
-                        # xyz_set[istep] only true for reuse_geom0
                         iatom.append(int(line[0]))
                         ixyz.append([float(x) for x in line[1:4]])
-                    if load_data and len(line) > 4:
+                    if ret_data and len(line) > 4:
                         idata.append([float(x) for x in line[4:]])
                 if not xyz_set[istep]:
                     atom[istep] = iatom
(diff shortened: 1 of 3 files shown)


sisl/io/xsf.py
```diff
@@ -391,6 +484,62 @@ def _setup(self, *args, **kwargs):
...
+           in case the XSF file has auxiliary data, return that as well.
+        step : int, default 0
+           the step/snapshot index (0-indexed) to load.
+           Please see ``read_geometries`` if you need to load several.
+        """
+        if step < 0:
+            step = self.read_num_steps() + step
+        return self._read_geom_many_step([step], load_data=data, squeeze=True)
+
+    def read_geometries(self, steps=None, one_geometry=False, load_data=False):
```
zerothi, line 502: A small comment, in other `read_geometry` variants, I allow keywords to control what to return, see e.g. `outSileSiesta.read_geometry`, I would prefer one such method.

jonaslb: Hmm okay you would like read_geometries to be the same method as read_geometry?

jonaslb: That would make good sense actually. Ill do that.

zerothi: Yes :)
Then arguments determine whether 1 or more geometries will be returned.

jonaslb:
Should be ready now @zerothi :)

sisl/io/xsf.py
```diff
@@ -384,12 +475,53 @@ class axsfSile(xsfSile):
...
+        else:
+            steps = np.asarray(steps).ravel()
+            negs = np.flatnonzero(steps < 0)
+            steps[negs] = self.read_num_steps() + steps[negs]
+        return self._read_geom_many_step(
+            steps, reuse_geom0=one_geometry, load_data=load_data, squeeze=squeeze
+        )
+
+    @sile_fh_open()
+    def read_num_steps(self):
```
zerothi, line 517: Hmm, I don't particularly like this name... Could we perhaps make it hidden for now? Is it really useful? I.e. it seems to me the overhead of doing `read_geometry` is minimal?

sisl/io/xsf.py
```diff
@@ -384,12 +475,53 @@ class axsfSile(xsfSile):
...
+        line = ' '
+        while line != '':
+            line = self.readline()
+            key = line.strip()
```
zerothi, line 521: is `strip` needed?

zerothi (approved):
I'll have a closer look soonish, but so far it looks good!

sisl/io/xsf.py
```diff
@@ -12,14 +12,23 @@
 __all__ = ['xsfSile', 'axsfSile']
 
 
+def _get_kw_index(key):
```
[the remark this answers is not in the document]
jonaslb: I wasn't sure if comments could be left inline though?

sisl/io/xsf.py
```diff
@@ -12,14 +12,23 @@
...
+    kl = key.split("#")[0].strip().split()
+    if len(kl) == 1:
+        return -1
+    return int(kl[1]) - 1
+
+
 @set_module("sisl.io")
 class xsfSile(Sile):
     """ XSF file for XCrySDen """
+    _has_steps = False
```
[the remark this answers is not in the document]
jonaslb: The other methods use `_md_steps is None` to determine whether to put indices on keywords (ie whether they are a xsf or axsf). Previously the xsfSile had a `_setup` method which called super but then didn't properly set the `_md_steps` afterwards anyway. I thought this was 'better' than having the overloaded method. But I can change it back though? It'll just call super and then `if _md_steps is None: _md_steps=0`

sisl/io/xsf.py
jonaslb: I know the name is "inelegant" but didn't know what else to do about it. I think it's nice to be able to get this information yeah. 
A possible alternative is that we read the number into `self._md_steps` upon instantiation instead?

sisl/io/xsf.py
jonaslb: It was needed due to `key.split()[0]` but I've changed that now.

jonaslb: I *have* changed it in the latest commit, I don't know why the change isn't coming up properly here in the PR diff view...

jonaslb: Ah hold on, I was looking in a different place. Now it's coming.

sisl/io/xsf.py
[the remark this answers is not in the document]
jonaslb: Nevermind, probably wont work in any of the relevant programs anyway

sisl/io/xsf.py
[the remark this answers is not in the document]
jonaslb: Im changing it back. Please see the comment I'm adding there.

sisl/io/xsf.py
```diff
@@ -386,11 +475,58 @@ class axsfSile(xsfSile):
...
-        # Correct number of steps
         super()._setup(*args, **kwargs)
-        if not hasattr(self, '_md_steps'):
+        if self._md_steps is None:
+            # The user didn't specify steps on instantiation.
+            # In read mode it doesn't really matter.
+            # In write mode it needs to be the correct number, so we should error out in this case.
+            # However sisl requires instantiation to be possible without additional args in its tests,
+            # so we need to assume that only one step is intended (as in the tests) in this case.
             self._md_steps = 1
```
jonaslb, line 485: @zerothi I put an explanation in a code comment. The referred test is `test_read_write_geometry` which works over all siles that can read and write geometries.

sisl/io/xsf.py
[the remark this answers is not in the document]
zerothi: Hmm, I am not sure what the specs say about this.. :(

sisl/io/xsf.py
zerothi: I would try and not do the instantiation read. That would almost always require reading the file twice.
Until we settle on a command naming convention for such a method, I think we should have it hidden. Users would then be required to read all geometries, then do `len(...)`, however, this is of minor inconvenience until the naming gets settled. E.g. the naming is useful in many other files, e.g. `outSileSiesta` and friends.

sisl/io/xsf.py
zerothi: Yeah, I saw that you check for the variable `md_steps`, perhaps we should then define a naming convention for a specific "steps" parameter. I.e. in FC runs one still have multiple steps, but I wouldn't consider them `md_steps`.

Could we name it `_geometry_steps`/`_ngeometry`/`_geometry_count` for clarity? This should then be applied to all files capable of having multiple geometries.

jonaslb: In this case axsf and xsf share all their code, so they check for whether the steps variable (whichever we call it) is None. Should this also be standardized then? Ie. that all siles capable of loading geometries must have `_geometry_count` set, either to `None` if they don't support it, an integer if they do support it but with a catch that it might be incorrect if the sile is in read-mode (because then that number hasn't been read)?

zerothi: hmmm. good, question, probably the cleanest way is to have `axsf` inherit the `xsf` and then decouple them entirely... What do you think?

It seems rather intrusive to have all siles have the `_geometry_count` variable if it ain't used...

sisl/io/xsf.py
jonaslb: I guess depending on what we settle on for `_md_steps` this method can be renamed to `_r_<whatever>` to keep the style of other hidden read methods.

zerothi: Yeah, good idea!

jonaslb:
I changed it so that the `xsfSile` doesn't have the `_geometry_count` and `_geometry_index` variables. The `write_x` methods also don't try to check if they are an axsf or xsf. Instead they call methods `_write_key` and `_write_comment` when relevant, which are overridden in the `axsfSile`.

I also renamed things like we discussed.

And last I made it so that it is required to pass `count` to `axsfSile` on creation if write mode is used.

I considered making `read_geometry` read all geometries by default instead of reading just the first one. I have also considered renaming the `data` argument in `xsfSile.read_geometry` to `load_data` to make it coherent with the new naming in axsfSile. Or maybe it should be `read_data`. What do you think?

zerothi:
@jonaslb any updates? I made a PR against your branch in your fork, could you have a look at it? Once cleared we can merge.

jonaslb:
Sorry, missed it, will comment over there

jonaslb:
I merged your PR into my working branch via github. It made a merge commit for some reason, I would have thought it could fast-foward. Should I try to make a rebased branch?

zerothi:
Yeah, a rebase + squash would be nice, lots of unrelated stuff :)

jonaslb:
Squashed and rebased it now, kept your commit separate and added a little detail after (some kwargs in the tests were unused after your changes)

zerothi:
Thanks for the PR!

PR #277, opened 2020-10-08
maint: remove ending blank line when writing xyz-files

tfrederiksen:
`sisl` writes an unnecessary blank line at the end of each `xyz` file.

I think in most cases this probably doesn't matter much, but I have encountered the following situation where it is actually unwanted: Sometimes I concatenate several snapshot `xyz` files into into a single sequence `xyz` file for easy visualization/inspection with `xcrysden`, e.g., 
```
cat snapshot*xyz > sequence.xyz
xcrysden --xyz sequence.xyz
```
However, this fails unless those extra blank lines written by `sisl` are removed. `xcrysden` expects exactly `N+2` lines per geometry.


zerothi:
Great! The reason I had this in the first place was because of old molden versions requiring this newline... :+1: 

I'll merge!

PR #284, opened 2020-12-10
WIP: first shot at frm/to in Geometry #282

### LINKED ISSUE #282: More convenient way to convert python objects to sisl
I think siles are great for standarizing the input of files into sisl, but there's no such infrastructure for python objects.

Maybe the creation and register of "converters" would help. Each converter would look something like this:

```python
class AseAtomsConverter:

    def from(ase_obj):
        ...code to convert ASE object into Geometry.
        return geometry

    def to(geometry):
         ...code to convert Geometry into ASE.
         return ase_obj
```

And it would be registered into the corresponding class (in this case, `Geometry`). Then, the `from` and `to` dispatchers would be implemented, and one could do:

```python
geometry.to("ase")
# or
geometry.to.ase()
```

One thing that I would particularly like is that I wouldn't need to call the `from` dispatcher. The class would check on call whether it is being passed an object that has been registered as "convertible" and use it. That is:

```python
sisl.Geometry(ase_obj)
```

would work as the current

```python
sisl.Geometry.fromASE(ase_obj)
```

**Edit**: Maybe `sisl.Geometry.from(ase_obj)` is enough so that we don't interfere in normal geometry creation

This would probably give the ability for a class to implement as many parsers as wanted without overpopulating with methods (which was one of your concerns for `Geometry`).
___________
PS: I would love a converter from strings because it would save me some repetitive code:

```python
sisl.Geometry("mygeom.fdf")
# would be equivalent to:
sisl.get_sile("mygeom.fdf").read_geometry()
```

zerothi:
- [x] Closes #282
 - [ ] Tests added

@pfebrer

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -1045,6 +1045,8 @@ class Geometry(SuperCellChild):
         idx_other : numpy.ndarray of int
              indices in `other` that are equivalent with `idx_self`
         """
+        # sanitize `other`
+        other = self.new(other)
         s_xyz = self.xyz + (_a.arrayd(offset) - _a.arrayd(offset_other)).reshape(1, 3)
         idx_self = []
         self_append = idx_self.append
@@ -2208,7 +2210,8 @@ class Geometry(SuperCellChild):
             With 'min', the routine will shift both the structures along the cell
             axis of `self` such that they coincide at the first atom, lastly one
             may use a specified offset to manually select how `other` is displaced.
-            NOTE: That `self.cell[axis, :]` will be added to `offset` always.
+            NOTE: That `self.cell[axis, :]` will be added to `offset` if `other` is
+            a geometry.
 
         See Also
         --------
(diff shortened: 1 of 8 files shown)


sisl/geometry.py
```diff
@@ -4418,6 +4419,72 @@ def __call__(self, parser, ns, value, option_string=None):
...
+        Z = aseg.get_atomic_numbers()
+        xyz = aseg.get_positions()
+        cell = aseg.get_cell()
+        return self._obj(xyz, atoms=Z, sc=cell, **kwargs)
+
+Geometry.frm.register("ase", GeometryFromAseDispatcher)
+# currently we can't ensure the ase Atoms type
+# to get it by type(). That requires ase to be importable.
+try:
+    from ase import Atoms ase ASE_Atoms
```
pfebrer, line 4453: You have a typo here :)

pfebrer:
I didn't think about the fact that `from` is a reserved word. It seemed like the most intuitive word to use, but `frm` is probably not that intuitive. Maybe we can find a different word for it.

Can we register all siles with the `read_geometry` attribute or is it too much overhead for loading the module? I'm saying this because in the end I think it would be great to be able to list in the most explicit way everything that can be converted to a geometry (or any other sisl object). This would probably be useful for documentation, and also for anything that can benefit from knowing the allowed types (e.g. the input fields of the graphical interface :) )

zerothi:
> I didn't think about the fact that `from` is a reserved word. It seemed like the most intuitive word to use, but `frm` is probably not that intuitive. Maybe we can find a different word for it.

Agreed, but I really have no other idea...

> Can we register all siles with the `read_geometry` attribute or is it too much overhead for loading the module? I'm saying this because in the end I think it would be great to be able to list in the most explicit way everything that can be converted to a geometry (or any other sisl object). This would probably be useful for documentation, and also for anything that can benefit from knowing the allowed types (e.g. the input fields of the graphical interface :) )

No, we don't need to. Currently the `GeometryFromFileDispatcher` takes care of this, if a `str` or `Path` it will automatically open the sile and do `Geometry.read` from that. So I think we shouldn't link stuff too hard, it does not aid in the understanding of the code, and if you want `siles` with certain methods, we already have that.

pfebrer:
Just some brainstorming: `load`, `create`, `build`, `construct`, `new`, use `read` for python objects to, `out_of`, `of`...

zerothi:
> Just some brainstorming: `load`, `create`, `build`, `construct`, `new`, use `read` for python objects too, `out_of`, `of`...

I like `new`, which clarifies that it is not related to the instance itself, rather a new object. i.e. `Geometry.new.ase(ase-object)`

`to` could still be used.

pfebrer:
By the way, do you think it makes sense that methods that have a `other` argument (meaning other geometry) could get sanitized as you do with `atoms`?

For example:
```
geom = Geometry(...)
geom.append("other_geom.fdf")
```

would automatically sanitize `"other_geom.fdf"` to `Geometry.new("other_geom.fdf")`

zerothi:
That would be a good idea I think :)

First approach, but failing due to wrong handling. It ain't easy to have a function that should accept objects, but figure out if the object should be "typefied" before looking up keys... Hmm...

sisl/geometry.py
```diff
@@ -132,6 +144,22 @@ def __init__(self, xyz, atoms=None, sc=None, names=None):
...
+                          obj_getattr=lambda obj, key: (_ for _ in ()).throw(
+                              KeyError((f"{obj}.new does not implement '{key}' "
+                                        f"dispatcher, are you using it incorrectly?"))),
+                          instance_dispatcher=TypeDispatcher)
+
+    # Define a dispatcher for converting Geometries
+    #  Geometry.to("run.fdf") will invoke Geometry.read("run.fdf")
+    to = ClassDispatcher("to",
+                         obj_getattr=lambda obj, key: (_ for _ in ()).throw(
+                             KeyError((f"{obj}.new does not implement '{key}' "
```
pfebrer, line 160: `{obj}.to` instead of `{obj}.new`?

zerothi: Thanks!

PR #285, opened 2020-12-13
enh: enabled reading VASP OUTCAR files

tfrederiksen:
This functionality enables the VASP user to 
* get the total energy (free energy as well as the energy extrapolated to sigma->0)
* get the cpu time consumed
* determine if the job terminated correctly
* determine if the relaxation accuracy criterion was met 

I tested this with version VASP 5.4.4.

### DIFF
--- a/sisl/io/__init__.py
+++ b/sisl/io/__init__.py
@@ -208,6 +208,7 @@ VASP (:mod:`~sisl.io.vasp`)
    eigenvalSileVASP
    chgSileVASP
    locpotSileVASP
+   outSileVASP
 
 
 .. _toc-io-wannier90:
--- a/sisl/io/vasp/__init__.py
+++ b/sisl/io/vasp/__init__.py
@@ -16,6 +16,7 @@ VASP files.
    eigenvalSileVASP
    chgSileVASP
    locpotSileVASP
+   outSileVASP
 
 """
 from .sile import *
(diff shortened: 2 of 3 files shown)


zerothi:
Great thanks a lot Thomas!!

I am inclined to change the method a bit. You are storing state variables which means that you can't reread the data. Once read, everything is parsed and you can't update this.  
I would like to change this to rereading the file on every read call. This ensures one to do a loop on the file checking whether anything has changed.

tfrederiksen:
Sounds good to me. Please go ahead (or let me know if you want me to give it a try).

zerothi:
> Sounds good to me. Please go ahead (or let me know if you want me to give it a try).

I am already on it. ;)
Please hold ;)

zerothi:
I have fixed it and pushed it, could you have a go, note that `read_energy` won't return numpy arrays, a bit sad. But it may be extended later on if needed (e.g. using pandas).

For now it holds a dictionary like the siesta output file, which I think is more consistent and allowing greater flexibility in terms of adding energies etc.

I think I got all the stuff there?

zerothi:
@tfrederiksen I have updated the energy dictionary a bit to reflect siesta's.

I think the names are a bit easier to read (no need for suffix `E`).

Also, the returned dictionary is a `PropertyDict` which basically means that values can be retrieved via;
```python
E = out.read_energy()
# get value from property
E.total
# get value from key
E["total"
```
either works.
I hope also like the new names.

PR #286, opened 2020-12-16
Added GUI build as a submodule

pfebrer:
I changed the hosting of the GUI online from firebase to github pages (still using the same domain, https://sisl-siesta.xyz).

For this reason, **I now have a branch that contains the build** of the GUI (and is updated everytime there's a push to master): https://github.com/pfebrer/sislGUI/tree/build.

In this PR I added a git submodule to link to that branch, as I think you wanted it this way and it's probably more convenient.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,7 @@
 [submodule "files"]
 	path = files
 	url = https://github.com/zerothi/sisl-files.git
+[submodule "sislGUI-build"]
+	path = sisl/viz/plotly/gui/build
+	url = https://github.com/pfebrer/sislGUI.git
+	branch = build


zerothi:
I thought you had convinced me otherwise ;)

But yeah, it would probably be easiest to have a build mechanism that just adds the build stuff into the zip/wheel.

Do you have any instructions for the build?

pfebrer:
> I thought you had convinced me otherwise ;)

Yes, but that is because I had the build published only to firebase (which wasn't accessible), so I had to alter the workflow to somehow make it available in github. Now I have the build published in a branch of the repo anyway in order to host it with github pages, so we can get it from there.

> But yeah, it would probably be easiest to have a build mechanism that just adds the build stuff into the zip/wheel.
> 
> Do you have any instructions for the build?

Do you mean for actually building the GUI or to incorporate the build into the package? If it's the latest, the only thing needed is to load the submodule that I added in this PR, so it's quite easy and robust. (The submodule points directly to the branch where the build is and you don't need to make any modification to it)

PR #291, opened 2021-01-18
Make github actions build the docs

pfebrer:
I'm trying to make github actions build the docs automatically and upload them to the `gh-pages` branch.

Mostly, it's working, but I have two doubts:
- How can I make the files in sisl-files available for the notebooks to use? I've tried to do `git submodule init` and `git submodule update`, but this is not giving me the actual files.
- For some reason, the rtd theme isn't working (see https://pfebrer.github.io/sisl/, which is hosted at https://github.com/pfebrer/sisl/tree/gh-pages-pol). Any hint as to why?

Thanks!

(I will squash all commits if this is to be merged, I've done a lot of testing)

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -1,0 +1,52 @@
+name: Build docs and upload them to gh-pages
+
+on:
+  push:
+    branches:
+      - gh-pages-docs
+  workflow_dispatch:
+  
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          lfs: true
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
+
+      - name: Install sisl
+        run: |
+          python -m pip install --upgrade pip
+          pip install Cython
+          pip install .[viz]
+        
+      - name: Install dependencies to build the docs
+        run: |
+          pip install IPython sphinx sphinx-gallery sphinx-rtd-theme nbsphinx ipykernel
+          pip install tqdm ipywidgets
+          pip install git+https://github.com/pfebrer/sisl-gui.git
+          pip install kaleido
+ 
+      - uses: r-lib/actions/setup-pandoc@v1 
+
+      - name: Build the docs
+        run: |
+          cd docs
+          SISL_FILES_TESTS=${GITHUB_WORKSPACE}/files/tests make html
+          cd ..
+        
+      # Deploy to github pages
+      - name: Deploy to github pages
+        uses: JamesIves/github-pages-deploy-action@132898
+        with:
+          BRANCH: gh-pages-pol
+          FOLDER: docs/build/html
+          GITHUB_TOKEN: ${{ github.token }}
+
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -1,9 +1,11 @@
-name: Build docs and upload them to gh-pages
+name: Build documentation + upload to gh-pages
 
 on:
   push:
     branches:
       - gh-pages-docs
+  schedule:
+    - cron: '37 10 * * SAT'
   workflow_dispatch:
   
 jobs:
@@ -14,39 +16,43 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
         with:
+          # The files submodule is required for viz documentation
           submodules: true
+          # the files submodule uses lfs
           lfs: true
 
-      - uses: actions/setup-python@v2
-        with:
-          python-version: '3.6'
+      # This should generally not be required, but in the end it is just easier
+      - name: Ensure fortran
+        run: |
+          sudo apt-get update
+          sudo apt-get install gfortran
 
-      - name: Install sisl
+      - name: Setup python environment
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: Install sisl + doc deps
         run: |
           python -m pip install --upgrade pip
-          pip install Cython
-          pip install .[viz]
+          python -m pip install Cython
+          python -m pip install .[viz]
+          python -m pip install -r docs/requirements.txt
         
-      - name: Install dependencies to build the docs
-        run: |
-          pip install IPython sphinx sphinx-gallery sphinx-rtd-theme nbsphinx ipykernel
-          pip install tqdm ipywidgets
-          pip install git+https://github.com/pfebrer/sisl-gui.git
-          pip install kaleido
- 
-      - uses: r-lib/actions/setup-pandoc@v1 
+      #- uses: r-lib/actions/setup-pandoc@v1 
 
       - name: Build the docs
         run: |
           cd docs
           SISL_FILES_TESTS=${GITHUB_WORKSPACE}/files/tests make html
+          rm -rf build/html/_sources
           cd ..
         
       # Deploy to github pages
       - name: Deploy to github pages
-        uses: JamesIves/github-pages-deploy-action@132898
+        uses: JamesIves/github-pages-deploy-action@4.1.5
         with:
-          BRANCH: gh-pages-pol
+          BRANCH: gh-pages-action
           FOLDER: docs/build/html
           GITHUB_TOKEN: ${{ github.token }}
 
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -39,7 +39,7 @@ jobs:
           python -m pip install .[viz]
           python -m pip install -r docs/requirements.txt
         
-      #- uses: r-lib/actions/setup-pandoc@v1 
+      - uses: r-lib/actions/setup-pandoc@v1 
 
       - name: Build the docs
         run: |
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -22,7 +22,7 @@ jobs:
           lfs: true
 
       # This should generally not be required, but in the end it is just easier
-      - name: Ensure fortran
+      - name: Ensure system dependencies
         run: |
           sudo apt-get update
           sudo apt-get install gfortran
@@ -30,9 +30,9 @@ jobs:
       - name: Setup python environment
         uses: actions/setup-python@v2
         with:
-          python-version: '3.8'
+          python-version: '3.9'
 
-      - name: Install sisl + doc deps
+      - name: Install sisl + documentation dependencies
         run: |
           python -m pip install --upgrade pip
           python -m pip install Cython
@@ -41,7 +41,7 @@ jobs:
         
       - uses: r-lib/actions/setup-pandoc@v1 
 
-      - name: Build the docs
+      - name: Build the documentation using the sisl-files as well
         run: |
           cd docs
           SISL_FILES_TESTS=${GITHUB_WORKSPACE}/files/tests make html
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -8,9 +8,14 @@ pyparsing>=1.5.7
 matplotlib
 ipykernel
 nbsphinx
-matplotlib
 dill>=0.3.2
 pathos
 xarray>=0.10.0
 scikit-image
 plotly
+sphinx-gallery
+sphinx-rtd-theme
+tqdm
+ipywidgets
+kaleido
+


docs/conf.py
```diff
@@ -147,11 +148,12 @@ def __getattr__(cls, name):
 if on_rtd:
     nbsphinx_allow_errors = True
 else:
-    nbsphinx_allow_errors = False
+    nbsphinx_allow_errors = True
```
pfebrer, line 151: I will set this back to `False`, I just wanted to see if everything else was fine.

.github/workflows/build_docs.yaml
```diff
@@ -0,0 +1,45 @@
+name: Build docs and upload them to gh-pages
+
+on:
+  push:
+    branches:
+      - gh-pages-docs
```
pfebrer, line 6: This is just while I'm testing, it should be changed to `master`, or even to listen to tagged commits.

.github/workflows/build_docs.yaml
```diff
@@ -0,0 +1,45 @@
...
+        run: |
+          cd docs
+          make html
+          cd ..
+        
+      # Deploy to github pages
+      - name: Deploy to github pages
+        uses: JamesIves/github-pages-deploy-action@132898
+        with:
+          BRANCH: gh-pages-pol
```
pfebrer, line 43: This would be changed to `gh-pages` if it is merged

docs/conf.py
```diff
@@ -15,6 +15,7 @@
 import os
 import pathlib
 import shlex
+import sphinx_rtd_theme
```
pfebrer, line 18: I did this trying to understand why the rtd theme wasn't working

docs/conf.py
```diff
@@ -147,11 +148,12 @@ def __getattr__(cls, name):
 if on_rtd:
     nbsphinx_allow_errors = True
 else:
-    nbsphinx_allow_errors = False
+    nbsphinx_allow_errors = True
     extensions += [
         'sphinx.ext.inheritance_diagram',
         'nbsphinx',
         'sphinx_gallery.load_style',
+        'sphinx_rtd_theme',
```
pfebrer, line 156: Same as the `sphinx_rtd_theme` import

zerothi:
Is there a way to manually trigger the build (say adding a tag to a commit).

I initially had continuous builds on rtfd, but it is quite annoying because you get tons of errors (sometimes) and it is basically a waste of CPU's in nearly all cases ;)

pfebrer:
Yes! Github actions give you the possibility to trigger workflows manually if your workflow yaml looks something like this:

```yaml
on:
  push:
    tags:
      - '*.*.*'
  workflow_dispatch:

...rest of workflow
```

So basically what I'm saying here is: run the workflow when tags that look like "*.*.*" are created, but also give the option to trigger it manually from the actions tab in github.com :)

docs/visualization/plotly/showcase/PdosPlot.ipynb
```diff
@@ -43,7 +44,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "plot = sisl.get_sile(\"../files/SrTiO3.PDOS\").plot(Erange=[-10,10])"
+    "plot = sisl.get_sile(\"../../../../files/tests/sisl/io/siesta/SrTiO3.PDOS\").plot(Erange=[-10,10])"
```
zerothi, line 47: this should be changed to the `env-var` that points to the `SISL_FILES_TESTS`, if not defined, then not possible.

docs/visualization/plotly/showcase/GridPlot.ipynb
```diff
@@ -53,16 +54,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "rho_file = \"../../../../files/tests/sisl/io/siesta/SrTiO3.RHO\"\n",
```
zerothi, line 57: same here (see further down)

docs/visualization/plotly/showcase/BandsPlot.ipynb
```diff
@@ -43,7 +44,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "bands_plot = sisl.get_sile(\"../files/SrTiO3.bands\").plot()"
+    "bands_plot = sisl.get_sile(\"../../../../files/tests/sisl/io/siesta/SrTiO3.bands\").plot()"
```
zerothi, line 47: same here (see further down)

docs/visualization/plotly/basic-tutorials/GUI with Python Demo.ipynb
```diff
@@ -393,7 +393,7 @@
     "class VeryCoolSession(Session):\n",
     "    \n",
     "    # Add the method to see PDOS, bands and RHO of a structure\n",
-    "    def see_results(self, struct, wdir='../files', tab_name=None):\n",
+    "    def see_results(self, struct, wdir='../../../../files/tests/sisl/io/siesta', tab_name=None):\n",
```
zerothi, line 396: same here (see further down)

docs/visualization/plotly/basic-tutorials/GUI with Python Demo.ipynb
```diff
@@ -332,7 +332,7 @@
     "session.add_tab(tab_name)\n",
     "\n",
     "# Get all the plots\n",
-    "plots = [Plot(f'../files/{struct}.{ext}') for ext in (\"PDOS\", \"RHO\", \"bands\")]\n",
+    "plots = [Plot(f'\"../../../../files/tests/sisl/io/siesta/{struct}.{ext}') for ext in (\"PDOS\", \"RHO\", \"bands\")]\n",
```
zerothi, line 335: same here (see further down)

zerothi (commented):
I am currently completely rewriting the documentation... So this will have to wait until that point. 

I think we should split this one up into two PR's

1. With _only_ github action
2. With the changes related to the `ipynb` for the tests (they don't relate to github actions)

In this way I can put in 2. much faster.

pfebrer:
Ok, I will split it. But just fyi, the github workflow is working great already, the only thing failing is the rtd theme :sweat_smile:

zerothi:
> Ok, I will split it. But just fyi, the github workflow is working great already, the only thing failing is the rtd theme

How long time does it take to build using workflow?

pfebrer:
The last one took 11min from which 5 min were used to install sisl and 4.30 min to build the docs.

pfebrer:
I solved the mistery of the theme! 

Github pages uses some framework that weirdly enough was blocking requests to all files but `index.html`. My browser's console was full of 404 responses when trying to get the static files.

The solution was simple, you need to have a `.nojekyll` empty file in the gh-pages branch. 

Now **everything is working fine** :tada: . You can check https://pfebrer.github.io/sisl/. So I don't think there's the need to split this PR. The workflow part can stay there and won't do anything until you create a version tag or manually trigger it.

What do you think?

zerothi:
I still think you should split it. The documentation is not working. Links to classes don't work. :)

So please split it up into two, one with only git workflow.

And another with the other fixes.

pfebrer:
Done: https://github.com/zerothi/sisl/pull/300

pfebrer:
By the way, this is what's causing the linking errors I think:

```
/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/sisl/__init__.py:docstring of sisl.rst:20: WARNING: autosummary: stub file not found 'sisl.PeriodicTable'. Check your autosummary_generate setting.
```

maybe you know how to solve it.

zerothi:
This is exactly one of the reasons that remote build systems are so difficult. It is immensely hard to debug since environments differ greatly. I know how this behaves on my setup, but not theirs. I would rather not debug this now...

pfebrer:
Ok, I will try a few things, you can turn the notifications of this PR off :sweat_smile:

zerothi:
Ok, but please wait a week, I am finishing the new documentation which fixes some things whilst changing other things. So don't waste work here! ;)

.github/workflows/build_docs.yaml
```diff
@@ -14,39 +13,47 @@ jobs:
...
+          python -m pip install .[viz]
         
       - name: Install dependencies to build the docs
         run: |
-          pip install IPython sphinx sphinx-gallery sphinx-rtd-theme nbsphinx ipykernel
-          pip install tqdm ipywidgets
-          pip install git+https://github.com/pfebrer/sisl-gui.git
-          pip install kaleido
+          python -m pip install -r docs/requirements.txt
+          python -m pip install git+https://github.com/pfebrer/sisl-gui.git
```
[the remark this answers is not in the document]
pfebrer: Oh, and please add kaleido to requirements as well, otherwise it won't be able to generate images from plotly plots :)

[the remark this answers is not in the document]
zerothi: Ok, it wasn't clear to me what was necessary. Are there anything else needed in `docs/requirements.txt`?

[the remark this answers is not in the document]
pfebrer: Hmm right now you would need also to include plotly, because I think [viz] does not install it, right? Apart from that, no more requirements.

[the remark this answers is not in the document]
zerothi: No, I think `[viz]` does everything, including ase.
```
        "viz": reduce(lambda a, b: a + b, viz.values()),
```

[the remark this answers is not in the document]
pfebrer: Aaah ok but this is because the branch is not rebased yet

zerothi:
Ok, I brought it up to speed with the latest `main`

Oh, and sorry, I had clearly thought I had pushed these to my own branch, I am trying to check this out!! :)

.github/workflows/build_docs.yaml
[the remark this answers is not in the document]
zerothi: Yes, just rebased ;)

zerothi:
Ok, I think this is soonish ready to go in...

pfebrer:
Have you checked if links work in the build? This was the only problem that I had with it (https://github.com/zerothi/sisl/pull/291#issuecomment-774485191).

zerothi:
> Have you checked if links work in the build? This was the only problem that I had with it ([#291 (comment)](https://github.com/zerothi/sisl/pull/291#issuecomment-774485191)).

Thanks for reminding me, I'll perhaps have a go tomorrow!

zerothi:
@pfebrer there is something wrong with the `viz` rendering. When I checkout the `gh-pages-action` branch and locally view the pages the viz rendering in the Wavefunction plots are just empty, with no warnings what-so-ever, have you seen this before? I didn't have this problem when doing it on my own? (see e.g. the sisl documentation online)

pfebrer:
> @pfebrer there is something wrong with the `viz` rendering. When I checkout the `gh-pages-action` branch and locally view the pages the viz rendering in the Wavefunction plots are just empty, with no warnings what-so-ever, have you seen this before? I didn't have this problem when doing it on my own? (see e.g. the sisl documentation online)

Yes I've seen it and it is very annoying because I don't know why it happens sometimes. I'll try to look for answers again.

pfebrer:
Actually, no I have never seen something like this. This is what you see, right?
image

zerothi:
Yes, exactly... And I checked the logs, there was nothing obvious! :(

pfebrer:
The javascript console shows what the error is. If you open `BandsPlot.html` in a browser you'll see this:

image

I'm trying to understand how to fix it.

pfebrer:
Hmm I thought the error was the file not found, but actually it is the last message.

Because if I check the locally built version, it works and the two first messages are there as well, just not the last one.

zerothi:
> Hmm I thought the error was the file not found, but actually it is the last message.
> 
> Because if I check the locally built version, it works and the two first messages are there as well, just not the last one.

I can't show this locally, could you? I don't follow what you write? ;)

pfebrer:
If you open `visualization/viz_module/showcase/BandsPlot.html` from your `gh-pages-action` branch in a browser and then check the browser console (press F12 on chrome based browsers), you don't see these messages?

zerothi:
no, I am on firefox, I did the same, then searched for `TypeError`, with nothing coming up... But still the plots don't show... :(
I don't know how to use the inspector ;(

pfebrer:
In the firefox console I get the same:

image

pfebrer:
Well not the same message, but the same error `window.MathJax.Hub` is `undefined`. Chrome tells me basically that someone is trying to access `.Config` on this undefined object.

zerothi:
Weird, I don't get that at all...

No errors what so ever...

zerothi:
See here:
Untitled

zerothi:
Ah, you pressed "Console", 2 secs

zerothi:
Then true, I get this:
Untitled

pfebrer:
Yes, exactly. I found the problem:

The last build that you did includes in the highlighted line Mathjax 3:

image

which seems to be incompatible with the `@jupyter-widgets` script (included in the line just above).

I've looked at previous builds that I did locally and they included `mathjax 2.7`. I copied the line and pasted it just above the `@jupyter-widgets` script tag:

image

Then everything works perfectly.

I don't know who should I complain to because I don't know who includes this lines (`sphinx`, `nbsphinx`...). 

I'm trying to reproduce this bug in a fresh environment locally and then I will try to fix it with a patch by adding some lines to `sisl/docs/conf.py`.

pfebrer:
Yaaay I found that the `sphinx.ext.mathjax` extension (which we load in `conf.py`) allows you to give the path to the mathjax version that you want. That's so kind of them! :)

I believe that if you include the following line wherever in `sisl/docs/conf.py`:

```python
mathjax_path = "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/latest.js?config=TeX-AMS-MML_HTMLorMML"
```

the problem should be solved. Let's see!

zerothi:
Pushed, lets see!

pfebrer:
:crossed_fingers:

pfebrer:
Working :partying_face: :partying_face: :partying_face: :partying_face:

zerothi:
It is now in main!!! Thanks for the work here!!!

zerothi:
For the record, it will only build once a week on saturdays ;)
It can also be triggered by a post event. I'll have to see about this to try it out once in a while.

pfebrer:
Saturday sounds like a good day to build documentation :) haha

Since I included `workflow_dispatch:`, you should be able to make it build by clicking the `Run workflow` button in the `Actions` tab after you select the workflow.

zerothi:
See that was informative ;)
Done! Lets see the new pages! :)
It couldn't be much easier.

zerothi:
I am going to clean the branches once this shows to complete correctly.

pfebrer:
Also if you make it listen to main, you can make it build only if something is modified in docs.

In this workflow for example I choose to only run it if something changes in the frontend directory:
https://github.com/pfebrer/sisl-gui/blob/75067dc85bcdab7d936b34d2a4720d30fa39a93e/.github/workflows/deploy-gui.yaml#L7-L8

zerothi:
> Also if you make it listen to main, you can make it build only if something is modified in docs.
> 
> In this workflow for example I choose to only run it if something changes in the frontend directory: https://github.com/pfebrer/sisl-gui/blob/75067dc85bcdab7d936b34d2a4720d30fa39a93e/.github/workflows/deploy-gui.yaml#L7-L8

Good idea. However, since the documentation is created from the doc-strings this may not be adequate.
For now I'll probably keep it like this, but otherwise it should check against doc-strings. I also looked into not building if no commits was made within the last 7 days. See my latest commit.

zerothi:
Erhm, the pages look horrible ;) How does it look on your end?
Screenshot_2021-11-12 sisl Toolbox for electronic structure calculations — sisl latest documentation

pfebrer:
Hehe, luckily I had already encountered this problem: https://github.com/zerothi/sisl/pull/291#issuecomment-774480310

pfebrer:
> Good idea. However, since the documentation is created from the doc-strings this may not be adequate.

Yes, totally.

> I also looked into not building if no commits was made within the last 7 days. See my latest commit.

That line of code will never get hit :smile: No commits in the last 7 days? That's impossible haha

zerothi:
> > Good idea. However, since the documentation is created from the doc-strings this may not be adequate.
> 
> Yes, totally.
> 
> > I also looked into not building if no commits was made within the last 7 days. See my latest commit.
> 
> That line of code will never get hit smile No commits in the last 7 days? That's impossible haha

Hehe, lets hope so. But I seem to remember a single week I didn't do anything ;)
Also, when I am on vacation, things rapidly decrease!!!!

pfebrer:
You don't need to add `.nojekyll` in the workflow! Just add it as a normal commit to the `gh-pages` branch. It will not get removed by workflow runs.

zerothi:
> You don't need to add `.nojekyll` in the workflow! Just add it as a normal commit to the `gh-pages` branch. It will not get removed by workflow runs.

I know, but I am playing with the thought of not having this staticly with all the history. The problem is that `git clone sisl` takes forever since it contains all commits on the gh-pages. If I can completely decouple the `gh-pages` and create it on the fly, then the repo size would be greatly reduced. In this way I can do whatever I want since each workflow is completely decoupled.

pfebrer:
Aaah ok :+1:

zerothi:
WORKING!!! Great!

Issue #292, opened 2021-01-22
Make viz.plotly with fewer dependencies:


zerothi:
**Describe the issue**

This discussion was initiated [here](https://github.com/zerothi/sisl/commit/0bc8b72f5c3a8618c8e5c782725183dc43560903#r46192426)

I have a few more ideas that needs thought:

I agree that plotly plotting should be entirely decoupled from the flask environment (GUI).

However, we discussed what to do with the GUI and @pfebrer suggested to make that a separately installed package.

I now thought that perhaps it would make sense to make it loadable from the configuration directory?

Pro's/cons

**GUI as a package**
- :heavy_check_mark: Easy to install
- :heavy_check_mark: Easy dependency management
- :x: cannot have multiple versions (for double checking)
- :x: may be difficult to revert to older versions

**GUI as a configuration**
- :heavy_check_mark: Easy to maintain multiple versions
- :heavy_check_mark: Easy to switch between versions
- :x: A bit more difficult to install
- :heavy_check_mark:/:x: dependency management in between

What do you think @pfebrer about having it as a configuration?

zerothi:
Once the dependencies are removed, lets take a look at the documentation!

pfebrer:
I don't understand what does this mean exactly:

> I now thought that perhaps it would make sense to make it loadable from the configuration directory?

Could you elaborate please :) What is the configuration directory? Is it e.g. `~/.sisl` or is it something inside the `sisl` directory? And what does it mean loading it from there, how would it work?

Thanks!

zerothi:
Remember we had `sisl/_environ` to accommodate environment variables read by sisl?

The idea was that the `SISL_CONFIGDIR` would host the layouts for the viz modules and other template files, if needed. So we could basically have the GUI stuff in a directory there (which is the default place to look for stuff) and then load it using `importlib` or similar?

It was just a thought whether you think this is more apt for a tool than a package?

pfebrer:
Aah ok, so for a release, you would get the current version of the GUI and copy it in the config dir.

This is basically the same as having it inside the package, right? How could dependencies be handled in this case?

I think that there would be no problem with versions if in the `sisl_gui` package I specify the minimum sisl version supported. Or would it?

zerothi:
> Aah ok, so for a release, you would get the current version of the GUI and copy it in the config dir.
Exactly.

> This is basically the same as having it inside the package, right? How could dependencies be handled in this case?
[...]

PR #293, opened 2021-01-22
Isolated the gui submodule from the rest of sisl.viz

pfebrer:
As discussed in https://github.com/zerothi/sisl/issues/292.

Now, the sisl.viz package can work without needing to load `sisl.viz.plotly.gui`. (Added a line to raise an error in `sisl.viz.plotly.gui.__init__` and ran `sisl.viz` tests succesfully).

All that's left to do now is to decide what to do with the `gui` module.


pfebrer:
Done!

pfebrer:
Sorry, now it's really done :sweat_smile:

pfebrer:
The gui part is already available at PyPi (and working):

`pip install sisl-gui`

:)

zerothi:
Great work! I am trying this out locally, I can't get the tests to run? However, it never runs the tests?

I just do `pytest --pyargs sisl`, how are you running tests?

zerothi:
Ok, I think I got it, you were missing `__init__.py` files in the `tests` folders ;)

I'll amend some changes (don't touch anything ;)

pfebrer:
:+1: Before merging, let me know. Because I need to change some things in the notebook about interacting with the GUI. (or maybe remove that notebook and put it in `sisl-gui` docs :thinking: )

zerothi:
Ok, so I pushed some changes to your branch, could you have a look?

Also, I still get errors, some I don't quite understand, do you get all tests to pass?

pfebrer:
The `test_plotables.py` were failing, but they do not make sense anymore. I just removed them. 

Then some documentation tests don't pass, but this can be easily solved with the `sisl/viz/plotly/_doc_updater.py` script. I will fix them.

Then there are some tests that in my case get skipped because it's not able to retrieve the files. I don't know if these fail for you. If you can tell me how to get the actual files in `sisl-files` I can check if they work. For now all I see when I check the `SrTiO3.bands` file is:

```
bands 
version https://git-lfs.github.com/spec/v1
oid sha256:3e0d2fe2fea9bcd67591fcf5f22625a07c90945703baf04621d5ae15f91e7bdc
size 143139
```

The rest of tests all pass. Are there any other tests failing for you?

zerothi:
> The `test_plotables.py` were failing, but they do not make sense anymore. I just removed them.

Ok.
 
> Then some documentation tests don't pass, but this can be easily solved with the `sisl/viz/plotly/_doc_updater.py` script. I will fix them.

Ok.

> Then there are some tests that in my case get skipped because it's not able to retrieve the files. I don't know if these fail for you. If you can tell me how to get the actual files in `sisl-files` I can check if they work. For now all I see when I check the `SrTiO3.bands` file is:
> 
> ```
> bands 
> version https://git-lfs.github.com/spec/v1
> oid sha256:3e0d2fe2fea9bcd67591fcf5f22625a07c90945703baf04621d5ae15f91e7bdc
> size 143139
> ```

But the file is already in sisl-files, so I think you are fetching it incorrectly? Could you see `sisl/io/siesta/tests/test_*.py` for details on fetching files from the `sisl-files` directories.

I'll check once you let me know ;)

zerothi:
Ok, something like this:
```patch
diff --git a/sisl/viz/plotly/plots/tests/test_bands.py b/sisl/viz/plotly/plots/tests/test_bands.py
index 7f0d0792d..44750e697 100644
--- a/sisl/viz/plotly/plots/tests/test_bands.py
+++ b/sisl/viz/plotly/plots/tests/test_bands.py
@@ -5,15 +5,18 @@ Tests specific functionality of the bands plot.
 Different inputs are tested (siesta .bands and sisl Hamiltonian).
 
 """
+import os.path as osp
+import itertools
+from functools import partial
+
 from xarray import DataArray
 import numpy as np
 import pytest
-import itertools
-from functools import partial
 
 import sisl
 from sisl.viz import BandsPlot
 from sisl.viz.plotly.plots.tests.conftest import PlotTester
+_dir = osp.join('sisl', 'io', 'siesta')
 
 
 pytestmark = [pytest.mark.viz, pytest.mark.plotly]
@@ -165,18 +168,16 @@ bands_plots["sisl_H_path"] = {
 
 # ---- From a non collinear calculation in SIESTA
 
+class TestBandsPlot(BandsPlotTester):
+    run_for = bands_plots
+
 
 def NC_init_func(sisl_files, **kwargs):
-    H = sisl.get_sile(sisl_files("fe_clust_noncollinear.TSHS")).read_hamiltonian()
+    H = sisl.get_sile(sisl_files(_dir, "fe_clust_noncollinear.TSHS")).read_hamiltonian()
     bz = sisl.BandStructure(H, [[0, 0, 0], [0.5, 0, 0]], 3, ["Gamma", "X"])
-
     return bz.plot(**kwargs)
 
 
-class TestBandsPlot(BandsPlotTester):
-    run_for = bands_plots
-
-
 class TestNCSpinBands(NCSpinBandsTester):
 
     run_for = {

```

But, another problem. The non-colinear matrices don't accept `spin` as argument. So your bands plots needs to differentiate between the two.
Also, your dictionaries for holding tests are not good. Better to do some explicit tests, why are you having these dictionaries?

pfebrer:
> Ok, something like this:

I tried something like this, but even then it isn't working for me. I don't understand how this storage of large files works. Do I need to do something with `Git LFS` or should `sisl_files` be able to handle files that look like this:

```
version https://git-lfs.github.com/spec/v1
oid sha256:3e0d2fe2fea9bcd67591fcf5f22625a07c90945703baf04621d5ae15f91e7bdc
size 143139
```

> But, another problem. The non-colinear matrices don't accept spin as argument. So your bands plots needs to differentiate between the two.

Ok, I will look into it, thanks!

> Also, your dictionaries for holding tests are not good. Better to do some explicit tests, why are you having these dictionaries?

This is best solution that I found to my problem.

My problem: I want to make sure that tests for a given kind of plot work whatever the entry point has been.

Therefore I will need to run the tests for e.g. `BandsPlot` multiple times. Keeping in mind that this should be easy to scale to "infinite" entry points, in my solution I wanted to:
- Write as little code as possible so that it was maintainable. This is achieved by putting all the machinery in `sisl/viz/plotly/plots/conftest.py`. I know that the code that I wrote works and I don't need to write it elsewhere.
- Give names to the different runs of the tests for a plot type so that I can easily identify them in the pytest report. I.e. they look something like `BandsPlot.test_whatever[from_bands_file]`. This is implemented with `pytest` fixtures and come from the dictionary keys. Again this is robust because I only wrote the code once, so I don't need to worry everytime I write a new test.
- Make sure that if the plot isn't initialized properly, the tests are skipped so that I can easily identify it (instead of having the report full of "Nonetype has no atribute X" meaningless errors). I don't want to think about this each time I add a new test, so again this is handled once in `sisl/viz/plotly/plots/conftest.py`.
- Make sure that general tests that all plot types should pass are also passed when the entry point is used. (also handled in `sisl/viz/plotly/plots/conftest.py`).

It may be unconventional, but it has saved me lots of headaches, as adding new entry points to test is much simpler. If you find another approach that achieves my four goals above, I'll be happy to change it.

zerothi:
> > Ok, something like this:
> 
> I tried something like this, but even then it isn't working for me. I don't understand how this storage of large files works. Do I need to do something with `Git LFS` or should `sisl_files` be able to handle files that look like this:
> 
> ```
> version https://git-lfs.github.com/spec/v1
> oid sha256:3e0d2fe2fea9bcd67591fcf5f22625a07c90945703baf04621d5ae15f91e7bdc
> size 143139
> ```

Hmm, yeah you probably need git-lfs installed.

> > Also, your dictionaries for holding tests are not good. Better to do some explicit tests, why are you having these dictionaries?
> 
> This is best solution that I found to my problem.
> 
> My problem: I want to make sure that tests for a given kind of plot work whatever the entry point has been.
> 
> Therefore I will need to run the tests for e.g. `BandsPlot` multiple times. Keeping in mind that this should be easy to scale to "infinite" entry points, in my solution I wanted to:
> 
> * Write as little code as possible so that it was maintainable. This is achieved by putting all the machinery in `sisl/viz/plotly/plots/conftest.py`. I know that the code that I wrote works and I don't need to write it elsewhere.
> * Give names to the different runs of the tests for a plot type so that I can easily identify them in the pytest report. I.e. they look something like `BandsPlotTester.test_whatever[from_bands_file]`. This is implemented with `pytest` fixtures and the names come from the dictionary keys. Again this is robust because I only wrote the code once, so I don't need to worry everytime I write a new test.
> * Make sure that if the plot isn't initialized properly, the tests are skipped so that I can easily identify it (instead of having the report full of "Nonetype has no atribute X" meaningless errors). I don't want to think about this each time I add a new test, so again this is handled once in `sisl/viz/plotly/plots/conftest.py`.
> * Make sure that general tests that all plot types should pass are also passed when the entry point is used. (also handled in `sisl/viz/plotly/plots/conftest.py`).
> 
> It may be unconventional, but it has saved me lots of headaches, as adding new entry points to test is much simpler. If you find another approach that achieves my four goals above, I'll be happy to change it.

I don't understand what you want to achieve, couldn't you create the plots (`BandsPlot`) through a fixture?

pfebrer:
> I don't understand what you want to achieve, couldn't you create the plots (BandsPlot) through a fixture?

In fact, I use a fixture. But I don't write it at each `test_x.py`, it's only written once in `conftest.py`:

https://github.com/zerothi/sisl/blob/eae8ee10bf20dac0a143da04044cb4a31a9f3cc6/sisl/viz/plotly/plots/tests/conftest.py#L175-L190

Here, `params` is the dictionary that you passed to `TesterClass.run_for`. The ids are the keys of the dictionary, that's why I see `TesterClass.test_function[dict_key]` when tests fail in `pytest`.

So, for each item in the dictionary I run the tests and check against the expected values for that specific entry point. After that, I "clean" the class so that a different entry can use it.

It would be the same if I wrote the fixture at each `test_x.py` file. But it would be the exact same code repeated multiple times.

pfebrer:
By the way, I make the class (i.e. `TestBandsPlot`) use this fixture in this line:

https://github.com/zerothi/sisl/blob/eae8ee10bf20dac0a143da04044cb4a31a9f3cc6/sisl/viz/plotly/plots/tests/conftest.py#L197-L200

Note that this is all taken care by the metaclass defined at the top of `conftest.py`, so it's done automatically when you define a class that inherits from `PlotTester`.

zerothi:
Ok. I am just wondering whether it is too complicated. I can't figure out what you do, and thus have a hard time debugging it... :)

pfebrer:
I think it's quite simple if you know what's going on. Let me explain:

When you write:

```python
class TestBandsPlot(PlotTester):

    _required_attrs = ["correct_value"]

    run_for = {
        "first_try": {
            "plot_file": "myfile.bands",
            "correct_value": 2
        },
        "second_try": {
            "init_func": sisl.get_sile("myotherfile.bands").plot,
            "correct_value": 4
        },
    }

    def test_if_correct(self):
        assert self.plot.value == self.correct_value
```

`TestBandsPlot` will run for each key in `TestBandsPlot.run_for` like so:
- All the keys of the dictionary item will be set as `TestBandsPlot` attributes. That is, you will be able to access the value of `run_for.first_try.correct_value` under `self.correct_value` when the tests are running for "first try".
- The plot will be initialized either using `"plot_file"` or `"init_func"` and stored under `self.plot`. This is done in `test_plot_initialization` implemented in `PlotTester`:
https://github.com/zerothi/sisl/blob/edde818455379fdab45afbfacd7ded0020cbbbb8/sisl/viz/plotly/plots/tests/conftest.py#L118-L137
If the plot doesn't succeed in being initialized, the rest of tests for that run are skipped. Why did I add this? Previously I was passing the already initialized plot, but then I was getting lots of failed tests that were incomprehensible when the plot simply failed to initialize.
- If the plot succesfully initializes, the rest of the tests of that class run.

What is it that you are having trouble debugging? Maybe it's because of some unnecessarily complex structures in my implementation of plots, not because of how tests run.

zerothi:
Ok thanks for explaining.


> ```python
> class TestBandsPlot(PlotTester):
> 
>     _required_attrs = ["correct_value"]
> 
>     run_for = {
>         "first_try": {
>             "plot_file": "myfile.bands",
>             "correct_value": 2
>         },
>         "second_try": {
>             "init_func": sisl.get_sile("myotherfile.bands").plot,
>             "correct_value": 4
>         },
>     }
> 
>     def test_if_correct(self):
>         assert self.plot.value == self.correct_value
> ```
> 
> `TestBandsPlot` will run **for each key in `TestBandsPlot.run_for`** like so:

I think this is better for parameterized runs.

```python
@pytest.mark.parametrize("attrs_assert":
[( # dict, dict for attrs assertionslamba for checking
{"plot_file": "myfile.bands"},
{"plot.value": 2}),
 # and this is not good, doing reading before actually starting the test is not encouraged since you cannot
# add the correct path, so it needs some more thought, here changed to lambda
({"init_func": lambda: sisl.get_sile("myotherfile.bands").plot},
{"plot.value": 4}))

This ain't prettier at first sight, but it is very easy to add stuff, and probably once the parametrize is added it becomes obvious how to do things.

The problem is you are basically implementing your own parametrization technique which is somewhat inconvenient to decipher ;)

Isn't the above something you could use instead?

**Comment by pfebrer** [comment]
Hmm I don't understand fully the approach that you are proposing. Could you rewrite my example using `pytest.mark.parametrize`?

**Comment by pfebrer** [comment]
Btw, does `sisl.get_sile("myotherfile.bands").plot` read something? I'm not calling the plotting function, just passing it.

**Comment by zerothi** [comment]
For now I am trying to fix your documentation.

1. I can't figure out how you say that your test succeeded? There were tons of missing documentations that ended up failing the test.
2. I can't figure out how you handle cases where plotly is not installed. 
3. With my latest commit I still get errors:
```bash
_____________________________________________________________________ TestPlotSubClass.test_documentation[FatbandsPlot] ______________________________________________________________________

self = <sisl.viz.plotly.plots.tests.test_plots.TestPlotSubClass object at 0x14edf1392850>

    def test_documentation(self):
    
        doc = self.PlotClass.__doc__
    
        # Check that it has documentation
        assert doc is not None, f'{self.PlotClass.__name__} does not have documentation'
    
        # Check that all params are in the documentation
        params = [param.key for param in self.PlotClass._get_class_params()[0]]
        missing_params = list(filter(lambda key: key not in doc, params))
>       assert len(missing_params) == 0, f"The following parameters are missing in the documentation of {self.PlotClass.__name__}: {missing_params}"
E       AssertionError: The following parameters are missing in the documentation of FatbandsPlot: ['aiida_bands']
E       assert 1 == 0
E         +1
E         -0

/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.10.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/plotly/tests/test_plot.py:39: AssertionError
_____________________________________________________________________________________ test_eigenstate_wf _____________________________________________________________________________________

setup = {'orb': <sisl.AtomicOrbital 2pzZ1 q0=0.0>, 'geom': <sisl.Geometry na=4, no=4, nsc=[3 3 1]>, 'H': <sisl.physics.Hamiltonian na=4, no=4, nsc=[3 3 1], dim=1, nnz=16, spin=unpolarized>}

    def test_eigenstate_wf(setup):
    
        H = setup.H
        geometry = H.geometry
    
        plot = H.eigenstate()[0].plot.wavefunction(geometry=geometry)
    
>       assert len(plot.data) > 0
E       assert 0 > 0
E        +  where 0 = len(())
E        +    where () = <sisl.viz.plotly.plots.grid.WavefunctionPlot object at 0x14edf0a9eeb0>.data

/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.10.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/plotly/tests/test_plotables.py:32: AssertionError
------------------------------------------------------------------------------------ Captured stdout call ------------------------------------------------------------------------------------
The plot has been initialized correctly, but the current settings were not enough to generate the figure.
 (Error: Grid.isosurface requires scikit-image to be installed)
____________________________________________________________________________________ test_hamiltonian_wf _____________________________________________________________________________________

setup = {'orb': <sisl.AtomicOrbital 2pzZ1 q0=0.0>, 'geom': <sisl.Geometry na=4, no=4, nsc=[3 3 1]>, 'H': <sisl.physics.Hamiltonian na=4, no=4, nsc=[3 3 1], dim=1, nnz=16, spin=unpolarized>}

    def test_hamiltonian_wf(setup):
        H = setup.H
        geometry = H.geometry
    
        # Check if it works for 3D plots
        plot = H.plot.wavefunction(2)
>       assert isinstance(plot.data[0], go.Isosurface)

/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.10.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/plotly/tests/test_plotables.py:42: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.10.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/plotly/plot.py:718: in __getattr__
    elif hasattr(self.figure, key):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <sisl.viz.plotly.plots.grid.WavefunctionPlot object at 0x14edf0a7a940>, key = 'figure'

    def __getattr__(self, key):
        """ This method is executed only after python has found that there is no such attribute in the instance
    
        So let's try to find it elsewhere. There are two options:
            - The attribute is in the figure object (self.figure)
            - The attribute is currently being shared with other plots (only possible if it's a childplot)
        """
        if key in ["figure", "shared_attr"]:
            pass
        elif hasattr(self.figure, key):
            return getattr(self.figure, key)
        else:
            #If it is a childPlot, maybe the attribute is in the shared storage to save memory and time
            try:
                return self.shared_attr(key)
            except (KeyError, AttributeError):
                pass
    
>       raise AttributeError(f"The attribute '{key}' was not found either in the plot, its figure, or in shared attributes.")
E       AttributeError: The attribute 'figure' was not found either in the plot, its figure, or in shared attributes.

/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.10.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/plotly/plot.py:727: AttributeError
```

zerothi:
Also, the more I read the tests, the more I think they should be split into duplicate code ;)

It is extremely difficult to maintain, and if you change something in the superclass tester and that shouldn't be used in a sub-classed testing class, then you need to make changes.  
I think it really needs restructuring. :)  
Subclassing in a single `test_*.py` is ok, but across many files is really hard :)
For instance, I really have no clue what you are actually testing ;)

Sorry to bash all your hard work!!!

pfebrer:
> I can't figure out how you say that your test succeeded? There were tons of missing documentations that ended up failing the test.

No, the documentation tests (some) are failing for me as well. I just said that doesn't worry me because there is a script to build the doc strings automatically using the info of the plot parameters.

> I can't figure out how you handle cases where plotly is not installed.

I don't at the moment, because the process is tightly bound to plotly right now. I'd say we can leave this like that and fix it in the refactoring when we implement the dispatchers for the plotting backend/plotting method + entry points. I'm really excited about that, because if we manage to reuse the same logic to process plots that are drawn in plotly, matplotlib and blender, while also letting the user define new ways of drawing, that would be awesome!

Apart from the obvious advantatges, that would mean that anyone can develop a GUI that uses our logic regardless of the plotting backend they choose/are forced to use. I.e. https://github.com/sozykinsa/GUI4dft uses matplotlib and it would be extremely cool to harvest all the 3D rendering/manipulating knowledge of blender by using it as a GUI for geometries and grids.

> With my latest commit I still get errors:

The `test_eigenstate_wf` and `test_hamiltonian_wf` tests I removed them here: https://github.com/zerothi/sisl/pull/293/commits/f511009ddd22f53d5a532a60c7959984fe8ed919. I don't know why you are still seeing them. Maybe you are testing an installation for which the files haven't been modified? :thinking: 

> Sorry to bash all your hard work, I really mean it well!!!

That's cool, but I don't understand how you want to do it exactly, so you'll have to do it (at least do it in one file to provide an example). Do you want to do it in this PR or in a separate one?

pfebrer:
Now all the docstrings are up to date! The other two failed tests, as I said, shouldn't be there anymore.

pfebrer:
By the way just so that you know:

The way the auto documentation works is. In the classes where you want to update the docs remove all the parameters from the docstring and write `%%configurable_settings%%` instead. I.e.:

```python
class BandsPlot(Plot):
    """"Bla bla bla

    Parameters
    ------------------
    %%configurable_settings%%
    """"
```

then just run the script:

```
python sisl/viz/plotly/_doc_updater.py
```

which should work but it gives me some import errors that I don't know how to interpret. So instead, for the moment I'm doing it like this:

```python
from sisl.viz import BandsPlot
from sisl.viz.plotly._doc_updater import fill_class_docs

fill_class_docs(BandsPlot)
```

zerothi:
> > I can't figure out how you say that your test succeeded? There were tons of missing documentations that ended up failing the test.
> 
> No, the documentation tests (some) are failing for me as well. I just said that doesn't worry me because there is a script to build the doc strings automatically using the info of the plot parameters.

Ok. We need those to be marked as xfail to remember them, but letting them fail for the moment (this is exactly what `xfail` is meant for.

> 
> > I can't figure out how you handle cases where plotly is not installed.
> 
> I don't at the moment, because the process is tightly bound to plotly right now. I'd say we can leave this like that and fix it in the refactoring when we implement the dispatchers for the plotting backend/plotting method + entry points. I'm really excited about that, because if we manage to reuse the same logic to process plots that are drawn in plotly, matplotlib and blender, while also letting the user define new ways of drawing, that would be awesome!

I agree this would be a major step forward to separate backends!

> Apart from the obvious advantatges, that would mean that anyone can develop a GUI that uses our logic regardless of the plotting backend they choose/are forced to use. I.e. https://github.com/sozykinsa/GUI4dft uses matplotlib and it would be extremely cool to harvest all the 3D rendering/manipulating knowledge of blender by using it as a GUI for geometries and grids.

:) Yes! :+1: 

> > With my latest commit I still get errors:
> 
> The `test_eigenstate_wf` and `test_hamiltonian_wf` tests I removed them here: [f511009](https://github.com/zerothi/sisl/commit/f511009ddd22f53d5a532a60c7959984fe8ed919). I don't know why you are still seeing them. Maybe you are testing an installation for which the files haven't been modified? 

Ok, I'll fix on my end ;)

> 
> > Sorry to bash all your hard work, I really mean it well!!!
> 
> That's cool, but I don't understand how you want to do it exactly, so you'll have to do it (at least do it in one file to provide an example). Do you want to do it in this PR or in a separate one?

Yeah, I will do something *eventually*, for now I'll concentrate on your 2 PR's ;)

zerothi:
Great, thanks! I'll merge!

zerothi:
@pfebrer I am now trying some more tests.

And some tests keep skipping:
```
SKIPPED [33] /opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.10.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/plotly/plots/tests/conftest.py:153: plot was not initialized so we can not run the test (plotly can be imported)
```
So plotly _can_ be imported. But I have no clue how to debug _why_ the plots were not initialized? Any hints?

zerothi:
Also where do you access `orbital.Z` ? I can't find anywhere but this is deprecated and should be `orbital.zeta`.
The reason is that we have to distinguish between atomic `Z` and orbital `zeta`. :)

Also, I have fixed some things. Could you have a look at master?

Now I get this spurious error:
```python
/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.10.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/plotly/plots/grid.py:1069: in scan
    breakpoints = np.linspace(*along_range, steps + 1)
<__array_function__ internals>:5: in linspace
    ???
```
which seems like a bug somewhere in `grid.py`?

I am now down to 15 skipping (down from 33) and 1 failing (the above).

zerothi:
Ok, I figured out the `orbital.Z` point. Now one can do both atomic number lookups (`atom.Z`) and orbital zeta (`orbital.zeta`)

The `grid.py` is still a problem.

pfebrer:
Cool that you are already modifying my code :)

`atom.Z` gives the same information as `atom.symbol`, doesn't it?

From my experience using the `PdosPlot` I've found categories extremely useful for selecting atoms. So if I want to select Z=2 I can say:

```python
pdos_plot.update_settings(requests=[
    {"atoms": AtomCategory(Z=2)},
])
```

and of course the best part is to create complex categories by combining them. So I think "Z" and "species" won't be very used in favour of categories. In fact, if orbital categories existed I would simply not create the df at all I think :)

> The grid.py is still a problem.

I will look into it!

zerothi:
> `atom.Z` gives the same information as `atom.symbol`, doesn't it?

`atom.Z` is an integer, whereas `atom.symbol` is the 2-letter designation of the element. So not exactly the same.

> From my experience using the `PdosPlot` I've found categories extremely useful for selecting atoms. So if I want to select Z=2 I can say:
> 
> ```python
> pdos_plot.update_settings(requests=[
>     {"atoms": AtomCategory(Z=2)},
> ])
> ```
> 
> and of course the best part is to create complex categories by combining them. So I think "Z" and "species" won't be very used in favour of categories. In fact, if orbital categories existed I would simply not create the df at all I think :)

Ok, I can't really figure out where the different things happen, but I agree, the categories seems very useful in this case ;)

 
> > The grid.py is still a problem.
> 
> I will look into it!

Great!

pfebrer:
> atom.Z is an integer, whereas atom.symbol is the 2-letter designation of the element. So not exactly the same.

Of course, I meant that it does not add extra "separation power" :)

> Ok, I can't really figure out where the different things happen, but I agree, the categories seems very useful in this case ;)

The atoms key is simply sanitized using the geometry. That's why it accepts categories or whatever the geometry can sanitize :) I know there's a lot of things in the code that I'm the only one who knows about, but I'm just happy you are finally able to modify some things in the code! 

PS: Having the documentation available will help here (I hope)

zerothi:
> > atom.Z is an integer, whereas atom.symbol is the 2-letter designation of the element. So not exactly the same.
> 
> Of course, I meant that it does not add extra "separation power" :)

True.

> 
> > Ok, I can't really figure out where the different things happen, but I agree, the categories seems very useful in this case ;)
> 
> The atoms key is simply sanitized using the geometry. That's why it accepts categories or whatever the geometry can sanitize :) I know there's a lot of things in the code that I'm the only one who knows about, but I'm just happy you are finally able to modify some things in the code!
> 
> PS: Having the documentation available will help here (I hope)

I hope I can get out some more information by looking at it ;)

PR #296, opened 2021-01-29
Method to calculate effective masses in BandsPlot

pfebrer:
I needed to calculate the effective masses of my bands, and I thought this would be a useful method to have in `BandsPlot`.

@zerothi I need your help to know if the calculation is correct. Especially regarding the values of **k**. Currently they come from `BandStructure.lineark()`. Is it right to use these values to fit `E` vs `k` or do I need to somehow modify `k` before fitting?

### DIFF
--- a/sisl/viz/plotly/plots/bands.py
+++ b/sisl/viz/plotly/plots/bands.py
@@ -704,6 +704,37 @@ class BandsPlot(Plot):
                 if spin in requested_spin:
                     self.draw_gap(custom_gap["from"], custom_gap["to"], color=custom_gap.get("color", None), gap_spin=spin)
 
+    def _sanitize_k(self, k):
+        """Returns the float value of a k point in the plot.
+
+        Parameters
+        ------------
+        k: float or str
+            The k point that you want to sanitize.
+            If it can be parsed into a float, the result of `float(k)` will be returned.
+            If it is a string and it is a label of a k point, the corresponding k value for that
+            label will be returned
+        
+        Returns
+        ------------
+        float
+            The sanitized k value.
+        """
+        san_k = None
+
+        try:
+            san_k = float(k)
+        except ValueError:
+            if k in self.bands.attrs["ticklabels"]:
+                i_tick = self.bands.attrs["ticklabels"].index(k)
+                san_k = self.bands.attrs["ticks"][i_tick]
+            else:
+                pass
+                # raise ValueError(f"We can not interpret {k} as a k-location in the current bands plot")
+                # This should be logged instead of raising the error
+        
+        return san_k
+
     def draw_gap(self, from_k, to_k=None, color=None, true_gap=False, gap_spin=0, save=False, **kwargs):
         """
         Function that draws a given gap in the plot.
@@ -749,16 +780,7 @@ class BandsPlot(Plot):
         # Parse the names of the kpoints into their numeric values
         # if a string was provided.
         for i, val in enumerate((from_k, to_k)):
-            try:
-                ks[i] = float(val)
-            except ValueError:
-                if val in self.bands.attrs["ticklabels"]:
-                    i_tick = self.bands.attrs["ticklabels"].index(val)
-                    ks[i] = self.bands.attrs["ticks"][i_tick]
-                else:
-                    return
-                    # raise ValueError(f"We can not interpret {val} as a k-location in the current bands plot")
-                    # This should be logged instead of raising the error
+            ks[i] = self._sanitize_k(val)
 
         if true_gap:
             # Take the energies from the gap information
@@ -867,3 +889,72 @@ class BandsPlot(Plot):
         plt.update_layout({"title": f"Delta K between bands {band1} and {band2}", 'xaxis_title': 'Delta k', 'yaxis_title': 'Energy [eV]'})
 
         return plt
+
+    def effective_mass(self, band, k, k_direction, band_spin=0, n_points=10):
+        """Calculates the effective mass from the curvature of a band in a given k point.
+
+        It works by fitting the band to a second order polynomial.
+
+        Notes
+        -----
+        Only valid if there are no band-crossings in the fitted range.  
+        The effective mass may be highly dependent on the `k_direction` parameter, as well as the
+        number of points fitted.
+
+        Parameters
+        -----------
+        band: int
+            The index of the band that we want to fit
+        k: float or str
+            The k value where we want to find the curvature of the band to calculate the effective mass.
+        band_spin: int, optional
+            The spin value for which we want the effective mass.
+        n_points: int
+            The number of points that we want to use for the polynomial fit.
+        k_direction: {"symmetric", "right", "left"}, optional
+            Indicates in which direction -starting from `k`- should the band be fitted. 
+            "left" and "right" mean that the fit will only be done in one direction, while
+            "symmetric" indicates that points from both sides will be used.
+        
+        Return
+        -----------
+        float
+            The efective mass, in atomic units.
+        """
+        from sisl.unit.base import units
+
+        # Get the band that we want to fit
+        band_vals = self.bands.sel(band=band, spin=band_spin)
+
+        # Sanitize k to a float
+        k = self._sanitize_k(k)
+        # Find the index of the requested k
+        k_index = abs(self.bands.k -k).values.argmin()
+
+        # Determine which slice of the band will we take depending on k_direction and n_points
+        if k_direction == "symmetric":
+            sel_slice = slice(k_index - n_points // 2, k_index + n_points // 2 + 1)
+        elif k_direction == "left":
+            sel_slice = slice(k_index - n_points + 1, k_index + 1)
+        elif k_direction == "right":
+            sel_slice = slice(k_index, k_index + n_points)
+        else:
+            raise ValueError(f"k_direction must be one of ['symmetric', 'left', 'right'], {k_direction} was passed")
+        
+        # Grab the slice of the band that we are going to fit
+        sel_band = band_vals[sel_slice] * units("eV", "Hartree")
+        sel_k = self.bands.k[sel_slice] - k
+
+        # Fit the band to a second order polynomial
+        polyfit = np.polynomial.Polynomial.fit(sel_k, sel_band, 2)
+
+        # Get the coefficient for the second order term
+        coeff_2 = polyfit.convert().coef[2]
+
+        # Calculate the effective mass from the dispersion relation.
+        # Note that hbar = m_e = 1, since we are using atomic units.
+        eff_m = 1 / (2 * coeff_2)
+
+        return eff_m
+
+


zerothi:
I think there are many ways of doing this. Here you are only doing it along a given path. But that forces the path to be linear. Others also take into account the full parabola curvature in 3D space. So perhaps you need to clarify what this one does and what it does not do.

pfebrer:
Hmm ok, I assumed that this would be clear since it's a unidimensional band structure plot. So it would be the effective mass in that particular direction, if that makes sense?

> But that forces the path to be linear

I don't understand what does this mean exactly. Is it possible that the calculation I added here makes no sense for some 1D band structures? If so, could you mention an example?

zerothi:
What if the position at which you want to calculate is at a point where you switch direction? i.e. at some `X` point in the bandstructure? In some cases you'll have different curvatures. Say `Gamma -> X - > K`. @ X.  
Then you don't necessarily have the same curvature from gamma -> X, as from X -> K.

pfebrer:
That's why I added the `k_direction` argument. If `k_direction` is "right" it will only fit the curvature from X -> K. And same for "left". Is this wrong?

zerothi:
> That's why I added the `k_direction` argument. If `k_direction` is "right" it will only fit the curvature from X -> K. And same for "left". Is this wrong?

Ah, ok. I don't know. I must admit, I don't have any experience in this.

sisl/viz/plotly/plots/bands.py
```diff
@@ -865,3 +888,67 @@ def _plot_Kdiff(self, band1, band2, E=None, offsetE=False):
         plt.update_layout({"title": f"Delta K between bands {band1} and {band2}", 'xaxis_title': 'Delta k', 'yaxis_title': 'Energy [eV]'})
 
         return plt
+
+    def effective_mass(self, band, k, band_spin=0, n_points=10, k_direction="symmetric"):
+        """
+        Calculates the effective mass from the curvature of a band in a given k point.
+
+        NOTE: The calculations done here are only valid if there are no band crossings in the
```
zerothi, line 896: I would change to this:
```python
Notes
-----
Only valid if there are no band-crossings in the fitted range.  
The effective mass may be highly dependent on the `k_direction` parameter.

File: sisl/viz/plotly/plots/bands.py [hunk 2]
```diff
@@ -865,3 +888,67 @@ def _plot_Kdiff(self, band1, band2, E=None, offsetE=False):
         plt.update_layout({"title": f"Delta K between bands {band1} and {band2}", 'xaxis_title': 'Delta k', 'yaxis_title': 'Energy [eV]'})
 
         return plt
+
+    def effective_mass(self, band, k, band_spin=0, n_points=10, k_direction="symmetric"):
+        """
```
zerothi, line 893: Remove newline

sisl/viz/plotly/plots/bands.py
```diff
@@ -865,3 +888,67 @@ def _plot_Kdiff(self, band1, band2, E=None, offsetE=False):
...
+            sel_slice = slice(k_index - n_points // 2, k_index + n_points // 2 + 1)
+        elif k_direction == "left":
+            sel_slice = slice(k_index - n_points + 1, k_index + 1)
+        elif k_direction == "right":
+            sel_slice = slice(k_index, k_index + n_points)
+        else:
+            raise ValueError(f"k_direction must be one of ['symmetric', 'left', 'right'], {k_direction} was passed")
+        
+        # Grab the slice of the band that we are going to fit
+        sel_band = band_vals[sel_slice] * units("eV", "Hartree")
```
zerothi, line 940: Should there be some check that the slice actually has a length of 4 or some minimum length? I.e. what if users have too few points?

sisl/viz/plotly/plots/bands.py
```diff
@@ -865,3 +888,67 @@ def _plot_Kdiff(self, band1, band2, E=None, offsetE=False):
...
+        NOTE: The calculations done here are only valid if there are no band crossings in the
+        range that is being fitted.
+
+        Parameters
+        -----------
+        band: int
+            The index of the band that we want to fit
+        k: float or str
+            The k value where we want to find the curvature of the band to calculate the effective mass.
+        band_spin: int, optional
```
zerothi, line 905: change to `spin`, I think the meaning is implicit. :)

zerothi (changes requested):
Great, some minor changes required :)

sisl/viz/plotly/plots/bands.py
pfebrer: Hehe, that's one of the problems that we'll have to address with how I currently have settings implemented.

Right now, the way I let the developer make use of plot settings is the following:

```python
class MyPlot(Plot):

    _parameters = (
        InputField(key="first_setting", ...)
    )

    def some_func(self, first_setting):
        # The first_setting argument here will receive the value of the setting by default,
        # unless another value is explicitly passed
```

After some trial and error on how to use settings, I was quite happy with this approach because it makes methods reusable from the "outside". E.g:

```python
class MyOtherPlot(Plot):

    def do_stuff(self):
        MyPlot.some_func(self, 4)
```

That is, I can still call the `some_func` function of `MyPlot` without needing to have the `first_setting` defined in `MyOtherPlot`.

It also lets `Plot` know which functions make use of `first_setting`, so when the setting is updated the appropiate functions are ran again.

However, this unfortunately leads to not being able to use some words as arguments. For example, in `BandsPlot`, `spin` is a setting. If I name this argument `spin` and no value is provided when the method is called, the value of the `spin` setting will be passed, which will result in an error :(

sisl/viz/plotly/plots/bands.py
pfebrer: Numpy will already raise an error (warning) which is quite clear if not enough (very few) points are passed to `np.polyfit`.

zerothi: Ok, fair enough ;)

sisl/viz/plotly/plots/bands.py
zerothi: Are you doing:
```python
class MyPlot(Plot):
    ...

class MyOtherPlot(Plot):
    ...
```
or 
```python
class MyPlot(Plot):
    ...

class MyOtherPlot(MyPlot):
    ...
```

The former I really think you should fix ;) Don't use stuff from non-subclassed classes. 

If it was a typo and you are actually doing the latter, then ok. :)

If you want to re-use settings it seems better to have another way of handling this.

zerothi: Wouldn't it be better if the `settings` have a hook (callback) that calls the appropriate class methods?
In that way you can *link* settings to different plots, and if they don't need linking, they get their own setting?

This would also leverage that problem of `band_spin` vs. `spin` vs. `eig_spin` ... (I think)?

pfebrer: > If you want to re-use settings it seems better to have another way of handling this.

The point was not to reuse settings, it was to be able to use a function without necessarily having all the settings defined. Silly example:

```
class BandsPlot(Plot):
    _parameters = (
        RangeInput(key="bands_range")
    )

    def draw_bands(self, bands_range):
        ...draw the bands within that range

class Band0Plot(BandsPlot):
    "A plot that draws only the first band"

    def draw_bands(self):
        BandsPlot.draw_bands(self, [0,0])
```

The point being, even though `Bands0Plot` is an extension of `BandsPlot`, it doesn't make sense to allow a `bands_range` value other than `[0,0]`, therefore it shouldn't be a setting. We still should be able to use the method to draw the bands.

> Wouldn't it be better if the settings have a hook (callback) that calls the appropriate class methods?
> In that way you can link settings to different plots, and if they don't need linking, they get their own setting?

Would you declare explicitly which method does each setting need to call when updated? That's what I wanted to avoid to make developement easier.

zerothi: > ```python
> class BandsPlot(Plot):
>     _parameters = (
>         RangeInput(key="bands_range")
>     )
> 
>     def draw_bands(self, bands_range):
>         ...draw the bands within that range
> 
> class Band0Plot(BandsPlot):
>     "A plot that draws only the first band"
> 
>     def draw_bands(self):
>         BandsPlot.draw_bands(self, [0,0])
> ```
> 
> The point being, even though `Bands0Plot` is an extension of `BandsPlot`, it doesn't make sense to allow a `bands_range` value other than `[0,0]`, therefore it shouldn't be a setting. We still should be able to use the method to draw the bands.

But here it would make sense to just have `BandsPlot` inherit from `Bands0Plot`.

> > Wouldn't it be better if the settings have a hook (callback) that calls the appropriate class methods?
> > In that way you can link settings to different plots, and if they don't need linking, they get their own setting?
> 
> Would you declare explicitly which method does each setting need to call when updated? That's what I wanted to avoid to make developement easier.

I still need to wrap my head around all of this (as you may see from my questions).
Since you still have to call a method explicitly, why not have a decorator taking care of this. That could hook in settings etc. for specific methods?

pfebrer: > But here it would make sense to just have BandsPlot inherit from Bands0Plot.

Why? `Band0Plot` is more specific than `BandsPlot` so the inheritance should be `BandsPlot` -> `Band0Plot`, no?

> Since you still have to call a method explicitly, why not have a decorator taking care of this. That could hook in settings etc. for specific methods?

Actually I'm using a decorator:

https://github.com/zerothi/sisl/blob/229c04680f41254673999629d349104274282fc8/sisl/viz/plotly/configurable.py#L892-L948

which is applied to every plot (or session) method here:

https://github.com/zerothi/sisl/blob/229c04680f41254673999629d349104274282fc8/sisl/viz/plotly/configurable.py#L405-L407

i.e. when you define a class that inherits from Configurable.

But you mean something more explicit like:

```python
class BandsPlot(Plot):
 
    _parameters = (
        RangeInput(key="bands_range")
    )
    
    @uses("bands_range")
    def some_func(self, bands_range):
        ...
```

right?

zerothi: > > But here it would make sense to just have BandsPlot inherit from Bands0Plot.
> 
> Why? `Band0Plot` is more specific than `BandsPlot` so the inheritance should be `BandsPlot` -> `Band0Plot`, no?
>

I would argue the other way around.
`Band0Plot` is your basic bands plot. Then you add functionality in `BandsPlot`. Hehe.

> But you mean something more explicit like:
> 
> ```python
> class BandsPlot(Plot):
>  
>     _parameters = (
>         RangeInput(key="bands_range")
>     )
>     
>     @uses("bands_range")
>     def some_func(self, bands_range):
>         ...
> ```
> 
> right?

Here you could do something like:
```python
@uses("bands_range")
def some_func(self):
    do whatever
```
and the decorator would add a `kwarg` equivalent to `bands_range`

zerothi: I really need to figure out why you have these settings implemented like this to better get an overview :)

pfebrer: > Band0Plot is your basic bands plot. Then you add functionality in BandsPlot. Hehe.

But `Band0Plot` is nothing without a general method that draws bands, which should be implemented in `BandsPlot` :thinking:

> @uses("bands_range")
> def some_func(self, **kwargs):
>     do whatever

```python
@uses("bands_range")
def some_func(self, **kwargs):
    do whatever
```

right?

> I really need to figure out why you have these settings implemented like this to better get an overview :)

Basically the only difference between decorators and the current implementation is that I thought it would be messy if there was decorators all around, and it would be more difficult for begginers to understand and therefore more difficult for them to create plots. But maybe it outweights the possibility of having bugs that are difficult to identify with the current implicit approach.

zerothi: You don't need the above. See e.g. this:
```python
from functools import wraps
def add_argument(*new_args):
    def true_decorator(func):
        @wraps(func)
        def func_add_argument(*args, **kwargs):
            for new_arg in new_args:
                arg = kwargs.pop(new_arg)
                print(new_arg, arg)
                # do something with arg
            return func(*args, **kwargs)
        return func_add_argument
    return true_decorator


@add_argument("temp")
def no_args():
    print("Hi from no_args")

no_args(temp="sthaoe")
```

In any case it is difficult. I think it has to be the simplest and most comprehensible one since that is easier to maintain ;)
But perhaps there are cases where your method is superior. :) I just need to get started on understanding.

pfebrer: Aah ok but then if you want to use the `"temp"` setting inside `no_args` you need to get the value of the setting yourself from `self.settings` (`self.settings["temp"]`)?. So you have to rely on the developer not forgetting to say what settings they are using for updates to run properly.

zerothi: Could you perhaps guide me towards one of the simpler classes where this is used so I can learn ;)

sisl/viz/plotly/plots/bands.py
pfebrer: What do you mean? No empty line between methods? :sweat_smile:

sisl/viz/plotly/plots/bands.py
pfebrer: In a call or here? :thinking:

sisl/viz/plotly/plots/bands.py
zerothi: In documentation. The first line is a special line for docs...

pfebrer: Aah ok, sorry! :sweat_smile:

pfebrer: I'll have to do a PR to correct this basically everywhere then :facepalm:

zerothi: Yeah, I have noticed ;) I corrected many in the initial merge. But stopped at some point. I will have to see if it actually matters, but I think it does for instance for `help` calls etc. Everywhere I find that the first line is _special_.

sisl/viz/plotly/plots/bands.py
zerothi: Perhaps we should make a call at some point. For now I really need to concentrate on fixing the documentation up and get plotly integrated with docs.

zerothi:
This will get merged once #293 is merged.

pfebrer:
Ok, by now I have left the argument as `band_spin` and removed the default from `k_direction`.

PR #300, opened 2021-02-06
Make plotly docs work

pfebrer:
This comes from https://github.com/zerothi/sisl/pull/291. Plotly docs should build now


pfebrer:
Btw you need to install sisl_gui from the repo for it to work at the moment:

```
pip install git+https://github.com/pfebrer/sisl-gui.git
```

Or discard the notebook that talks about the GUI

zerothi:
The point was that no workflow files should be present... Isn't this exactly the same with a different title? ;)

pfebrer:
Yep it sneaked, now it's fine!

zerothi (approved):

PR #304, opened 2021-02-09
Fixing bug in spin treatment for PdosPlot

pfebrer:
There were two bugs:
- `pdos_plot.split_DOS(on="spin")` didn't work properly. The same bug caused `pdos_plot.split_DOS(...)` to cancel the ability to use the spin parameter in requests anymore.
- PDOS calculated from the hamiltonian was always interpreted to be spin polarized.

The first bug was not detected by the tests because there is no test that involves spin. Probably I should generate the `SrTiO3.PDOS` file from a spin polarized calculation. I will do it tomorrow, and then I will also need to change something in the `test_pdos.py` file.

The second bug sneaked through tests due to a magical compensation by the first bug.

Also, as @arsalan-akhtar reported this bug, he also noted (I already had it in mind for a long time) that it is usual to have "spin down DOS" with a negative value. I introduced a `scale` parameter to the request dict so that the user now can multiply the DOS by whatever number they want:

```python
pdos_plot.update_settings(requests=[
    {"spin": [0]},
    {"spin": [1], "scale": -1}
])
```

in the future there should probably be an "invert_spin_down" setting that automatically sets up the scale values, but for now this should be fine.

**PS**: I'm in the process of thinking about the things of your PR, but this bug was quite important and it took me a while to realize what was going on so I wanted to fix it while it was fresh in my mind.

### DIFF
--- a/sisl/viz/plotly/input_fields/dropdown.py
+++ b/sisl/viz/plotly/input_fields/dropdown.py
@@ -155,10 +155,10 @@ class SpinSelect(DropdownInput):
     }
 
     _options = {
-        Spin.UNPOLARIZED: [],
+        Spin.UNPOLARIZED: [{"label": "Total", "value": 0}],
         Spin.POLARIZED: [{"label": "↑", "value": 0}, {"label": "↓", "value": 1}],
-        Spin.NONCOLINEAR: [{"label": val, "value": val} for val in ("x", "y", "z")],
-        Spin.SPINORBIT: []
+        Spin.NONCOLINEAR: [{"label": val, "value": val} for val in ("sum", "x", "y", "z")],
+        Spin.SPINORBIT: [{"label": val, "value": val} for val in ("sum", "x", "y", "z")]
     }
 
     def __init__(self, *args, only_if_polarized=False, **kwargs):
--- a/sisl/viz/plotly/input_fields/queries.py
+++ b/sisl/viz/plotly/input_fields/queries.py
@@ -260,15 +260,12 @@ class OrbitalQueries(QueriesInput):
         # Get the tadatframe
         df = self.orb_filtering_df
 
-        # If there is no spin in this df, we are going to introduce it to avoid problems.
-        if "spin" not in df:
-            df["spin"] = 0
-
         # Filter the dataframe according to the constraints imposed by the kwargs,
         # if there are any.
         if kwargs:
-            query = ' & '.join([f'{self._keys_to_cols.get(k, k)}=={repr(v)}' for k, v in kwargs.items()])
-            df = df.query(query)
+            query = ' & '.join([f'{self._keys_to_cols.get(k, k)}=={repr(v)}' for k, v in kwargs.items() if self._keys_to_cols.get(k, k) in df])
+            if query:
+                df = df.query(query)
 
         # If + is in key, it is a composite key. In that case we are going to
         # split it into all the keys that are present and get the options for all
@@ -276,11 +273,38 @@ class OrbitalQueries(QueriesInput):
         # the possible combinations of the keys.
         keys = [self._keys_to_cols.get(k, k) for k in key.split("+")]
 
-        options = df.drop_duplicates(subset=keys)[keys].values.squeeze()
+        # Spin values are not stored in the orbital filtering dataframe. If the options
+        # for spin are requested, we need to pop the key out and get the current options
+        # for spin from the input field
+        spin_in_keys = "spin" in keys
+        if spin_in_keys:
+            spin_key_i = keys.index("spin")
+            keys.remove("spin")
+            spin_options = self.get_param("spin").options
 
-        # If there is only one option, squeeze converts it to a number, so
-        # we need to make sure there is at least 1d
-        options = np.atleast_1d(options)
+            # We might have some constraints on what the spin value can be
+            if "spin" in kwargs:
+                spin_options = set(spin_options).intersection(kwargs["spin"])
+
+        # Now get the unique options from the dataframe
+        if keys:
+            options = df.drop_duplicates(subset=keys)[keys].values.astype(np.object)
+        else:
+            # It might be the only key was "spin", then we are going to fake it
+            # to get an options array that can be treated in the same way.
+            options = np.array([[]], dtype=np.object)
+
+        # If "spin" was one of the keys, we are going to incorporate the spin options, taking into
+        # account the position (column index) where they are expected to be returned.
+        if spin_in_keys:
+            options = np.concatenate([np.insert(options, spin_key_i, spin, axis=1) for spin in spin_options])
+
+        # Squeeze the options array, just in case there is only one key
+        # There's a special case: if there is only one option for that key,
+        # squeeze converts it to a number, so we need to make sure there is at least 1d
+        if options.shape[1] == 1:
+            options = options.squeeze()
+            options = np.atleast_1d(options)
 
         return options
 
@@ -293,7 +317,7 @@ class OrbitalQueries(QueriesInput):
 
         return filtered_df.index
 
-    def _split_query(self, query, on, only=None, exclude=None, query_gen=None, **kwargs):
+    def _split_query(self, query, on, only=None, exclude=None, query_gen=None, ignore_constraints=False, **kwargs):
         """
         Splits a query into multiple queries based on one of its parameters.
 
@@ -315,6 +339,13 @@ class OrbitalQueries(QueriesInput):
             request that this method has come up with and gets a chance to do some modifications.
 
             This may be useful, for example, to give each request a color, or a custom name.
+        ignore_constraints: boolean or array-like, optional
+            determines whether constraints (imposed by the query that you want to split) 
+            on the parameters that we want to split along should be taken into consideration.
+
+            If `False`: all constraints considered.
+            If `True`: no constraints considered.
+            If array-like: parameters contained in the list ignore their constraints.
         **kwargs:
             keyword arguments that go directly to each new request.
 
@@ -335,10 +366,15 @@ class OrbitalQueries(QueriesInput):
         # because these will be our constraints. If a parameter is set to None or not
         # provided, we have no constraints for that parameter.
         constraints = {}
-        for key in on:
-            val = query.get(key, None)
-            if val is not None:
-                constraints[key] = val
+        if ignore_constraints is not True:
+
+            if ignore_constraints is False:
+                ignore_constraints = ()
+
+            for key in filter(lambda key: key not in ignore_constraints, on):
+                val = query.get(key, None)
+                if val is not None:
+                    constraints[key] = val
 
         # Knowing what are our constraints (which may be none), get the available options
         values = self.get_options("+".join(on), **constraints)
--- a/sisl/viz/plotly/plots/pdos.py
+++ b/sisl/viz/plotly/plots/pdos.py
@@ -20,8 +20,9 @@ class PdosPlot(Plot):
     pdos_file: pdosSileSiesta, optional
         This parameter explicitly sets a .PDOS file. Otherwise, the PDOS file
         is attempted to read from the fdf file
-    tbt_nc: optional
-        TBtrans output for plotting PDOS from NEGF calculations
+    tbt_nc: tbtncSileTBtrans, optional
+        This parameter explicitly sets a .TBT.nc file. Otherwise, the PDOS
+        file is attempted to read from the fdf file
     Erange: array-like of shape (2,), optional
         Energy range where PDOS is displayed.
     nE: int, optional
@@ -39,7 +40,12 @@ class PdosPlot(Plot):
         Erange).
     requests: array-like of dict, optional
         Here you can ask for the specific PDOS that you need.
-        TIP: Queries can be activated and deactivated.
+        TIP: Queries can be activated and deactivated.   Each item is a
+        dict. Structure of the expected dicts:{         'name':
+        'species':          'atoms':          'orbitals':          'spin':
+        'normalize':          'color':          'linewidth':          'dash':
+        'split_on':          'scale': The final DOS will be multiplied by
+        this number. }
     root_fdf: fdfSileSiesta, optional
         Path to the fdf file that is the 'parent' of the results.
     results_path: str, optional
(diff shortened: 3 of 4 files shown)


zerothi:
This is perfect.
I much prefer smaller PR's that fixes individual problems, so definitely the right way of doing things!

I would suggest you put the scale also for the first spin channel. It may be useful for other things as well. So why not have it both places? :)

sisl/viz/plotly/plots/pdos.py
```diff
@@ -177,6 +183,13 @@ class PdosPlot(Plot):
                         "isSearchable": True,
                         "options": [{"value": option, "label": option} for option in ("species", "atoms", "Z", "orbitals", "spin", "n", "l", "m", "zeta")]
                     }
+                ),
+
+                FloatInput(
+                    key="scale", name="Scale",
+                    default=1,
+                    params={"min": None},
+                    help="The final DOS will be multiplied by this number. Note that it can also be a negative number."
```
zerothi, line 192: Just remove `Note that it can also be a negative number` it is a scale factor, so no need for extended documentation. :)

sisl/viz/plotly/plots/pdos.py
```diff
@@ -39,7 +40,12 @@ class PdosPlot(Plot):
         Erange).
     requests: array-like of dict, optional
         Here you can ask for the specific PDOS that you need.
-        TIP: Queries can be activated and deactivated.
+        TIP: Queries can be activated and deactivated.   Each item is a
+        dict. Structure of the expected dicts:{         'name':
+        'species':          'atoms':          'orbitals':          'spin':
+        'normalize':          'color':          'linewidth':          'dash':
+        'split_on':          'scale': The final DOS will be multiplied by
+        this number. Note that it can also be a negative number. }
```
zerothi, line 48: Also here, remove `Note that it can also be a negative number`

sisl/viz/plotly/plots/pdos.py
```diff
@@ -313,11 +326,14 @@ def _after_read(self):
         """
         from xarray import DataArray
 
-        # Normalize self.PDOS to do the same treatment for both spin-polarized and spinless simulations
-        self.spin_polarized = len(self.PDOS.shape) == 3
+        # Check if the PDOS contains spin resolution (there should be three dimensions,
+        # and the first one should be the spin components)
+        self.spin_polarized = self.PDOS.squeeze().ndim == 3
```
zerothi, line 331: Are you also checking for non-colinear?
The last dimension may be either 1, 2 or 4?

zerothi (changes requested):

sisl/viz/plotly/plots/pdos.py
pfebrer: Do you have an example of a non-colinear PDOS file? What are each of the 4 spin channels in the non-colinear case?

pfebrer:
The scale can be used in any request! What I wrote there was just an example of how one may use it.

sisl/viz/plotly/plots/pdos.py
zerothi: See [here](https://github.com/zerothi/sisl/blob/293969f45ef21af2d63a04ca60a7dda5ab673008/sisl/io/siesta/pdos.py#L102)
and the following lines.

pfebrer: Aah ok, is the sum channel just the sum of x, y and z? Or is it more complex?

zerothi: No, it is more complex. The total DOS (sum) may be 4, while all the rest may be 0. You can think of it like a polarized calculation (here x and y are always 0). For a spin polarized up=1, down=1 one would have sum=2 and z = up - down = 0.
For up=1, down=0.5 one would have sum=1.5 and z=0.5.

So it is the *net* spin of that orbital. However one will always have `|x**2+y**2+z**2|**0.5 <= sum`.

pfebrer: Ok, thanks! I have a non-colinear PDOS file and I'm trying to see what I need to change to make it work now :)

pfebrer:
Ok, let's see if you agree with the implementation:

Normally, if no spin was specified, I just aggregated the spin axis (either doing the mean or a sum, depending on `normalize`). However this doesn't make sense with non-colinear spins. Therefore, what I did is that when no spin is passed and the PDOS contains non-colinear information `"spin": ["sum"]` is added to the request so that it only takes that contribution by default.

So, if I grab a non-colinear PDOS file and split the PDOS on species:

```python
plot = sisl.get_sile("non-colinear_SrTiO3.PDOS").plot()
plot.split_DOS(on="species", name="$species")
```
I get the PDOS by species, using only the "sum" spin channel.

image

You can see that the spin parameter of the request has been added:

image

Also, of course if a value is passed for the spin parameter, that value is respected:

```python
plot.split_DOS(on="species", name="$species", spin=["x"])
```
image

Now, the only problem I saw with that implementation is that one desired to split the requests that they got further along spin, that would not be possible because there would always be the constraint of the spin channel being "sum":

image

So I added the argument `ignore_constraints` so that one can say: for the parameter I want to split on, please ignore the constraints of the existing query:

```python
plot.split_requests("Sr", on="spin", ignore_constraints=True)
```

image

One can choose to ignore also only specific constraints:

```python
plot.split_DOS(on="species", l=[0], name="$species, l=0")
plot.split_requests(0, on="spin+l", ignore_constraints=["spin"])
```

image

You can see here the constraint about l being 0 is still respected, vs:

```python
plot.split_requests(0, on="spin+l", ignore_constraints=True)
```

image

Let me know if there's something that you don't like about the implementation and the way I tried to made spin splitting of the requests possible.

sisl/viz/plotly/plots/pdos.py
zerothi: > By the way, I didn't use your `read_data(as_dataarray=True)` because it seemed like there wasn't an atom index, so I didn't understand. For two orbitals that have the same parameters but are in a different atom, is the DOS information summed?

No, true there isn't. But that could readily be added.
```python
a = sile.read_data(as_dataarray=True)
geom = a.attrs.geometry
atom_idx = geom.o2a(np.arange(geom.no))
```
done.
I would actually like to reduce number of coordinates since the stored matrix in `xarray` is a dense matrix. So adding a new coordinate needs careful thinking.
I am

sisl/viz/plotly/input_fields/dropdown.py
```diff
@@ -155,9 +155,9 @@ class SpinSelect(DropdownInput):
     }
 
     _options = {
-        Spin.UNPOLARIZED: [],
+        Spin.UNPOLARIZED: [{"label": "Total", "value": 0}],
         Spin.POLARIZED: [{"label": "↑", "value": 0}, {"label": "↓", "value": 1}],
-        Spin.NONCOLINEAR: [{"label": val, "value": val} for val in ("x", "y", "z")],
+        Spin.NONCOLINEAR: [{"label": val, "value": val} for val in ("sum", "x", "y", "z")],
         Spin.SPINORBIT: []
```
zerothi, line 161: spin-orbit is the same as noncolinear

sisl/viz/plotly/input_fields/queries.py
```diff
@@ -315,6 +337,13 @@ def _split_query(self, query, on, only=None, exclude=None, query_gen=None, **kwa
             request that this method has come up with and gets a chance to do some modifications.
 
             This may be useful, for example, to give each request a color, or a custom name.
+        ignore_constraints: boolean or array-like, optional
+            determines whether constraints (existing in the query) on the parameters that we want 
+            to split along should be taken into consideration.
+
+            If `False`: all constraints considered.
+            If `True`: no constraints considered.
+            If array-like: parameters contained in the list ignore their constraints.
```
zerothi, line 346: while this seems to solve the problem, I still don't get why ;)
I need to look into your request handles. Perhaps they could be restructured for the same effect with less clutter.

In particular it seems odd that the original one added l=1 when l=0 was requested explicitily.

sisl/viz/plotly/plots/pdos.py
```diff
@@ -313,24 +326,33 @@ def _after_read(self):
...
 
-        # Normalize self.PDOS to do the same treatment for both spin-polarized and spinless simulations
-        self.spin_polarized = len(self.PDOS.shape) == 3
+        # Check if the PDOS contains spin resolution (there should be three dimensions,
+        # and the first one should be the spin components)
+        self.spin = sisl.Spin.UNPOLARIZED
+        if self.PDOS.squeeze().ndim == 3:
+            self.spin = {
+                2: sisl.Spin.POLARIZED,
+                4: sisl.Spin.NONCOLINEAR
```
zerothi, line 335: and spin-orbit

zerothi (changes requested):
For now this is fine. It seems to me the requests needs to figure this out automatically.
But lets start with this. :)

sisl/viz/plotly/input_fields/queries.py
pfebrer: What do you mean the original one? `ignore_constraints=True` is not "the original one", the default value of `ignore_constraints` is `False`.

What seems odd about ignoring a constraint if the user told us to ignore it? hahaha

zerothi: It just seems obvious that `l=0` would ignore different `l` values otherwise.

Again, you have to see me as a new user, and a new user would not see this connection :)

sisl/viz/plotly/plots/pdos.py
pfebrer: Soif I understand correctly, this implementation also makes sense for spin-orbit and there is no way to distinguish between spin-orbit and non-colinear from the shape of the PDOS array?

zerothi: exactly :)

sisl/viz/plotly/input_fields/queries.py
pfebrer: Yes, that's why the **default behavior is to ignore the other `l` values**, therefore applying the constraint of `l=0` because the request that we are splitting imposes that constraint on `l`.

Only when you say `ignore_constraints=True` is it that `l` becomes free to be any value when you split on it.

I don't understand what the problem is :confused:

zerothi: The problem is that users (including me) don't understand what `ignore_constraints` may be used for?

And to me it seems that ignore_constraints should always be `False`, when is `True` useful?

Also, from the doc's in `split_query` there is no other mention of `constraints`. So it is not clear what it refers to? Is it used together with `on` or something else? And also, how is it connected to `kwargs`.
:)

pfebrer: > when is True useful

Mainly in non-colinear spin PDOS, that's why I introduced it. Automatically you get a request that has `spin=["sum"]` unless you specify a channel, because that's what you want (you don't want "sum" + "x" + "y", + "z"). But then if you split that on `spin`, you have the constraint that spin can only be `"sum"`, so basically you wouldn't split anything. That's why you need to ignore that constraint if you want the `x`, `y` and `z` channels for that request.

It may also be useful when you have a request like `{"atoms": [0, 1, 5, 8], "n": [3], "l": [0], "zeta": [1]}` and you want to see the DOS of that same request across `l`, for example. 

But the main point is to overcome the fact that splitting `spin=["sum"]` doesn't result in `spin=["x", "y", "z"]`.

> Also, from the doc's in split_query there is no other mention of constraints. So it is not clear what it refers to? Is it used together with on or something else? And also, how is it connected to kwargs.
> :)

In the `PdosPlot.split_requests` method:

> ignore_constraints: boolean or array-like, optional
>     
> determines whether constraints (**existing in the request to be splitted**) 
> on the parameters that we want to split along should be taken into consideration.
> If `False`: all constraints considered.
> If `True`: no constraints considered.
> If array-like: parameters contained in the list ignore their constraints.

How would you say it? Maybe "imposed by the request to be splitted"?

I don't know, this is a complex use case, and you will only come across it if you need to use it, where you will already understand what your problem is. Otherwise you don't even realize it's there.

sisl/viz/plotly/plots/pdos.py
pfebrer: Then I don't know what to do here, or if I should do anything :sweat_smile:

zerothi: Are true. Here it is ok.

sisl/viz/plotly/input_fields/dropdown.py
zerothi: But what about here? why the `Spin.SPINORBIT` box? I guess this should have the same as noncolinear?

sisl/viz/plotly/input_fields/queries.py
zerothi: It seems that I am missing the implementation details. I still have no idea what you are doing ;)

I'll have to read in the queries file regarding these details, I guess?

From the looks of it, it is very complicated :)
Why wouldn't users just add `spin=["sum", "x", ...]` explicitly when they want so?

zerothi:
Is it complete now?

sisl/viz/plotly/input_fields/dropdown.py
pfebrer: Yes, I will add it, but for now it will have no effect on anything I believe.

sisl/viz/plotly/input_fields/queries.py
pfebrer: Yes I agree that the `OrbitalQueries` input is not simple, but that's the way I managed to handle all the corner cases and make queries behave like the user would expect. (note that is not used only by `PdosPlot`, also by `FatbandsPlot` and I'm quite happy with the consistency it achieved).

Maybe you find an easier way of handling all this :)

pfebrer:
I just have to add the tests for spin polarized and non-colinear PDOS files. Then it will be ready

sisl/viz/plotly/input_fields/queries.py
zerothi: Ok. There is just lots to it :) And new for me, so lots of questions. It is very difficult to generalize and make things simple in a GUI interface like this :)
So all good :)

pfebrer:
Done!

zerothi:
I'll merge in (with rebasing)

Issue #266, opened 2020-09-16
Plotting/GUI CLI should they be separate or combined?


zerothi:
We need to figure out whether we need two CLI for plotting and for GUI handling.

I am starting to more advocate one CLI for both.
```bash
sgui ase
sgui matplotlib/plot
sgui plotly
```
something like this.

Since we are in doubt of the placement of individual functionality, users will without a doubt also!! :)

Please chip in here! ;)
@pfebrer96 @jonaslb

_Originally posted by @zerothi in https://github.com/zerothi/sisl/pull/221#discussion_r487800251_

jonaslb:
I haven't really dived much into the visualization/plotting module of Pol, so my insight there is a bit limited.
But in the end I think plotting graphs is quite different from viewing/editing geometries so in that sense I guess it would make sense to have them be separate clis.

Also it's true (Nick's comment) that the editing features in ase gui are very limited. I suppose it's more for getting a quick look at a structure in a gui that can give more or less all the geometry information one needs (which is quite often nice imo).

pfebrer96:
> But in the end I think plotting graphs is quite different from viewing/editing geometries so in that sense I guess it would make sense to have them be separate clis.

I have the same opinion. I would do `splot struct.bands` when I want to get a quick plot of a bands file, no need for a GUI. In the current implementation of `splot` in the viz module you can also pass the settings as `--arguments` and even use the presets that you have previously defined for yourself, as well as saving the plot or writing it to an image. So I would say that is about all the functionality you would need for just getting quick plots without the overhead of a GUI (quick doesn't mean simple and bad, they can be as beautiful and complex as the presets you have defined).

zerothi:
But what about end-users who don't really know the difference?

I am a bit worried that users will confuse the two.  
Could we perhaps have a middle ground `sdisplay` or `sview`, or something similar? Since all the functionality is passed down to sub-parsers I don't think we need two cli's.

My point being that while matplotlib plots are minor, they still do provide some level of interaction (change of limits etc.) so in that respect it _may_ be considered a GUI.

Would you be very much against having `sgui` as the cli for everything?

pfebrer96:
Ok, I understand that it is hard to objectively draw a hard line between what is a GUI and what is not. The main problem is probably that we leave in a very nice era where plots themselves are interactive :)

[...]

PR #312, opened 2021-03-09
ENH: Detaching plotly from the processing and allowing other drawers.

pfebrer:
First of all, sorry to give you more work to review :)

Second: The idea is actually working and I'm excited to have this implemented!

PR introduction
-------
Basically, as we were discussing in https://github.com/zerothi/sisl/issues/266#issuecomment-726183388 and the following comments, this PR is a first sketch to have three different steps (not necessarily with these names): 
- **Entry**: There are different entries that read from different sources. They all normalize the data for the next stage. Already implemented.
- **Processing**: Processes the data according to settings (which can be updated) to generate the data ready to plot. 
- **Draw**: Different drawers can be implemented for the same plot.

The entry and draw phases would allow the user to register new methods.

Here is the diagram that I draw initially to explain the concept:

image

In this PR, the separation of the drawing parts from the rest is done so that we can have this distinction.

What has been done
---------

- I have removed everything related to plotly from the main files.
- I have moved all the plots to `sisl.viz`
- I created a new submodule `sisl.viz.drawers` where the different drawers are implemented. Right now you can find a folder for `plotly` and another for `matplotlib`.
- There is a file `sisl/viz/drawers/plotly/drawer.py` that contains the generic things for `plotly` figures. Same for matplotlib.
- `sisl/viz/drawers/plotly/drawers/*` contains the specific drawers for each plot (right now only bands), which are then registered to the corresponding plotting class (in the same file). Same for matplotlib.
- When a subclass of `Plot` is defined (i.e. in `Plot.__init_subclass__`), it receives a drawers manager, which will be waiting for new drawers to be registered. This drawers manager is defined in `sisl.viz.drawers._plot_drawers`. I just did it as simple as I could, waiting for you to reimplement it as you find suitable.
- **Hasn't been done but needs to be done**: The `PlotEngine` in `sisl.viz._plotables` doesn't make sense anymore because `drawer` is just a regular setting that you can update on the fly. 

**The only specific plot to which the changes have been implemented for now is the bands plot.** This is because I wanted to have a proof of concept so that you can give your opinion and propose different approaches. After we agree with what's the best approach, I can do it with the rest of plots. Presumably it is a significant amount of work so it's better to first have a consensus :)

How it works
-----------------
```python
import sisl
# Import the visualization module
import sisl.viz
# Import the drawers that you want, otherwise they are not loaded
# This could be somehow automatized with environment variables
import sisl.viz.drawers.plotly
import sisl.viz.drawers.matplotlib

# The plot is drawn with the default drawer (plotly for now)
plot = sisl.get_sile("myfile.bands").plot()
plot.show()
```
image

```python
plot.update_settings(drawer="matplotlib")
plot.show()
```

image

The settings of `BandsPlot` of course work regardless of the drawer. If you want to render subplots, this is also drawer-aware:

```python
# Draw some subplots
plot = sisl.get_sile("myfile.bands").plot(
    bands_color=["red", "orange", "green", "purple"], gap=[True, False, False, True], subplots=["bands_color", "gap"],
    gap_color="black", rows=2,
)
plot.show()
```
image

```python
plot = sisl.get_sile("myfile.bands").plot(
    bands_color=["red", "orange", "green", "purple"], gap=[True, False, False, True], subplots=["bands_color", "gap"],
    gap_color="black", rows=2, drawer="matplotlib"
)
```
image

By the way: I have tried to uninstall plotly and the bands work with the `matplotlib` drawer :)

Looking forward to your feedback!


zerothi:
Looks amazing! Thanks!

Yes, more work!! ;)
I'll probably have to post-pone this until after TS workshop, time is limited ;)
But just a comment, could we name it `backend` instead of `drawer`? I am imagining this could be useful for other stuff as well :)

And great you only started with one. Then we can have an iteration there!

pfebrer:
> But just a comment, could we name it backend instead of drawer?

I was thinking about that but since you can write as many different drawers that use the same plotting framework for a given plot, `backend` seems too restrictive because it seems to be synonim of plotting framework in this case.

So `backend` seemed both too generic in the general picture (data processing can also be seen as a backend) and too restrictive in the picture of plotting (might give the idea that the mapping "plotting framework -> backend" should be one to one, and it's not the case). That's why I chose not to use it.

> I am imagining this could be useful for other stuff as well :)

What do you have in mind? :thinking:

zerothi:
> > But just a comment, could we name it backend instead of drawer?
> 
> I was thinking about that but since you can write as many different drawers as you want that use the same plotting framework for a given plot, `backend` seems too restrictive because it seems to be synonim of plotting framework in this case.

True, but you would still have to name the `drawer` differently for each of them, `d1` vs `d2` etc.

> So `backend` seemed both too generic in the general picture (data processing can also be seen as a backend) and too restrictive in the picture of plotting (might give the idea that the mapping "plotting framework -> backend" should be one to one, and it's not the case). That's why I chose not to use it.
> 
> > I am imagining this could be useful for other stuff as well :)
> 
> What do you have in mind?

For exactly the general picture you mention, there is no need to make users think this is restricted to drawing stuff.  
Some drawers may not have a direct API back end and one would have to resort to temporary files + shell commands. While that may be interpreted as a drawer, I would still consider it a backend.

I kind of like the generality here. It is what matplotlib uses as words for different frameworks. So it is also a matter of choosing a common word. So I have to say, backend would be the proper word :)

pfebrer:
Hey Nick, do you think we can work on this before the SIESTA school? I would like to present it there :sweat_smile:

zerothi (commented):
Yes, i think we should finish this.

However, we still have to decide on "backend" name. I would still like the name to be backend.

zerothi:
Could you also rebase?

pfebrer:
Great, thank youuuu! I will rename it to backend.

pfebrer:
Now the path to a given drawer is `sisl/viz/drawers/plotly/drawers/bands.py`.

I understand that the first "drawers" could be replaced by `backends`, but the second one?

Also how would you rename the classes? `MatplotlibBandsDrawer` -> `MatplotlibBandsBackend`?

zerothi (commented):
I'll probably have some more stuff. But this `drawer` vs. `backend` seems like the most pressing thing?

sisl/viz/drawers/matplotlib/drawers/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .bands import *
+
+__all__ = []
```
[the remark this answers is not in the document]
pfebrer: They are registered on the plot class, so there's no point in exposing them. This import is just so that the backends get registered.

pfebrer:
Ok, the renaming is done. The next things are: 

- how should the sisl-provided backends be loaded. Should we just load them if the needed packages are there? I.e. if `matplotlib` is available, matplotlib backends are loaded on `import sisl.viz`.
- do you think it's a good idea to have abstract classes to define the methods that each particular backend should have?. E.g. an abstract `BandsBackend` that bands backends have to inherit from.

sisl/viz/drawers/matplotlib/drawers/__init__.py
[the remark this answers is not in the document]
zerothi: What about sub-classing them?

zerothi:
> * how should the sisl-provided backends be loaded. Should we just load them if the needed packages are there? I.e. if `matplotlib` is available, matplotlib backends are loaded on `import sisl.viz`.

I think we shouldn't load unless a user wants too do it, i.e. lazy load when doing `plot(object)` and/or manual imports.
The reason is that sisl may be used for post-processing and loading GUI's is a waste of time.
Better be safe than sorry.

> * do you think it's a good idea to have abstract classes to define the methods that each particular backend should have?. E.g. an abstract `BandsBackend` that bands backends have to inherit from.

Definitely!!! :)

pfebrer:
I'm not saying to load them on `import sisl`, I'm saying to load them on `import sisl.viz`. I think at that point is safe to assume that they want the visualization features and half a second is not a significant time

pfebrer:
I propose that at that point everything is loaded by default and if you want you can block the auto loading with env variables. But I think the average user would want everything to be loaded at that point without any extra configuration.

zerothi:
> I'm not saying to load them on `import sisl`, I'm saying to load them on `import sisl.viz`. I think at that point is safe to assume that they want the visualization features and half a second is not a significant time

This is fine for me.

zerothi:
@pfebrer could you unmark "draft"

pfebrer:
By the way, in this PR or a subsequent one I would like to finally decide what the CLI for plotting should look like, now that everything will be unified.

sisl/viz/drawers/matplotlib/drawers/__init__.py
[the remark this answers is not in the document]
pfebrer: Then you can explicitly import them, no? I don't think someone who wants to subclass `MatplotlibBandsBackend` will do

```python
from sisl.viz.drawers.matplotlib import *

class MyBackend(MatplotlibBandsBackend):
    ...
```

That would not be very good practice...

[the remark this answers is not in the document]
pfebrer: But yes we can expose them if you want.

zerothi:
> By the way, in this PR or a subsequent one I would like to finally decide what the CLI for plotting should look like, now that everything will be unified.

See [here](https://github.com/zerothi/sisl/issues/266#issuecomment-726068775) I think we should go for `sisl plot/gui`

zerothi:
In a subsequent one! :)

sisl/viz/drawers/matplotlib/drawers/__init__.py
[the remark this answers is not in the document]
zerothi: No, lets wait. Easier to add than remove. :)

pfebrer:
Ok, I added a `backends/templates` directory that basically contains all the abstract classes.

Templates are registered on the backends manager with the `register_template` method and when you attempt to register a backend that doesn't inherit from the template an error is raised.

Any suggestion/comments about the organization and implementation of this?

zerothi:
I'll merge this now.

pfebrer:
Nooooooooooo! This only contains the changes for the bands plot :scream:

zerothi:
Ok, I can't run the tests?

1. The test imports are wrong? Referring to the old imports?
2. The sisl/viz/plotly/tests/__init__.py should be deleted

zerothi:
I haven't done anything yet... ;)
Seems there are still issues?

pfebrer:
Yeah the tests are also not modified. That's why it was a WIP and a draft :sweat_smile:

zerothi:
Ah, ok. I didn't get what was happening. Thought it was more or less complete.

In that case, then I would suggest `viz/backend/matplotlib/backend` gets renamed to something appropriate for the specific backend, say `viz/backend/matplotlib/_plots`, the underscore highlights that these are private imports and makes it easier to follow.

pfebrer:
In fact tests up until now were only testing the plotly representation, so I have to find a way to separate tests for different backends, while testing also the common processing part.

pfebrer:
Ok, makes sense to rename that :+1:

pfebrer:
Do you plan to do a release for the SIESTA school and that's why you were going to merge?

zerothi:
> Do you plan to do a release for the SIESTA school and that's why you were going to merge?

Yeah, if time allowed I could do that? But you require everything ready end of this week, no?

pfebrer:
So I did the first sketch on how to test multiple backends. Could you check the `test_bands.py` file (https://github.com/pfebrer/sisl/blob/detach-plotly/sisl/viz/plots/tests/test_bands.py) and see if you agree of how tests should work? Basically there are two types of tests: 
- Tests for the data processing stage (they don't depend on the backend).
- Tests for the rendering stage (they depend on the backend).

I set up a fixture for the `backend` argument, which is parametrized to iterate over all the available backends. This fixture is used for example in the `test_gap_in_figure` method. Then, it depends on the backend-defined `_test_is_gap_drawn`, which returns whether the gap is drawn or not. This is implemented in the `BandsBackend` template to raise an error:

https://github.com/pfebrer/sisl/blob/fc133438b699147840a6cbdba64943da5481ee4f/sisl/viz/backends/templates/_plots/bands.py#L24-L28

So, basically tests will fail if the backend doesn't implement it (the code for plotting can still run though!).  Then, each backend implements its own version:
- Plotly: https://github.com/pfebrer/sisl/blob/fc133438b699147840a6cbdba64943da5481ee4f/sisl/viz/backends/plotly/_plots/bands.py#L80-L81
- Matplotlib: https://github.com/pfebrer/sisl/blob/fc133438b699147840a6cbdba64943da5481ee4f/sisl/viz/backends/matplotlib/_plots/bands.py#L94-L95

I thought that all methods demanded by the testing framework could be prefixed with `_test`.

This is for now the way I've thought the testing could work, do you have other suggestions/comments?

sisl/viz/plots/tests/test_bands.py
```diff
@@ -16,13 +16,16 @@
 import numpy as np
```
zerothi, line 16: Instead of importing xarray here, you should do something like is done in `sisl/io/siesta/tests/test_bands.py` with importorskip

sisl/viz/plots/tests/test_bands.py
```diff
@@ -16,13 +16,16 @@
...
 
 
 pytestmark = [pytest.mark.viz, pytest.mark.plotly]
 _dir = osp.join('sisl', 'io', 'siesta')
 
+@pytest.fixture(params=BandsPlot.get_class_param("backend").options)
+def backend(request):
+    return request.param
 
 class BandsPlotTester(PlotTester):
```
zerothi, line 30: `pytest` will only test classes with _prefixed_ names `Test<>`

zerothi:
The test has these lines *in the middle of the file*:
```
gr = sisl.geom.graphene()
H = sisl.Hamiltonian(gr)
bz = sisl.BandStructure(H, [[0, 0, 0], [2/3, 1/3, 0], [1/2, 0, 0]], 9, ["Gamma", "M", "K"])
```
this is not good since every time you import that file it will rerun it.
Generally this should be put in a fixture so that it is runned once.

zerothi:
In https://github.com/pfebrer/sisl/blob/fc133438b699147840a6cbdba64943da5481ee4f/sisl/viz/backends/plotly/_plots/bands.py#L83 you register the backend.
Is this because you need to ensure that the backend is implemented for various backends. 
This could potentially be done automatically since one can use `class.__bases__` to get a list of all bases.
Then one could do `hasattr` to see if the method exists, if so, then use it?
I think this would make it easier?

zerothi:
I have to say I am not particularly fond of 
```
class TestBandsPlot(BandsPlotTester):
    run_for = bands_plots
```
why not just being explicit? This is, to me, just obfuscating code calls. You are basically building a new testing framework on top of pytest.

I would much prefer explicitness in tests. But ok.

In general you are simply too smart for me ;)
And if I have a hard time following, my bet is that it will be very long before anybody can pick this up and develop it further. ;)

pfebrer:
I already planned to change the tests organization, because I knew it was probably my lack of knowledge about `pytest` that made me develop it in this way. But I wanted to do it in a separate PR, to focus here on the multiple backends part.

However, I have given it a try. Everything is now handled with fixtures. See `sisl/viz/plots/tests/conftest.py` and `sisl/viz/plots/tests/test_bands.py`. I still have a helper class `_TestPlot` that automatically defines the `plot` (checking that there was no error initializing) and `test_attrs` (checking that everything needed to test is there). All you have to do then is to define a class inheriting from it and define the `init_func_and_attrs` fixture returning a tuple with the initializing function and the values for testing. See `TestBandsPlot`.

Let me know if there is still anything that you don't like if you have time. Also, if you could comment on the approach to test multiple backends it would be great :) Thanks!

`sisl/viz/plots/tests/conftest.py`



Let me know what you find is still not good if you have time.

pfebrer:
I'm almost done with moving all the implementations to this branch. Now all plots have a `plotly` and a `matplotlib` backend, and `GridPlot` and `GeometryPlot` have also a `blender` backend.

The only things left would be to update all the tests and the documentation.

pfebrer:
Whenever you have time, could you check the `docs/visualization/viz_module/Add new backends.ipynb` notebook? (https://github.com/pfebrer/sisl/blob/detach-plotly/docs/visualization/viz_module/diy/Adding%20new%20backends.ipynb) I tried my best to explain how new backends could be added and I think it is not that difficult to understand in the end. But it might just be that I've been thinking about it a lot so it is natural to me :sweat_smile:

I will continue to update the documentation, and when I finish I will update the tests.

zerothi:
> > In https://github.com/pfebrer/sisl/blob/fc133438b699147840a6cbdba64943da5481ee4f/sisl/viz/backends/plotly/_plots/bands.py#L83 you register the backend.
> > Is this because you need to ensure that the backend is implemented for various backends.
> > This could potentially be done automatically since one can use class.**bases** to get a list of all bases.
> 
> I hadn't thought about this because I was kind of emulating the behavior of the brillouin zone apply. So what you are proposing is that if I do:
> 
> ```python
> from sisl.viz.backends.templates._plots.BandsBackend
> 
> class MyBandsBackend(BandsBackend):
>     pass
> ```
> 
> `MyBandsBackend` becomes automatically registered? I see the problem that you could not create intermediate classes between the template and the final registerable thing because it would get registered without asking.

I think one could simply hide the backends by prefixing with `_` or `__`, that should be relatively simple to bypass, by Python design these are generally local.

> > Then one could do hasattr to see if the method exists, if so, then use it?
> > I think this would make it easier?
> 
> I don't understand which methods exactly do you want to check here. Could you make and example?

For instance some of the backends may not have `draw_gap` associated and then you would need to check for `attr` existence. But agreed, it is probably not that important since the back-end it-self should do everything, and hence know which methods are available. So effectively not important! :)

pfebrer:
But I'm not sure I want to hide them. See the notebook on how to extend the backends, where I explain how general users can extend the backends. For that reason, it probably doesn't make sense to hide them, does it?

> For instance some of the backends may not have draw_gap associated and then you would need to check for attr existence.

The backend template is an abstract class with some abstract methods. In this case, since `draw_gap` is a compulsory method it is marked as an `abstractmethod`. Other methods which are not essential but add extra functionality are implemented in the template to raise a `NotImplementedError`. I think this plus the documentation on each template is enough to help debugging and avoid errors in production, and it's cleaner than checking for attributes, no? This is also explained in the notebook on how to add backends.

zerothi:
> I'm almost done with moving all the implementations to this branch. Now all plots have a `plotly` and a `matplotlib` backend, and `GridPlot` and `GeometryPlot` have also a `blender` backend.
> 
> The only things left would be to update all the tests and the documentation/tutorials.

Great! :fireworks: 

> Whenever you have time, could you check the `docs/visualization/viz_module/Add new backends.ipynb` notebook? (https://github.com/pfebrer/sisl/blob/detach-plotly/docs/visualization/viz_module/diy/Adding%20new%20backends.ipynb) I tried my best to explain how new backends could be added and I think it is not that difficult to understand in the end. But it might just be that I've been thinking about it a lot so it is natural to me 
> 
> I will continue to update the documentation, and when I finish I will update the tests.

Great you did this.  
I think we should focus on merging your work here, then later we should consider letting the backends automatically register.

I am not particularly fond of the `register` method. This requires you to register a specific plot (here `BandsPlot` to a specific `backend`. I would rather go the other way around and say, you create a new backend, and from that one attach a BandsPlot.
I think this makes it much more obvious the relationship between backend. The backend should be the _top level_ thing, and not associated with a specific plot.

Here some ideas:
```python
from sisl.viz import backends

# Create new backend class
# This will use plotly_myown plots, but fall back to plotly backends for stuff not being overwritten
new_backend = backends.register("plotly_myown", cls=<optional name of class>, fallback="plotly")

# create bandsplot (this is not totally clear how it is best handled), this needs some thought
class MyBandsplot(PlotlyBands):
    # do whatever
    pass

# The above should be registered to the backend in some way, perhaps a register, or a dispatch method with a default
# then the dispatch can take care of fallbacks...
new_backend.Bands = MyBandsplot

# there should be ways to define the default backend
# 1: using contexts:
with new_backend:
    # this should use `MyBandsplot`
    bands.plot()

# 2: being explict
backends.set_default("plotly_myown")
bands.plot()
```
Hmm.. I see it is not easy, but as it is I think the backends are not backend specific, they are plot specific.

pfebrer:
Thanks for taking a look!

> but as it is I think the backends are not backend specific, they are plot specific.

Yes, indeed they are. So if I understand correctly, what you propose is that there is one big `plotly` backend that implements everything for all plots? This is interesting, but I thought about it and I discarded it because:
- The backend might get very big and maybe hard to understand with methods that are meant for different plots mixed together.
- Some extensions of plots (i.e. `BandsPlot` -> `FatbandsPlot`) need the same methods but slightly modified. This would not be possible if everything is put in the same class (main reason I thought backends had to be plot specific).
- If backends are plot specific you can have different defaults for each plot type. (E.g. suppose you love the `plotly` drawings but for some particular plot type the `matplotlib` representation is much better)

Maybe you have some ideas on how these problems should be overcomed.

> ```python
> with new_backend:
>     bands.plot()
> ```

Using contexts seems like a great idea! In the case of backends being plot specific, `backend` is a setting just like any other. In that sense, this concept could be applied to presets:
```python
my_preset = Preset(backend="plotly", bands_color="red", ...)

with my_preset:
    bands.plot()
```

zerothi:
> Thanks for taking a look!
> 
> > but as it is I think the backends are not backend specific, they are plot specific.
> 
> Yes, indeed they are. So if I understand correctly, what you propose is that there is one big `plotly` backend that implements everything for all plots? This is interesting, but I thought about it and I discarded it because:
> 
> * The backend might get very big and maybe hard to understand with methods that are meant for different plots mixed together.

I see the backend it self as a *container* for the different plots, i.e. each backend contains a `BandsPlot`, `FatbandsPlot` etc.

> * Some extensions of plots (i.e. `BandsPlot` -> `FatbandsPlot`) need the same methods but slightly modified. This would not be possible if everything is put in the same class (main reason I thought backends had to be plot specific).

I don't see why not?
```python
from sisl.viz import backends
backend = backends.get("plotly")
class FatbandsPlot(backend.get("BandsPlot")): # or simply backend.Bands?
      # do specific stuff to the fatbands plot
backend.Fatbands = FatbandsPlot
```
perhaps this needs some more careful thought, but the idea is simple I think?

> * If backends are plot specific you can have different defaults for each plot type. (E.g. suppose you love the `plotly` drawings but for some particular plot type the `matplotlib` representation is much better)

That is not a problem, no?
```python
with backends("plotly"):
    bands.plot(backend="matplotlib")
```
?
Each plot may still choose a backend, but the used plotting scheme is determined from the object _in_ the backend.

> 
> Maybe you have some ideas on how these problems should be overcomed.
> 
> > ```python
> > with new_backend:
> >     bands.plot()
> > ```
> 
> Using contexts seems like a great idea! In the case of backends being plot specific, `backend` is a setting just like any other. In that sense, this concept could be applied to presets:
> 
> ```python
> my_preset = Preset(backend="plotly", bands_color="red", ...)
> 
> with my_preset:
>     bands.plot()
> ```

I would say even in the case of my idea are contexts nice:
```python
with my_preset:
    bands.plot()
    fatbands.plot()
```

pfebrer:
Ok, so I think I misunderstood. You would still have plot specific backends then. 

The only difference with your approach is that instead of having one backend manager for each plot class you have one single global backends manager?

zerothi:
> Ok, so I think I misunderstood. You would still have plot specific backends then.
> 
> The only difference with your approach is that instead of having one backend manager for each plot class you have one single global backends manager?

Yes, I think this is much cleaner no?

pfebrer:
Yes it might be. I have to think about the details though :sweat_smile: 

One thing that I'm not sure I understand: The plot itself cannot be contained inside the backend, because then if you change backends you would need to reinitialize the plot. So when you say:

```python
from sisl.viz import backends
backend = backends.get("plotly")
class FatbandsPlot(backend.get("BandsPlot")): # or simply backend.Bands?
      # do specific stuff to the fatbands plot
backend.Fatbands = FatbandsPlot
```

what could be registered here is the class that **draws the fatbands plot**, but not the full `FatbandsPlot` class.

zerothi:
> Yes it might be. I have to think about the details though 
> 
> One thing that I'm not sure I understand: The plot itself cannot be contained inside the backend, because then if you change backends you would need to reinitialize the plot. So when you say:
> 
> ```python
> from sisl.viz import backends
> backend = backends.get("plotly")
> class FatbandsPlot(backend.get("BandsPlot")): # or simply backend.Bands?
>       # do specific stuff to the fatbands plot
> backend.Fatbands = FatbandsPlot
> ```
> 
> what could be registered here is the class that **draws the fatbands plot**, but not the full `FatbandsPlot` class.

I don't understand this? Whenever you call a `*.plot()` you would instantiate the plot, no?
What is the difference between the class that **draws the fatbands plot** and the `FatbandsPlot`? Why are they not the same?

pfebrer:
`FatbandsPlot` takes care of the entry point, the processing, and the rendering. Then when something is updated the process is repeated from the point it needs to be repeated, but not necessarily from the very beggining.

This is what allows you to do:

```python
band_structure = sisl.BandStructure(...)
bands = band_structure.plot() # Calculates the eigenvalues
bands.update_settings(bands_color="blue") # Doesn't recalculate the eigenvalues
```

zerothi:
> `FatbandsPlot` takes care of the entry point, the processing, and the rendering. Then when something is updated the process is repeated from the point it needs to be repeated, but not necessarily from the very beggining.
> 
> This is what allows you to do:
> 
> ```python
> band_structure = sisl.BandStructure(...)
> bands = band_structure.plot.fatbands() # Calculates the eigenvalues and the weights
> bands.update_settings(groups=[{"atoms": 1}, {"atoms": 3}]) # Doesn't recalculate the eigenvalues and the weights
> ```

I don't see the problem? The class returned from `.plot.fatbands()` has the capability of changing the plot on the fly? Is that object not the `FatbandsPlot`?

pfebrer:
But what if you do:

```python
bands.update_settings(backend="matplotlib")
```

Is the class capable of changing the rendering backend? As I understood with your approach the backends manager contains all the plots and plots are not aware of the backend manager and how to change their backend

zerothi:
> But what if you do:
> 
> ```python
> bands.update_settings(backend="matplotlib")
> ```
> 
> Is the class capable of changing the rendering backend? As I understood with your approach the backends manager contains all the plots and plots are not aware of the backend manager and how to change their backend

Hmm. I see. With the above then it would require that the `bands` object is another class that contains the plotting backend.
I don't think it is too difficult since that object may re-instantiate the required back-end and pass the values to draw again. However, this of course requires the same interfaces? But seems like your interfaces are still forced to be the same?

pfebrer:
> Hmm. I see. With the above then it would require that the bands object is another class that contains the plotting backend.
> I don't think it is too difficult since that object may re-instantiate the required back-end and pass the values to draw again. However, this of course requires the same interfaces? But seems like your interfaces are still forced to be the same?

Yes, exactly. This is exactly what the framework does right now. The interfaces are the same on purpose. I think then it's much simpler to explain what you need to do to add a new backend and the design is much more modular, approachable (if explained properly) and less prone to bugs.

So, if you do what you are proposing here, really the only difference with the current implementation would be how the backends are managed and where the plots have to go to access them. At this point, I don't care. We can try both and see what seems cleaner and allows more functionality.

zerothi:
> Yes, exactly. This is exactly what the framework does right now. The interfaces are the same on purpose. I think then it's much simpler to explain what you need to do to add a new backend and the design is much more modular, approachable (if explained properly) and less prone to bugs.

Then I don't see a big problem in the method I suggest.

> So, if you do what you are proposing here, really the only difference with the current implementation would be how the backends are managed and where the plots have to go to access them. At this point, I don't care. We can try both and see what seems cleaner and allows more functionality.

Yes, but _how_ the backends would be managed is much simpler since you directly engage with the backend, rather than having to re-use the `"plotly_myown"` for registering stuff. It would also allow you to overwrite, locally, some default backend. 

I think you'll find that the *single* backend is a bit easier to extend than what you have now.

Perhaps once you are done here, we can have a chat and we can go through some of the code together, then it will be easier for me, I hope ;)

Lets stick with the way you have it now, then change it later.

pfebrer:
> and I don't think my approach would be less prone to bugs, I think rather the other way around (could be wrong though!).

I meant `same interface -> less bugs than different interfaces`, I wasn't comparing the way you manage backends.

> Yes, but how the backends would be managed is much simpler since you directly engage with the backend, rather than having to re-use the "plotly_myown" for registering stuff. It would also allow you to overwrite, locally, some default backend.

I think until I see the implementation and how a user can interact with it I won't be able to tell whether it looks smoother to me or not :sweat_smile: But the best part is that it is not that difficult to reorganize.

> Perhaps once you are done here, we can have a chat and we can go through some of the code together, then it will be easier for me, I hope ;)

Sure, that would be great!

sisl/viz/backends/matplotlib/backend.py
```diff
@@ -91,22 +94,24 @@ class MatplotlibMultiplePlotBackend(MatplotlibBackend, MultiplePlotBackend):
 
 class MatplotlibSubPlotsBackend(MatplotlibMultiplePlotBackend, SubPlotsBackend):
 
-    def draw(self, backend_info, rows, cols, childs, **make_subplots_kwargs):
+    def draw(self, backend_info):
+        childs = backend_info["child_plots"]
+        rows, cols = backend_info["rows"], backend_info["cols"]
 
-        self.figure, self.axes = plt.subplots(rows, cols)
+        self.figure, self.ax = plt.subplots(rows, cols)
```
zerothi, line 101: Why this name change? Axes is well describing multiple axis? (axis is singular?)

sisl/viz/backends/plotly/backend.py
```diff
@@ -1,6 +1,7 @@
 from collections import defaultdict
 import itertools
 from functools import partial
+from sisl.viz.plots.tests.test_bands import backend
```
zerothi: This is a no go, nothing in the tests should be imported in the main code.

sisl/viz/backends/plotly/backend.py
```diff
@@ -430,11 +434,13 @@ def draw(self, drawer_info, rows, cols, childs, **make_subplots_kwargs):
...
-    def draw(self, drawer_info, childs, get_frame_names):
-        frames_layout = self._build_frames(childs, None, get_frame_names)
+    def draw(self, backend_info):
+        childs = backend_info["child_plots"]
+        frame_names = backend_info["frame_names"]
+        frames_layout = self._build_frames(childs, None, frame_names)
         self.update_layout(**frames_layout)
 
-    def _build_frames(self, childs, ani_method, get_frame_names):
+    def _build_frames(self, childs, ani_method, frame_names):
```
zerothi, line 442: I think the common name is children, not childs

sisl/viz/backends/plotly/backend.py
```diff
@@ -1,7 +1,6 @@
 from collections import defaultdict
 import itertools
 from functools import partial
-from sisl.viz.plots.tests.test_bands import backend
```
zerothi: Ok, I see it was a typo. ;)

pfebrer: Yeah it was automatically imported by VS studio. Text editors are so smart nowadays :eyes: hahaha

sisl/viz/backends/plotly/backend.py
pfebrer: True

sisl/viz/backends/matplotlib/backend.py
pfebrer: Because it makes it easier to generalize methods, all matplotlib plots have the `self.ax` attribute and then an optional `ax_indices` can be passed. See:

https://github.com/pfebrer/sisl/blob/ccc1fcef193a4fe51ddda17533e3159174de106c/sisl/viz/backends/matplotlib/backend.py#L30-L53

I have not thought much about it though. I will think about it again tomorrow :)

zerothi: I think it would be best to use `axes` (plural) and `axis` (singular) and adhere to this naming scheme (everywhere).

pfebrer: In `matplotlib` they always use `ax` (https://matplotlib.org/stable/tutorials/introductory/usage.html#sphx-glr-tutorials-introductory-usage-py), so I wanted to stick with `matplotlib` conventions for the `matplotlib` backend.

zerothi: Not internally. One thing is the user scripts, another thing is internally and how they adhere to objects.

zerothi: At least from the parts I read, see e.g. https://github.com/matplotlib/matplotlib/blob/9e0bb9c72121a7c3ba85c55368ae402bbfdd7782/lib/matplotlib/figure.py#L193.

pfebrer: Hmm ok, but then it should always be `axes` (as they do internally from what I understand. Because in fact even if it's "singular" it contains the `x` axis and the `y` axis.

zerothi: True :)

pfebrer:
Ok, I finished updating the documentation and all tests are passing now. I'd say this is finally ready :) 

Whenever you want we can discuss some details of it.

Note: I have to figure out why are some warnings appearing when I run the tests, because they don't seem to appear when running the code on the notebook.

zerothi:
@pfebrer I have just rebased, could you please hold your branch for now (locally). There were no conflicts, but I am trying to fix the warnings!

zerothi:
When I run the tests I get lots of "print-out errors", but not formal errors?

```
viz/plots/tests/test_plots.py::TestPlotSubClass::test_save_and_load[PdosPlot]
  info:0: SislInfo:
  
  The plot has been initialized correctly, but the current settings were not enough to generate the figure.
  Error: Could not read or generate data for PdosPlot from any of the possible sources.
  Here are the errors for each source:
  	- TB trans: TypeError.expected str, bytes or os.PathLike object, not NoneType
  	- hamiltonian: TypeError.expected str, bytes or os.PathLike object, not NoneType
  	- siesta output: TypeError.expected str, bytes or os.PathLike object, not NoneType
```
is this on purpose, or should this test actually have failed?

sisl/viz/plots/geometry.py
```diff
@@ -445,7 +465,77 @@ def xyz(coeffs):
...
+        Parameters
+        ------------
+        geometry: sisl.Geometry
+            the geometry for which you want the projected coords
+        xyz: array-like of shape (natoms, 3), optional
+            the 3D coordinates that we want to project.
+            otherwise they are taken from the geometry. 
+        axis: {0,1,2, "x", "y", "z", "a", "b", "c"} or array-like of shape 3, optional
+            the direction to be displayed along the X axis. 
+            If it's an int, it will interpreted as the index of the cell axis.
```
zerothi, line 486: In your `_sanitize_axis` routine you interpret (0, 1, 2) as Cartesian coordinates?
I was confused whether `cell` meant lattice or Cartesian cell?
Please clarify?

zerothi:
Ok, I have amended another commit. Please ensure that you re-fetch your branch (I have overwritten it).

Sorry for doing it like this, but it cleared many warnings and added some things needed.

sisl/viz/plots/geometry.py
pfebrer: This is an old docstring that should have been changed. Initially `0` meant the first lattice vector, but since I realised that it was strange that users needed to project along a lattice vector I decided to change it. Now, as you understood, `0` is the same as `"x"` as it's always more comfortable to write integers than to write strings. Any opinion on whether this is fine or not?

zerothi: My other implementation here `sisl/utils/misc.py:direction` could solve this kind of easily. I would suspect integer indices to respect lattice vectors. Typing an `x` vs. a `0` is not a problem I think

pfebrer:
> is this on purpose, or should this test actually have failed?

This warning is to be expected. These tests are just to check that the basic functionality of the `Plot` class has not been broken by the child class. The plots are initialized, but since they have no data to plot, they show this warning. I agree that this is annoying though when running the tests. Maybe we could filter the warnings in the test functions?

sisl/viz/plots/geometry.py
pfebrer: Hmm ok, the problem is that it's not an `x`, it's an `"x"` :) And when you do it lots of times it becomes annoying to write the `"`.

> I would suspect integer indices to respect lattice vectors.

Yeah me too, that's why this was the initial idea. But for the representations I can't see many cases where you would want to project the lattice vectors into the plot axes to visualize real space data.

zerothi:
> > is this on purpose, or should this test actually have failed?
> 
> This warning is to be expected. These tests are just to check that the basic functionality of the `Plot` class has not been broken by the child class. The plots are initialized, but since they have no data to plot, they show this warning. I agree that this is annoying though when running the tests. Maybe we could filter the warnings in the test functions?

Yes, I think we should filter, I tried to manage this in some of my tests, and this is still a work-in-progress. But it makes it much easier to figure out if something goes wrong. In fact pytest can check where a specific warning has been raised, see [here](https://docs.pytest.org/en/6.2.x/warnings.html#warns)

sisl/viz/plots/geometry.py
zerothi: But isn't this in the GUI drop-down? I think `"x"` is fine to force people to do. It makes the code readable. Integers are not really good either. Better to force good habits than remove 2 characters. :)

Perhaps now there isn't much reason, but perhaps later...

zerothi: I have multiple times found that what makes it easier isn't always what users expect, and once it is in, it is hard to change. Always think in terms of functionality and what one would expect (without reading the documentation). Then if there is an ambiguity it can be discussed.

pfebrer: Yes in the graphical interfaces this is clearly not a problem. It is also not a problem in scripts. But it is a problem in notebooks or other interactive environments that are very dynamic. In those cases you are changing code really fast and the displayed axes is a very recurrent thing to change.

I don't know, maybe `[0,1,2,3,4,5]` could stand for `["a", "b", "c", "x", "y", "z"]`? This would not cause problems for people who haven't read the documentation (they wouldn't expect `3,4,5` to be accepted), but allow people who have read it to be faster.

zerothi: I would really advocate in not doing this! Once you have typed the quotation marks you are only going to change x->a or a->x. So not a problem. Using your suggested approached is even more error prone, who would know/guess this combination of numbers?  
Again, it should be intuitive above all, the feel will be much simpler.  
Another example would be pandas.DataFrame, many of the methods have so many possibilities that I have to look up the documentation even for simple things I thought I could do by memory. This is because it is not natural or intuitive. Is asking for an initial 2 characters really a problem?

I don't really get your point given the above; even if you would like to be very dynamic, you still have typed the quotation marks once, and subsequent changes are only 1 character anyway?

zerothi: Also remember that users are *always* less efficient and less knowledgeable than you, they are not familiar with every piece of the software. So it should be *easy* to use, not to be fast. I have made many mistakes in this also. ;)

pfebrer: > I don't really get your point given the above; even if you would like to be very dynamic, you still have typed the quotation marks once, and subsequent changes are only 1 character anyway?

The most common changes I make are:
- Change from 2D to 3D.
- Plot more grids/geometries, then I have to write again the axes that I want.

Writing `axes=["x", "y", "z"]` is annoying compared to `axes=[0,1,2]` (I'm experiencing it myself when writing it here, for the xyz version you need to press/unpress shift 6 times)

> So it should be easy to use, not to be fast.

I understand this, but if the point of the software is to allow quick visualization, then I think "quick to use" is also a feature that one would expect.

I'm worried that people will just use `0,1,2` as shortcuts for `"x", "y", "z"` when they have an (almost) orthogonal cell, not because they want the projection of the axes, but because they want the cartesian representation and it's more convenient to write integers. This will then not behave in the same way when the cell is not orthogonal.

zerothi: I am still not convinced ;)  
For rapid changes, a temporary variable can make everything much easier:
```python
axes = 'xyz'
#axes = [0,1,2]
```
Allow `axes` argument to handle strings as a list, say: `axes='xyz'` this would be even more natural and just as short, agreed `['x', 'y', 'z']` is too much. The other is more natural!

pfebrer: Hmm maybe the issues that I have could be solved by allowing multicharacter strings also? I mean `axes="xyz"` would be the same as `axes=["x", "y", "z"]`. I think my speed thrive would be satisfied with that :) hahahah

zerothi: > > So it should be easy to use, not to be fast.
> 
> I understand this, but if the point of the software is to allow quick visualization, then I think "quick to use" is also a feature that one would expect.

I really don't think users are in this category ;)
Only very high level users will be irritated about this.

> I'm worried that people will just use `0,1,2` as shortcuts for `"x", "y", "z"` when they have an (almost) orthogonal cell, not because they want the projection of the axes, but because they want the cartesian representation and it's more convenient to write integers. This will then not behave in the same way when the cell is not orthogonal.

That is fine, there has to be rules. And you and I both thought that the lattice vectors were natural for integer indices, and so a large fraction of users will probably also have that thought. Of course there will be users that think the other way around, but regardless you'll get some users who don't expect the outcome of the arguments.

pfebrer: Does this solution make sense to you?

pfebrer: Ok, I see that we commented the same at the same time, then great!

zerothi: yes, great. I think this is better. Also consistency in sisl and elsewhere is key, I use integers as lattice vector matches.

zerothi:
I am done pushing to your branch, so perhaps you could have a go, and the we can merge this?

pfebrer:
I removed all the warnings that are due to the failed initialization of the plot, now I only get some warnings for using `MonkhorstPack.apply.average.PDOS()`:
```
  dep:0: SislDeprecation:
  
  use Hamiltonian.eigenstate(...).PDOS() instead [>=0.13.0]
```
I have two questions:
- Why don't these warnings appear when I run the code in a notebook? This is the code that I run in the tests:
```python
gr = sisl.geom.graphene()
H = sisl.Hamiltonian(gr)

H.plot.pdos(Erange=[-5,5])
```
- How should I go about calculating the PDOS using the monkhorst pack object if `Hamiltonian.PDOS()` is deprecated?

zerothi:
> I removed all the warnings that are due to the failed initialization of the plot, now I only get some warnings for using `MonkhorstPack.apply.average.PDOS()`:
> 
> ```
>   dep:0: SislDeprecation:
>   
>   use Hamiltonian.eigenstate(...).PDOS() instead [>=0.13.0]
> ```
> 
> I have two questions:
> 
> * Why don't these warnings appear when I run the code in a notebook? This is the code that I run in the tests:
> 
> ```python
> gr = sisl.geom.graphene()
> H = sisl.Hamiltonian(gr)
> H.construct([(0.1, 1.44), (0, -2.7)])
> 
> H.plot.pdos(Erange=[-5,5])
> ```

Probably because the default level of the warnings module is too low.

> * How should I go about calculating the PDOS using the monkhorst pack object if `Hamiltonian.PDOS()` is deprecated?

One should pass a wrapper: `mp.apply.average.eigenstate(wrap=lambda es: es.PDOS())`, a bit clumsy, we should think of a way  to incorporate class calls as deferred methods in dispatchers.

pfebrer:
It took me a while to make geometries and grids play well with each other in all cases. This is what I ended up doing for the different possible values of axes:
- `{"x", "y", "z"}`: The cartesian coordinates are displayed, for both `GeometryPlot` and `GridPlot`. 
- `{"a", "b", "c"}`: The fractional coordinates are displayed, for both `GeometryPlot` and `GridPlot`. Same for `{0,1,2}`.
- `np.array of shape (3, )`: The coordinates are projected into that direction for `GeometryPlot`. If two directions are passed, the coordinates are not projected to each axis separately. The displayed coordinates are then the coefficients of the linear combination to get that point. Not implemented in `GridPlot`.

zerothi:
Sounds like a good compromise!

pfebrer:
Good, then I just have to finish some details and update the documentation.

pfebrer:
I'd say it's finished now. There might still be some inconsistent things, but we can figure them out along the way, as we get some feedback :)

pfebrer:
Do you plan to release a version soon? I would like to parallelize the PDOS and (fat)bands calculations (I will do it today) and agree on a CLI.

zerothi:
No, we can wait for that. Definitely the cli would be great to have in the next release.

pfebrer:
Good!

PR #316, opened 2021-03-19
bug: (PdosPlot) TBT.nc file doesn't contain species information

pfebrer:
When I submitted https://github.com/zerothi/sisl/pull/311 I didn't know that the `*TBT.nc` files did not contain the species information. 

So here I added that the geometry should be read from the `root_fdf` if available. Also I added the possibility to pass a geometry to overwrite whichever one has been read.

### DIFF
--- a/sisl/viz/plotly/input_fields/sisl_obj.py
+++ b/sisl/viz/plotly/input_fields/sisl_obj.py
@@ -1,4 +1,5 @@
 """ This input field is prepared to receive sisl objects that are plotables """
+from pathlib import Path
 
 import sisl
 
@@ -31,10 +32,12 @@ class SislObjectInput(InputField):
 
     def __init__(self, key, *args, **kwargs):
 
-        if 'dtype' not in kwargs:
+        super().__init__(key, *args, **kwargs)
+
+        if self.dtype is None:
             raise ValueError(f'Please provide a dtype for {key}')
 
-        valid_key = forced_keys.get(kwargs['dtype'], None)
+        valid_key = forced_keys.get(self.dtype, None)
 
         if valid_key is not None and not key.endswith(valid_key):
 
@@ -44,16 +47,15 @@ class SislObjectInput(InputField):
                 f'If there are multiple settings that accept {kwargs["dtype"]}, please use *_{valid_key}'
             )
 
-        super().__init__(key, *args, **kwargs)
-
 
 class GeometryInput(SislObjectInput):
 
+    dtype = (sisl.Geometry, "sile (or path to file) that contains a geometry")
     _dtype = (str, sisl.Geometry, *sisl.get_siles(attrs=['read_geometry']))
 
     def parse(self, val):
 
-        if isinstance(val, str):
+        if isinstance(val, (str, Path)):
             val = sisl.get_sile(val)
         if isinstance(val, sisl.io.Sile):
             val = val.read_geometry()
--- a/sisl/viz/plotly/plots/pdos.py
+++ b/sisl/viz/plotly/plots/pdos.py
@@ -6,7 +6,7 @@ from ..plotutils import find_files, random_color
 from ..input_fields import (
     TextInput, SileInput, SwitchInput, ColorPicker, DropdownInput, CreatableDropdown,
     IntegerInput, FloatInput, RangeInput, RangeSlider, OrbitalQueries,
-    ProgramaticInput, Array1DInput, ListInput
+    ProgramaticInput, Array1DInput, ListInput, GeometryInput
 )
 from ..input_fields.range import ErangeInput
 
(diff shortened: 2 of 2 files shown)


sisl/viz/plotly/plots/pdos.py
```diff
@@ -23,6 +23,9 @@ class PdosPlot(Plot):
     tbt_nc: tbtncSileTBtrans, optional
         This parameter explicitly sets a .TBT.nc file. Otherwise, the PDOS
         file is attempted to read from the fdf file
+    geometry: str or Geometry or asciiSileBigDFT or cubeSile or gotSileGULP or hamiltonianSile or omxSileOpenMX or pdbSile or refSileScaleUp or restartSileScaleUp or tshsSileSiesta or onlysSileSiesta or ncSileSiesta or xvSileSiesta or fdfSileSiesta or outSileSiesta or pdosSileSiesta or deltancSileTBtrans or tbtsencSileTBtrans or phtsencSilePHtrans or tbtncSileTBtrans or tbtavncSileTBtrans or phtncSilePHtrans or phtavncSilePHtrans or tbtprojncSileTBtrans or phtprojncSilePHtrans or carSileVASP or chgSileVASP or locpotSileVASP or winSileWannier90 or xsfSile or axsfSile or xyzSile, optional
```
zerothi, line 26: perhas change to `Sile` instead of listing all possibilities?

Anything that has `read_geometry` would work?

sisl/viz/plotly/plots/pdos.py
```diff
@@ -287,7 +296,12 @@ def _read_TBtrans(self, root_fdf, tbt_nc):
         self.PDOS = tbt_sile.DOS(sum=False).data.T
         self.E = tbt_sile.E
```
zerothi, line 298: Here instead it would be a wiser choice to do something like this:
```python
geom = self.get_sile("root_fdf").read_geometry(output=True)
self.geometry = tbt_sile.read_geometry(atom=geom.atoms)
```
I think this should work. This is better since the cdf reading of the geometry does some nice things ;)

sisl/viz/plotly/plots/pdos.py
```diff
@@ -287,7 +296,12 @@ def _read_TBtrans(self, root_fdf, tbt_nc):
...
 
-        self.geometry = tbt_sile.read_geometry().sub(tbt_sile.a_dev)
+        try:
+            self.geometry = self.get_sile("root_fdf").read_geometry(output=True)
+        except (FileNotFoundError, TypeError):
+            self.geometry = tbt_sile.read_geometry()
+
+        self.geometry = self.geometry.sub(tbt_sile.a_dev)
```
zerothi, line 304: Hmm.. I am not sure doing the `sub` is always a good idea. The problem is that some users may expect to work on full geometry indices and not a subset. Also, some of the tbt_nc sile stuff returns data in the *global* picture i.e. with buffer atoms etc. Why are you subbing?

zerothi (changes requested):

sisl/viz/plotly/plots/pdos.py
pfebrer: > Anything that has read_geometry would work?

Yes

This string is obviously generated automatically, but I will change it because I agree it doesn't make sense to write all of the possibilities.

sisl/viz/plotly/plots/pdos.py
pfebrer: Ok, this is just because in the tests that I did `tbt_sile.DOS` returns only data for orbitals in the scattering region. Are there cases where the DOS is returned for all the atoms? (I wrote this entry point quickly because I wanted to use the PdosPlot for my first tbtrans PDOS calculation, I didn't know all the different possibilities, and I still don't)

> The problem is that some users may expect to work on full geometry indices and not a subset.

Yes, I agree. I'll have to think of some way to handle the mapping from "full geometry" to "only scattering" indices (or maybe this is already in `tbt_sile`?). Although this has not bothered me at all because I'm always using categories (or dicts) to select the atoms. Knowing this, I'm not sure if this is a high priority thing that should be fixed in this PR or not, what do you think?

zerothi: This is one of the parts where sisl is not really consistent... :(

Some of the quantities, DOS returns for the device atoms only.
Other quantiities such as sparse matrices (bond-current and density matrices) are returned in the full geometry.

I initially did the DOS, thinking that users would always only think about the device region.
However, when I did the sparse matrices it seemed more natural to return everything, since that is the matrices from siesta anyways. Perhaps I should make them only return the device region....

pfebrer: I don't know what to say, I have no opinion on what would be best :sweat_smile:

sisl/viz/plotly/plots/pdos.py
pfebrer: Done!

zerothi:
Ok, I think I will just approve this one.

However, you messed up the `files` sub-repo. Could you rebase on the current master?

pfebrer:
Done!

zerothi:
Could you also do a rebase + squash?

pfebrer:
Also done!

sisl/viz/plotly/input_fields/sisl_obj.py
```diff
@@ -44,16 +47,15 @@ def __init__(self, key, *args, **kwargs):
...
 
+    dtype = (sisl.Geometry, "sile (or path to file) that contains a geometry")
     _dtype = (str, sisl.Geometry, *sisl.get_siles(attrs=['read_geometry']))
 
     def parse(self, val):
 
-        if isinstance(val, str):
+        if isinstance(val, (str, Path)):
             val = sisl.get_sile(val)
         if isinstance(val, sisl.io.Sile):
```
zerothi, line 60: oh, I found something here. Here you want to check for `sisl.io.BaseSile` since the `sisl.io.Sile` is only ascii siles.

That would enable you to also read from cdf files etc.

zerothi: Haven't you had problems reading geometries from non-ascii files?

zerothi: I did it ;)

pfebrer: Thanks! I didn't know this

Issue #301, opened 2021-02-08
sisl.viz._plotables


zerothi:
@pfebrer

Your `_plotables` looks very similar to the Dispatch methods.

Could you clarify the needs of this contrary to the regular dispatch methods? And could we perhaps combine them to be based on the same base-classes?

pfebrer:
Yes, you are right. I think it definitely can be implemented with dispatchers.

Some comments:
- `PlotEngine` is probably not needed anymore with the vision of implementing multiple backends in the same plot class (the backend will end up being a regular setting probably).
- It is very important that a default can be specified, because it looks dumb to `sisl.get_sile("file.bands").plot.bands()` (or same with a geometry).
- The `__dir__` method is important to me because it's used for autocompletion in IDEs, and it's a nice way of not needing to check the documentation to know what are your possibilities. I think in general dispatchers could really benefit from it, so maybe it can be incorporated in the base class?

pfebrer:
Do you want me to do it or are you doing it already? hahah

pfebrer:
This is solved with #317

PR #317, opened 2021-03-19
maint: Using dispatchers for plot handlers

pfebrer:
This addresses https://github.com/zerothi/sisl/issues/301 to use dispatchers for plot handlers in the viz module.

Also, since different plot backends will be introduced in https://github.com/zerothi/sisl/pull/312 and the backend will simply be another setting, there's no need for `PlotEngine` anymore.

Overall, handling of plotables is much cleaner now.

### DIFF
--- a/sisl/viz/_plotables.py
+++ b/sisl/viz/_plotables.py
@@ -1,264 +1,90 @@
 """
 This file provides tools to handle plotability of objects
 """
-
-from copy import copy
-from functools import wraps
+from sisl._dispatcher import ClassDispatcher, AbstractDispatch, ObjectDispatcher
 
 __all__ = ["register_plotable"]
 
 
-class PlotEngine:
-    """
-    Stores and takes care of calling all methods of a plotting engine.
-    """
+class ClassPlotHandler(ClassDispatcher):
+    """Handles all plotting possibilities for a class"""
 
-    def __init__(self, name, handler, default=None, methods={}):
-
-        self._raw_methods = {}
-        self._handler = handler
-
-        for name, method in methods.items():
-            self._add(name, method)
-
-        self._name = name
-        self._default = default
-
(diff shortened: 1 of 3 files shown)


sisl/viz/_plotables.py
```diff
@@ -1,264 +1,89 @@
...
-        # into the engine
-        @wraps(function)
-        def real_function(*args, **kwargs):
-            return function(self._handler._obj, *args, **kwargs)
+class ClassPlotHandler(ClassDispatcher):
+    """Handles all plotting possibilities for a class"""
 
-        return real_function
+    def __init__(self, *args, **kwargs):
+        kwargs["instance_dispatcher"] = ObjectPlotHandler
```
zerothi, line 13: Better to only add this in case user put it in.

Something like:
```python
if not "instance_dispatcher" in kwargs:
    kwargs["instance_dispatcher"] = ObjectPlotHandler
```
ensures users can do what-ever.

zerothi (changes requested):
Looks great. Lots of deletions which seems to clean up the implementation.

Great! Just a small note, then I'll get it in. It sounded like you wanted this in before #312 ?

sisl/viz/_plotables.py
pfebrer: Done :+1:

pfebrer:
> It sounded like you wanted this in before #312 ?

Yees, functionally it doesn't matter since the API is exactly the same, but it's nice to have it cleaned up before proceeding :)

pfebrer:
All good now!

zerothi:
Could you do a rebase and squash to only get 1 commit? I can also do it when merging, but perhaps you want to try it out? ;)

pfebrer:
Done :)

zerothi (approved):

Issue #319, opened 2021-04-08
Conversion between spin dimensions


tfrederiksen:
**Describe the feature**
Suppose one has constructed (or read from file) a `Hamiltonian` (or `DensityMatrix` etc) with `dim=1` (unpolarized) and want to map it over to a `dim=2` (or higher) object. Is there an easy (high-level) way of doing this?

It could perhaps also be useful if one could go the other way around, obtaining an object with `dim=1` by selecting one of the spin components from a `dim=2` object.

I notice that the function `copy([dtype])` enables the possibility of changing `dtype`. Maybe one could enable here also an option to change `dim`?

zerothi:
Hmm, currently there isn't one.

One can do like this:
```python
H = Hamiltonian(...)
...
csr = [H.tocsr(i) for i in range(H.dim)]
S = None
if not H.orthogonal:
    S = csr.pop()
# Now extend as much as you please
if len(csr) == 1:
    # make it a polarized case
    csr.append(csr[0])

Hpol = H.fromsp(H.geometry, csr, S=S)
```
This is currently the easiest way.

- Going from unpolarized to polarized is trivial, a simple copy
- unpolarized to non-collinear? In that case the off-diagonals should just be 0, I guess.
- from polarized/non-collinear to unpolarized, what to do here? average between the two components, or some way of choosing the fraction of the diagonal spin-box terms?

tfrederiksen:
> Hmm, currently there isn't one.
> 
> One can do like this:
> 
> ```python
> H = Hamiltonian(...)
> ...
> csr = [H.tocsr(i) for i in range(H.dim)]
> S = None
> if not H.orthogonal:
>     S = csr.pop()
> # Now extend as much as you please
> if len(csr) == 1:
>     # make it a polarized case
>     csr.append(csr[0])
> 
> Hpol = H.fromsp(H.geometry, csr, S=S)
> ```
> 
> This is currently the easiest way.

Thanks, this is very instructive. But perhaps still a bit too "low-level" from the user perspective?

>     * Going from unpolarized to polarized is trivial, a simple copy

Not sure even this is completely trivial. Do you mean something like this?
```python
if H.orthogonal:
    Hpol = H.fromsp(H.geometry, [csr, csr])
else:
    Hpol = H.fromsp(H.geometry, [csr, csr], S=H.S)
```

>     * unpolarized to non-collinear? In that case the off-diagonals should just be 0, I guess.

Yes, I agree.

[...]

PR #320, opened 2021-04-10
Introduced transform of sparse matrices for easy spin class conversion etc

tfrederiksen:
This branch strives to enable "high-level" conversion of spin dimension as discussed in #319.

### DIFF
--- a/sisl/physics/sparse.py
+++ b/sisl/physics/sparse.py
@@ -1214,10 +1214,17 @@ class SparseOrbitalBZSpin(SparseOrbitalBZ):
     def convert(self, spin, dtype=None):
         """ Create a new matrix of the specified spin type through the following conversion rules:
 
-        unpolarized -> (polarized, non-colinear, spinorbit): Duplicate up/down components
-        (polarized, non-colinear, spinorbit) -> unpolarized: Average up/down components
-        non-colinear -> spinorbit: set spin components 4-7 to zero
-        spinorbit -> non-colinear: ignore spin components 4-7 (should one rather average?)
+        Up-conversion:
+        * unpolarized -> (polarized, non-colinear, spinorbit): Copy unpolarized value to both up and down components
+        * polarized -> (non-colinear, spinorbit): Copy up and down components
+        * non-colinear -> spinorbit: Copy first four spin components
+        * all other new spin components are set to zero
+
+        Down-conversion:
+        * (polarized, non-colinear, spinorbit) -> unpolarized: Set unpolarized value to a mix 0.5*up + 0.5*down
+        * (non-colinear, spinorbit) -> polarized: Keep up and down spin components
+        * spinorbit -> non-colinear: Keep first four spin components
+        * all other spin components are dropped
 
         Parameters
         ----------
--- a/sisl/physics/sparse.py
+++ b/sisl/physics/sparse.py
@@ -1210,6 +1210,62 @@ class SparseOrbitalBZSpin(SparseOrbitalBZ):
 
         return new
 
+
+    def convert(self, spin, dtype=None):
+        """ Create a new matrix of the specified spin type through the following conversion rules:
+
+        unpolarized -> (polarized, non-colinear, spinorbit): Duplicate up/down components
+        (polarized, non-colinear, spinorbit) -> unpolarized: Average up/down components
+        non-colinear -> spinorbit: set spin components 4-7 to zero
+        spinorbit -> non-colinear: ignore spin components 4-7 (should one rather average?)
+
+        Parameters
+        ----------
+        spin : str, sisl.Spin
+            the desired new spin dimension
+        dtype : np.dtype, optional
+            data type contained in the matrix. See details of `Spin` for default values.
+        """
+        if not isinstance(spin, Spin):
+            spin = Spin(spin, dtype)
+
+        if spin == self._spin:
+            # same type, no conversion needed
+            return self.copy()
+
+        if dtype is None:
+            dtype = spin.dtype
+
+        # determine transformation matrix
+        n = self._spin.spins
+        m = spin.spins
+        matrix = np.eye(m, n, dtype=dtype)
+
+        if spin.is_unpolarized:
+            # average up and down components
+            matrix[0, 0] = 0.5
+            matrix[0, 1] = 0.5
+
+        else:
+            # at least two spin components
+            if self._spin.is_unpolarized:
+                # set up and down components to unpolarized value
+                matrix[0, 0] = 1.
+                matrix[1, 0] = 1.
+
+        if not self.orthogonal:
+            # include overlap matrix
+            matrix2 = np.zeros((m + 1, n + 1), dtype=dtype)
+            matrix2[:m, :n] = matrix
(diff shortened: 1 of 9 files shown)


sisl/sparse_geometry.py
```diff
@@ -133,6 +133,10 @@ def copy(self, dtype=None):
         new._csr = self._csr.copy(dtype=dtype)
         return new
 
+    def convert(self, scale):
```
zerothi, line 136: add `dtype` here as well, and also.
I think what you want is:
```python
new = self.copy()
new._csr = self._csr.convert(scale, dtype)
return new
```
?
The current implementation returns a `SparseCSR` matrix and not the same class.

zerothi: Hmm, come to think about it. This won't work for spin stuff since the spin value is determined at initialization time...
Hmm...
Perhaps leave this part out so we can play a bit?

sisl/sparse.py
```diff
@@ -1352,6 +1352,37 @@ def copy(self, dims=None, dtype=None):
 
         return new
 
+    def convert(self, scale, dtype=None):
+        """ Convert sparse matrix data to a different dimension by mixing with specified coefficients
+
```
zerothi, line 1357: We should add an example her, probably something similar to the `test_convert` test would suffice to clarify its use and what it does.

zerothi (changes requested):
Great thomas! Looks good!

A few things we should consider, especially the sparse_geometry needs some revision. It isn't exactly clear how we should do this efficiently and reliably...

tfrederiksen:
Thanks, Nick. I will look at this asap but have a busy week ahead. Please hang on.

sisl/sparse_geometry.py
tfrederiksen: Leave what out? Not exactly sure what you have in mind.

sisl/sparse.py
tfrederiksen: I agree. Was also thinking the variable could be called `coefficients` instead of `scale`. What do you think is more descriptive?

tfrederiksen: or `transformation_matrix`?

tfrederiksen: Maybe this low-level function for `SparseCSR` should even be changed to `transform`? We can still use `convert` for the high-level spin changes.

zerothi:
> Thanks, Nick. I will look at this asap but have a busy week ahead. Please hang on.

No worries, take your time :) I am too busy this week!

sisl/sparse_geometry.py
zerothi: Since it isn't exactly trivial to have the `convert` for the sparse_geometry classes you can just remove everything here. I don't think a user would suspect a Hamiltonian.convert to return a SparseCSR matrix?

sisl/sparse.py
zerothi: I think you are correct. `transform` is the better word! Perhaps that would allow the argument to be named `matrix` ? `.transform(matrix=...)` wouldn't that suffice in telling that matrix is the transformation matrix?

Then it could be implemented like this:

```python
matrix = np.asarray(matrix)
new._D = old._D.dot(matrix.T)
```
which means that the matrix is LHS operated.

sisl/sparse_geometry.py
tfrederiksen: OK, got it. Indeed the function should not change class.

sisl/sparse.py
tfrederiksen: Good, let's go for this.

tfrederiksen:
The low-level `SparseCSR.transform` seems to work well.

Based on it I wrote the new function `SparseOrbitalBZSpin.convert(spin)` that can create the object for another spin type. But I am unsure what the conversion rules should be.

Am I on the right track?

tfrederiksen:
Would it be relevant to include in `convert` also the option to switch between orthogonal and nonorthogonal models?

sisl/physics/sparse.py
```diff
@@ -1210,6 +1210,66 @@ def trs(self):
...
+            the desired new spin dimension
+        dtype : np.dtype, optional
+            data type contained in the matrix. See details of `Spin` for default values.
+        """
+        if not isinstance(spin, Spin):
+            spin = Spin(spin, dtype)
+
+        if spin == self._spin:
+            # same type, no conversion needed
+            return self.copy()
```
zerothi, line 1241: dtype argument needed.

zerothi (changes requested):
Looks great, some small comments, and perhaps a streamlining of the method. I think `transform` should be general enough to not have the `convert` method, what do you think?

sisl/physics/sparse.py
```diff
@@ -1210,6 +1210,66 @@ def trs(self):
 
         return new
 
+
+    def convert(self, spin, dtype=None):
+        """ Create a new matrix of the specified spin type through the following conversion rules:
```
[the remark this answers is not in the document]
tfrederiksen: OK, good idea to stick with `transform` also here and allow the user to pass a matrix directly.

To generalize further, perhaps one also would like to add a constant in the transformation, ie., `T(x) = A.x + b` where `b` is an optional `m-vector?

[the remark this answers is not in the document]
zerothi: As for the `b`, I think this should be covered elsewhere, like `H = H + [2, 1]` or similarly.
While currently this does not work, it really should ;)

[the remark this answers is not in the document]
zerothi: Ok, I found the error, it really does work now! One can do `H= H + [2, 1]` to add different values on different spin-components.

sisl/sparse.py
```diff
@@ -1377,6 +1377,39 @@ def tocsr(self, dim=0, **kwargs):
...
                           shape=shape, **kwargs)
 
+    def transform(self, matrix, dtype=None):
+        """ Apply a linear transformation :math:`R^n \rightarrow R^m` to the :math:`n`-dimensional elements of the sparse matrix.
+
+        Parameters
+        ----------
+        matrix : array_like
+            transformation matrix of shape :math:`m \times n`
+        dtype : `numpy.dtype`, optional
```
zerothi, line 1387: here one should not have the backtick's then the documentation renders better.

sisl/physics/sparse.py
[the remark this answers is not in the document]
tfrederiksen: fixed

sisl/sparse.py
tfrederiksen: fixed

sisl/physics/sparse.py
tfrederiksen: fixed

zerothi:
@tfrederiksen I did a few minor changes. I think the `matrix` should be the first argument, this goes better in line with the SparseCSR and for future, if this gets implemented for the SparseGeometryBZ class (not spin class).

In fact, this made be realize it could be useful to look at the dynamical matrix in terms of 9 values in the non-sparse direction?
Then `transform` could be used in this, directly. Hmm...

PR #333, opened 2021-05-12
More intuitive behavior of Geometry.swapaxes

juijan:
Calling `.swapaxes(..., swap='xyz')` on geometry creates unintuitive results, because only the Cartesian coordinates of atomic positions are being swapped. 

Example
``` python 
import sisl
geom = sisl.geom.zgnr(10).swap_axes(0, 1, swap='xyz')
geom.write(....)
```

With this change the Cartesian coordinates of the lattice vectors and the cell origin are being swapped as well.

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -2181,14 +2181,18 @@ class Geometry(SuperCellChild):
            the xyz (Cartesian) axis are swapped.
            Both may be in `swap`.
         """
-        xyz = np.copy(self.xyz)
-        if 'xyz' in swap:
-            xyz[:, axis_a] = self.xyz[:, axis_b]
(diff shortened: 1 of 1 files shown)


zerothi:
Thanks for the pr, but the intent is that if xyz is passed then only coordinates are swapped, not lattice vectors. For both to be swapped, simply use cell+xyz as argument.

So it is designed like this by intent?

zerothi:
@juijan as it stands I am reluctant to merge this. The fact is that when one has orthogonal lattice vectors, it seems natural to also swap lattice vectors, as you mention here. But for skewed lattices, it is not so obvious that is what one wants.  
At least when I have used this I wanted to only swap the atomic coordinates.

PR #339, opened 2021-05-21
GridPlot: grid isosurface was always calculated based on real values

pfebrer:
@tfrederiksen discovered the bug that `imag`, `mod`, `deg_phase`, `rad_phase` representations in 3D were all wrong.

This was basically because I was not modifying the values of the grid before running the `isosurface` method. Therefore, it was calculating **always** isosurfaces based on the real values.

The bug popped out luckily because when trying to calculate isosurfaces for the phase representations, the value was not inside of the range of real values of the grid, so `marching_cubes` complain.

So, message: **3D REPRESENTATIONS OTHER THAN `real` ARE WRONG BEFORE THIS BUG FIX**

This bug also made me realize that `marching_cubes` always converts the grid to `float` before calculating isosurfaces. Therefore it might be helpful to allow an argument in `Grid.isosurface` to let the user specify that they want the isosurface of the imaginary values, the modulus or the phase. What do you think?

PD: Sorry for the bug :(

### DIFF
--- a/sisl/viz/plotly/plots/grid.py
+++ b/sisl/viz/plotly/plots/grid.py
@@ -19,97 +19,105 @@ class GridPlot(Plot):
     Versatile visualization tool for any kind of grid.
 
     Parameters
-    -------------
+    ------------
     grid: Grid, optional
-        A sisl.Grid object. If provided, grid_file is ignored.
-    grid_file: Sile, optional
-        a sile that can read grids, e.g. `_gridSileSiesta` or `chgSileVASP` and friends.
+    	A sisl.Grid object. If provided, grid_file is ignored.
+    grid_file: cubeSile or rhoSileSiesta or ldosSileSiesta or rhoinitSileSiesta or rhoxcSileSiesta or drhoSileSiesta or baderSileSiesta or iorhoSileSiesta or totalrhoSileSiesta or stsSileSiesta or stmldosSileSiesta or hartreeSileSiesta or neutralatomhartreeSileSiesta or totalhartreeSileSiesta or gridncSileSiesta or ncSileSiesta or fdfSileSiesta or tsvncSileSiesta or chgSileVASP or locpotSileVASP, optional
+    	A filename that can be return a Grid through `read_grid`.
     represent:  optional
-        The representation of the grid that should be displayed
+    	The representation of the grid that should be displayed
     transforms:  optional
-        Transformations to apply to the whole grid.             It can be a
-        function, or a string that represents the path             to a
-        function (e.g. "scipy.exp"). If a string that is a single
-        word is provided, numpy will be assumed to be the module (e.g.
-        "square" will be converted into "np.square").              Note that
-        transformations will be applied in the order provided. Some
-        transforms might not be necessarily commutable (e.g. "abs" and
-        "cos").
-    transform_bc: optional
-        Transform the boundary conditions.
+    	Transformations to apply to the whole grid.             It can be a
+    	function, or a string that represents the path             to a
+    	function (e.g. "scipy.exp"). If a string that is a single
+    	word is provided, numpy will be assumed to be the module (e.g.
+    	"square" will be converted into "np.square").              Note that
+    	transformations will be applied in the order provided. Some
+    	transforms might not be necessarily commutable (e.g. "abs" and
+    	"cos").
     axes:  optional
-        The axis along you want to see the grid, it will be averaged along
-        the other ones
+    	             The axis along you want to see the grid, it will be
+    	reduced along the other ones, according to the the `reduce_method`
+    	setting.
     plot_geom: bool, optional
-        If True the geometry associated to the grid will also be plotted
+    	If True the geometry associated to the grid will also be plotted
     geom_kwargs: dict, optional
-        Extra arguments that are passed to geom.plot() if plot_geom is set to
-        True
+    	Extra arguments that are passed to geom.plot() if plot_geom is set to
+    	True
     zsmooth:  optional
-        Parameter that smoothens how data looks in a heatmap.
-        'best' interpolates data, 'fast' interpolates pixels, 'False'
-        displays the data as is.
+    	Parameter that smoothens how data looks in a heatmap.
+    	'best' interpolates data, 'fast' interpolates pixels, 'False'
+    	displays the data as is.
     interp: array-like, optional
-        Interpolation factors to make the grid finer on each axis.See the
-        zsmooth setting for faster smoothing of 2D heatmap.
+    	Interpolation factors to make the grid finer on each axis.See the
+    	zsmooth setting for faster smoothing of 2D heatmap.
+    transform_bc:  optional
+    	             The boundary conditions when a cell transform is applied
+    	to the grid. Cell transforms are only             applied when the
+    	grid's cell doesn't follow the cartesian coordinates and the
+    	requested display is 2D.
     nsc: array-like, optional
-        number of times the grid should be repeated
+    	Number of times the grid should be repeated
     offset: array-like, optional
-        The offset of the grid along each axis. This is important if you are
-        planning to match this grid with other geometry related plots.
+    	The offset of the grid along each axis. This is important if you are
+    	planning to match this grid with other geometry related plots.
     cut_vacuum: bool, optional
-        Whether the vacuum should not be taken into account for displaying
-        the grid.             This is essential especially in 3D
-        representations, since plotly needs to calculate the
-        isosurfaces of the grid.
+    	Whether the vacuum should not be taken into account for displaying
+    	the grid.             This is essential especially in 3D
+    	representations, since plotly needs to calculate the
+    	isosurfaces of the grid.
     trace_name: str, optional
-        The name that the trace will show in the legend. Good when merging
-        with other plots to be able to toggle the trace in the legend
+    	The name that the trace will show in the legend. Good when merging
+    	with other plots to be able to toggle the trace in the legend
     x_range: array-like of shape (2,), optional
-        Range where the X is displayed. Should be inside the unit cell,
-        otherwise it will fail.
+    	Range where the X is displayed. Should be inside the unit cell,
+    	otherwise it will fail.
     y_range: array-like of shape (2,), optional
-        Range where the Y is displayed. Should be inside the unit cell,
-        otherwise it will fail.
+    	Range where the Y is displayed. Should be inside the unit cell,
+    	otherwise it will fail.
     z_range: array-like of shape (2,), optional
-        Range where the Z is displayed. Should be inside the unit cell,
-        otherwise it will fail.
+    	Range where the Z is displayed. Should be inside the unit cell,
+    	otherwise it will fail.
     crange: array-like of shape (2,), optional
-        The range of values that the colorbar must enclose. This controls
-        saturation and hides below threshold values.
+    	The range of values that the colorbar must enclose. This controls
+    	saturation and hides below threshold values.
     cmid: int, optional
-        The value to set at the center of the colorbar. If not provided, the
-        color range is used
+    	The value to set at the center of the colorbar. If not provided, the
+    	color range is used
     colorscale: str, optional
-        A valid plotly colorscale. See https://plotly.com/python/colorscales/
+    	A valid plotly colorscale. See https://plotly.com/python/colorscales/
+    reduce_method:  optional
+    	             The method used to reduce the dimensions that will not
+    	be displayed in the plot.
     isos: array-like of dict, optional
-        The isovalues that you want to represent.             The way they
-        will be represented are of course dependant on the type of
-        representation:                 - 1D representations: Just a scatter
-        mark                 - 2D representations: A contour (i.e. a line)
-        - 3D representations: A surface                Each item is a dict.
-        Structure of the expected dicts:{         'name': The name of the iso
-        query. Note that you can use $isoval$ as a template to indicate where
-        the isoval should go.         'val':          'frac': If val is not
-        provided, this is used to calculate where the isosurface should be
-        drawn.                     It calculates them from the minimum and
-        maximum values of the grid like so:                     If iso_frac =
-        0.3:                     (min_value-----
-        ISOVALUE(30%)-----------max_value)                     Therefore, it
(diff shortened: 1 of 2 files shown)


pfebrer:
I will add tests for this, don't merge it yet.

sisl/viz/plotly/plots/grid.py
```diff
@@ -512,6 +512,8 @@ def _set_data(self, axes, nsc, interp, trace_name, transforms, represent, cut_va
         elif represent.endswith("phase"):
             values = np.angle(grid.grid, deg=represent.startswith("deg"))
```
zerothi, line 513: What happens if `represent` isn't any of these?

sisl/viz/plotly/plots/grid.py
```diff
@@ -512,6 +512,8 @@ def _set_data(self, axes, nsc, interp, trace_name, transforms, represent, cut_va
...
 
+        grid.grid = values
+
         #Remove the leftover dimensions
         values = np.squeeze(values)
```
zerothi, line 519: what about using `plot_func = getattr(self, f"_plot{values.ndim}D")` ?
That would also raise an issue for erroneous data?

zerothi (changes requested):
It looks like `values` isn't defined in all cases?

Secondly, one may be interested in plotting both real+imaginary parts in one plot.
In particular one could then use 4 colors: real positive/negative and imag positive/negative

That is quite useful in some cases.

PS. No worries about bugs, we all make them, and better to fix them. :)

sisl/viz/plotly/plots/grid.py
pfebrer: Good point, the obvious thing to do here would be
```
else:
    raise ValueError()
```
but since parameters are defined with input fields and input fields parse values before passing them to the plot I think it would be a good idea to raise the error when parsing, to avoid unnecessary calculations (the code would run until here, which might take a while, even though it's clearly deemed to fail). 

Also, raising the error on parse would make life much more easy for GUIs to know if there's a problem with a specific setting, for example, as they could check in real time if the setting is valid or not before trying to update the plot.

By the way, you have made me realise that there should probably be a new error type, something like `SislVizSettingValueError` for the same reason.

sisl/viz/plotly/plots/grid.py
pfebrer: Done! (not pushed yet)

pfebrer:
> Secondly, one may be interested in plotting both real+imaginary parts in one plot.
> In particular one could then use 4 colors: real positive/negative and imag positive/negative

This can already be done with the general functionality of `MultiplePlot` (https://zerothi.github.io/sisl/docs/latest/visualization/plotly/combining-plots/Intro%20to%20multiple%20plots.html). It allows you to do:

```python
grid.plot(represent=["imag", "real"], varying="represent")
```

This would draw both the imaginary and real contributions in the same plot, while

```python
grid.plot(represent=["imag", "real"], subplots="represent")
```

would create a subplot for each of them (warning, 3D subplots are not possible yet), and

```python
grid.plot(represent=["imag", "real"], animate="represent")
```

would create animation where the first step displays the real part and the second step displays the imaginary part. 

So I'd say it is unnecessary to implement it inside `GridPlot`. It would also make the code more complex. But I'm open to discuss it :)

PD: I don't like the keyword `varying` at all, if you have a better suggestion it would be great!

zerothi:
> > Secondly, one may be interested in plotting both real+imaginary parts in one plot.
> > In particular one could then use 4 colors: real positive/negative and imag positive/negative
> 
> This can already be done with the general functionality of `MultiplePlot` (https://zerothi.github.io/sisl/docs/latest/visualization/plotly/combining-plots/Intro%20to%20multiple%20plots.html). It allows you to do:
> 
> ```python
> grid.plot(represent=["imag", "real"], varying="represent")
> ```
> 
> This would draw both the imaginary and real contributions in the same plot, while
> 
> ```python
> grid.plot(represent=["imag", "real"], subplots="represent")
> ```
> 
> would create a subplot for each of them (warning, 3D subplots are not possible yet), and
> 
> ```python
> grid.plot(represent=["imag", "real"], animate="represent")
> ```
> 
> would create an animation where the first step displays the real part and the second step displays the imaginary part.
> 
> So I'd say it is unnecessary to implement it inside `GridPlot`. It would also make the code more complex. But I'm open to discuss it :)
> 
> PD: I don't like the keyword `varying` at all, if you have a better suggestion it would be great!

Ok! :)

What about `variables`?

zerothi:
Don't know if there are multiple variables, or always 1, in the latter case `variable`?

pfebrer:
```python
grid.plot(setting1=[0,1], setting2=[2,3] varying=["setting1", "setting2"])
```

is allowed. It creates a plot with `(setting1, setting2) = (0, 2)` and another one where `(setting1, setting2) = (1, 3)`.

Also
```python
grid.plot(varying={"setting1": [0,1], "setting2": [2,3]})
```

is equivalent.

So, `variables`? I was looking for some word that makes it clear that they are going to be drawed in the same plot. I find `variables` better than `varying`, but still not ideal :) Also it is important that the name does not clash with names that people could give to settings, and in that respect `variables` seems more dangerous to me. Taking into account that I think I like `variable` better (as an adjective, not as a noun :smile: ).

tfrederiksen:
Have you considered having a single parameter here to control what kind of representation to be produced, e.g. `style : {"single", "multiple", "subplots", "animation"}`? To me this seems more intuitive from the user perspective, but perhaps it presents some limitations?

pfebrer:
Yes, but then it is extremely difficult to determine which are the settings that need to be iterated. Unless you have some idea about it that I didn't think about. You can't check for lists, because there might be settings that accept lists as a value.

zerothi:
Perhaps it would be useful to use tuples or callable in `setting` which are then elements in the varying space.
This is a bit unintuitive but follows the pandas design where list and tuples have different meanings. It would save an argument.

pfebrer:
**Now related to the bug fix:**

While writing the tests I found that maybe the order in which things are applied during the processing of the grid is not optimal. Here is the order:
- Transform the grid (i.e. apply functions such as `np.cos`).
- Cut the grid according to `x_range`, `y_range` and `z_range`.
- Interpolate values to a grid of different shape if `interp` is passed.
- Average dimensions that are not going to be plotted.
- Choose representation (`real`, `imag`, ...)

Could you comment on which order do you think it's best? I'm specially concerned about whether the representation should be chosen at the beggining or at the end.

Maybe it makes sense to allow the user to choose the order of processing?

zerothi:
> Have you considered having a single parameter here to control what kind of representation to be produced, e.g. `style : {"single", "multiple", "subplots", "animation"}`? To me this seems more intuitive from the user perspective, but perhaps it presents some limitations?

I think @tfrederiksen has a point, something like:
```
style={`animate': {"var1": [1, 2, 3], 'style': {"multiple": {"var2": [1, 2, 3]}}}, "multiple": {"var3": [2, 3, 4}}
```
the dictionary levels would be double nested (no animation within an animation. Whereas there could be as many of the others as needed?

1-keyword here I think would be best... I know it is a bit complicated.. But...
```
plot(multiple={"var3": [1, 2, 3]}, animate={"var1": [1, 2, 3]})
```
is already quite verbose, so the chances of this being automated in some way is pretty high.

Alternatively you could create objects which determine how to mix them (they should be hashable, mergeable and iterable), and the contained values could even be grabbed from the `__call__` method?
```
class Animate:
    variable: str
    values: list-like
class Multiple
    variable: str
    values: list-like

var3 = Multiple("var3", [1, 2, 3])
var1 = Animate("var1", [1, 2, 3])
...plot(style=var3 & var1) # meaning both at same time
...plot(style=var3 | var1) # meaning the product of the two combinations
```
Ok, perhaps this last idea isn't as great... Just throwing in some ideas.

pfebrer:
I moved the discussion about multiple plots to https://github.com/zerothi/sisl/issues/340 so that we can focus here on the bug that the PR solves and other things related to `GridPlot` :)

So, in the last push I:
- Moved choosing the representation to the beggining of the processing chain
- Added the `reduce_method` setting, which accepts `{"average", "sum"}` to determine how to reduce the undisplayed dimensions. I did this because sometimes I was annoyed by the plot averaging the axis when I wanted to see the total contribution (to analyze charge accumulation). Complaints suggestions about the setting's name?
- Added tests for all different representations in 1D, 2D and 3D.

Nick: You added the `skip_skimage` part at the top of the `test_grid.py` file. For me it skips tests even though `skimage` is available and `skip_skimage` is set to `pytest.mark.skip(False, reason="scikit-image (skimage) not available")`. Does it work properly for you?

zerothi:
> I moved the discussion about multiple plots to #340 so that we can focus here on the bug that the PR solves and other things related to `GridPlot` :)

Good! Thanks!

> So, in the last push I:
> 
> * Moved choosing the representation to the beggining of the processing chain
> * Added the `reduce_method` setting, which accepts `{"average", "sum"}` to determine how to reduce the undisplayed dimensions. I did this because sometimes I was annoyed by the plot averaging the axis when I wanted to see the total contribution (to analyze charge accumulation). Complaints/suggestions about the setting's name?

Couldn't we re-use the `apply` naming scheme to allow function arguments or str`s to be fetched from `numpy`? That would also allow more functionality without complicating the interface. My problem with these very long interfaces is that they are extremely complex to document and use + there are many corner cases one could get into. Generally I think these should be really simple and try and offload the hard work to simpler methods.

> * Added tests for all different representations in 1D, 2D and 3D.

> Nick: You added the `skip_skimage` part at the top of the `test_grid.py` file. For me it skips tests even though `skimage` is available and `skip_skimage` is set to `pytest.mark.skip(False, reason="scikit-image (skimage) not available")`. Does it work properly for you?

I see, there was a bug `skip -> skipif`, added and merged, please rebase if needed.

pfebrer:
> Couldn't we re-use the apply naming scheme to allow function arguments or strs to be fetched from numpy`?

I'm not sure I fully understand what you are proposing, but if I understand correctly, you want the value of this setting to be passed to `Grid.apply` ? The problem is that these are methods that reduce the dimensionality of the grid, so they can't be treated by the `apply` method unless we improve it. Also, to be able to generalize, we should impose the constraint that the reducing function accepts the `axis` argument, otherwise we will not be able to support everything that can be done.

I think it could be a great idea if we find a nice way to generalize dimension reduction in the `Grid.apply` method.

Did you mean something different?

zerothi:
Ok, I am a bit lost now, we moved some of the discussion elsewhere, so where are we here? ;)

pfebrer:
Yes my bad, I fixed some other things about the grid plot in this PR. The main thing to discuss is the order in which things are processed.

Right now it is:
- Choose representation (`real`, `imag`, ...)
- Transform the grid (i.e. apply functions such as `np.cos`).
- Cut the grid according to `x_range`, `y_range` and `z_range`.
- Interpolate values to a grid of different shape if interp is passed.
- Reduce dimensionality with `average` or `sum`.

Also, I needed you to clarify what you meant here:

> Couldn't we re-use the apply naming scheme to allow function arguments or strs to be fetched from numpy`? That would also allow more functionality without complicating the interface. My problem with these very long interfaces is that they are extremely complex to document and use + there are many corner cases one could get into. Generally I think these should be really simple and try and offload the hard work to simpler methods.

If you could give an example that would be nice :)

zerothi:
> Right now it is:
> 
> * Choose representation (`real`, `imag`, ...)
> * Transform the grid (i.e. apply functions such as `np.cos`).
> * Cut the grid according to `x_range`, `y_range` and `z_range`.
> * Interpolate values to a grid of different shape if interp is passed.
> * Reduce dimensionality with `average` or `sum`.

I think I would just swap the 2 last ones. Interpolation and reducing dimensionality seems to be problematic. I think having more consistent numbers by averaging over *true* numbers, then interpolating would be better. Otherwise I think it looks ok.

sisl/viz/plotly/plots/grid.py
```diff
@@ -609,6 +618,34 @@ def _transform_grid(grid, transform):
...
+    
+    @staticmethod
+    def _get_representation(grid, represent):
+        """Returns a representation of the grid
+        
+        Parameters
+        ------------
+        grid: sisl.Grid
+            the grid for which we want return
+        represent: {"real", "imag", "mod", "deg_phase", "rad_phase"}
```
zerothi, line 630: perhaps a default `phase` which is using `radians`?

zerothi:
> > Couldn't we re-use the apply naming scheme to allow function arguments or strs to be fetched from numpy`? That would also allow more functionality without complicating the interface. My problem with these very long interfaces is that they are extremely complex to document and use + there are many corner cases one could get into. Generally I think these should be really simple and try and offload the hard work to simpler methods.
> 
> If you could give an example that would be nice :)

My point is that your `_set_data` is extremely cumbersome to understand :) Its # of arguments is 15 which is very hard to understand and follow. While many of the flags are passed automatically, there may be better ways of doing these complicated things. I.e. not necessarily all in 1 go.

So it was more the _idea_ rather than anything now. :)

sisl/viz/plotly/plots/grid.py
pfebrer: Actually, currently anything that ends with `"phase"` and doesn't begin with `"deg"` is interpreted as `"rad_phase"`. I should probably be more strict and accept only "phase", "rad_phase" and "deg_phase", right? :sweat_smile:

zerothi: Yeah ;)

pfebrer:
> My point is that your _set_data is extremely cumbersome to understand :)

Totally, that's why I'm trying to put each part into a method, to improve readability. The arguments are there because these are the settings that `_set_data` uses. It's the way of telling the `Plot` class that `_set_data` should be ran if these settings are updated. Also it allows other classes to pass some settings:

https://github.com/pfebrer/sisl/blob/2cd243353f09b3611ae49d9fb99598b4ff2c8a92/sisl/viz/plotly/plots/grid.py#L1500

> I think I would just swap the 2 last ones.

Good!

pfebrer:
Everything is ready now, all tests pass, ready to merge :+1:

zerothi (approved):

PR #345, opened 2021-05-28
bug: origo not taken into account for plots

pfebrer:
This bug was also discovered by @tfrederiksen here: https://discord.com/channels/742636379871379577/845235607461691402/845284162994176081

Basically the `origo` of the supercell was not taken into account for plotting.

With this solution, the axes of the cell are plotted from `origo` instead of (0,0,0) and the coordinates of the atoms are plotted as they are. Grid plots have also been modified so that their axes ranges take origo into account.

It has been suggested that `origo` should be the (0,0,0) point. In that case, atoms coordinates should be displaced (coords -= `origo`), and the modification to axes display and grid plots is not needed. Let me know what you prefer.

Here is the display of a grid that Thomas sent me, with this coordinates and origo:
```python
print(grid.geometry.xyz)
print(grid.origo)
```
```
[[ 0.          0.          0.        ]
 [ 0.          0.          2.56000131]
 [ 0.          0.          5.12000261]
 [ 0.          0.          7.68000339]
 [ 0.          0.         10.2400047 ]
 [ 0.          0.         12.800006  ]
 [ 0.          0.         14.70770427]
 [ 0.          0.         15.94428144]
 [ 0.          0.         19.36000892]
 [ 0.          0.         21.92001023]
 [ 0.          0.         24.48001154]
 [ 0.          0.         27.04001284]
 [ 0.          0.         29.60001415]
 [ 0.          0.         32.16001493]]
[-5.00000215 -5.00000215 -5.00000215]
```
Before fixing the bug:
origo_not_implemented

And after fixing the bug (in the above mentioned way):
origo_implemented

### DIFF
--- a/sisl/viz/plotly/plots/geometry.py
+++ b/sisl/viz/plotly/plots/geometry.py
@@ -442,7 +442,7 @@ class GeometryPlot(Plot):
         if unique:
             points = np.unique(points, axis=0)
 
-        return np.array([xyz(coeffs) for coeffs in points])
+        return self.geometry.origo + np.array([xyz(coeffs) for coeffs in points])
 
     def _get_atoms_bonds(self, bonds, atom, geom=None, sanitize_atom=True):
         """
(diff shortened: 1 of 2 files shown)


zerothi:
@tfrederiksen was there some confusing on the name convention, should it be `origin` and not `origo`?

I think we could change it without much problems since this hasn't been thoroughly used in many places?

zerothi (approved):

zerothi:
I am going to merge, but let me know if we should change origo with origin?

pfebrer:
That means you agree with 

> With this solution, the axes of the cell are plotted from origo instead of (0,0,0) and the coordinates of the atoms are plotted as they are. Grid plots have also been modified so that their axes ranges take origo into account.

instead of:

> It has been suggested that origo should be the (0,0,0) point. In that case, atoms coordinates should be displaced (coords -= origo), and the modification to axes display and grid plots is not needed.

?

zerothi:
My idea of the `origo` is that the atoms are *already* displaced, whereas the unit-cell isn't. So the origo should only refer to the unit-cell when needed. Perhaps we need a different discussion on what it really should do...

PR #349, opened 2021-06-11
Brillouin zone xarray Dataset dispatch

pfebrer:
I was trying to parallelize a fatbands calculation using the `dataarray` dispatch. However I was failing because it looked something like this:

```python
weights_store = []

def _weights_from_eigenstate(eigenstate):

    weights = eigenstate.norm2(sum=False)
    weights_store.append(weights)
    return eigenstate.eig

bz.apply.dataarray.eigenstate(wrap=_weights_from_eigenstate)
```

and the problem was that `weights_store` was not available to the processes, so it ended up empty. Therefore, I realized that I needed to return everything at the wrapping function and let the dispatch handle it. I saw that the `dataarray` dispatch has some support for datasets, but it was not what I needed. The dispatch that I added can be used like this:

```python
import sisl
import numpy as np

# Construct a hamiltonian for graphene
gr = sisl.geom.graphene()
H = sisl.Hamiltonian(gr)
# And a band structure
bz = sisl.BandStructure(H, [[0, 0, 0], [2/3, 1/3, 0], [1/2, 0, 0]], 9, ["Gamma", "M", "K"])

# The wrapping function to calculate weights
def _weights_from_eigenstate(eigenstate):
    """Function that will wrap the loop to calculate the fatbands weights"""
    weights = eigenstate.norm2(sum=False)
    
    return np.array((eigenstate.eig, *weights.real)).T

# And now run the calculation to get the dataset
dataset = bz.apply.dataset.eigenstate(
        coords=("band",), # The coordinates coming from the brillouin zone
        wrap=_weights_from_eigenstate, # The wrapping function to return all the values
        datavars=[
            "eigval", # First data in the array are the eigenvalues
            ("weight", "orb", int(bz.parent.no)), # After that, there are bz.parent.no chunks of data 
            # that contain information about the weights. It should all be packed into the same data variable
            # while adding the new "orb" coordinate
        ], 
)
```

which gives you:
```
<xarray.Dataset>
Dimensions:  (band: 2, k: 9, orb: 2)
Coordinates:
  * k        (k) int32 0 1 2 3 4 5 6 7 8
  * band     (band) int32 0 1
  * orb      (orb) int32 0 1
Data variables:
    eigval   (k, band) float64 -8.1 8.1 -7.774 7.774 ... -1.977 1.977 -2.7 2.7
    weight   (k, band, orb) float64 0.5 0.5 0.5 0.5 0.5 ... 0.5 0.5 0.5 0.5 0.5
Attributes:
    bz:       BandStructure{nk: 9,\n SuperCell{nsc: [3 3 1],\n  A=[2.130, 1.2...
    parent:   Hamiltonian{non-zero: 8, orthogonal: True,\n Spin{unpolarized, ...
```

and (the reason why I did this) the same calculation can be done in parallel now:
```python
with bz.apply.renew(pool=True) as parallel:
    dataset = parallel.dataset.eigenstate(
        coords=("band",), datavars=["eigval", ("weight", "orb", int(bz.parent.no))], 
        wrap=_weights_from_eigenstate
    )
```

Let me know if you think it could be useful (you can also reimplement it to make it better, I don't care as long as it has the same functionality). Otherwise I can register it in the fatbands plot of course, that's the beauty of dispatchers :), but I thought it might be useful in other cases as well.

Cheers!

### DIFF
--- a/sisl/physics/_brillouinzone_apply.py
+++ b/sisl/physics/_brillouinzone_apply.py
@@ -334,42 +334,43 @@ class AverageApply(BrillouinZoneParentApply):
 class XArrayApply(NDArrayApply):
     def __str__(self, message="xarray"):
         return super().__str__(message)
+    
+    @staticmethod
+    def _fix_coords_dims(nk, array, coords, dims, prefix="v"):
+        if coords is None and dims is None:
+            # we need to manually create them
+            coords = [('k', _a.arangei(nk))]
+            for i, v in enumerate(array.shape[1:]):
+                coords.append((f"{prefix}{i+1}", _a.arangei(v)))
+        elif coords is None:
+            coords = [('k', _a.arangei(nk))]
+            for i, v in enumerate(array.shape[1:]):
+                coords.append((dims[i], _a.arangei(v)))
+            # everything is in coords, no need to pass dims
+            dims = None
+        elif isinstance(coords, dict):
+            # ensure coords has "k" as dimensions
+            if "k" not in coords:
+                coords["k"] = _a.arangei(nk)
+                dims = list(dims)
+                # ensure we have dims first
+                dims.insert(0, "k")
+        else:
+            # add "k" correctly
+            coords = list(coords)
+            coords.insert(0, ('k', _a.arangei(nk)))
+            for i in range(1, len(coords)):
+                if isinstance(coords[i], str):
+                    coords[i] = (coords[i], _a.arangei(array.shape[i]))
+            # ensure dims is not used, everything is in coords
+            # and since it is a list, it should also be the correct order
+            # TODO add check that dims and coords match in order
+            # or convert dims to list and prepend "k"
+            dims = None
+        return coords, dims
 
     def dispatch(self, method):
         """ Dispatch the method by returning a DataArray or data-set """
-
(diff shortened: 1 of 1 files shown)


pfebrer:
Nevermind for now, I think this can be made more intuitive with a different implementation. But I'd like to focus on merging https://github.com/zerothi/sisl/pull/312 first

zerothi:
I believe this can already be done with:
```
d = band_struct.apply.renew(unzip=True).xarray.eigenstate(wrap=wrap, coords=[('eig',), ("band", "orb")])
```
?

zerothi:
But, yes, currently one cannot name the variables, I'll have to amend the `xarray` class here. Would this work?

zerothi:
Oh, actually

```python
def wrap(es):
    return es.eig, es.norm2(sum=False)
d = band_struct.apply.renew(unzip=True).xarray.eigenstate(wrap=wrap, coords=['eig', ("band", "orb")], name=['eigenstate', 'norm2'])
```
already works in the latest commit (the latest commit allows `'eig'` rather than `('eig',)`)

pfebrer:
Great! Thank you, that totally works! Somehow I was playing with unzip but didn't manage to get this behavior.

Issue #352, opened 2021-07-06
Computation of matrix elements


tfrederiksen:
We are trying to compute some general matrix elements of the form `A_ij = < psi_i | A | psi_j >` for a Hamiltonian object `A` and where `psi_i` and `psi_j` are not necessarily referring to the same `EigenstateElectron` object.

Maybe the sisl function `expectation` could be generalized to allow for an optional `right` state?

For instance to enable something like this:
```python
import sisl
g = sisl.geom.graphene()
H = sisl.Hamiltonian(g)
H.construct([[0.1, 1.5], [0, 2.7]])
s1 = H.eigenstate([2/3, 1/3, 0])
s2 = H.eigenstate([1/3, 2/3, 0])
A = s1.expectation(H.Hk(), right=s2, diag=False)
```
What do you think about this? I am willing to give such an implementation a try.

zerothi:
Sounds like an idea, perhaps we should convert the `inner` function to accept the matrix `A` and then remove `expectation`? They surely are quite similar. Also, I think the `inner` should not have the `align` keyword. What do you think?

tfrederiksen:
You mean something like:
```python
s1.inner(A=H.Hk(), right=s2, diag=False)
```
where the default for the first argument would be the identity matrix (simple inner product)?

I do not really understand the `align` function - what is the purpose?

zerothi:
> ```python
> s1.inner(A=H.Hk(), right=s2, diag=False)
> ```
> 
> where the default for the first argument would be the identity matrix (simple inner product)?

Yes, something like this. But perhaps having `A` argument being on the right, so that it is *more* optional with a matrix.

> I do not really understand the `align` function - what is the purpose?

The `align` option may be used to minimize the phase difference to the lowest value. I implemented these in the functions but have come to dislike them since users can do this themselves. So I plan on removing all `align=` arguments in the `State` classes.  
Users should then prefer:
```python
s1 = ...
s2 = ...
s2 = s2.align(s1)
...
```
instead.

zerothi:
woops, pressed wrong button.

PR #353, opened 2021-07-06
Generalized inner product for ElectronState

tfrederiksen:
As discussed in #352 this branch

- [x] generalizes the `inner` product to allow for an optional operator (matrix) to be passed,
- [x] removes the `align` keyword in `inner`
- [x] removes the `expectation` function

### DIFF
--- a/sisl/physics/electron.py
+++ b/sisl/physics/electron.py
@@ -1646,6 +1646,7 @@ class _electron_State:
             ndim = M.ndim
 
         # TODO, perhaps check that it is correct... and fix multiple transposes
+        left = self.state
         if right is None:
             right = self.state
         else:
@@ -1655,16 +1656,25 @@ class _electron_State:
                 right = right.state
             if len(right.shape) == 1:
                 right = right.reshape(1, -1)
+            if diag:
+                if len(left) != len(right):
+                    warn(f"{self.__class__.__name__}.inner matrix product is non-square, only the first {min(len(left), len(right))} diagonal elements will be returned")
+                    if len(left) < len(right):
+                        right = right[:len(left)]
+                    else:
+                        left = left[:len(right)]
+                elif not np.allclose(left, right):
+                    warn(f"{self.__class__.__name__}.inner only diagonal elements requested, although left.state != right.state")
 
         if diag:
             if ndim == 2:
-                Mij = einsum('ij,ji->i', conj(self.state), M.dot(right.T))
+                Mij = einsum('ij,ji->i', conj(left), M.dot(right.T))
             elif ndim == 1:
-                Mij = einsum('ij,j,ij->i', conj(self.state), M, right)
+                Mij = einsum('ij,j,ij->i', conj(left), M, right)
         elif ndim == 2:
-            Mij = dot(conj(self.state), M.dot(right.T))
+            Mij = dot(conj(left), M.dot(right.T))
         elif ndim == 1:
-            Mij = einsum('ij,j,kj->ik', conj(self.state), M, right)
+            Mij = einsum('ij,j,kj->ik', conj(left), M, right)
 
         return Mij
 
--- a/sisl/physics/electron.py
+++ b/sisl/physics/electron.py
@@ -1615,7 +1615,7 @@ class _electron_State:
         S = self.Sk()
         return conj(self.state) * S.dot(self.state.T).T
 
-    def inner(self, right=None, matrix=None, diagonal=True):
+    def inner(self, right=None, matrix=None, diag=True):
         r""" Return the inner product by :math:`\mathbf M_{ij} = \langle\psi_i| \mathbf M |\psi'_j\rangle`
 
         Parameters
@@ -1625,7 +1625,7 @@ class _electron_State:
            product with itself. This object will always be the left :math:`\langle\psi_i|`.
         matrix : array_like, optional
            a matrix that expresses the operator `M`. Defaults to the overlap matrix `S`.
-        diagonal : bool, optional
+        diag : bool, optional
            only return the diagonal matrix :math:`\mathbf M_{ii}`.
 
         Raises
@@ -1645,7 +1645,7 @@ class _electron_State:
 
         # TODO, perhaps check that it is correct... and fix multiple transposes
         if right is None:
-            if diagonal:
+            if diag:
                 return einsum('ij,ji->i', conj(self.state), M.dot(self.state.T))
             return dot(conj(self.state), M.dot(self.state.T))
 
@@ -1655,7 +1655,7 @@ class _electron_State:
 
             # Same as State.inner
             # In the current implementation we require no overlap matrix!
-            if diagonal:
+            if diag:
                 if self.shape[0] != right.shape[0]:
                     return np.diag(dot(conj(self.state), M.dot(right.state.T)))
                 return einsum('ij,ji->i', conj(self.state), M.dot(right.state.T))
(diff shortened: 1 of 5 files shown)


sisl/physics/electron.py
```diff
@@ -1615,14 +1615,16 @@ def norm2(self, sum=True):
         S = self.Sk()
         return conj(self.state) * S.dot(self.state.T).T
 
-    def inner(self, right=None, diagonal=True, align=False):
-        r""" Return the inner product by :math:`\mathbf M_{ij} = \langle\psi_i|\psi'_j\rangle`
+    def inner(self, right=None, matrix=None, diagonal=True, align=False):
```
zerothi, line 1618: is `matrix` too long, or should we just use `A`, or `M`?

Also, some places I use `diagonal` others `diag`, which do you prefer? Numpy has the `np.diag` routine, so perhaps `diag` is ok.

tfrederiksen: I think `matrix` is fine (when next to `right`). I can change `diagonal -> diag`, sounds sufficiently descriptive to me.

zerothi: ok, sounds good.

sisl/physics/electron.py
```diff
@@ -1637,32 +1637,36 @@ def inner(self, right=None, diagonal=True, align=False):
...
         # TODO, perhaps check that it is correct... and fix multiple transposes
         if right is None:
-            if diagonal:
-                return einsum('ij,ji->i', conj(self.state), S.dot(self.state.T))
-            return dot(conj(self.state), S.dot(self.state.T))
-
+            right = self.state
         else:
-            if "FakeSk" in S.__class__.__name__:
+            if "FakeSk" in right.__class__.__name__:
```
tfrederiksen, line 1652: I am unsure about these checks. How to ensure the inner product refers to quantities with the same overlap matrix?

sisl/physics/tests/test_hamiltonian.py
```diff
@@ -589,16 +589,23 @@ def test_eig4(self, setup):
             assert np.allclose(e, es.eig)
             assert np.allclose(v, es.state.T)
             assert np.allclose(es.norm2(), 1)
-            assert np.allclose(es.inner(diagonal=False) - np.eye(len(es)), 0)
+            assert np.allclose(es.inner(diag=False), np.eye(len(es)))
 
-            assert es.inner(es.sub(0)).shape == (1, )
-            assert es.inner(es.sub(0), diagonal=False).shape == (len(es), 1)
+            assert es.inner(es.sub(0)).shape == (len(es), )
```
tfrederiksen, line 594: Note that here the number of bras and kets differ - what is then the meaning of asking just for the diagonal part (the default)?

sisl/physics/electron.py
zerothi: Yeah, this is generally not the case, the easiest is to not allow `right=not None`, as is it.

Then we can revise later.

sisl/physics/tests/test_hamiltonian.py
zerothi: Agreed... This is not clear. However, currently it just returns the first `n` diagonal components where `n` is the lowest dimension of the `left/right` state vectors. Agreed it does not make sense... Perhaps we should issue a warning in these cases, but then again, users would probably force `diag=False` in those cases since that would be what they want... Hmmm...

tfrederiksen: Sounds good to issue a warning when `diag=True` and `right` is provided.

zerothi: something like this:
```python
 warnings.warn(f"{self.__class__.__name__}.inner matrix product is non-square, only the first {min(len(self), len(right))} diagonal elements will be returned")
```

sisl/physics/electron.py
tfrederiksen: I think this check is not working. At least I could not invoke the error with this non-orthogonal example:
```python
import sisl

geom = sisl.geom.graphene()
H = sisl.Hamiltonian(geom, orthogonal=False)
H.construct([[0.1, 1.5], [[0., 1.], [2.7, 0.1]]])
print(H.tocsr(-1))

es1 = H.eigenstate([0.1] * 3)
es2 = H.eigenstate([0.2] * 3)
ds = abs(es1.Sk() - es2.Sk()).toarray().sum()
assert ds > 0

print(es1.inner(es2, diag=False))
```
which returns
```python
  (0, 0)	1.0
  (0, 1)	0.1
  (0, 5)	0.1
  (0, 9)	0.1
  (1, 1)	1.0
  (1, 0)	0.1
  (1, 10)	0.1
  (1, 14)	0.1
[[ 0.92761498+0.20932275j -0.03659905+0.16218892j]
 [-0.06346039+0.2812251j   0.9656287 +0.21790081j]]


```

zerothi: Yeah, let me have a go!

tfrederiksen: Maybe it would be better to make `inner` work also for non-ortho models? I think this would be an equally important use case.

tfrederiksen: What would actually be the general approach? To expand `right` to the whole supercell and then do the inner product with "unfolded" S_ij?

zerothi: Yeah, checking for the overlap matrix is a mess. I think we should just forget that... I think there is no general approach in these cases, not even lowdin would work since two different overlap matrices would mean two different basis'.

So, I don't really know :(

zerothi:
Could you check the latest addition?

sisl/physics/electron.py
tfrederiksen: My point is that maybe we should avoid altogether having the operator `A`(or overlap `S`) folded onto a specific k-point, ie., to do something like
```python
A_{nk,mk'} = \sum_{i in primitive cell} \sum_{j in extended cell} <nk|i><i|A|j><j|mk'>
```
where `<i|nk>` is the left state vector in the primitive cell (at `k` in the BZ) and `<j|mk'>` the right state vector in the *extended* cell (at `k'`). Wouldn't that be correct/exact?

zerothi: But wouldn't you still require the phase factors for the operator in the extended cell? Perhaps that needs some testing? ;)

One can get the full *extended* overlapmatrix using `H.Sk(format='sc:' + format)` where `sc` denotes the retrieval of the matrix in supercell format with shape `(no, no * n_s)`

tfrederiksen:
The latest addition works fine here!

zerothi:
Perhaps before merging, would it make sense to change `right` to `ket`, it clarifies the bra/ket notation?

tfrederiksen:
> Perhaps before merging, would it make sense to change `right` to `ket`, it clarifies the bra/ket notation?

Good idea! Fixed in 96d93ce.

zerothi:
Will merge! :+1:

tfrederiksen:
There is still no warning if the basis is non-orthogonal!

zerothi:
I will add the warning and merge, thanks!

sisl/physics/electron.py
tfrederiksen: Can one also get the state vector(s) in the *extended* cell in a similar way?

zerothi: Not currently, I planned on doing this because I had a use for it. It should be rather trivial by adding some tiling functionality to the state objects. This belongs to a different issues/pr. :)

PR #355, opened 2021-07-07
mnt: enabled tiling an eigenstate

### LINKED ISSUE #354: Can one also get the state vector(s) in the *extended* cell in a similar way?
Can one also get the state vector(s) in the *extended* cell in a similar way?

_Originally posted by @tfrederiksen in https://github.com/zerothi/sisl/pull/353#discussion_r665276788_

zerothi:
Using the Bloch expansion we can get the same state for
a new lattice by expanding it.

 - [x] Closes #354 
 - [x] Tests added

@tfrederiksen could you have a look at this

### DIFF
--- a/sisl/physics/state.py
+++ b/sisl/physics/state.py
@@ -305,7 +305,7 @@ class State(ParentContainer):
         sub.info = self.info
         return sub
 
-    def tile(self, reps, axis, normalize=False):
+    def tile(self, reps, axis, normalize=False, offset=0):
         r"""Tile the state vectors for a new supercell
 
         Tiling a state vector makes use of the Bloch factors for a state by utilizing
@@ -314,7 +314,10 @@ class State(ParentContainer):
 
            \psi_{\mathbf k}(\mathbf r + \mathbf T) \propto e^{i\mathbf k\cdot \mathbf T}
 
-        where :math:`\mathbf T = i\mathbf a_0 + j\mathbf a_1 + k\mathbf a_2`.
+        where :math:`\mathbf T = i\mathbf a_0 + j\mathbf a_1 + l\mathbf a_2`. Note that `axis`
+        selects which of the :math:`\mathbf a_i` vectors that are translated and `reps` corresponds
+        to the :math:`i`, :math:`j` and :math:`l` variables. The `offset` moves the individual states
+        by said amount, i.e. :math:`i\to i-\mathrm{offset}`.
 
         Parameters
         ----------
@@ -325,6 +328,8 @@ class State(ParentContainer):
         normalize: bool, optional
            whether the states are normalized upon return, may be useful for
            eigenstates
+        offset: float, optional
+           the offset for the phase factors
 
         See Also
         --------
@@ -346,7 +351,7 @@ class State(ParentContainer):
         # with T being
         #   i * a_0 + j * a_1 + k * a_2
         # We can leave out the lattice vectors entirely
-        phase = exp(k[axis] * _a.aranged(reps) * 2j * _pi)
+        phase = exp(k[axis] * (_a.aranged(reps) - offset) * 2j * _pi)
 
         state *= phase.reshape(1, -1, 1)
         state.shape = (len(self), -1)
--- a/sisl/physics/state.py
+++ b/sisl/physics/state.py
@@ -346,7 +346,7 @@ class State(ParentContainer):
         where :math:`\mathbf T = i\mathbf a_0 + j\mathbf a_1 + l\mathbf a_2`. Note that `axis`
         selects which of the :math:`\mathbf a_i` vectors that are translated and `reps` corresponds
         to the :math:`i`, :math:`j` and :math:`l` variables. The `offset` moves the individual states
-        by said amount, i.e. :math:`i\to i-\mathrm{offset}`.
+        by said amount, i.e. :math:`i\to i+\mathrm{offset}`.
 
         Parameters
         ----------
@@ -380,7 +380,7 @@ class State(ParentContainer):
         # with T being
         #   i * a_0 + j * a_1 + k * a_2
         # We can leave out the lattice vectors entirely
-        phase = exp(2j*_pi * k[axis] * (_a.aranged(reps) - offset))
+        phase = exp(2j*_pi * k[axis] * (_a.aranged(reps) + offset))
 
         state *= phase.reshape(1, -1, 1)
         state.shape = (len(self), -1)
--- a/sisl/physics/state.py
+++ b/sisl/physics/state.py
@@ -305,6 +305,72 @@ class State(ParentContainer):
         sub.info = self.info
         return sub
 
+    def tile(self, reps, axis, normalize=False):
+        r"""Tile the state vectors for a new supercell
+
+        Tiling a state vector makes use of the Bloch factors for a state by utilizing
+
+        .. math::
+
+           \psi_{\mathbf k}(\mathbf r + \mathbf T) \propto e^{i\mathbf k\cdot \mathbf T}
+
+        where :math:`\mathbf T = i\mathbf a_0 + j\mathbf a_1 + k\mathbf a_2`.
+
+        Parameters
+        ----------
+        reps : int
+           number of repetitions along a specific lattice vector
+        axis : int
+           lattice to tile along
+        normalize: bool, optional
+           whether the states are normalized upon return, may be useful for
+           eigenstates
+
+        See Also
+        --------
+        Geometry.tile
+        """
+        # the parent gets tiled
+        parent = self.parent.tile(reps, axis)
+        # the k-point gets reduced
+        k = _a.asarrayd(self.info.get("k", [0]*3))
+
+        # now tile the state vectors
+        state = np.tile(self.state, (1, reps)).astype(np.complex128, copy=False)
+        # re-shape to apply phase-factors
+        state.shape = (len(self), reps, -1)
+
+        # Tiling stuff is trivial since we simply
+        # translate the bloch coefficients with:
+        #   exp(i k.T)
+        # with T being
+        #   i * a_0 + j * a_1 + k * a_2
+        # We can leave out the lattice vectors entirely
+        phase = exp(k[axis] * _a.aranged(reps) * 2j * _pi)
+
+        state *= phase.reshape(1, -1, 1)
+        state.shape = (len(self), -1)
+
+        # update new k; when we double the system, we halve the periodicity
+        # and hence we need to account for this
+        k[axis] = (k[axis] * reps % 1)
+        while k[axis] > 0.5:
+            k[axis] -= 1
+        while k[axis] <= -0.5:
+            k[axis] += 1
+
+        # this allows us to make the same usable for StateC classes
+        s = self.copy()
+        s.parent = parent
+        s.state = state
+        # update the k-point
+        s.info = dict(**self.info)
+        s.info.update({'k': k})
+
+        if normalize:
+            return s.normalize()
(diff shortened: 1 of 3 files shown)


tfrederiksen:
How does one use `tile` to construct a given state in the supercell?

I was thinking something like:
```python
es = H.eigenstate(k)
es_ext = np.concatenate([np.exp(2j * np.pi * k.dot(isc)) * es.state for _, isc in H.geom.sc], axis=1)
```
where one could then compute matrix elements (in this case band energies) in "real space" for the operator/matrix like
```python
ev1 = np.diag(es.state.conj().dot(H.tocsr().dot(es_ext.T)))
ev2 = es.inner(matrix=H.Hk(k))
assert np.allclose(ev1, ev2)
```

zerothi:
> How does one use `tile` to construct a given state in the supercell?
I think you can just do:
```python
es = H.eigenstate(...)
es_nsc = es
for axis, nsc in enumerate(H.geometry.sc.nsc):
   es_nsc = es_nsc.tile(nsc, axis)

# possibly one should re-arrange data in some other way to make them *diagonal*
es.inner(es_nsc.state.T, H.Hk(...)
```
and then you have the full eigenstate vectors in a geometry as large as the supercell, and the phase-factors will run from `[0 ; nsc-1]` instead of `[-nsc//2; nsc//2]`. But I think this is just a wrap-around and would be equivalent (at the top of my head, could be wrong)?

> I was thinking something like:
> 
> ```python
> es = H.eigenstate(k)
> es_ext = np.concatenate([np.exp(2j * np.pi * k.dot(isc)) * es.state for _, isc in H.geom.sc], axis=1)
> ```
> 
> where one could then compute matrix elements (in this case band energies) in "real space" for the operator/matrix like
> 
> ```python
> ev1 = np.diag(es.state.conj().dot(H.tocsr().dot(es_ext.T)))
> ev2 = es.inner(matrix=H.Hk(k))
> assert np.allclose(ev1, ev2)
> ```

Also, these *big* calculations will quickly take up lots of memory, for a system with 1000 orbitals and `nsc = [3, 3, 1]` you will find that doing this for a single `ev1` takes up `1000*1000 * 3 * 3` reals (68 MB). 

Wouldn't the above method suffice?

zerothi:
I have now amended the code to have an `offset` parameter, this will enable you to do `es.tile(3, 1, offset=1)` which then uses the phases `[-1,0,1]`.

tfrederiksen:
Nice! Could we have a property to an eigenstate object, like `es.state_sc`, that automatically calls `tile` with the correct offsets to have the state vectors in the supercell?

zerothi:
> Nice! Could we have a property to an eigenstate object, like `es.state_sc`, that automatically calls `tile` with the correct offsets to have the state vectors in the supercell?

Hmm. I am a little bit hesitant since you can't really use it as you imagine it, i.e. the problem of re-ordering. And if one has a hard time figuring out what it actually means, then probably not so useful.. ?

> Reordering seems also necessary before one can do the dot product `H.tocsr().dot(es.state_sc)`

Yeah, this makes it rather problematic. I have now added a `translate` method so that one can do:
```python
es_exts = map(lambda x: es.translate(x[1]), H.geometry.sc)
np.concatenate((es.inner(ext) for ext in es_exts), axis=1)
```
would this suffice?

It seems this is a bit more natural?

zerothi:
> I have now amended the code to have an `offset` parameter, this will enable you to do `es.tile(3, 1, offset=1)` which then uses the phases `[-1,0,1]`.

Actually, the offset should not be subtractive, I'll fix this, the above should be `es.tile(3, 1, offset=-1)`

tfrederiksen:
> ```python
> es_exts = map(lambda x: es.translate(x[1]), H.geometry.sc)
> np.concatenate((es.inner(ext) for ext in es_exts), axis=1)
> ```
> 
> would this suffice?
> 
> It seems this is a bit more natural?

I would really still like to see this as something the user can invoke directly from the object. I think `H.tocsr(-1).dot(es.state_sc)` is what we need in `ìnner` to handle the non-ortho case and  bra/ket at different k-points.

Couldn't you make it a property?
```python
    @property
    def state_sc(self):
        r""" Eigenstate(s) in the extended supercell """
        es_exts = map(lambda x: es.translate(x[1]), H.geometry.sc)
        return np.concatenate((es.inner(ext) for ext in es_exts), axis=1)
```

zerothi:
> I would really still like to see this as something the user can invoke directly from the object. I think `H.tocsr(-1).dot(es.state_sc)` is what we need in `ìnner` to handle the non-ortho case and bra/ket at different k-points.
> 
> Couldn't you make it a property?
> 
> ```python
>     @property
>     def state_sc(self):
>         r""" Eigenstate(s) in the extended supercell """
>         es_exts = map(lambda x: es.translate(x[1]), H.geometry.sc)
>         return np.concatenate((es.inner(ext) for ext in es_exts), axis=1)
> ```

I guess you mean:
```python
@property
def state_sc(self):
    return np.concatenate([self.translate(isc).state for _, isc in H.geometry.sc], axis=1)
```
?

However, it still isn't natural to use, I think. It requires bookkeeing to figure out where what belongs, and since you can directly use the above in an `es.inner(state_sc)` then its use is rather limited.

Also, perhaps another way to use the state would be:
```python
@property
def state_sc(self):
    return np.stack([self.translate(isc).state for _, isc in H.geometry.sc])
```
which would return `(n_s, len(es), es.shape[1])`, this would have the correct order by doing, `state_sc.reshape(-1, es.shape[1])`, I think

tfrederiksen:
Sorry, yes, I copied the wrong piece of code.

The book-keeping is already taken care of by adhering to the supercell ordering, no?

From my perspective, being able to evaluate `A.tocsr().dot(es.state_sc)` seems important (and, from the user perspective, should be as simple as that).

Regarding the wording, my logic was that as `state` provides the state vectors in the primitive cell, `state_sc` would denote the corresponding state vectors with Bloch phases and ordering corresponding to the supercell.

zerothi:
Ok, I have added what should work.

Before merging, one final question:

Currently the `ket` argument of `inner` is transposed internally, does this make sense? Wouldn't one expect one to pass the correct order?

sisl/physics/state.py
```diff
@@ -305,6 +305,38 @@ def sub(self, idx):
...
+        """
+        k = _a.asarrayd(self.info.get("k", [0]*3))
+        sc = self.parent.sc
+        if np.allclose(k, 0):
+            def translate(isc):
+                return self.state
+        else:
+            def translate(isc):
+                return self.state * exp(2j*_pi * k @ isc)
+        return np.concatenate([translate(isc) for _, isc in sc], axis=0)
```
tfrederiksen, line 338: Shouldn't it be `axis=1` here? Such that `self.state_sc.shape == (len(state), len(state) * state.parent.n_s)`?

The first index should naturally be the band index as for `self.state`. The number of bands dnoes not change by extending to the supercell. The second index the orbital index which grows.

zerothi: No, that would leave the ordering wrong. You would have to *increase* the number of bands to `nbands * n_s` to get this correct. Alternatively it should return `(n_s, nbands, ncoeffs)`?

zerothi: or `(nbands, n_s, ncoeffs)`

zerothi: actually, I am here more in favour of `(n_s, nbands, ncoeffs)`

zerothi: I mean you want to do `es.inner(es.state_sc)`, right, then you would have to increase the number of bands, and not number of coefficients, no?

tfrederiksen: I do not understand - why does the number of bands change? I thought we were just representing the *same* states in a larger region of basis orbitals.

zerothi: Agreed, we are representing the same states in a larger region. But you said you wanted to be able to do `es.inner(es.state_sc)`, the kets represents a band, so if you add more coefficients, it won't work, since the matrix dimensions does not fit.  
Your point is why I think this method was not so obvious to use in the first place.

It becomes unnatural since doing `axis=1` would mean the ordering is wrong, `(nbands, ncoeffs * n_s).T == (ncoeffs * n_s, nbands)` would not enable MM with a square `nbands` matrix, say `H.Hk(...)`.

tfrederiksen: My understanding is this:

We have `es.state[n, i]` where `n` is a band index and `i` an orbital in the primitive cell. And `H.tocsr()` has indices `[i, J]` with `i` an orbital index in the primitive cell and `J` one in the supercell.

To compute general matrix elements we need `<s1| A |s2> = s1.inner(matrix=A.tocsr, ket=s2.state_sc)` to compute general matrix elements between the Bloch states `s1, s2`.

What I have in mind is thus that `s2.state_sc` should have indices `[m, J]` such that the general inner product `<s1(n)| A |s2(m)>` takes the form `[n, i] . [i,J] . ([m, J].T)`.

zerothi: Yes ok, so what you want is to use a non-square matrix as the operator. However, I don't really follow what you then get, since `[i, J] . [m, J].T -> [i, m]` and becomes a square matrix again, and you loose the supercell information? No?

Note: generally you shouldn't use `tocsr` for this, rather use `H.Hk(format='sc:csr')`, this will ensure to be working for NC and SOC as well.

zerothi: It seems to me that this is a greater problem than what this PR proposes to do, I think I should delete the `state_sc`, and then another PR can take care of that, then the discussion can continue there, and more ftitingly so I think. :)

tfrederiksen: What I want is basically the matrix elements between primitive cell band indices `n, m` but where the operator `A` is given in "real space". I do not want any phases applied to `A`.

As far as I see this is also the correct approach to do the inner product between states at different k-points with non-ortho basis, i.e.,
```
<k1(n) | k2(m)> = k1.state[n, i].S[i, J].((k2.state_sc[m, J]).T)
```

tfrederiksen: > becomes a square matrix again, and you loose the supercell information? No?

That's the point, I do not want to retain information about the supercell.

zerothi: lets continue discussion in #356

zerothi:
This is now merged (the last commit here isn't and will be followed by #356)

zerothi:
For completeness sake, the code left out of this PR is:
```python
   @property
    def state_sc(self):
        r""" Calculate the state coefficients in the auxiliary supercell
        The format of this state vector is usable for `inner` as the ``ket`` argument.
        The order of the state vectors is equivalent to the auxiliary supercell order.
        See Also
        --------
        inner : these state vectors may be used in the inner product
        tile : equivalent form along a single given axis
        translate : a single translation of the state vectors
        Examples
        --------
        The returned state vectors can be formulated like this:
        >>> state = State(...)
        >>> state_sc = np.stack([state.translate(isc)
        ...                      for _, isc in state.parent.sc])
        >>> assert state_sc.shape == (len(state) * state.parent.n_s, state.shape[1])
        """
        k = _a.asarrayd(self.info.get("k", [0]*3))
        sc = self.parent.sc
        if np.allclose(k, 0):
            dtype = self.state.dtype
            def translate(isc):
                return self.state
        else:
            dtype = np.complex128
            def translate(isc):
                return self.state * exp(2j*_pi * k @ isc)
        state = np.empty((sc.n_s, ) + self.shape, dtype=dtype)
        for i, (_, isc) in enumerate(sc):
            state[i] = translate(isc)
        return state.reshape(-1, self.shape[1])
```

PR #361, opened 2021-07-26
init commit of siestabarriers

arsalan-akhtar:
Hi @zerothi  

just start working on merging NEB stuff in sisl toolbox i just open the pull for u to just check if u are agreed with this way  ...
cheers


arsalan-akhtar:
by the way, i already have talked to Miguel regarding bsse for neb we have an Idea how to do and check it now ... i"ll work on that later ...

setup.py
```diff
@@ -472,12 +472,12 @@ def readme():
 # in the correct place unless we use 'package_dir' and this trick.
 # 1. Here we list files as they should appear in packages for end-users
 # 2. In 'package_dir' we defer the package name to the local file path
-packages += map(lambda x: f"sisl_toolbox.{x}", find_packages("toolbox"))
+packages += map(lambda x: f"toolbox.{x}", find_packages("toolbox"))
```
zerothi, line 475: Why are you changing this?

setup.py
```diff
@@ -472,12 +472,12 @@ def readme():
...
 
 # Also ensure we have all "pxd" files
 package_data = {p: ["*.pxd"] for p in packages}
 # Add toolbox data
-package_data["sisl_toolbox.siesta.minimizer"] = ["basis.yaml", "pseudo.yaml"]
+package_data["toolbox.siesta.minimizer"] = ["basis.yaml", "pseudo.yaml"]
```
zerothi, line 480: Don't change here as well.

toolbox/siestabarriers/SiestaBarriers.py
```diff
@@ -0,0 +1,245 @@
+# -*- coding: utf-8 -*-
+########################################################################################
+# Copyright (c), The SiestaBarriers authors. All rights reserved.                       #
+#                                                                                      #
+# SiestaBarriers is hosted on GitHub at https://github.com/.................. #
+# For further information on the license, see the LICENSE.txt file                     #
+########################################################################################
+
+from __future__ import absolute_import
```
zerothi, line 9: In python 3.6 and beyond this is not needed.

toolbox/siestabarriers/SiestaBarriers.py
```diff
@@ -0,0 +1,245 @@
...
+__author__ = "Arsalan Akhatar"
+__copyright__ = "Copyright 2021, SIESTA Group"
+__version__ = "0.1"
+__maintainer__ = "Arsalan Akhtar"
+__email__ = "arsalan_akhtar@outlook.com," #+ \" miguel.pruneda@icn2.cat "
+__status__ = "Development"
+__date__ = "Janurary 30, 2021"
+
+
+class SiestaBarriers(SiestaBarriersBase):
```
zerothi, line 35: There is no documentation? :)

toolbox/siestabarriers/SiestaBarriers.py
```diff
@@ -0,0 +1,245 @@
...
+                 host_fdf_name = None,
+                 host_structure = None ,
+                 initial_relaxed_path = None,
+                 initial_relaxed_fdf_name = None,
+                 initial_structure = None,
+                 final_relaxed_path = None,
+                 final_relaxed_fdf_name = None,
+                 final_structure = None,
+                 image_direction = None ,
+                 trace_atom_initial_position = None,
```
zerothi, line 50: In particular all these keywords don't make sense to the uninitiated.

toolbox/siestabarriers/SiestaBarriers.py
```diff
@@ -0,0 +1,245 @@
...
+                         neb_results_path,
+                         relax_engine,
+                         interpolation_method,
+                         tolerance_radius,
+
+                         )
+        
+
+        self.host_path = host_path
+        print(self.host_path)
```
zerothi, line 101: I guess you'll remove the `print` later?

toolbox/siestabarriers/SiestaBarriers.py
```diff
@@ -0,0 +1,245 @@
...
+        #super(SiestaBarriersBase,self).__init__()
+        #super().__init__()
+        
+        
+
+        print ("---------------------------")
+        print ("Wellcome To SiestaBarriers")
+        print ("      Version : {}".format(__version__))
+        print ("---------------------------")
+
```
zerothi, line 134: Print should be put in a log so users can easily control output to terminal.

toolbox/siestabarriers/SiestaBarriers.py
```diff
@@ -0,0 +1,245 @@
...
+                                                    final_relaxed_fdf_name = self.final_relaxed_fdf_name,
+                                                    final_structure = self.final_structure,
+                                                    flos_path = self.flos_path,
+                                                    interpolation_method = self.interpolation_method,
+                                                    ghost = self.ghost,
+                                                    trace_atom_initial_position = self.trace_atom_initial_position,
+                                                    trace_atom_final_position = self.trace_atom_final_position,
+                                                    number_of_images = self.number_of_images
+                                                    )
+        if self.neb_scheme == "exchange":
```
zerothi, line 155: use `elif` instead.

toolbox/siestabarriers/SiestaBarriers.py
```diff
@@ -0,0 +1,245 @@
...
+        
+
+        print ("---------------------------")
+        print ("Wellcome To SiestaBarriers")
+        print ("      Version : {}".format(__version__))
+        print ("---------------------------")
+        
+        self.setup() 
+
+        if self.neb_scheme == "vacancy-exchange":
```
zerothi, line 137: It seems you are "hosting" class differences in a main class, which seems rather strange. Wouldn't it me more natural to have the class describe the intent with proper inheritance?

toolbox/siestabarriers/SiestaBarriersBase.py
```diff
@@ -0,0 +1,318 @@
+# -*- coding: utf-8 -*-
+########################################################################################
+# Copyright (c), The SiestaBarriers authors. All rights reserved.                       #
+#                                                                                      #
+# SiestaBarriers is hosted on GitHub at https://github.com/.................. #
+# For further information on the license, see the LICENSE.txt file                     #
+########################################################################################
+
+from __future__ import absolute_import
```
zerothi, line 9: same here, not needed in py3

toolbox/siestabarriers/SiestaBarriersBase.py
```diff
@@ -0,0 +1,318 @@
...
+
+#from SiestaBarriers.Utils.utils_siesta import print_siesta_fdf,read_siesta_fdf
+
+from .Utils.utils_siesta import print_siesta_fdf,read_siesta_fdf
+#from Utils.utils_general import AtomIndex
+#from SiestaBarriers.Utils.utils_siesta import print_siesta_fdf
+from .Utils.utils_siesta import print_siesta_fdf
+import sys
+import inspect
+pacakge_dir= inspect.getabsfile(print_siesta_fdf).split("utils_siesta.py")[0]+"flos/"
```
zerothi, line 19: do you use `pacakge_dir` anywhere? Otherwise remove what you don't need.

toolbox/siestabarriers/SiestaBarriersBase.py
```diff
@@ -0,0 +1,318 @@
...
+#pacakge_dir= inspect.getabsfile(print_siesta_fdf).split("neb_base.py")[0]+"flos/"
+
+
+__author__ = "Arsalan Akhatar"
+__copyright__ = "Copyright 2021, SIESTA Group"
+__version__ = "0.1"
+__maintainer__ = "Arsalan Akhtar"
+__email__ = "arsalan_akhtar@outlook.com," #+ \" miguel.pruneda@icn2.cat "
+__status__ = "Development"
+__date__ = "Janurary 30, 2021"
```
zerothi, line 29: Date is probably too much. Is it really needed?

toolbox/siestabarriers/SiestaBarriersBase.py
```diff
@@ -0,0 +1,318 @@
...
+                print("NOT IMPLEMENTED")
+                #sys.exit()
+            else:
+                print("NEB image scheme is: {}".format(self.neb_scheme))
+                #return self.neb_scheme
+        if self.ghost == True:
+            print(" NOTE: The Ghost Support Basis is True You Have To Provide The Basis Set Of Ghost Specie/s via PAO.Basisblock! ")
+        else:
+            print(" No Ghost Support Basis")
+    def is_none_scheme(self):
```
zerothi, line 147: All these would be superfluous if you used a class scheme.

toolbox/siestabarriers/SiestaBarriersIO.py
```diff
@@ -0,0 +1,274 @@
...
+        self.neb_results_path = neb_results_path
+
+    def write_all_images_sisl(self, fname = 'images',folder_name = 'all' , out_format = 'xyz'):
+        """
+        """
+        if self.relaxed == True:
+            print ("Generating All Images After Relaxation")
+            folder_name = folder_name + '_relaxed'
+            if os.path.isdir(folder_name):
+                print (" The All Image Folder Relaxed is there Already PASS")
```
zerothi, line 80: Again, print-statements should really be avoided. `logging` is better

toolbox/siestabarriers/SiestaBarriersIO.py
```diff
@@ -0,0 +1,274 @@
...
+                print (" Check The Folder: '{}' ".format(folder_name))
+                pass
+            else:
+                os.mkdir(folder_name)
```
zerothi, line 84: If you can, please use `pathlib` instead, some operations exists there which are easier to use. Some don't, in which case you do need to use `os`.

toolbox/siestabarriers/Utils/flos/neb_analysis.lua
```diff
@@ -0,0 +1,209 @@
...
+end
+
+function siesta_update_Analysis(image_number)
+
+   if not siesta.IONode then
+      -- only allow the IOnode to perform stuff...
+      return
+   end
+   -- Move about files so that we re-use old DM files
+   local PDOS = label .. ".PDOS"
```
zerothi, line 114: I really don't think this should be done in `flos` / `lua`. This *must* be the user's responsibility. Consider when new files are added (which are not tracked here), then one would assume this script to do the correct thing, but it doesn't.

zerothi (changes requested):
I haven't gone through the rest of the files, but there are quite some things that needs adaptation. The structure of the code seems to *want* to be class based, but in fact it is just a container with `if` statements. Now, either choice could be ok, but classes would allow easier extensibility.

Generally I would discourage CamelCase module names. So I would advice against this.

Is there a reason for `siestabarriers` rather than `siesta.neb`? or `siesta.barrier`? From reading the module name `neb` should be sufficiently descriptive. Or does this do anything beyond NEB?

setup.py
arsalan-akhtar: It couldn't find toolbox folder i think probably u changed the name from `sisl_toolbox` ----> `toolbox`...

setup.py
arsalan-akhtar: same reason... couldn't find the folders...

toolbox/siestabarriers/SiestaBarriers.py
arsalan-akhtar: will do.... ;)

toolbox/siestabarriers/SiestaBarriers.py
arsalan-akhtar: - the image_direction will be just for Exchange Paths so i put it None...
- the relaxed one is bcz:
 first user gives the pristine  case structure and initial and final position of atoms so it will generate an unrelaxed  path for NEB from initial to final ... then there is method where (`prepare_endpoint_relax`) generate initial and final structure to be relaxed) after relaxing the endpoints user will provide the results folders of initial and final configuration and regenerate the  Relaxed initial Path for NEB and neb folder which contains the inputs.fdf and images* with *.DM.initial and *.DM.final to save the time to calculating the fixed endpoints(initial and final config...) scf ...
 so at first user just need to provide the pristine fdf and intial and final position of specie/s ...

toolbox/siestabarriers/SiestaBarriersBase.py
arsalan-akhtar: jajajaj it was fast copy paste from the source ...

setup.py
zerothi: Please remove it from this pr and make a new issue. I don't have problems with this, so I think you either do something wrong or misinterpret what this does?

setup.py
zerothi: Same as above please remove from this pr

toolbox/siestabarriers/Utils/flos/neb_analysis.lua
arsalan-akhtar: this comment i don't understand it ?!

setup.py
arsalan-akhtar: humm i don't know may be but it doesn't find the toolbox then to install it unless there is another way to install sisl (so when u just pip install ... it should install the toolbox ... )

zerothi: And I am pretty sure it does. I have checked with conda. But I'll check some more, in any case it does not belong in this pr.

zerothi: Also, it works seemleasly for me

toolbox/siestabarriers/Utils/flos/neb_analysis.lua
zerothi: All the copying of files, it shouldn't be done in this major lua script. If anything it should be a separate thing in flos.

arsalan-akhtar: ok ill fix this things in flos pr ...

setup.py
arsalan-akhtar: okay found my mistake u r correct things are fine probably the doc have some type in doc (http://zerothi.github.io/sisl/docs/latest/toolbox/index.html) is:
`import sisl_toolbox.siesta.atom`
while it should be:
`import toolbox.siesta.atom`

arsalan-akhtar:
> I haven't gone through the rest of the files, but there are quite some things that needs adaptation. The structure of the code seems to _want_ to be class based, but in fact it is just a container with `if` statements. Now, either choice could be ok, but classes would allow easier extensibility.
> 
> Generally I would discourage CamelCase module names. So I would advice against this.
> 
> Is there a reason for `siestabarriers` rather than `siesta.neb`? or `siesta.barrier`? From reading the module name `neb` should be sufficiently descriptive. Or does this do anything beyond NEB?

now i'have something like this :

**Example case**

`from toolbox.siesta.barriers import VacancyExchange `
`import sisl`
`host= sisl.get_sile('./MgO-sheet.fdf').read_geometry()`
`A=VacancyExchange(host_structure=host,`
`                 ghost=True,`
`                 interpolation_method='idpp',`
`                 trace_atom_initial_position=2,`
`                 trace_atom_final_position=18,`
`                 number_of_images=17)`
`A.set_initial_relaxed_path("/home/aakhtar/Calculations/2020/siesta/SiestaBarriers/VacancyExchange/new-0/results/")`
`A.set_final_relaxed_path("/home/aakhtar/Calculations/2020/siesta/SiestaBarriers/VacancyExchange/new-8/results/")`
`A.set_initial_relaxed_fdf_name("input.fdf")`
`A.set_final_relaxed_fdf_name("input.fdf")`
`A.set_relaxed(False)`
`A.Generate_Vacancy_Exchange_Images()`
`A.IO.prepare_NEB()`

**Put code here**

i removed the intermediate class (SiestaBarrier) ... now for each barrier type will have its one class to initiate (VacancyExchange,Exchange,Manual,...)
 like this:
class VacancyExchange(SiestaBarriersBase):
    """
    Init Object for VacancyExchange
    INPUTs:
           host_structure = Sisl Stucture Object
           number_of_images = # of images for NEB
           trace_atom_initial_position = initial Atom Position/Index
           trace_atom_final_position = final Atom Position/Index
           interpolation_method = idpp
    """

    def __init__(self,
                 host_structure ,
                 number_of_images  ,
                 trace_atom_initial_position ,
                 trace_atom_final_position ,
                 interpolation_method = 'iddp',
                 ghost = True ,
                 ):

        super().__init__(
                 relaxed = False,
                 host_structure = host_structure ,
                 number_of_images = number_of_images,
                 initial_relaxed_path = None,
                 initial_relaxed_fdf_name = None ,
                 final_relaxed_path = None ,
                 final_relaxed_fdf_name = None,
                 trace_atom_initial_position = trace_atom_initial_position ,
                 trace_atom_final_position = trace_atom_final_position ,
                 atol = 1e-2,
                 rtol = 1e-2,
                )
        self.host_structure = host_structure
        self.number_of_images = number_of_images
        self.interpolation_method = interpolation_method
        self.ghost = ghost

setup.py
zerothi: No, no typo it works with that module name

arsalan-akhtar:
hey @zerothi ... 
u removed the tag property from sisl.Atom object?!!!
is there anyway to set tag for specie to have a different name like for O & O_surface :
`%block ChemicalSpeciesLabel`
` 1 12 Mg`
` 2 8 O`
` 3 8  O_surface`
`4  8 O_kicked`
`%endblock ChemicalSpeciesLabel`

setup.py
zerothi: This is the way python ensures renaming folders gets translated to the correct module name.

toolbox/siesta/barriers/BarriersBase.py
```diff
@@ -52,7 +51,7 @@ class SiestaBarriersBase():
     ghost_info                   :
     """
     def __init__(self,
-                 host_path = None,
+                 host_path = "./",
```
zerothi, line 54: If you use `pathlib.Path` it would be best to simply do: `host_path=''` since `pathlib.Path('')` is the current directory (by default).

toolbox/siesta/barriers/BarriersBase.py
```diff
@@ -117,14 +118,27 @@ def __init__(self,
         self.relax_engine = relax_engine
 
         self.tolerance_radius = tolerance_radius
+        self.atol = atol
+        self.rtol = rtol
         
+        self.wellcome()
+        self.setup()
+
+    def wellcome(self):
```
zerothi, line 127: typo

toolbox/siesta/barriers/BarriersBase.py
```diff
@@ -136,14 +150,15 @@ def setup(self):
         if self.neb_scheme is not None:
             if self.neb_scheme not in neb_schemes_available:
                 print("NOT IMPLEMENTED")
-                #sys.exit()
+                sys.exit()
             else:
                 print("NEB image scheme is: {}".format(self.neb_scheme))
                 #return self.neb_scheme
         if self.ghost == True:
```
zerothi, line 157: Again, proper documentation might clear lots of my confusion ;)
Is `ghost` really needed? Shouldn't it me trivially handled by sisl?

toolbox/siesta/barriers/BarriersIO.py
```diff
@@ -70,7 +71,7 @@ def __init__(self,
         self.final_relaxed_fdf_name = final_relaxed_fdf_name
         self.neb_results_path = neb_results_path
 
-    def write_all_images_sisl(self, fname = 'images',folder_name = 'all' , out_format = 'xyz'):
+    def Write_All_Images(self, fname = 'images',folder_name = 'all' , out_format = 'xyz'):
```
zerothi, line 74: I don't particularly like CamelCase. However, it is your choice ;)
It just goes against all sisl commands. So if you want it, you should really use CamelCase everywhere.

toolbox/siesta/barriers/Utils/utils.py
```diff
@@ -41,28 +63,28 @@ def AtomIndex(Positions,AtomPosition,rtol,atol):
...
-        else:
-            print("Something is Wrong in InitialAtomPosition or FinalAtomPosition")
-    elif type(Position) == type(1):
-        print ("It's index")
-        out = 'index'
-    else:
-        print("Something is Wrong in InitialAtomPosition")
-
-    return out
+#def is_frac_or_cart_or_index(Position):
```
zerothi, line 66: Delete this if not used, there are other ways to test for arrays/indices

toolbox/siesta/barriers/interstitial.py
```diff
@@ -0,0 +1,206 @@
...
+             if self.initial_relaxed_fdf_name == None or self.final_relaxed_fdf_name == None :
+                 sys.exit("intial/final relaxed fdf not provided")
+
+             #self.host_structure = read_siesta_fdf(self.host_path,self.host_fdf_name)
+             self.initial_structure = read_siesta_XV(self.initial_relaxed_path,self.initial_relaxed_fdf_name)
+             self.final_structure = read_siesta_XV(self.final_relaxed_path,self.final_relaxed_fdf_name)
+             
+             self.check_ghost()
+             self.test = pre_prepare_ase_after_relax(self.initial_structure['XV'],self.final_structure['XV'],self.ghost)
+             initial = sisl.Geometry.toASE(self.test['initial'])
```
zerothi, line 95: sisl now prefers `Geometry.to.ase()` instead, you should have experienced tons of warnings ;)

zerothi (changes requested):
You are still not using class-based objects.
You are hiding everything in the `...Base` class which then still has all the stuff.

If you could document all the flags in the Base class, then I could help you out with the `ManualBarrier` class so you can see how I think you'll get a cleaner and easier to maintain code?

Also, perhaps `toolbox/siesta/barrier` (without the plural `s`)? But if you really want it, retain it ;)

toolbox/siesta/barriers/BarriersBase.py
arsalan-akhtar: Didn't expect that u went through pull so fast i just push it to be safe and work on it later. Jjaajajjaja but okay ill document the the base class tmrw then if u show me for the manual barrier then i could follow... Sorry 😅

Yes we need the ghost need ...cz in vacancy we put two ghost on each point ... So in both initial and final there will be a ghost on top of specie ... basically We first generate the path without any ghost later we put back the ghost in initial & final position...
 
Also i found out sisl have now Atomghost .. i have to update the coroseponding parts ..

toolbox/siesta/barriers/BarriersIO.py
arsalan-akhtar: So i think i don't understand this camel way vs not 🐪 ...

arsalan-akhtar:
@zerothi haven't fixed  ur other comments yet ... will do after u show me the BarrierManual example .... tanx...

zerothi:
I have made a PR against your branch, please have a look there, I am quite honest, so don't be discouraged.
Perhaps you may use a co-worker and ask them what would be natural in the interface for the NEB classes. As it is, it is too complicated, I think. :)

Let me know if you have questions!

zerothi:
This has gone stale for some time, I would be happy to pick it up later, but for now we'll close this PR.

Issue #360, opened 2021-07-18
ComplexWarning on spin_moment()


pfebrer:
When I use `spin_moment`, I get a numpy `ComplexWarning`. 

This is annoying but it is not super bad. However, when I run a parallel calculation on the BZ, I get a `ComplexWarning` for each k point, since numpy can't realize it has already issued this warning. Then it gets really bad :sweat_smile: 

Substituting https://github.com/zerothi/sisl/blob/bba363b1cedc7f3f544597a01e2eabfb0c6a9a2f/sisl/physics/electron.py#L390 by

```python
s[i, 2] = (D[0, 0] - D[1, 1]).real
```

removes the warning, but I don't know if it's the correct fix.

zerothi:
Thanks, correct fix, could you please make a pr?

PR #363, opened 2021-08-03
maint: avoid ComplexWarning on spin moment calculation

pfebrer:
Fixes https://github.com/zerothi/sisl/issues/360


zerothi:
Thanks!

PR #367, opened 2021-08-18
Parallel calculation of PDOS and (fat)bands

pfebrer:
Now if `pathos` is available, PDOS and (fat)bands calculated from a hamiltonian are computed in parallel over k-points using the functionality of the `BrillouinZone` applies.

If someone doesn't want parallelization they can control it through the `SISL_NPROCS` env variable by setting it to `1`, so I didn't add any settings for it.

### DIFF
--- a/sisl/viz/plots/bands.py
+++ b/sisl/viz/plots/bands.py
@@ -287,7 +287,7 @@ class BandsPlot(Plot):
 
     @property
     def spin_moments(self):
-        return self.bands_data["spin_moments"].sel(spin=0)
+        return self.bands_data["spin_moments"]
 
     def _after_init(self):
         self.spin = sisl.Spin("")
@@ -419,6 +419,11 @@ class BandsPlot(Plot):
             self.bands_data = xr.Dataset({"E": self.bands_data})
             self.bands_data.attrs = attrs
 
+        # If the calculation is not spin polarized it makes no sense to
+        # retain a spin index
+        if "spin" in self.bands_data and not self.spin.is_polarized:
+            self.bands_data = self.bands_data.sel(spin=self.bands_data.spin[0], drop=True)
+
         # Inform the spin input of what spin class are we handling
         self.get_param("spin").update_options(self.spin)
         self.get_param("custom_gaps").get_param("spin").update_options(self.spin)
@@ -471,9 +476,11 @@ class BandsPlot(Plot):
         self.spin_texture = False
         if spin is not None and len(spin) > 0:
             if isinstance(spin[0], int):
-                filtered_bands = filtered_bands.sel(spin=spin)
+                # Only use the spin setting if there is a spin index
+                if "spin" in filtered_bands.coords:
+                    filtered_bands = filtered_bands.sel(spin=spin)
             elif isinstance(spin[0], str):
-                if not hasattr(self, "spin_moments"):
+                if "spin_moments" not in self.bands_data:
                     raise ValueError(f"You requested spin texture ({spin[0]}), but spin moments have not been calculated. The spin class is {self.spin.kind}")
                 self.spin_texture = True
 
@@ -563,7 +570,9 @@ class BandsPlot(Plot):
             if requested_spin is None:
                 requested_spin = [0, 1]
 
-            for spin in self.bands.spin:
+            avail_spins = self.bands_data.get("spin", [0])
+
+            for spin in avail_spins:
                 if spin in requested_spin:
                     from_k = custom_gap["from"]
                     to_k = custom_gap["to"]
@@ -642,7 +651,8 @@ class BandsPlot(Plot):
             ks[i] = self._sanitize_k(val)
 
         VB, CB = self.gap_info["bands"]
-        Es = [self.bands.sel(k=k, band=band, spin=gap_spin, method="nearest") for k, band in zip(ks, (VB, CB))]
+        spin_bands = self.bands.sel(spin=gap_spin) if "spin" in self.bands.coords else self.bands
+        Es = [spin_bands.sel(k=k, band=band, method="nearest") for k, band in zip(ks, (VB, CB))]
         # Get the real values of ks that have been obtained
         # because we might not have exactly the ks requested
         ks = [np.ravel(E.k)[0] for E in Es]
@@ -765,7 +775,11 @@ class BandsPlot(Plot):
         from sisl.unit.base import units
 
         # Get the band that we want to fit
-        band_vals = self.bands.sel(band=band, spin=band_spin)
+        bands = self.bands
+        if "spin" in bands.coords:
+            band_vals = bands.sel(band=band, spin=band_spin)
+        else:
+            band_vals = bands.sel(band=band)
 
         # Sanitize k to a float
         k = self._sanitize_k(k)
@@ -784,7 +798,7 @@ class BandsPlot(Plot):
 
         # Grab the slice of the band that we are going to fit
         sel_band = band_vals[sel_slice] * units("eV", "Hartree")
-        sel_k = self.bands.k[sel_slice] - k
+        sel_k = bands.k[sel_slice] - k
 
         # Fit the band to a second order polynomial
         polyfit = np.polynomial.Polynomial.fit(sel_k, sel_band, 2)
--- a/sisl/viz/plots/bands.py
+++ b/sisl/viz/plots/bands.py
@@ -6,18 +6,25 @@ from functools import partial
 import itertools
 
 import numpy as np
+import xarray as xr
 
 import sisl
+from sisl.messages import warn
 from ..plot import Plot, entry_point
-from ..plotutils import call_method_if_present, find_files
+from ..plotutils import find_files
 from ..input_fields import (
-    TextInput, SwitchInput, ColorPicker, DropdownInput,
-    IntegerInput, FloatInput, RangeInput, RangeSlider,
-    QueriesInput, ProgramaticInput, FunctionInput, SileInput,
-    PlotableInput, SpinSelect, AiidaNodeInput, BandStructureInput
+    TextInput, SwitchInput, ColorPicker,
+    FloatInput, RangeSlider,
+    QueriesInput, FunctionInput, SileInput,
+    SpinSelect, AiidaNodeInput, BandStructureInput
 )
 from ..input_fields.range import ErangeInput
 
+try:
+    import pathos
+    _do_parallel_calc = True
+except:
+    _do_parallel_calc = False
 
 class BandsPlot(Plot):
     """
@@ -123,14 +130,6 @@ class BandsPlot(Plot):
         AiidaNodeInput(key="aiida_bands", name="Aiida BandsData node",
             default=None,
             help="""An aiida BandsData node."""
-        ),
-
-        FunctionInput(key="eigenstate_map", name="Eigenstate map function",
-            default=None,
-            positional=["eigenstate", "plot"],
-            returns=[],
-            help="""This function receives the eigenstate object for each k value when the bands are being extracted from a hamiltonian.
-            You can do whatever you want with it, the point of this function is to avoid running the diagonalization process twice."""
         ),
 
         FunctionInput(key="add_band_data", name="Add band data function",
@@ -281,6 +280,14 @@ class BandsPlot(Plot):
             return [childPlot.get_setting("bands_file").name for childPlot in self.child_plots]
 
         return cls.animated("bands_file", bands_files, frame_names = _get_frame_names, wdir = wdir, **kwargs)
+    
+    @property
+    def bands(self):
+        return self.bands_data["E"]
+
+    @property
+    def spin_moments(self):
+        return self.bands_data["spin_moments"].sel(spin=0)
 
     def _after_init(self):
         self.spin = sisl.Spin("")
@@ -308,7 +315,7 @@ class BandsPlot(Plot):
             tick_info["ticklabels"].append(label)
 
         # Construct the dataarray
-        self.bands = xr.DataArray(
+        self.bands_data = xr.DataArray(
             bands,
             coords={
                 "spin": np.arange(0, bands.shape[0]),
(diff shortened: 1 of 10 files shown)


zerothi:
Please remove commits changing `files` submodules needs a bit more care...

Also, it should already be there in the latest files commit?

sisl/viz/plots/bands.py
```diff
@@ -3,21 +3,24 @@
...
-    QueriesInput, ProgramaticInput, FunctionInput, SileInput,
-    PlotableInput, SpinSelect, AiidaNodeInput, BandStructureInput
+    TextInput, SwitchInput, ColorPicker,
+    FloatInput, RangeSlider,
+    QueriesInput, FunctionInput, SileInput,
+    SpinSelect, AiidaNodeInput, BandStructureInput
 )
 from ..input_fields.range import ErangeInput
 
+_do_parallel_calc = importlib.util.find_spec("pathos") is not None
```
zerothi, line 23: why not just `try ... except`? using an entire module for this seems not needed?

sisl/viz/plots/bands.py
```diff
@@ -281,6 +284,14 @@ def _get_frame_names(self):
...
 
         return cls.animated("bands_file", bands_files, frame_names = _get_frame_names, wdir = wdir, **kwargs)
+    
+    @property
+    def bands(self):
+        return self.bands_data["E"]
+
+    @property
+    def spin_moments(self):
+        return self.bands_data["spin_moments"].sel(spin=0)
```
zerothi, line 294: This just seems weird? spin-moments but selecting spin=0?

sisl/viz/plots/bands.py
```diff
@@ -347,17 +356,15 @@ def _read_from_H(self, band_structure, eigenstate_map):
 
         # Also, in this wrapper we will get the spin moments in case it is a non_colinear
         # or spin-orbit calculation
-        if self.spin.is_noncolinear or self.spin.is_spinorbit:
-            self.spin_moments = []
-        elif hasattr(self, "spin_moments"):
-            del self.spin_moments
+        calc_spin_moments = not self.spin.is_diagonal
```
zerothi, line 359: couldn't you just append to `extra_vars` to clean out this?

sisl/viz/plots/bands.py
```diff
@@ -366,38 +373,39 @@ def bands_wrapper(eigenstate, spin_index):
...
+            if not self.spin.is_diagonal:
                 spin_kwarg = {}
-
-            spin_bands = band_structure.apply.dataarray.eigenstate(
-                wrap=partial(bands_wrapper, spin_index=spin_index),
-                **spin_kwarg,
-                coords=('band',),
-            )
+            
+            with band_structure.apply(pool=_do_parallel_calc, unzip=True) as parallel:
```
zerothi, line 394: you should probably be careful about this when the blas/lapack library is threaded. Then you'll saturate the machine quite fast?

sisl/viz/plots/fatbands.py
```diff
@@ -186,6 +186,10 @@ class FatbandsPlot(BandsPlot):
 
     )
 
+    @property
+    def weights(self):
+        return self.bands_data["weight"]
```
zerothi, line 191: perhaps it should also be named `weights` in the data? for consistency?

sisl/viz/plots/fatbands.py
```diff
@@ -257,29 +261,26 @@ def _read_from_H(self):
...
 
-            plot.weights[spin_index].append(weights)
+            return weights.real
 
         # We make bands plot read the bands, which will also populate the weights
         # thanks to the above step
         bands_read = False; err = None
         try:
-            super()._read_from_H(eigenstate_map=_weights_from_eigenstate)
+            super()._read_from_H(extra_vars=[{"coords": ("band", "orb"), "name": "weight", "getter": _weights_from_eigenstate}])
```
zerothi, line 283: here, name?

sisl/viz/plots/pdos.py
```diff
@@ -2,20 +2,22 @@
...
-    TextInput, SileInput, SwitchInput, ColorPicker, DropdownInput, CreatableDropdown,
-    IntegerInput, FloatInput, RangeInput, RangeSlider, OrbitalQueries,
-    ProgramaticInput, Array1DInput, ListInput, GeometryInput
+    TextInput, SileInput, SwitchInput, ColorPicker, DropdownInput,
+    IntegerInput, FloatInput, OrbitalQueries,
+    Array1DInput, GeometryInput
 )
 from ..input_fields.range import ErangeInput
 
+_do_parallel_calc = importlib.util.find_spec("pathos") is not None
```
zerothi, line 19: again here, `try ... except`

zerothi (changes requested):

sisl/viz/plots/bands.py
zerothi: perhaps the `try except` could carry a warning for threaded blas/lapack such that users are aware of potential risks? This could even be checked against OMP_NUM_THREADS == 1 (in which case no warning is required).

sisl/viz/plots/fatbands.py
pfebrer: I'm not sure, because for me the logic way to call the variable is `weight`. It feels strange to give a variable a plural name, doesn't it?

In fact I was thinking quite the opposite: that the `spin_moments` variable should be renamed to `spin_moment`.

sisl/viz/plots/bands.py
pfebrer: Is there some way of knowing whether the library is threaded?

sisl/viz/plots/fatbands.py
zerothi: I don't really care ;) but you are using `weights` and `weight` here?
I tend to agree that plural is not aiding any understanding and hence I also try to avoid plural s

sisl/viz/plots/bands.py
zerothi: no, not really. But I think standard OS' are using a threaded library. In particular for conda installations this is true (MKL by default).

sisl/viz/plots/bands.py
pfebrer: Yes!

sisl/viz/plots/bands.py
pfebrer: Maybe `self.bands_data.sel(spin=0)["spin_moments"]` is more understandable?

Basically the spin coordinate is always there so that I don't have to treat each case differently, even though in the non-colinear case it doesn't make much sense.

Here is how the dataset looks for a non-colinear fatbands plot:
```
<xarray.Dataset>
Dimensions:       (axis: 3, band: 56, k: 301, orb: 28, spin: 1)
Coordinates:
  * orb           (orb) int32 0 1 2 3 4 5 6 7 8 9 ... 19 20 21 22 23 24 25 26 27
  * band          (band) int32 0 1 2 3 4 5 6 7 8 ... 47 48 49 50 51 52 53 54 55
  * axis          (axis) <U1 'x' 'y' 'z'
  * k             (k) float64 0.0 0.007169 0.01434 0.02151 ... 2.14 2.147 2.155
  * spin          (spin) int64 0
Data variables:
    E             (spin, k, band) float64 -24.83 -24.83 -24.81 ... 19.67 19.67
    spin_moments  (spin, k, band, axis) float64 -0.2438 0.1371 ... 0.9072 0.1044
    weight        (spin, k, band, orb) float64 3.708e-05 0.0001171 ... 0.217
Attributes:
    ticks:       [0.         0.78860842 1.69921498 2.15451827]
    ticklabels:  ['M', '$\\Gamma$', 'K', 'M']
    bz:          BandStructure{nk: 301,\n SuperCell{nsc: [5 5 1],\n  A=[4.600...
    parent:      Hamiltonian{non-zero: 6476, orthogonal: False,\n Spin{spin-o...
```

sisl/viz/plots/fatbands.py
pfebrer: `weight` is the name of the variable in the dataset, `weights` is what feels natural to me as an attribute if I want all the weights of the orbitals.

sisl/viz/plots/bands.py
pfebrer: The `import` statement already uses `importlib`, so I don't think this adds much overhead, no? Anyway, I'll change it if you prefer it.

pfebrer:
> Please remove commits changing files submodules needs a bit more care...

I don't understand how the submodule works exactly, sometimes things get pushed, sometimes they don't :confused:

sisl/viz/plots/fatbands.py
zerothi: ok, fine.

sisl/viz/plots/bands.py
zerothi: Perhaps spin shouldn't be there for non-diagonal spins?

sisl/viz/plots/bands.py
zerothi: I would prefer try except, it is basically what you do, so no need to complicate things. True about the existing overhead anyways!

zerothi:
> > Please remove commits changing files submodules needs a bit more care...
> 
> I don't understand how the submodule works exactly, sometimes things get pushed, sometimes they don't

Yeah, I generally do a commit only with the hash to ensure I do stuff correctly.

sisl/viz/plots/bands.py
pfebrer: And in non-polarized calculations?

zerothi: yeah, always a problem... ;)
In my PDOS returns I don't return spin-dimension for non-polarized. This is also annoying since users have to change code to handle corner cases. But I don't know how we should do it more easily...

pfebrer: Ok, then I'll have to look at it carefully, because this implies a handful of modifications in the code and in tests, basically adding checks to discern between the two cases.

zerothi: Ok, hmm.. I don't really know. If this is not something the user really touches then ok. However, I think we need to make it "easy to understand for users" and having a spin-index for a non-polarized calculation and non-collinear is a bit problematic. I should probably change the spin index in the total, x, y, z returns to spin-components for clarity and reduction of confusion? Any suggestions here?

pfebrer: > I should probably change the spin index in the total, x, y, z returns to spin-components for clarity and reduction of confusion? Any suggestions here?

I don't understand where do you mean to change it :sweat_smile:

zerothi: For instance the dataarray returned from sisl.io.siesta.pdos, this contains a spin index, however it is named so it is not ambiguous.

zerothi: All I mean is that there are other places in sisl where I should reconsider what to return and more importantly, _how_.

pfebrer:
Yes, I thought of this too, but then a user will get this warning even if everything they want to do is to plot a geometry or plot the bands from a .bands file :(

zerothi:
> Yes, I thought of this too, but then a user will get this warning even if everything they want to do is to plot a geometry or plot the bands from a .bands file :(

true... hmm... Perhaps could we make it a decorator to make it portable and use the same strings, I think the warning system should catch that as "same warning"?

pfebrer:
Which may cause unnecesary confusion and will probably be seen as a drawback for using the visualization module: "I don't use it because it says it might break my computer, I'd rather be safe" :sweat_smile:

zerothi:
> Which may cause unnecesary confusion and will probably be seen as a drawback for using the visualization module: "I don't use it because it says it might break my computer, I'd rather be safe"

They will so too if it is needlessly slow if they are oversubscribing. So no good solution.... :(

But probably the decorator is the best way?

pfebrer:
> true... hmm... Perhaps could we make it a decorator to make it portable and use the same strings, I think the warning system should catch that as "same warning"?

Yes that definitely makes sense, although it is not until the point where I wrote it now that you are 100% sure the calculation is going to happen. Otherwise they will get the warning because the entry point has been attempted but it's not going to work because there's no Hamiltonian.

pfebrer:
Probably the warning could be issued when you do `MonkhorstPack.apply.renew(pool=True)`? As this will also be a problem for users that do the parallellization on their own

pfebrer:
And what about using a context manager that sets `OMP_NUM_THREADS` to 1 temporarily? That would work, wouldn't it?

zerothi:
> And what about using a context manager that sets `OMP_NUM_THREADS` to 1 temporarily? That would work, wouldn't it?
> 
> Something similar to https://stackoverflow.com/a/34333710/11991536

Hmm. True. I don't know of the best place... In principle BLAS is used many, many places.
I have to say I don't really think the warning will be too much of a problem. Otherwise as you say the context should take care of sharing resources the best possible way. But then again that should issue a warning telling the user that sisl is overriding this?

You could for instance imagine a user doing `SISL_NPROCS=2 ; OMP_NUM_THREADS=2` with 4 cores in total. This shouldn't issue a warning since the total core-count is <=4.

If it isn't a warning but an `info` then it doesn't sound as severe?

zerothi:
Perhaps we should make a 2nd round of this warning, so feel free to leave it out here. Otherwise lets keep it as is and get it in.

pfebrer:
> But then again that should issue a warning telling the user that sisl is overriding this?

I was thinking it could set the `OMP_NUM_THREADS` variable only if it's not defined.

> Perhaps we should make a 2nd round of this warning, so feel free to leave it out here. Otherwise lets keep it as is and get it in.

I'll remove it for now then.

pfebrer:
I've removed all spin indices for spin kinds other than unpolarized and all the tests pass, so I'd say this is ready.

pfebrer:
I found a bug, don't merge yet!

pfebrer:
Now it's all good! Added some more tests to make sure this bug is catched in the future :)

zerothi (approved):

PR #368, opened 2021-08-26
Allow per-category atom styles in GeometryPlot

pfebrer:
As discussed [here in discord](https://discord.com/channels/742636379871379577/845235607461691402/854659603014615050) this PR introduces a way to specify atom styles per category. We discussed also bond styles, but this will come some time in the future :sweat_smile:

Mainly, I **removed the `atoms_color` and `atoms_size` settings and introduced an `atoms_style` setting** that accommodates not only colors and sizes but also other properties such as opacity. It also leaves space for new controls such as arrows [as discussed here](https://discord.com/channels/742636379871379577/845235607461691402/845315231210799135) to be introduced seemlessly. All this while allowing a very fine per-category control of the properties. It will also allow more control from GUIs. I'm quite happy with how it turned up :smile: 

Here I show how the API works with this PR so that we can discuss it and improve it if needed. When (if) we agree, I will prepare everything for merging (i.e. update tests and docs):

Assume we have created a geometry and stored it under `ribbon`.

```python
import sisl
import sisl.viz

ribbon = sisl.geom.graphene_nanoribbon(9)
```

Then we want to plot the geometry specifying the colors and sizes of the atoms.

**The old way**

```python
ribbon.plot(axes="xyz", atoms_color="green", atoms_size=[0.6, 0.8])
```

**The new way**

```python
ribbon.plot(axes="xyz", atoms_style={"color": "green", "size": [0.6, 0.8]})
```
image

So, to get the old behavior, one only needs to include the `"color"` and `"size"` keys in `atoms_style`. More keys could be introduced in the future. For example, `"opacity"` can also be controlled in this way now:

```python
ribbon.plot(axes="xyz", atoms_style={"color": "green", "size": [0.6, 0.8]}, "opacity": [1, 0.3])
```

image

But of course, as I said, there's much more to this new `atoms_style` setting. You can give different style specifications to different selections of atoms. Basically, all you need to do is to pass a list of styles, including (optionally) the `"atoms"` key to select the atoms to which these styles will be applied:

```python
ribbon.plot(
    axes="xyz", 
    atoms_style=[
        {"color": "green", "size": [0.6, 0.8], "opacity": [1, 0.3]},
        {"atoms": [0,1], "color": "orange"}
    ]
)
```

image

Notice how the first specification doesn't contain `"atoms"`, so it applies to all atoms. Also note that the properties that were not specified for `[0, 1]` are "inherited" from the previous specifications. If some atom is selected in more than one specification, the last one remains, that's why the color is finally set to orange for `[0,1]`. 

You don't need to include general styles. For atoms that don't have styles specified the defaults are used:

```python
ribbon.plot(axes="xyz", atoms_style=[{"atoms": [0,1], "color": "orange"}])
```
image

Finally, `"atoms"` accepts anything that `Geometry` can sanitize, so it can accept categories:

```python
ribbon_plot = ribbon.plot(axes="xyz", atoms_style=[
    {"atoms": {"fx": (None, 0.4)}, "color": "orange"},
    {"atoms": sisl.geom.AtomOdd(), "opacity":0.3},
])
ribbon_plot
```

image

And of course, as always this works for 1D, 2D and 3D:

```python
ribbon_plot.update_settings(axes="xy")
```
image

And all plotting backends:

```python
ribbon_plot.update_settings(backend="matplotlib")
```

image

```python
ribbon_plot.save("ribbon.plot")
```
*open blender*
```python
import sisl
import sisl.viz

ribbon_plot = sisl.viz.plotutils.load("ribbon.plot")
ribbon_plot.update_settings(axes="xyz", backend="blender")
```

image

(I don't know why blender now doesn't use the opacity, I will look into it)

As I said, looking forward to your suggestions on things that could be handled differently!

*PD*: I will ask people who I know have used `GeometryPlot` to participate in this discussion, so that we have a bigger number of informed opinions :) (@tfrederiksen you might be interested in discussing this API)


pfebrer:
Ok, here is the first implementation of arrows:

```python
sisl.geom.graphene().plot(axes="xy", atoms_style={"arrow": [[0,1,0], [1,1,0]]})
```

image

I implemented a generic `draw_arrows` method in the base `Backend` that just uses lines to draw arrows. So in all backends that have a `draw_line` implemented drawing this arrows will be possible. Of course if for some backend there is a better way to draw the arrows and someone wants to implement it the method can be overwritten.

```python
sisl.geom.graphene().plot(axes="xy", atoms_style={"arrow": [[0,1,0], [1,1,0]]}, backend="matplotlib")
```

image

As with all other properties, there's support to draw arrows based on atom selection:

```python

```

zerothi:
Looks great!!!

Perhaps
```python
ribbon.plot(
    axes="xy", show_cell=False,
    atoms_style=[
        {"color": "skyblue", "opacity": [1, 0.3], "arrow": [2, 1, 0]},
        {"atoms": [0,1], "color": "orange"},
        {"atoms": {"fx": (None, 0.4)}, "arrow": [-2, 0, 0]},
    ],
    arrow_style={"scale": 1.5, "color": "orange", "width": 3, "arrowhead_angle": 3.14 / 4, "arrowhead_scale": 0.2}
)
```

could be changed to:

```python
ribbon.plot(
    axes="xy", show_cell=False,
    atoms_style=[
        {"color": "skyblue", "opacity": [1, 0.3], "arrow": [2, 1, 0]},
        {"atoms": [0,1], "color": "orange"},
        {"atoms": {"fx": (None, 0.4)}, "arrow": {'data': [-2, 0, 0], 'style': {"scale": 1.5, "color": "orange", "width": 3, "arrowhead_angle": 3.14 / 4, "arrowhead_scale": 0.2}}}
    ],
)
```
seems like you really want the style, per atom-style as well?

Alternatively
```python
ribbon.plot(
    axes="xy", show_cell=False,
    atoms_style=[
        {"color": "skyblue", "opacity": [1, 0.3], "arrow": [2, 1, 0]},
        {"atoms": [0,1], "color": "orange"},
        {"atoms": {"fx": (None, 0.4)}, "arrow": ([-2, 0, 0], {"scale": 1.5, "color": "orange", "width": 3, "arrowhead_angle": 3.14 / 4, "arrowhead_scale": 0.2})}
    ],
   'arrow_style': {**default_options}
)
```
where the `"arrow"` could be an optional tuple, I don't know... :(

pfebrer:
Note that one can always post process the plot to get very specific displays for their use case. So you can use the arrow coordinates to split them into separate arrows and style them as you wish.

pfebrer:
Now plotly also supports 3D arrows:

```python
import sisl
import sisl.viz

ribbon = sisl.geom.graphene_nanoribbon(9)

ribbon.plot(
    axes="xyz", show_cell=False,
    atoms_style=[
        {"color": "skyblue", "opacity": [1, 0.3], "arrow": [2, 1, 0]},
        {"atoms": [0,1], "color": "orange"},
        {"atoms": {"fx": (None, 0.4)}, "arrow": [-1, 0, 2]},
    ],
    arrow_style={"scale": 1, "color": "red", "width": 4, "arrowhead_scale": 0.3}
)
```
image

They are drawn with a combination of a `scatter3d` (for the lines) and a `Cone` trace (for the arrow heads).

zerothi:
> Yes, that would be great. The problem is that with the current approach (single line for all arrows) it is not possible to give a different style for each arrow. Drawing a line for each arrow could be the solution but it would slow down the rendering a lot. Therefore other approaches are needed (which exist but are backend specific).

It was not to give individual styles for each arrow, but for each atom selection. That would fit nicely with your implementation (I think) https://github.com/pfebrer/sisl/commit/ee2c1ca2467efec10611ec9d9395f5d2cae2d08f#diff-09452270d8e15eb236b9dc19b111c0d8142daf685247f8e5ed4f165b740623c3R68

If you extend the *main* `arrow_style` with the `atom_style='arrow_style'` using a `ChainMap` you would have it right up?

```python
arrow_style = ChainMap(backend_info['atoms_props']['arrow_style'], backend_info["arrow_style"])
```
then use that?

> I like your last proposal. It seems like it would be backwards compatible with the current implementation. So, I would prefer to keep it simple for now and then increase the complexity if there is demand for the features :)

Agreed.

zerothi:
As a side-note, it seems that many places in the code you are constantly recreating dictionaries and overwriting keys in local temporary values. I think perhaps the `ChainMap` would be way more efficient, and also much easier. But I could be wrong?

pfebrer:
Perhaps you thought that there was a loop like:

```python
for style in atoms_style:
    self.draw_atoms(style)
    self.draw_arrows(style)
```

? That would be an interesting idea actually :thinking: Although that will mean there are more things to be drawn and could slow down rendering (specially in animations), but it is possible that there are not so many different styles that this becomes problematic.

zerothi:
> Perhaps you thought that there was a loop like:
> 
> ```python
> for style in atoms_style:
>     self.draw_atoms(style)
>     self.draw_arrows(style)
> ```
> 
> ? That would be an interesting idea actually  Although that will mean there are more things to be drawn and could slow down rendering (specially in animations), but it is possible that there are not enough different styles so that this becomes problematic.

I think the ability to customize outweighs the performance. Generally users will probably use the same styles and then it will draw 2-5 times (per atomic specie). How much slower is this compared to everything in one go? ;)

zerothi:
> > As a side-note, it seems that many places in the code you are constantly recreating dictionaries and overwriting keys in local temporary values. I think perhaps the ChainMap would be way more efficient, and also much easier. But I could be wrong?
> 
> I have never used `ChainMap`, so I don't know how it works. What is the advantage? Could you give an example of some place where it could be better?

A `ChainMap` is basically a dictionary with *levels*. Meaning that it will search incrementally through the levels of the dictionaries for a matching key, and return the first one encountered. And when you try to add a new key it will only add that for the current level (not changing the prior levels).

Example:
```python
d1 = {'hello': 1, 'world': 2}
d2 = {'hello': 2, 'other': 3}
cm = ChainMap(d2, d1)
print(cm["hello"])
print(cm["world"])
cm["world"] = 4
print(cm["world"], d1["world"]) # no overwriting in `d1`
```
will print:
```
2
2
4, 2
```
hope this makes sort of sence, see doc [here](https://docs.python.org/3/library/collections.html#collections.ChainMap)

zerothi:
Oh I forgot an example in your code. The pdos.py code uses lots of dicts with requests etc. It could be handy there?

pfebrer:
> I think the ability to customize outweighs the performance. Generally users will probably use the same styles and then it will draw 2-5 times (per atomic specie).

Hmm but everything in atoms can already be customized (color, size, opacity) by drawing one single scatter. Even the marker shape, which in plotly could be done like this:

```python
sisl.geom.graphene().plot(
    axes="xy"
).update_traces({"marker_symbol": ["star", "x"]}, selector={"name": "Atoms"})
```
image

I agree it would be nice to customize arrows individually, but drawing the arrows per atom selection doesn't improve the situation much. You still have **only one style per atom selection**. In practice, I think people would like that the color of each arrow displayed the norm of the vector or something like that. The `draw_arrows` function (backend agnostic) is available for the users if they want to draw arrows styled for their specific case:

```python
ribbon_plot = sisl.geom.graphene_nanoribbon(9).plot(axes="xy")

ribbon_plot.draw_arrows(
    ribbon[{"fx": (None, 0.4)}][:, :2], np.array([-2,0]), line={"color": "red"}
)

ribbon_plot.draw_arrows(
    ribbon[{"fx": (0.4, None)}][:, :2], np.array([2,0]), line={"color": "green"}
)

ribbon_plot
```

image

In terms of performance I'm not sure exactly how things change. I'm quite busy with other stuff right now and this would be a big change in how the drawing works, which would need some careful thinking. Since we agree on the API that users see, the internal workings will not mean any difference to how it is used.  Would it be OK to leave it like this for now and then change it if necessary?

pfebrer:
> A ChainMap is basically a dictionary with levels. Meaning that it will search incrementally through the levels of the dictionaries for a matching key, and return the first one encountered. And when you try to add a new key it will only add that for the current level (not changing the prior levels).

I will try to use it if I see a good case for it :)

> Oh I forgot an example in your code. The pdos.py code uses lots of dicts with requests etc. It could be handy there?

Hmm yes I can see how it can be useful, mainly for the defaults. I will try to change it when I have more time :smile:

zerothi:
> > I think the ability to customize outweighs the performance. Generally users will probably use the same styles and then it will draw 2-5 times (per atomic specie).
> 
> Hmm but everything in atoms can already be customized (color, size, opacity) by drawing one single scatter. Even the marker shape, which in plotly could be done like this:
> 
> ```python
> sisl.geom.graphene().plot(
>     axes="xy"
> ).update_traces({"marker_symbol": ["star", "x"]}, selector={"name": "Atoms"})
> ```
> 
> image
> 
> I agree it would be nice to customize arrows individually, but drawing the arrows per atom selection doesn't improve the situation much. You still have **only one style per atom selection**. 

And that is all I think is usable :)  
An arrow style per atom selection.

> In practice, I think people would like that the color of each arrow displayed the norm of the vector or something like that. The `draw_arrows` function (backend agnostic) is available for the users if they want to draw arrows styled for their specific case:
> 
> ```python
> ribbon_plot = sisl.geom.graphene_nanoribbon(9).plot(axes="xy")
> 
> ribbon_plot.draw_arrows(
>     ribbon[{"fx": (None, 0.4)}][:, :2], np.array([-2,0]), line={"color": "red"}
> )
> 
> ribbon_plot.draw_arrows(
>     ribbon[{"fx": (0.4, None)}][:, :2], np.array([2,0]), line={"color": "green"}
> )
> 
> ribbon_plot
> ```
> 
> image
> 
> In terms of performance I'm not sure exactly how things change. I'm quite busy with other stuff right now and this would be a big change in how the drawing works, which would need some careful thinking. Since we agree on the API that users see, the internal workings will not mean any difference to how it is used. Would it be OK to leave it like this for now and then change it if necessary?

I think so since the *global* settings would still be applicable in the new setup.

pfebrer:
> And that is all I think is usable :)
> An arrow style per atom selection.

Do you have something in mind? It's hard for me to think about a case where I would want a group of arrows all with one color and another group all with another color. 

Something that is not covered either by the current API/implementation is if you want to draw more than one arrow per atom (to display vector decompositions or more than one vector field at the same time). I see this as something that could come up more easily than coloring arrows by group.

zerothi:
> > And that is all I think is usable :)
> > An arrow style per atom selection.
> 
> Do you have something in mind? It's hard for me to think about a case where I would want a group of arrows all with one color and another group all with another color.

Say for atoms where you want to color arrows pointing in one direction (along some vector) with one color, and another color in the opposite direction.
For forces this may be quite natural to more easily visualise which atoms moves *with* some vector, and which atoms moves *against* the same vector.

> 
> Something that is not covered either by the current API/implementation is if you want to draw more than one arrow per atom (to display vector decompositions or more than one vector field at the same time). I see this as something that could come up more easily than coloring arrows by group.

Yeah ok, this could actually also be nice, but perhaps not so important right now...

decerio:
Hi!

I wasn't able to follow in detail all the conversation due to my lack of
knowledge in this field, but I'll leave a couple of comments, in case they
might be useful:


> In practice, I think people would like that the color of each arrow
> displayed the norm of the vector or something like that.
>
> I've always thought that would be an interesting feature.

Concerning having one arrow style per atom selection, in case one would
like to implement the last idea (the color of each arrow displays its norm,
e.g. bond current, bond length etc.), one should define as many atom
selections as atoms in the system, right? And just one question, is one
able to plot more than one arrow per atom?

I have my one super rudimentary code for matplotlib to create an array of
bond-lengths from a geometry and plot them assigning them colors depending
on the norm. I guess this is not possible to implement in plotly...

pfebrer:
> Something that is not covered either by the current API/implementation is if you want to draw more than one arrow per atom (to display vector decompositions or more than one vector field at the same time). I see this as something that could come up more easily than coloring arrows by group.

Maybe I should allow arbitrary `*_arrow` arguments in `atoms_style`? Then I could draw as many as the user wants and it could be a good way to distinguish. `arrow_style` could then accept different styles for each `*_arrow`.

```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style=[
        {"forces_arrow": [1, 0, 0], "efield_arrow": [2,0,0]}
    ],
    arrows_style={
        "forces": {"color": "red", ...},
        "efield": {"color": "green", ...}
    }
)
```

Of course you could overwrite the arrow values of each one in later atom selections as with any other property. You could use this for your doing something like what you wanted to do as well, so it would solve two things in one :)

> For forces this may be quite natural to more easily visualise which atoms moves with some vector, and which atoms moves against the same vector.

I agree, but this means you have to know in advance which arrows will be "positive" and which ones "negative", which is not ideal. I see this case of coloring the same as coloring individual arrows by their norm. This is something that should be done internally by checking the values and coloring accordingly. While it is not possible, I would say that using `draw_arrows` yourself would be the solution, as you have to write the logic to find the negative and the positive ones anyway.

zerothi:
> > Something that is not covered either by the current API/implementation is if you want to draw more than one arrow per atom (to display vector decompositions or more than one vector field at the same time). I see this as something that could come up more easily than coloring arrows by group.
> 
> Maybe I should allow arbitrary `*_arrow` arguments in `atoms_style`? Then I could draw as many as the user wants and it could be a good way to distinguish. `arrow_style` could then accept different styles for each `*_arrow`.
> 
> ```python
> sisl.geom.graphene_ribbon(9).plot(
>     atoms_style=[
>         {"forces_arrow": [1, 0, 0], "efield_arrow": [2,0,0]}
>     ],
>     arrows_style={
>         "forces": {"color": "red", ...},
>         "efield": {"color": "green", ...}
>     }
> )
> ```
> 
> Of course you could overwrite the arrow values of each one in later atom selections as with any other property. You could use this for doing something like what you wanted to do as well, so it would solve two things in one :)

Hmm, what about:
```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style=[
         {'arrows': [
       {'name': 'force', 'data': [[1, 0,0]], 'style': {...}},
       {'name': 'efield', 'data': [[2, 0,0 ]], 'style': {...}}
   ]}
    ],
    arrows_style={
        # default values
        "force": {"color": "red", ...},
        "efield": {"color": "green", ...}
    }
)
```

> > For forces this may be quite natural to more easily visualise which atoms moves with some vector, and which atoms moves against the same vector.
> 
> I agree, but this means you have to know in advance which arrows will be "positive" and which ones "negative", which is not ideal. I see this case of coloring the same as coloring individual arrows by their norm. This is something that should be done internally by checking the values and coloring accordingly. While it is not possible, I would say that using `draw_arrows` yourself would be the solution, as you have to write the logic to find the negative and the positive ones anyway.

That is quite trivial:
```python
proj = force.dot([1, 0, 0])
positive = (proj>0).nonzero()[0]
negative = (proj < 0).nonzero()[0]
atoms_style = [
  {'atoms': positive, 'arrows': {..., 'data': force[positive]}},
  {'atoms': negative, 'arrows': {..., 'data': force[negative]}},
]
```
?

I am not sure whether we talk by each other here. I am not necessarily talking about a callback function for retrieving the colors of individiual arrows...

pfebrer:
> I have my one super rudimentary code for matplotlib to create an array of
> bond-lengths from a geometry and plot them assigning them colors depending
> on the norm. I guess this is not possible to implement in plotly...

Do you use `LinesCollection`? Exactly, coloring a single line with different colors is a difficult issue and it doesn't have a simple workaround in `plotly` in 2D. In `3D` however it is simpler (due to the way the low-level frameworks handle things) so you can directly provide multiple colors for a 3D line in `plotly`. Matplotlib (as far as I know) has the same issue and that's why you need to use `LinesCollection`. In `BondLenghMap` I overcome this by drawing very close dots instead of a line, which you can then color individually :sweat_smile:

pfebrer:
> ```
> sisl.geom.graphene_ribbon(9).plot(
>     atoms_style=[
>          {'arrows': [
>        {'name': 'force', 'data': [[1, 0,0]], 'style': {...}},
>        {'name': 'efield', 'data': [[2, 0,0 ]], 'style': {...}}
>    ]}
>     ],
>     arrows_style={
>         # default values
>         "force": {"color": "red", ...},
>         "efield": {"color": "green", ...}
>     }
> )
> ```

This seems too crowded for me, specially if you want to specify arrows for multiple atom selections :( I think it all boils down to the fact that you want to include arrow styles in `atoms_style` and I think it is not necessary :sweat_smile: 

> I am not sure whether we talk by each other here. I am not necessarily talking about a callback function for retrieving the colors of individiual arrows...

I know what you mean, but still this is coloring based on values, which can be done internally once the values are known. What I was thinking is that in the future (when I figure out how to efficiently color arrows) you could do something like this:

```python
arrows_style={
    "force": {"color": NegPosColoring(proj=[1,0,0])}
}
```

The advantage of this is the same advantage that `AtomCategory` has over atom indices. You don't need to know **anything** about the values, but you can still create the classifier. I think that is very powerful. Also that would play much better with what I have in mind for the future of `atom_styles`. You won't need to pass actual values, but specify what is the property that you want to use:

```python
atoms_style={"color": AtomData("Z"), "size": AtomData("z"), "arrow": AtomData("forces")}
```

This would know how to get some data from the geometry (`Z` and `z` are trivial, but forces would need some help from the fdf that was used to read the geometry).

The names are not important at all, I hope I was able to transmit the concept that I have in mind. The main driver of this concept is that I see this as the only way that complex customization can happen in a GUI. Although I see it as a useful thing as well for programming.

decerio:
>
> Do you use LinesCollection? Exactly, coloring a single line with
> different colors is a difficult issue and it doesn't have a simple
> workaround in plotly in 2D. In 3D however it is simpler (due to the way
> the low-level frameworks handle things) so you can directly provide
> multiple colors for a 3D line in plotly. Matplotlib (as far as I know)
> has the same issue and that's why you need to use LinesCollection. In
> BondLenghMap I overcome this by drawing very close dots instead of a
> line, which you can then color individually 😅
>
No, I mean plotting each line with a different color (one color per line).
I guess this is just one arrow style per arrow, so that we have as many
styles as number of arrows. I use quiver for that and just do a colormap
according to bond-lengths saved in an array.

Regarding coloring a single line with different colors, yes, I've only used
LineCollection for that purpose.

>

pfebrer:
> No, I mean plotting each line with a different color (one color per line).
> I guess this is just one arrow style per arrow, so that we have as many
> styles as number of arrows.

That's exactly what I'm trying to avoid because having hundreds/thousands of lines is deadly for the rendering performance :turtle:.  (at least in plotly)

> I use quiver for that and just do a colormap
> according to bond-lengths saved in an array.

I saw that possibility, probably `quiver` uses a `LinesCollection` with a colormap internally, no? That is why I have hope on coloring arrows individually, it just needs more careful thought (since different implementations are needed for each backend).

pfebrer:
By the way @zerothi with the approach I was proposing you can already do what you wanted to achieve:

```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style=[
        {"atoms": [0, 1], "negforces_arrow": [-1, 0, 0]}
        {"atoms": [2, 3], "posforces_arrow": [1,0,0]}
    ],
    arrows_style={
        "negforces": {"color": "blue", ...},
        "posforces": {"color": "red", ...}
    }
)
```

zerothi:
> By the way @zerothi with the approach I was proposing you can already do what you wanted to achieve:
> 
> ```python
> sisl.geom.graphene_ribbon(9).plot(
>     atoms_style=[
>         {"atoms": [0, 1], "negforces_arrow": [-1, 0, 0]}
>         {"atoms": [2, 3], "posforces_arrow": [1, 0, 0]}
>     ],
>     arrows_style={
>         "negforces": {"color": "blue", ...},
>         "posforces": {"color": "red", ...}
>     }
> )
> ```

you are right. It makes no sense. Rather I would prefer:
```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style=[
        {"atoms": [0, 1], "arrow": {'data':[-1, 0, 0], 'style': {...}}
        {"atoms": [2, 3], "arrow": {...}}
    ],
})
```
which is also not too much different than your proposal. :)

pfebrer:
I don't understand why don't you like having an arbitrary number of arrows and specifying their style separately :( I guess you are thinking that if you want multiple arrows you will do something like this:

```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style=[
        {"arrow": {'data': forces, 'style': {"color": "blue"}} # The forces arrows
        {"arrow": {'data': efield, 'style': {"color": "blue"}} # The efield arrows.
    ],
})
```
This would make the behavior of `arrow` inconsistent with all the other properties. This:
```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style=[
        {"color": "blue"}
        {"color": "green"} 
    ],
})
```
doesn't draw the atoms twice (one in blue and one in green), but just once in green because it was the last specification. However, you want arrows to be drawn regardless. Maybe if we choose to go for this approach simply there should be a separate `arrows` setting that handles arrows, instead of trying to put them in `atoms_style`.

zerothi:
wouldn't multiple atoms be allowed:
```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style=[
        {"atoms": [0, 1], "arrow": {'data':[-1, 0, 0], 'style': {'color': 'blue', ...}},
        {"atoms": [0, 1], "arrow": {'data':[1, 0, 0], 'style': {'color': 'red', ...}},
        {"atoms": [2, 3], "arrow": {...}}
    ],
})
```
I am just not particularly fond of having them disconnected from the dictionary that defines them. That seems weird to me... ;)

pfebrer:
Yes, but what you want this to say is:
```python
atoms_style = [
    {...draw these arrows in atoms [0,1]},
    {...draw these other arrows in atoms [0,1]},
    {...draw these other arrows in atoms [2,3]}
]
```
while with all other styling properties this works like:
```python
atoms_style = [
    {...if no further arrows are specified for atoms [0,1], draw these},
    {...if no further arrows are specified for atoms [0,1], draw these},
    {...if no further arrows are specified for atoms [2,3], draw these}
]
```
I agree that arrows should behave like the first example. So I think that atomic styles and arrows, even though they are something that is specified per atom, they are two separate things that should be treated differently.

Also, I agree now that it is better if everything for the arrows (data and styles) is defined in one place. This will also be better for GUIs. Putting it inside `atoms_style`, everything seemed too nested. But if we define a new `arrows` setting I think it's perfectly fine:

```python
sisl.geom.graphene_ribbon(9).plot(
    atoms_style= ...however you want to style atoms,
    arrows = [
        {"atoms": [0, 1], 'data': [-1, 0, 0], 'color': 'blue', 'name': 'Negative forces'}, # Draw these arrows
        {"atoms": [0, 1], ...}, # And these
        {"atoms": [2, 3], ...} # And these
    ],
)
```

What do you think? I believe can be much more smoothly handled by GUIs and it will be less confusing to users: styles are styles and arrows are arrows :)

zerothi:
> Also, I agree now that it is better if everything for the arrows (data and styles) is defined in one place. This will also be better for GUIs. Putting it inside `atoms_style`, everything seemed too nested. But if we define a new `arrows` setting I think it's perfectly fine:
> 
> ```python
> sisl.geom.graphene_ribbon(9).plot(
>     atoms_style= ...however you want to style atoms,
>     arrows = [
>         {"atoms": [0, 1], 'data': [-1, 0, 0], 'color': 'blue', 'name': 'Negative forces'}, # Draw these arrows
>         {"atoms": [0, 1], ...}, # And these
>         {"atoms": [2, 3], ...} # And these
>     ],
> )
> ```
> 
> What do you think? I believe this can be much more smoothly handled by GUIs and it will be less confusing to users: styles are styles and arrows are arrows :)

Yeah, I like this better, it doesn't get too nested!

pfebrer:
Tomorrow I'll update the docs, and this will be ready :+1: .

pfebrer:
All ready to merge :) I added a section for blender in the `sisl.viz` docs and I think with the installation method detailed there it should be quite easy to get started with it. Let's see if somebody starts using it :+1:

docs/visualization/viz_module/blender/Getting started.rst
```diff
@@ -0,0 +1,129 @@
...
+3. **Create an environment with that python version** and install sisl (*skip if you have it already*).
+In this case, we will use conda as the environment manager, since it lets us very easily select the python version.
+You probably don't need the exact micro version. In our case asking for ``3.9`` is enough::
+    
+    conda create -n blender-python python=3.9
+
+Then install all the packages you want to use in blender::
+
+    conda activate blender-python
+    pip install sisl[viz]
```
zerothi, line 46: the pip community is preferring `python -m pip install` because this ensures the pip module that is used corresponds to the python running. I'll merge and update.

zerothi:
Oh, I am not happy about images in the repo...  
I don't want the repo to explode, the CI's are generally depending on fast downloads and having a huge repo with lots of binary files always becomes the bottleneck. For the documentation we need to think differently about this. Especially in this case for images in MB sizes...

docs/visualization/viz_module/blender/Getting started.rst
pfebrer: True, sorry!

pfebrer:
You are right, I hadn't even realised that the images where 1MB. Would it make sense that docs is a submodule? Do CI workflows  need the docs? Becaus e it makes sense that docs/tutorials have images, doesn't it? How do other packages handle the situation?

Otherwise the images could be stored in `sisl-files` and get fetched to build the documentation? They could also not get fetched and just provide external links for the image path

pfebrer:
> Otherwise the images could be stored in sisl-files

I think this could be the solution. But the tests use this also so the CIs have to download this anyway no? At least the developers/users won't need to download them.

By the way, the images can be resized from 1MB to 200kB without loosing too much definition.

zerothi:
Good idea. However, I think a better approach would be to create another repository only for documentation. Hmm...
Could we for now remove the images and then just put a placeholder? We should probably see what others are doing...

pfebrer:
I've made the blender backend much better! I think this will be a great addition before the release :)

Previously drawing big systems took a long time, but now the performance is much better.

**More importantly, animations work!!** You can use the same interface as always and you'll get now the animation in blender. This is a gif I rendered of a rotating nanoribbon with very quickly changing colors and size of the atoms (don't look at it for too long):



Code:

```python
import sisl
import sisl.viz
import numpy as np

# Build all the properties that we will animate
frames = 20
rots = np.linspace(0, 360, frames)
scales = [*np.linspace(1, 0.2, int(frames/2)), *np.linspace(0.2, 1, int(frames/2))]
colors = ["red", "orange", "green", "blue", "red"]*int(frames/4)

# Build the ribbon geometry
ribbon = sisl.geom.graphene_nanoribbon(9)
# Define the rotation parameters (vector and origin)
v_rot = [1,0,0]
orig_rot = ribbon.cell[1]/2 

# Build all plots
plots = []
for rot, scale, color in zip(rots, scales, colors):
    
    geom_plot = ribbon.rotate(rot, v_rot, orig_rot).plot(
        atoms_scale=scale, atoms_style={"color": color},
        backend="blender", nsc=[9,1,1], show_cell=False
    )
    
    plots.append(geom_plot)

# And generate the animation
ani = sisl.viz.Animation(plots=plots, backend="blender", interpolated_frames=10) 

# Show it
ani.show()
```

And here an animation of a fake vibration: (fulfilling finally @tfrederiksen's request!).

https://user-images.githubusercontent.com/42074085/132590460-73b9529a-10d7-4f3d-b563-44a1cb34930b.mp4

Code:

```python
import sisl 
import numpy as np

# Build the oscillation using a sine signal
sin = np.sin(np.linspace(0, 2*3.14, 20))
bonds = sin + 3
data = - sin.reshape(-1, 1).dot(np.reshape([[0,0,3], [2,0,0]], (1, -1))).reshape(-1, 2, 3)
arrow_colors = ["red", "green", "blue", "red"] * 5

plots = []
for bond, frame_data, color in zip(bonds, data, arrow_colors):
    
    geom_plot = sisl.geom.graphene(bond).plot(
        backend="blender",
        arrows={"data": frame_data, "color": color},           
    )
    
    plots.append(geom_plot)

sisl.viz.Animation(plots=plots, backend="blender").show()
```

Bonds are yet not animatable unfortunately, because it needs careful thought (number of bonds and bonded atoms can change throughout the animation).

pfebrer:
By the way, yes I can remove the images for now. Probably the tutorial will not be as clear and harder to read, but it's ok if we can add the images later.

Can I do this (and fix some tests that I know now are failing) in a separate PR? I'm having a lot of trouble rebasing to master now (I don't know why) and I don't want to mess everyhting up.

pfebrer:
Ok, now I created lots of conflicts :( I think it is all because of the pep8 clean, and I'm not sure how to fix them all at once.

pfebrer:
The images are not there anymore.  I believe everything is ready to merge :+1:

zerothi:
Amazing job Pol!

zerothi:
I'll have to check the log to ensure the images are not in the history as well. This is the important part since if they are added, then removed, they are still part of the repo. One has to squash the two commits that add and remove them together in one.

zerothi:
I'll rebase and squash the two commits:
The commands to do this is:
```bash
$> git checkout geomplot_additions
$> rebase_cmt=$(git merge-base master HEAD)
$> git rebase -i $rebase_cmt
```
this will yield:
```
pick 37fd99790 enh: sketch of atoms_style in GeometryPlot
pick 5fa9a8409 enh: first implementation of arrows in GeometryPlot
pick 7f04d3b4a enh: 3D arrows in plotly
pick 4b30c9979 enh: bonds and cell drawn by blender
pick cb7e84bb7 enh: final implementation of arrows and tests passing
pick f363ab42b bug: arrows were failing in blender
pick 221b9d417 docs: Added blender docs
pick a5ee96655 docs: added docs for styles and arrows
pick 956a6a12c enh: much faster blender
pick 629f2d8fd maint: removed images from docs and tests pass
```
change to
```

pick 37fd99790 enh: sketch of atoms_style in GeometryPlot
pick 5fa9a8409 enh: first implementation of arrows in GeometryPlot
pick 7f04d3b4a enh: 3D arrows in plotly
pick 4b30c9979 enh: bonds and cell drawn by blender
pick cb7e84bb7 enh: final implementation of arrows and tests passing
pick f363ab42b bug: arrows were failing in blender
pick 221b9d417 docs: Added blender docs
squash 629f2d8fd maint: removed images from docs and tests pass
pick a5ee96655 docs: added docs for styles and arrows
pick 956a6a12c enh: much faster blender
```
this will merge the addition and removal of images into one commit, thus completely removing the files from the history.

zerothi:
Advice, generally one should make a commit for each file that is added in a distinct commit. This means that one can move about the squashed commits to the location where needed in the history.

pfebrer:
If you want, I can try to remove the images from history myself. Let's see if I don't mess up :)

pfebrer:
It should be fine now :+1: Let me know if there's still something wrong

pfebrer:
I've pushed opacity in at the last minute hehe. It is also animatable. I promise I'm done now.

pfebrer:
Demo:

https://user-images.githubusercontent.com/42074085/132663353-d7e92d1a-d094-4b31-8399-4e2ef3e70c17.mp4

Code:

```python
import sisl
import sisl.viz

plots = []
for color, opacity, scale in zip(["red", "orange", "green"], [1, 0.2, 1], [0.5, 1, 0.5]):
    geom_plot = sisl.geom.graphene().plot(backend="blender", 
        atoms_style={"color": color, "opacity": opacity},
        atoms_scale=scale
    )
    
    plots.append(geom_plot)
    
sisl.viz.Animation(plots=plots, backend="blender", interpolated_frames=50).show()
```

zerothi:
> I've pushed opacity in at the last minute hehe. It is also animatable. I promise I'm done now.

Sure you are ;)

Great work!

zerothi:
@pfebrer no, it does not work. The images are still in the history. It looks as though you did a merge which doesn't work...

Let me know when I can have a go at it.

zerothi:
Actually it seems you have just removed the commit that removed the files?

Since you have the history very weird know, I would suggest you first clean up the history... ;)

pfebrer:
I know what happened. I had fixed it but then I forced pushed from another instance of the repository that didn't have the history fixed 😅 I can fix it again when I get home in 1h. All that I did was to reset --soft 956a6a1 , which is the commit before I introduced the images and then cherry-pick the subsequent commits modifying the ones that had something to do with the images, it was quite logic and straightforward in this way.

zerothi:
> I know what happened. I had fixed it but then I forced pushed from another instance of the repository that didn't have the history fixed  I can fix it again when I get home in 1h. All that I did was to reset --soft [956a6a1](https://github.com/zerothi/sisl/commit/956a6a12cceb8a69e383f226c5aa5e92fbf6c58a) , which is the commit before I introduced the images and then cherry-pick the subsequent commits modifying the ones that had something to do with the images, it was quite logic and straightforward in this way.

Ok, great, I'll wait till you say go ;)

pfebrer:
Go! :)

pfebrer:
image

Isn't this the history? Or there are more things hidden?

zerothi:
No, try and do a gitk and it should be clear.
The github interface does not show all commits, in this case it looks like you did a merge which results in hidden commits.

zerothi:
Perhaps wait, I'll check if my repo is fetched.... Sorry so far

zerothi:
Sorry, about this, indeed my local repo wasn't updated and I kept seeing the same thing... It is ready! Thanks!

zerothi:
I got errors when running locally, something I am missing? Before you made some changes, I could get it to work?

```python
___________________________________________________________________________ TestMultiplePlot.test_update_settings ____________________________________________________________________________

self = <sisl.viz.tests.test_plot.TestMultiplePlot object at 0x14aeb27b5100>

    def test_update_settings(self):
    
        kw = MultiplePlot._kw_from_cls(self._cls)
    
        geom = sisl.geom.graphene()
    
        multiple_plot = geom.plot(show_cell=["box", False, False], axes=[0, 1], **{kw: "show_cell"})
        assert len(multiple_plot.children) == 3
    
>       prev_data_lens = [len(plot.data) for plot in multiple_plot]

/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.12.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/tests/test_plot.py:174: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/gnu/9.3.0/python/packages/3.8.5/sisl-dev/0/lib/python3.8/site-packages/sisl-0.12.0.dev0-py3.8-linux-x86_64.egg/sisl/viz/tests/test_plot.py:174: in <listcomp>
    prev_data_lens = [len(plot.data) for plot in multiple_plot]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <sisl.viz.plots.geometry.GeometryPlot object at 0x14aeb2700310>, key = 'data'

    def __getattr__(self, key):
        """ This method is executed only after python has found that there is no such attribute in the instance
    
        So let's try to find it elsewhere. There are two options:
            - The attribute is in the backend object (self._backend)
            - The attribute is currently being shared with other plots (only possible if it's a childplot)
        """
        if key in ["_backend", "_get_shared_attr"]:
            pass
        elif hasattr(self._backend, key):
            return getattr(self._backend, key)
        else:
            #If it is a childPlot, maybe the attribute is in the shared storage to save memory and time
            try:
                return self._get_shared_attr(key)
            except (KeyError, AttributeError):
                pass
    
>       raise AttributeError(f"The attribute '{key}' was not found either in the plot, its backend, or in shared attributes.")
E       AttributeError: The attribute 'data' was not found either in the plot, its backend, or in shared attributes.
```
I get this error 3 times: `test_update_settings(self)` in `TestAnimation` and `TestSubPlots` and `TestMultiplePlot`

PR #373, opened 2021-09-07
fix: added missing : at end of if statement

sofiasanz:
There were ":" missing at the end of two `if` statements which have been added in this minor fix.


zerothi:
Thanks! However, I was in the process of fixing this + some larger rework of an error message. So I think this is redundant now... :( Sorry for being so slow :)

PR #374, opened 2021-09-09
bug: geometry plot tests were failing

pfebrer:
This should fix the problems with the tests. Sorry, I couldn't test the last commit because repo installation was in a different  environment than the one I use for testing :smile: Thanks to tests we will not release a buggy visualization module :)


pfebrer:
Do they pass for you as well now?

zerothi:
Yes.

Issue #375, opened 2021-09-10
viz and import statements in tests


zerothi:
**Describe the issue**
@pfebrer I have a task for you ;)

See [here](https://app.travis-ci.com/github/zerothi/sisl/jobs/536604850).

The tests fails because of imports, now I can't figure out exactly where all the tests are. But could you please put a 
```python
pytest.importorskip("plotly", reason="plotly not available")
```
where necessary so they won't fail?

Thanks!

EDIT: yes, I meant to write `plotly`

pfebrer:
All tests that depend on `skimage` are already marked.

The problem is that if you don't have plotly you will basically not test almost anything. So I would say we either install `plotly` in the CI workflows or I can try to rewrite the tests so that they don't need plotly. It makes complete sense to rewrite the tests without plotly in this new design. I can do it if you can wait 1 day.

zerothi:
> All tests that depend on `skimage` are already marked.
> 
> The problem is that if you don't have plotly you will basically test almost nothing. So I would say we either install `plotly` in the CI workflows or I can try to rewrite the tests so that they don't need plotly. It makes complete sense to rewrite the tests without plotly in this new design. I can do it if you can wait 1 day.

Since plotly isn't required by the user, it shouldn't be required in the tests. This is my point. Exactly the same issue is apparent for `dill` and `pathos`. These are optional dependencies, but since `pytest` imports all test files at test-time, it will fail the entire suite. This means that users can't test on their own. So we better have a CI test doing the same as a minimal user.

I think the tests that require plotly should just be skipped with importerskips, or something similar, whatever you prefer. I'll hold ;)

zerothi:
note I have just `autopep8` the code, so please pull :)

pfebrer:
Ok :+1:  I'm almost done :sweat_smile:

PR #376, opened 2021-09-13
tests: removed plotly requirement to test sisl.viz

pfebrer:
Closes https://github.com/zerothi/sisl/issues/375.

From now on, having loaded backends (`plotly`, `matplotlib`...) is not required to run the viz tests. However, if they are there, they will be tested, as detailed in the description of the commit in this PR.

Also `pathos` and `dill` are no longer required to use `sisl.viz` and for running tests. The `tqdm` dependency has also been dropped (this was a bug, not even documented).

However, `xarray` and `pandas` are still required to run the tests. I wanted to make pytest avoid the whole `sisl.viz` directory if they are not available, but I didn't know how to do it.


zerothi:
> From now on, having loaded backends (`plotly`, `matplotlib`...) is not required to run the viz tests. However, if they are there, they will be tested, as detailed in the description of the commit in this PR.

Great! 

> Also `pathos` and `dill` are no longer required to use `sisl.viz` and for running tests. The `tqdm` dependency has also been dropped (this was a bug, not even documented).

Great! 

> However, `xarray` and `pandas` are still required to run the tests. I wanted to make pytest avoid the whole `sisl.viz` directory if they are not available, but I didn't know how to do it.

I think this is perfect! Perhaps some of the history stuff can then be tested. :) but agreed it could be ok in the long run, not worth pursuing! Thanks!

zerothi:
still failing:
```
==================================== ERRORS ====================================
2771_ ERROR collecting builds/sisl-venv/lib/python3.6/site-packages/sisl/viz/plots/tests/test_bands.py _
2772ImportError while importing test module '/home/travis/build/zerothi/sisl/builds/sisl-venv/lib/python3.6/site-packages/sisl/viz/plots/tests/test_bands.py'.
2773Hint: make sure your test modules/packages have valid Python names.
2774Traceback:
2775../builds/sisl-venv/lib/python3.6/site-packages/_pytest/python.py:578: in _importtestmodule
2776    mod = import_path(self.fspath, mode=importmode)
2777../builds/sisl-venv/lib/python3.6/site-packages/_pytest/pathlib.py:524: in import_path
2778    importlib.import_module(module_name)
2779/opt/python/3.6.7/lib/python3.6/importlib/__init__.py:126: in import_module
2780    return _bootstrap._gcd_import(name[level:], package, level)
2781<frozen importlib._bootstrap>:994: in _gcd_import
2782    ???
2783<frozen importlib._bootstrap>:971: in _find_and_load
2784    ???
2785<frozen importlib._bootstrap>:955: in _find_and_load_unlocked
2786    ???
2787<frozen importlib._bootstrap>:665: in _load_unlocked
2788    ???
2789../builds/sisl-venv/lib/python3.6/site-packages/_pytest/assertion/rewrite.py:170: in exec_module
2790    exec(co, module.__dict__)
2791../builds/sisl-venv/lib/python3.6/site-packages/sisl/viz/plots/tests/test_bands.py:20: in <module>
2792    from xarray import DataArray
2793E   ModuleNotFoundError: No module named 'xarray'
2794_ ERROR collecting builds/sisl-venv/lib/python3.6/site-packages/sisl/viz/plots/tests/test_fatbands.py _
```

sisl/viz/plots/tests/test_fatbands.py
```diff
@@ -12,6 +12,7 @@
 import sisl
 from sisl.viz.plots.tests.test_bands import TestBandsPlot as _TestBandsPlot
 
+from xarray import DataArray
```
zerothi, line 15: I think this is the offending line.

zerothi: I think I've fixed this, I'll have a go

pfebrer:
If pandas and xarray are not there all sisl.viz tests should be skipped, that's what I meant here:

> However, xarray and pandas are still required to run the tests. I wanted to make pytest avoid the whole sisl.viz directory if they are not available, but I didn't know how to do it.

I don't know how to do it.

pfebrer:
Not only tests should be skipped, but also sisl.viz should never be imported. Otherwise you'll get the same error because xarray is imported in some places throughout the visualization module

zerothi:
> Not only tests should be skipped, but also sisl.viz should never be imported. Otherwise you'll get the same error because xarray is imported in some places throughout the visualization module

Yes, and this is why we should *lazy load* xarray and the necessary modules. Because `pytest` forces the imports of the test files.

One could do something like this:
```python
pytestmark = [pytest.mark.importorskip("xarray")

try:
     import xarray
except: pass
```
in the involved tests. But I presume this needs more careful thought at it should be handled higher I think.

zerothi:
I think it is ready, the final tests failing are my fault. I'll fix these...

PR #381, opened 2021-09-16
doc: no errors in notebooks

pfebrer:
With this small change I get no errors building the documentation


zerothi:
Thanks! Fixed!

PR #383, opened 2021-10-01
Some fixes so that the GUI can work properly

pfebrer:
This morning I've been adding the possibility to use atom categories in the GUI so that you can do some fairly complex selection of atoms without having to use python. 

For example, here I select atoms to provide styles for them, but this will be also probably useful for PDOS and fatbands or any other place where you need to select atoms.

image

image

I needed to do some changes in sisl for it to work. I would say that the changes you need to check if you agree are:
- Added an `AtomTag` category, which is pretty useful because it works with regular expressions.
- Fixed (?) the `AtomIndex` category. It wasn't working and I needed it so that the selection of atoms in the GUI is only made through categories, which facilitates the design.
- Modified `AtomNeighbours` so that the keyword `neigbour` accepts a dictionary which is then converted to a category.

The rest involve only `sisl.viz`


sisl/geom/category/_kind.py
```diff
@@ -52,6 +53,39 @@ def __eq__(self, other):
         return False
 
 
+@set_module("sisl.geom")
+class AtomTag(AtomCategory):
```
zerothi, line 57: Also add `AtomTag` to `__all__`

sisl/geom/category/_kind.py
```diff
@@ -109,13 +143,12 @@ def make_partial(a, b):
...
-        if len(idx) == 0:
-            operator = []
-        else:
+        operator = []
+        if len(idx) > 0:
             @wraps(op.contains)
             def func_wrap(a, b):
                 return op.contains(b, a)
-            operator.append((func_wrap, b))
+            operator.append((func_wrap, idx))
```
zerothi, line 151: Thanks for catching this bug!

sisl/geom/category/_neighbours.py
```diff
@@ -71,6 +71,8 @@ def __init__(self, *args, **kwargs):
             name = f" ∈ [{self._min};{self._max}]"
 
         self._in = kwargs.get("neighbour", None)
+        if isinstance(self._in, dict):
+            self._in = AtomCategory(**self._in)
```
zerothi, line 75: Good idea!

zerothi (changes requested):
Just a minor change needed

zerothi:
Also for indexes, you could consider using `strseq` and `lstranges` for parsing `"2-5,10,13:19" == `range(2, 6) + [10] + range(13, 20)`

pfebrer:
Does the implementation of `AtomSeq` make sense or you were thinking in doing it in some other way?

(don't merge yet, I want to still make some changes)

zerothi:
> Does the implementation of `AtomSeq` make sense or you were thinking in doing it in some other way?
> 
> (don't merge yet, I want to still make some changes)

Looks like a very good addition. At first I thought it was easier, but your point about first finalizing the object when you know the geometry length makes sense. A bit annoying since I would have liked it to be directly parseable in `AtomIndex`, that is easier in terms of `dict` parsing with `index=...`, now one has to differentiate with `seq=...` and `index=...`.  
For now this is good, but perhaps in the future it would be nice to have it in `AtomIndex`?

pfebrer:
I'm writing the tests and I realized that you can not have standalone negative numbers using your parsers. `AtomSeq("-2")` is interpreted as "from the beggining to 2". Do you use this some place in the code currently? I think support for `-` as a range should be dropped to avoid this problem, or have some way of differentiate when a negative number is being passed. E.g. with parenthesis `AtomSeq("(-2)")`.

zerothi:
> I'm writing the tests and I realized that you can not have standalone negative numbers using your parsers. `AtomSeq("-2")` is interpreted as "from the beggining to 2". Do you use this somewhere in the code currently? I think support for `-` as a range should be dropped to avoid this problem, or otherwise have some way of differentiating when a negative number is being passed. E.g. with parenthesis `AtomSeq("(-2)")`.

I have that because of command line utilities. Users who are not python knowledgeable would use - as range specifier (by experience). So yes it is used... 

I see that it does not go well with negative numbers... :(

pfebrer:
Hmm yes I figured this was the reason. However, I think no one would expect `-2` to mean `0-2`. Perhaps you could not assume the beggining for ranges defined with `-`? Maybe use some special character for "start"? `_-2` `?-2` `*-2`, idk.

pfebrer:
Or even the word start itself directly. `start-2`, `4-end`. I don't think "normal" people is used to empty meaning "start" or "end"

pfebrer:
Ok, so this is ready to merge. I didn't do anything about the standalone negative numbers, I kept it as it is.

zerothi:
See #384, I have fixed your request, so perhaps you could check the tests again.

zerothi:
> Ok, so this is ready to merge. I didn't do anything about the standalone negative numbers, I kept it as it is.

I thought many changes were required, I fixed everything and will merge! Thanks!

Issue #147, opened 2019-10-21
Siesta read wfs and wfsx files


zerothi:
**Describe the feature**
Title says it all. 

Probably this should only be in fdf since then we can read the Hamiltonian and associate it simultaneously with the gauge.

zerothi:
Given this some thought and files should implement this `read_state` method

pfebrer96:
This would be very nice to have :)

zerothi:
I'll reopen since it is not stable for real use. It is slow and only provides one way of dealing with it.

zerothi:
> What do I have to do to test changes in fortran sources? Reinstall sisl again?

Yes, it is fastest locally, and fortran codes are quite fast to compile, so I don't think it should be a problem? A little tedious, but ok...

zerothi:
I believe @pfebrer has done everything here :)
I'll close this.

Issue #295, opened 2021-01-27
static code analysis


zerothi:
We should implement pylint for static code analysis

zerothi:
@pfebrer

zerothi:
This is now added in pyproject.toml, but is currently done manually.

PR #388, opened 2021-10-20
enh: optimizing the reading of eigenstates from WFSX file

pfebrer:
Addresses https://github.com/zerothi/sisl/issues/147 by making python keep track of the fortran file unit.

Now if one wants to iterate over all eigenstates the whole file is only read once, in contrast to the previous implementation, which read the file from the beginning at each iteration. This quick code snippet shows the difference in performance for iterating over all eigenstates for the file `spin_unpolarized.bands.WFSX`, which contains 95 k points:

```python
import sisl

wfsx = sisl.get_sile("spin_unpolarized.bands.WFSX")
geom = sisl.geom.diamond()

def iterate(which):
    it = {"new": wfsx.yield_eigenstate, "old": wfsx.old_yield_eigenstate}[which]
    
    for state in it(parent=geom):
        pass
    
%timeit iterate("new")
%timeit iterate("old")
```
```
1.87 ms ± 90.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
35.7 ms ± 97.1 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

I have some doubts:
- I think the spin and k indices are not written on the WFSX file as the old implementation assumes (`spin-loop[k-loop]`). I believe it's the opposite (`k-loop[spin-loop]`). You can test this by trying to read a polarized WFSX file, which will fail on the second eigenstate. I attach one so that you can test it. If this is true, I need to make some changes to the old implementation.
- You already mentioned in https://github.com/zerothi/sisl/issues/147 that the reading is unstable. I think I stumble upon this when reading the polarized file attached here. The binary check raises an error at `ik = 68`. Why is this? :sweat_smile: 

Files (too big to attach): https://drive.google.com/drive/folders/1lgqckfkpq0mZo-3ucBxtXvz5fMKlXCQv?usp=sharing

### DIFF
--- a/sisl/io/siesta/binaries.py
+++ b/sisl/io/siesta/binaries.py
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 from numbers import Integral
-from itertools import product, groupby
+from itertools import product
 from collections import deque
 import numpy as np
 from numpy import pi
@@ -14,10 +14,10 @@ except Exception as e:
     print(e)
     found_module = False
 
-from ..sile import add_sile, SileError
+from ..sile import add_sile, SileError, SileWarning
 from .sile import SileBinSiesta
 from sisl._internal import set_module
-from sisl.messages import warn, SislError
+from sisl.messages import info, warn, SislError
 
 from ._help import *
 import sisl._array as _a
@@ -27,6 +27,7 @@ from sisl.sparse import _ncol_to_indptr
 from sisl.unit.siesta import unit_convert
 from sisl.physics.sparse import SparseOrbitalBZ
 from sisl.physics import Hamiltonian, DensityMatrix, EnergyDensityMatrix
+from sisl.physics import BrillouinZone
 from sisl.physics.overlap import Overlap
 from sisl.physics.electron import EigenstateElectron
 
@@ -41,12 +42,6 @@ __all__ += ['tsgfSileSiesta']
 _Bohr2Ang = unit_convert('Bohr', 'Ang')
 _Ry2eV = unit_convert('Ry', 'eV')
 _eV2Ry = unit_convert('eV', 'Ry')
-
-
-def _bin_check(obj, method, message):
-    ierr = _siesta.io_m.iostat_query()
-    if ierr != 0:
-        raise SileError(f'{obj!s}.{method} {message} (ierr={ierr})')
 
 
 def _toF(array, dtype, scale=None):
@@ -142,9 +137,9 @@ class onlysSileSiesta(SileBinSiesta):
     def read_supercell(self):
         """ Returns a SuperCell object from a TranSiesta file """
         n_s = _siesta.read_tshs_sizes(self.file)[3]
-        _bin_check(self, 'read_supercell', 'could not read sizes.')
+        self._bin_check('read_supercell', 'could not read sizes.')
         arr = _siesta.read_tshs_cell(self.file, n_s)
-        _bin_check(self, 'read_supercell', 'could not read cell.')
+        self._bin_check('read_supercell', 'could not read cell.')
         nsc = np.array(arr[0], np.int32)
         # We have to transpose since the data is read *as-is*
         # The cell in fortran files are (:, A1)
@@ -162,9 +157,9 @@ class onlysSileSiesta(SileBinSiesta):
         sc = self.read_supercell()
 
         na = _siesta.read_tshs_sizes(self.file)[1]
-        _bin_check(self, 'read_geometry', 'could not read sizes.')
+        self._bin_check('read_geometry', 'could not read sizes.')
         arr = _siesta.read_tshs_geom(self.file, na)
-        _bin_check(self, 'read_geometry', 'could not read geometry.')
+        self._bin_check('read_geometry', 'could not read geometry.')
         # see onlysSileSiesta.read_supercell for .T
         xyz = arr[0].T * _Bohr2Ang
         lasto = arr[1]
@@ -216,14 +211,14 @@ class onlysSileSiesta(SileBinSiesta):
 
         # read the sizes used...
         sizes = _siesta.read_tshs_sizes(self.file)
-        _bin_check(self, 'read_overlap', 'could not read sizes.')
+        self._bin_check('read_overlap', 'could not read sizes.')
         # see onlysSileSiesta.read_supercell for .T
         isc = _siesta.read_tshs_cell(self.file, sizes[3])[2].T
-        _bin_check(self, 'read_overlap', 'could not read cell.')
+        self._bin_check('read_overlap', 'could not read cell.')
         no = sizes[2]
         nnz = sizes[4]
         ncol, col, dS = _siesta.read_tshs_s(self.file, no, nnz)
-        _bin_check(self, 'read_overlap', 'could not read overlap matrix.')
+        self._bin_check('read_overlap', 'could not read overlap matrix.')
 
         # Create the Hamiltonian container
         S = Overlap(geom, nnzpr=1)
@@ -255,7 +250,7 @@ class onlysSileSiesta(SileBinSiesta):
         Ef : fermi-level of the system
         """
         Ef = _siesta.read_tshs_ef(self.file) * _Ry2eV
-        _bin_check(self, 'read_fermi_level', 'could not read fermi-level.')
+        self._bin_check('read_fermi_level', 'could not read fermi-level.')
         return Ef
 
 
@@ -270,15 +265,15 @@ class tshsSileSiesta(onlysSileSiesta):
 
         # read the sizes used...
         sizes = _siesta.read_tshs_sizes(self.file)
-        _bin_check(self, 'read_hamiltonian', 'could not read sizes.')
+        self._bin_check('read_hamiltonian', 'could not read sizes.')
         # see onlysSileSiesta.read_supercell for .T
         isc = _siesta.read_tshs_cell(self.file, sizes[3])[2].T
-        _bin_check(self, 'read_hamiltonian', 'could not read cell.')
+        self._bin_check('read_hamiltonian', 'could not read cell.')
         spin = sizes[0]
         no = sizes[2]
         nnz = sizes[4]
         ncol, col, dH, dS = _siesta.read_tshs_hs(self.file, spin, no, nnz)
-        _bin_check(self, 'read_hamiltonian', 'could not read Hamiltonian and overlap matrix.')
+        self._bin_check('read_hamiltonian', 'could not read Hamiltonian and overlap matrix.')
 
         # Check whether it is an orthogonal basis set
         orthogonal = np.abs(dS).sum() == geom.no
@@ -362,7 +357,7 @@ class tshsSileSiesta(onlysSileSiesta):
                               csr.ncol, csr.col + 1,
                               _toF(h, np.float64, _eV2Ry), _toF(s, np.float64),
                               isc)
-        _bin_check(self, 'write_hamiltonian', 'could not write Hamiltonian and overlap matrix.')
+        self._bin_check('write_hamiltonian', 'could not write Hamiltonian and overlap matrix.')
 
 
 @set_module("sisl.io.siesta")
@@ -374,10 +369,10 @@ class dmSileSiesta(SileBinSiesta):
 
         # Now read the sizes used...
         spin, no, nsc, nnz = _siesta.read_dm_sizes(self.file)
-        _bin_check(self, 'read_density_matrix', 'could not read density matrix sizes.')
+        self._bin_check('read_density_matrix', 'could not read density matrix sizes.')
 
         ncol, col, dDM = _siesta.read_dm(self.file, spin, no, nsc, nnz)
-        _bin_check(self, 'read_density_matrix', 'could not read density matrix.')
+        self._bin_check('read_density_matrix', 'could not read density matrix.')
 
         # Try and immediately attach a geometry
         geom = kwargs.get('geometry', kwargs.get('geom', None))
@@ -448,7 +443,7 @@ class dmSileSiesta(SileBinSiesta):
         nsc = DM.geometry.sc.nsc.astype(np.int32)
 
         _siesta.write_dm(self.file, nsc, csr.ncol, csr.col + 1, _toF(dm, np.float64))
-        _bin_check(self, 'write_density_matrix', 'could not write density matrix.')
+        self._bin_check('write_density_matrix', 'could not write density matrix.')
 
 
 @set_module("sisl.io.siesta")
@@ -460,9 +455,9 @@ class tsdeSileSiesta(dmSileSiesta):
 
         # Now read the sizes used...
         spin, no, nsc, nnz = _siesta.read_tsde_sizes(self.file)
-        _bin_check(self, 'read_energy_density_matrix', 'could not read energy density matrix sizes.')
+        self._bin_check('read_energy_density_matrix', 'could not read energy density matrix sizes.')
         ncol, col, dEDM = _siesta.read_tsde_edm(self.file, spin, no, nsc, nnz)
-        _bin_check(self, 'read_energy_density_matrix', 'could not read energy density matrix.')
+        self._bin_check('read_energy_density_matrix', 'could not read energy density matrix.')
 
         # Try and immediately attach a geometry
         geom = kwargs.get('geometry', kwargs.get('geom', None))
@@ -515,7 +510,7 @@ class tsdeSileSiesta(dmSileSiesta):
         Ef : fermi-level of the system
         """
         Ef = _siesta.read_tsde_ef(self.file) * _Ry2eV
-        _bin_check(self, 'read_fermi_level', 'could not read fermi-level.')
+        self._bin_check('read_fermi_level', 'could not read fermi-level.')
         return Ef
 
     def write_density_matrices(self, DM, EDM, Ef=0., **kwargs):
@@ -567,7 +562,7 @@ class tsdeSileSiesta(dmSileSiesta):
         _siesta.write_tsde_dm_edm(self.file, nsc, DMcsr.ncol, DMcsr.col + 1,
                                   _toF(dm, np.float64),
                                   _toF(edm, np.float64, _eV2Ry), Ef * _eV2Ry)
-        _bin_check(self, 'write_density_matrices', 'could not write DM + EDM matrices.')
+        self._bin_check('write_density_matrices', 'could not write DM + EDM matrices.')
 
 
 @set_module("sisl.io.siesta")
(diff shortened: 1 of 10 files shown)


zerothi:
Thanks for your comments. Indeed this shows some inconsistencies in the entire SiestaBin suite!!! :)

I agree this might be fitting in the binary sile for siesta, for coherence! Thanks for noticing.

Regarding the `read_eigenstate`, this is probably my mistake!!! ;)
The idea should be that all methods have a similar interface, so it should definitely be `k` and not `ik`. So this needs changing. I just recall I did this *quickly* because somebody wanted to use the wavefunctions. So I think it is ok to break API to make it stable.

pfebrer:
> Regarding the read_eigenstate, this is probably my mistake!!! ;)

No, no. It wasn't there. I added it and then removed it. But couldn't it be useful in some cases to be able to get an eigenstate by its index? It would use the old subroutines and therefore avoid going back and forth between fortran and python.

pfebrer:
I added the possibility to read the information (not values) of all k points at once. It is useful for debugging but could be also useful for other things (you can get a `BandStructure` / `BrillouinZone` object easily for example, or find the best match to your desired k value).

Since all information is read in one call this is significantly faster than iterating over the eigenstates:

```python
import sisl
import numpy as np

wfsx = sisl.get_sile("spin_unpolarized.bands.WFSX")
geom = sisl.geom.diamond()
def all_fortran():
    ks = wfsx.read_file_info(parent=geom)

def python_loop():
    for state in wfsx.yield_eigenstate(parent=geom):
        state.info
        
%timeit all_fortran()
%timeit python_loop()
```
```
335 µs ± 9.52 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
1.99 ms ± 112 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
Here is an example of what is returned by `read_file_info`:
```
{'k': array([[0.     , 0.     , 0.     ],
        [0.     , 0.0714 , 0.0714 ],
        [0.     , 0.1428 , 0.1428 ],
        [0.     , 0.2142 , 0.2142 ],
        [0.     , 0.2856 , 0.2856 ],
        [0.     , 0.357  , 0.357  ],
        [0.     , 0.4284 , 0.4284 ],
        [0.     , 0.4998 , 0.4998 ],
        [0.     , 0.5712 , 0.5712 ],
        [0.     , 0.6426 , 0.6426 ],
        [0.     , 0.714  , 0.714  ],
        [0.     , 0.7854 , 0.7854 ],
        [0.     , 0.8568 , 0.8568 ],
        [0.     , 0.9282 , 0.9282 ],
        [0.     , 0.9996 , 0.9996 ],
        [0.     , 1.071  , 1.071  ],
        [0.     , 1.1424 , 1.1424 ],
        [0.     , 1.2138 , 1.2138 ],
        [0.     , 1.2852 , 1.2852 ],
        [0.     , 1.3566 , 1.3566 ],
        [0.     , 1.428  , 1.428  ],
        [0.     , 1.4994 , 1.4994 ],
        [0.     , 1.5708 , 1.5708 ],
        [0.     , 1.6422 , 1.6422 ],
        [0.     , 1.7136 , 1.7136 ],
        [0.     , 1.785  , 1.785  ],
        [0.08925, 1.785  , 1.87425],
        [0.1785 , 1.785  , 1.9635 ],
        [0.26775, 1.785  , 2.05275],
        [0.357  , 1.785  , 2.142  ],
        [0.44625, 1.785  , 2.23125],
        [0.5355 , 1.785  , 2.3205 ],
        [0.62475, 1.785  , 2.40975],
        [0.714  , 1.785  , 2.499  ],
        [0.80325, 1.785  , 2.58825],
        [0.8925 , 1.785  , 2.6775 ],
        [0.952  , 1.785  , 2.618  ],
        [1.0115 , 1.785  , 2.5585 ],
        [1.071  , 1.785  , 2.499  ],
        [1.1305 , 1.785  , 2.4395 ],
        [1.19   , 1.785  , 2.38   ],
        [1.2495 , 1.785  , 2.3205 ],
        [1.309  , 1.785  , 2.261  ],
        [1.3685 , 1.785  , 2.2015 ],
        [1.428  , 1.785  , 2.142  ],
        [1.4875 , 1.785  , 2.0825 ],
        [1.547  , 1.785  , 2.023  ],
        [1.6065 , 1.785  , 1.9635 ],
        [1.666  , 1.785  , 1.904  ],
        [1.7255 , 1.785  , 1.8445 ],
        [1.785  , 1.785  , 1.785  ],
        [1.69575, 1.69575, 1.69575],
        [1.6065 , 1.6065 , 1.6065 ],
        [1.51725, 1.51725, 1.51725],
        [1.428  , 1.428  , 1.428  ],
        [1.33875, 1.33875, 1.33875],
        [1.2495 , 1.2495 , 1.2495 ],
        [1.16025, 1.16025, 1.16025],
        [1.071  , 1.071  , 1.071  ],
        [0.98175, 0.98175, 0.98175],
        [0.8925 , 0.8925 , 0.8925 ],
        [0.80325, 0.80325, 0.80325],
        [0.714  , 0.714  , 0.714  ],
        [0.62475, 0.62475, 0.62475],
        [0.5355 , 0.5355 , 0.5355 ],
        [0.44625, 0.44625, 0.44625],
        [0.357  , 0.357  , 0.357  ],
        [0.26775, 0.26775, 0.26775],
        [0.1785 , 0.1785 , 0.1785 ],
        [0.08925, 0.08925, 0.08925],
        [0.     , 0.     , 0.     ],
        [0.1071 , 0.1071 , 0.1071 ],
        [0.2142 , 0.2142 , 0.2142 ],
        [0.3213 , 0.3213 , 0.3213 ],
        [0.4284 , 0.4284 , 0.4284 ],
        [0.5355 , 0.5355 , 0.5355 ],
        [0.6426 , 0.6426 , 0.6426 ],
        [0.7497 , 0.7497 , 0.7497 ],
        [0.8568 , 0.8568 , 0.8568 ],
        [0.9639 , 0.9639 , 0.9639 ],
        [1.071  , 1.071  , 1.071  ],
        [1.1781 , 1.1781 , 1.1781 ],
        [1.2852 , 1.2852 , 1.2852 ],
        [1.3923 , 1.3923 , 1.3923 ],
        [1.4994 , 1.4994 , 1.4994 ],
        [1.6065 , 1.6065 , 1.6065 ],
        [1.7136 , 1.7136 , 1.7136 ],
        [1.8207 , 1.8207 , 1.8207 ],
        [1.9278 , 1.9278 , 1.9278 ],
        [2.0349 , 2.0349 , 2.0349 ],
        [2.142  , 2.142  , 2.142  ],
        [2.2491 , 2.2491 , 2.2491 ],
        [2.3562 , 2.3562 , 2.3562 ],
        [2.4633 , 2.4633 , 2.4633 ],
        [2.5704 , 2.5704 , 2.5704 ],
        [2.6775 , 2.6775 , 2.6775 ]]),
 'weight': array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
        1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
        1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
        1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
        1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
        1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]),
 'nwf': array([[8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
         8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
         8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
         8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8,
         8, 8, 8, 8, 8, 8, 8, 8]], dtype=int32)}
```

On the 505 MB file that I attached, reading the info takes 140 ms. So I'd say this is very usable in practice even for huge WFSX files.

pfebrer:
Ok, all this is because I want to allow using WFSX files for plotting wavefunctions, (fat)bands and PDOS.

For the wavefunctions and (fat)bands I will just add entry points in the respective plots. For the PDOS, do you think it would be useful to provide a method outside the plotting suite as well? E.g. : `wfsxSileSiesta.read_PDOS()`. 

Or maybe in a more abstract way provide apply dispatchers that would work similarly as the `BrillouinZone` applies? (I like this idea)

pfebrer:
Can I generate a `BandStructure` object from a list of k points?

pfebrer:
Also, one question for PDOS. If I want to generate the PDOS from the `WFSX` file, I need to read the overlap matrix as well, right?

zerothi:
> Can I generate a `BandStructure` object from a list of k points? I want to be able to plot bands and fatbands just from the `WFSX` file, but I am missing some points sweat_smile

Yes, I can't see why not. Except that the `bandstructure` object requires a parent with cell information.

> Also, one question for PDOS. If I want to generate the PDOS from the `WFSX` file, I need to read the overlap matrix as well, right?

Yes, you need the overlap matrix as well.

pfebrer:
> Yes, I can't see why not. Except that the bandstructure object requires a parent with cell information.

So just pass the list of points to the `point` argument and set `division` to 1? It would be nice to have something to detect corners for this case, wouldn't it? :smile:

zerothi:
> > Yes, I can't see why not. Except that the bandstructure object requires a parent with cell information.
> 
> So just pass the list of points to the `point` argument and set `division` to 1? It would be nice to have something to detect corners for this case, wouldn't it? smile

Hmm, I see. It may not be so viable after all... Is there a particular reason for not just using the `BrillouinZone` object directly? I.e. what methods from `BandStructure` do you need?

pfebrer:
I think I just needed `BandStructure.lineark` to plot the bands proportionally. I mean that dx in the bands plot is a constant amount.

And `BandStructure.lineartick` to get the ticks, but this probably not needed since there are no ticks in the WFSX file anyway.

zerothi:
> I think I just needed `BandStructure.lineark` to plot the bands proportionally. I mean that dx in the bands plot is a constant amount.
> 
> And `BandStructure.lineartick` to get the ticks, but this is probably not needed since there are no ticks in the WFSX file anyway.

Latest commit has a `linspace_bz` which converts BZ object k-points into linear distances (accepts any BZ-object).

pfebrer:
Nice, thank you!

While implementing PDOS calculation from WFSX I realized that it was a bit slower than I thought. Could you take a look at the last commit? Is there some reason to not do everything in one numpy call as I did now?

I tested both approaches:

```python
import numpy as np
conj = np.conj

def PDOS(E, eig, state, S, distribution):
    # Get the weights for all eigenvalues and all energies
    eigenvalue_weights = distribution(E - eig.reshape(-1, 1).repeat(len(E), axis=1))

    overlaps = conj(state.T) * S.dot(state.T)
    return np.dot(overlaps.real, eigenvalue_weights)

def PDOS_old(E, eig, state, S, distribution):
    PDOS = (conj(state[0]) * S.dot(state[0])).real.reshape(-1, 1) \
                * distribution(E - eig[0]).reshape(1, -1)
    for i in range(1, len(eig)):
        PDOS[:, :] += (conj(state[i]) * S.dot(state[i])).real.reshape(-1, 1) \
                    * distribution(E - eig[i]).reshape(1, -1)
    
    return PDOS

# Timing code
eigenstate = ...

E = np.array(np.linspace(0, 3, 100))
Sk = eigenstate.Sk()
distribution = sisl.get_distribution("gaussian")

%timeit PDOS_py(E, eigenstate.eig, eigenstate.state, Sk, distribution)
%timeit PDOS_old(E, eigenstate.eig, eigenstate.state, Sk, distribution)
```

on a very small eigenstate (only 8 wfs) it gives me these timings:

```
30.1 µs ± 1.04 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
150 µs ± 3.3 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
And I guess this difference will get larger with a higher number of wavefunctions.

Are you worried about memory usage here?

zerothi:
Could we perhaps limit this PR to the reading of the eigenstates, you are adding too many things that ought to be in a separate MR ;)

pfebrer:
Hmm ok, but I wanted to make sure that the API is useful for creating the plots, which was my initial motivation.

Maybe when I finish with the plots I can just keep the changes in the wfsx sile if you want.

zerothi:
> Hmm ok, but I wanted to make sure that the API is useful for creating the plots, which was my initial motivation.
> 
> Maybe when I finish with the plots I can just keep the changes in the wfsx sile if you want.

It is fine for now, but generally it would be better to create a new branch off of this one and test it there. Smaller merge requests that are self-contained is much better and easier to manage.

pfebrer:
This is ready for review! After the review, all that's left is to add tests.

I squashed everything into two commits so that you can check it easier: one for WFSX reading and the other to incorporate it into the plots.

sisl/io/siesta/binaries.py
```diff
@@ -14,10 +14,10 @@
     print(e)
     found_module = False
 
-from ..sile import add_sile, SileError
+from ..sile import add_sile, SileError, SileWarning
```
pfebrer, line 17: By the way, I caught this one (`SileWarning` was used in a method but not imported) because in VS code I have the Pylance extension which does static code analysis on the files you open. It reinforces the need for #295 to effortlessly make the code cleaner (remove unused imports) and less buggy :)

zerothi: Nice catch, yeah at some point I had an analyser, however it annoyingly didn't catch some things, and too many of others. Definitely we should have a static analyzer!

pfebrer: `flake8` seems to be a quite good static code analyzer.

You can test it very quickly. To install:

```
python -m pip install flake8
```

Then to check for undeclared variables (error `F821`):
```
flake8 --select F821 path/to/sisl
```

There are quite a few :)

zerothi: damn ;)

zerothi: FYI, I already have configure flake8, but just never ran it periodically... :(

zerothi: I fixed everything except a `display` call in your viz modules:
```
sisl/viz/plot.py:1314:9: F821 undefined name 'display'
```
I have tried to fix this in 608616c6f1

pfebrer: Yes, that works! :+1:

zerothi:
I have pushed a new commit which you can check, also added a test for this.

zerothi:
> Ok, all this is because I want to allow using WFSX files for plotting wavefunctions, (fat)bands and PDOS.
> 
> For the wavefunctions and (fat)bands I will just add entry points in the respective plots. For the PDOS, do you think it would be useful to provide a method outside the plotting suite as well? E.g. : `wfsxSileSiesta.read_PDOS()`.
> 
> Or maybe in a more abstract way provide `apply` dispatchers that would work similarly as the `BrillouinZone` applies? (I like this idea)

No, I don't think we should plaster the file with this, it can be done with `f = wfsxSileSiesta(parent=Hamiltonian) ; for state in f.yield_eigenstate(): state.PDOS(...)`

I clearly made a mistake in the Hamiltonian object to add DOS and PDOS methods there. Better to have these decoupled.

zerothi:
Note that now each method does _not_ have the `parent` argument. Instead the creation of the object has this interface:

```python
wfsxSileSiesta(..., parent=?, geometry=?, sc=?)
```
where the `sc` is automatically fetched from `parent` if supplied. This makes calls more strict and easier. :)

pfebrer:
Great! I'll submit another PR in a moment then because in the plots I was using `yield_eigenstate(parent=...)` :)

pfebrer:
Also, could you add spin-polarized and noncolinear spin versions of the `.bands.WFSX` file if you have time? I would like to add tests for all the spin possibilities in plots

zerothi:
> Great! I'll submit another PR in a moment then because in the plots I was using `yield_eigenstate(parent=...)` :)

Great, I have a helper script `pep8.sh` that I think you should run before MR's, quite often it catches indentation problems in some of the plotting regions.

pfebrer:
> Great, I have a helper script pep8.sh that I think you should run before MR's, quite often it catches indentation problems in some of the plotting regions.

Yes, I'll do that

pfebrer:
> Generally one should limit the return types as dictionaries
> as it imposes naming conventions on certain variables
> that may not be optimal for the end user. In these cases
> namedtuple returns are better since users can do whatever they
> wish with them.

I remember why I returned dictionaries. Dictionaries allow returning additional items in the future without breaking backwards compatibility. If you return a tuple, you are basically stuck with the choice you made in the first place. What if there's an extra size that you think could be useful to return in the future? If you make the tuple larger, working code will get errors when updating the sisl version, which can be annoying.

zerothi:
> I remember why I returned dictionaries. Dictionaries allow returning additional items in the future without breaking backwards compatibility. If you return a tuple, you are basically stuck with the choice you made in the first place. What if there's an extra size that you think could be useful to return in the future? If you make the tuple larger, working code will get errors when updating the sisl version, which can be annoying.

Yes, but dictionaries is really annoying to work with when dealing with *simple data*. I agree that the flexibility is good. But I am thinking about the end users who are needing to write code. And using dictionaries is in my opinion a needless hurdle. It obfuscates code and complicates things.
```python
for out in ...iter:
    no_u = out["no_u"]
    data = out["data"]
```
could be done with
```python
for no_u, data in ...iter:
    ...
```
it is clear and meaningful. Also when you see other big, succesfull code projects (numpy, scipy). They only ever return dictionaries when you know you have variable data associated. And here we must have some kind of consistency.

If we later wishes to return something else, then this could be solved by adding flags to the sile itself. `wfsxSile.VERSION` or something like that.
Secondly, how often do you think this code will change? I don't think it will actually change a lot during the next several years ;)

zerothi:
Actually, this is why I would prefer namedtuple returns, it has a direct `_todict` method allowing you to use it directly as a dict.

pfebrer:
> Secondly, how often do you think this code will change? I don't think it will actually change a lot during the next several years ;)

No, but it may change 10 years from now and then break codes whose authors may not be maintaining anymore. 

> Actually, this is why I would prefer namedtuple returns, it has a direct _todict method allowing you to use it directly as a dict.

I think this could be the best option.

zerothi:
> > Secondly, how often do you think this code will change? I don't think it will actually change a lot during the next several years ;)
> 
> No, but it may change 10 years from now and then break codes whose authors may not be maintaining anymore.
> 
> > Actually, this is why I would prefer namedtuple returns, it has a direct _todict method allowing you to use it directly as a dict.
> 
> I think this could be the best option.

You can see that that is what I have done in this PR

pfebrer:
Aah ok yes, I was looking at `read_info`, not `read_sizes` :+1:

PR #397, opened 2021-11-08
tests: avoid too many matplotlib figures

### LINKED ISSUE #396: sisl.viz tests open up *tons* of matplotlib figures
**Describe the issue**

```shell
viz/plots/tests/test_grid.py::TestGridPlot::test_representation[matplotlib-complex_grid-1-imag]
  /opt/gnu/11.2.0/python/packages/3.9.7/sisl-dev/0/lib/python3.9/site-packages/sisl/viz/backends/matplotlib/backend.py:29: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
    self.figure, self.axes = plt.subplots()

```

perhaps it would be a good idea to close them using some kind of fixture. If more tests are added I fear it will consume too much. In any case it is a good idea to clean. :)

pfebrer:
Fixes #396

### DIFF
--- a/sisl/viz/backends/matplotlib/backend.py
+++ b/sisl/viz/backends/matplotlib/backend.py
@@ -26,7 +26,7 @@ class MatplotlibBackend(Backend):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.figure, self.axes = plt.subplots()
+        self.figure, self.axes = self._init_figure()
         self._init_axes()
 
     def draw_on(self, axes, axes_indices=None):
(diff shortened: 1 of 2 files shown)


pfebrer:
I also moved figure initialization to a separate method so that anyone that wants to subclass `MatplotlibBackend` can choose how to initialize the figure. (e.g. not use `pyplot`)

zerothi:
Ok, I think this is acceptable :)

Thanks! I'll try it out immediately.

PR #402, opened 2021-11-25
Use PeriodicTable to convert atomic labels to atomic numbers in xsf file

ahkole:
The XSF file format allows the use of atomic labels instead of atomic numbers. The xsfSile has been modified to use PeriodicTable().Z() to convert these labels to the corresponding atomic numbers if these labels are used in an xsf file.


zerothi (approved):
Great, I'll rewrite it slightly to not create the object on every line.
Thanks for extending the functionality.

PR #403, opened 2021-11-25
Fixed construction of non-orthogonal (2-atom) hcp cell

ahkole:
There was a minor error in the construction of the non-orthogonal hcp cell. The distance (in the xy-plane) of the second atom in the cell to the origin is a/sqrt(3) not a/sqrt(2). Because of this error the second atom wasn't exactly in the center of the triangle spanned by the atoms of the bottom layer (as it should be). This has been changed in this version to produce the correct hcp cell.


zerothi (approved):
Thanks a lot for discovering this! :)

PR #404, opened 2021-11-30
Enabled reading zipped *.fdf.gz files

tfrederiksen:
I was trying to use sisl to read some zipped SIESTA fdf files and found that this was not working for two reasons:
* reading files *via* `gzip` should be using `mode=rt` (text mode) to handle comparisons like `line.startswith('#')` (with `mode=r` the strings are of `binary` type),
* include-files like `%include OTHER.fdf` means that sisl should also check for `OTHER.fdf.gz`. 

This branch should fix the above.

### DIFF
--- a/sisl/io/siesta/fdf.py
+++ b/sisl/io/siesta/fdf.py
@@ -7,6 +7,7 @@ import numpy as np
 import scipy as sp
 from os.path import isfile
 import itertools as itools
+import gzip
 
 from ..sile import add_sile, get_sile_class, sile_fh_open, sile_raise_write, SileError
 from .sile import SileSiesta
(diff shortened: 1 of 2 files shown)


zerothi (approved):
Thanks for finding this!

zerothi:
@tfrederiksen I will amend your correction with something like this which should handle corner cases were users are reading binary files

```diff
diff --git a/sisl/io/sile.py b/sisl/io/sile.py
index 6790b8aa..2156265a 100644
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -555,7 +555,12 @@ class Sile(BaseSile):
 
     def _open(self):
         if self.file.suffix == ".gz":
-            self.fh = gzip.open(str(self.file), mode='rt')
+            if self._mode == 'r':
+                # assume the file is a text file and open in text-mode
+                self.fh = gzip.open(str(self.file), mode='rt')
+            else:
+                # assume this is opening in binary or write mode
+                self.fh = gzip.open(str(self.file), mode=self._mode)
         else:
             self.fh = self.file.open(self._mode)
         self._line = 0
```

tfrederiksen:
OK. Maybe it would also be a good idea to include some tests with the `.gz` format?

zerothi:
Agreed!

zerothi:
Done, added a test, but still not a test for binary stuff. Currently not having any files read in binary mode directly from python.

PR #412, opened 2021-12-09
Enabled plotting from WFSX files

pfebrer:
Basically, this PR enables using the WFSX files for visualization. These visualizations can now be created from WFSX files:

- **Bands** (No extra requirements)
- **Fatbands** (Needs geometry with basis and Hamiltonian to use the overlap matrix). Equivalent to the SIESTA Util `fat`.
- **PDOS** (Needs geometry with basis and Hamiltonian to use the overlap matrix). This one is I guess equivalent to the SIESTA Util `mprop`? I've never used it, could you confirm?
- **Wavefunction** (Needs geometry with basis). Equivalent to how `denchar` plots wavefunctions.


sisl/viz/plots/bands.py
```diff
@@ -433,41 +434,52 @@ def _read_from_wfsx(self, root_fdf, wfsx_file, extra_vars=()):
         self.geometry = fdf.read_geometry(output=True)
 
         # Get the wfsx file
-        wfsx_sile = self.get_sile(wfsx_file or "wfsx_file")
+        wfsx_sile = self.get_sile(wfsx_file or "wfsx_file", parent=self.geometry)
```
zerothi, line 437: It is not clear to me in this PR whether PDOS or `eigenstate` will be called with parent being the geometry? If so you'll get wrong values...

This goes for both fatbands and PDOS.

sisl/viz/plots/grid.py
```diff
@@ -1525,17 +1530,22 @@ def _set_data(self, i, geometry, grid, k, grid_prec, nsc):
         if self.geometry is None:
             raise ValueError('No geometry was provided and we need it the basis orbitals to build the wavefunctions from the coefficients!')
 
+        # Get the spin class for which the eigenstate was calculated.
+        spin = sisl.Spin()
+        if self.eigenstate.parent is not None:
```
zerothi, line 1535: if `eigenstate.parent is None` then basically nothing can be used here?

zerothi (changes requested):
Just some minor things related to how the parent thing works. I.e. the overlap matrix is required for all of these things.

sisl/viz/plots/bands.py
pfebrer: I didn't know `norm2` used the overlap matrix :scream: Indeed I was not getting the hamiltonian for fatbands. For the PDOS I already set the hamiltonian as the parent. Thanks for noticing!

sisl/viz/plots/grid.py
pfebrer: Hmm do you mean because there is no geometry? The geometry is determined in the lines above. It can be passed explicitly as a setting (the `geometry` setting), just in case we could not automatically find it. After going through the multiple ways the geometry can be retrieved, if there was still no geometry, an error is raised (line 1531). 

So at this point we already know we have a geometry, we are just trying to see if we know what the spin class of the calculation was.

sisl/viz/plots/bands.py
zerothi: Everytime you are doing something with the eigenvectors you generally need the overlap matrix due to the eigenvectors not being orthogonal. It is only for the wavefunction where the overlap is not needed (it is implicitly calculated during the construction of the real space grid).

sisl/viz/plots/grid.py
zerothi: Yeah ok, I would just try never to use a `geometry` as a parent in the eigenstate approach, *only* for wavefunctions this would be ok.

sisl/viz/_plotables.py
```diff
@@ -12,11 +12,10 @@
 class ClassPlotHandler(ClassDispatcher):
     """Handles all plotting possibilities for a class"""
 
-    # TODO fix handling with default type_dispatcher=None
```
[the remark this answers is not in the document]
zerothi: Your fix LGTM!

sisl/viz/plots/bands.py
pfebrer: Fixed, now fatbands from `WFSX` files require the `Hamiltonian` to be set as parent. And if it doesn't find one, then it fails.

sisl/viz/plots/grid.py
zerothi: For bands it is ok. :)

pfebrer:
This is ready!

zerothi (approved):

PR #415, opened 2021-12-14
Support for discontinous bands plots

pfebrer:
Adds support for the recently added feature in `BandStructure`.

Discontinous band structures are also supported for spin texture
and fatbands plots.

As an example of discontinous fatbands:

```python
import sisl

gr = sisl.geom.graphene()
H = sisl.Hamiltonian(gr)
H[0,0] = -2

bz = sisl.BandStructure(H, [[0, 0, 0], [2/3, 1/3, 0], None, [2/3, 1/3, 0], [1/2, 0, 0]], 30, ["Gamma", "M","M", "K"])

bands_plot = bz.plot.fatbands()
bands_plot.split_groups(on="atoms", name="Atom $atoms")
```

newplot - 2021-12-14T015844 155

### DIFF
--- a/sisl/viz/backends/plotly/_plots/fatbands.py
+++ b/sisl/viz/backends/plotly/_plots/fatbands.py
@@ -1,6 +1,8 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
+import numpy as np
+
 from ....plots import FatbandsPlot
 from .bands import PlotlyBandsBackend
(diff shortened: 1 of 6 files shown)


zerothi (approved):

zerothi:
It seems like you are doing manual jumps, but I don't understand whether this is necessary? I.e. in matplotlib a `np.nan` will yield a discontinuity, so is it necessary with all this boiler plate code, rather than just `insert_jump`?

pfebrer:
The only thing that is divided in chunks are fatbands in `plotly` because they do not handle `np.nan` in the desired way.

The rest (bands, spin texture and fatbands in `matplotlib`) is passed directly with the NaNs because as you say they are already understood as discontinuities :)

PR #421, opened 2022-01-12
Function to build heteroribbons

### LINKED ISSUE #420: Function to produce armchair heteroribbons.
I was tired of having to think about how to build each heteroribbon and I've written a function to produce only **valid heteroribbons** (with no dangling bonds) specifying the **width, units and shift of each section**. Initially, I thought that might be a useful thing to put in `sisl_toolbox`, but seeing that these structures are getting popular I thought that maybe it's worth it to add it to `sisl.geom`.

<details>
<summary>Here's the (pretty long) code.</summary>

```python
import sisl
import math
import numpy as np

def graphene_heteroribbon(Ws, units, shifts=None, **kwargs):
    """Builds a graphene nanoribbon with multiple widths.
    
    The function makes sure that only valid ribbons, that is
    ribbons with no dangling bonds, are being produced.
    Explanatory errors are raised if that's not possible. Only the 
    connection at the periodic boundary remains unchecked.
    
    It only works for armchair ribbons for now.
    
    Parameters
    ----------
    Ws: array-like of int
        The width of each section of the heteroribbon.
    units: array-like of int
        The number of units of each section. Note that a "unit" is
        not a unit cell, but half of it. I.e. a zigzag string.
        
        There are two situations in which you can choose the starting
        configuration of your border ("open", "closed"). At the beggining of
        the heteroribbon and on an even section after an odd one. In those
        cases, an integer imaginary number here will indicate that you want
        the open configuration (e.g. 2j indicates that you want 2 units, starting
        with an open configuration). If you pass an imaginary number on an invalid
        situation, an error will be raised.
    shifts: array-like of int, optional
        How each section must be shifted from the previous one.
        
        The exact meaning of the value depends on the sections that are
        being joined so that only valid ribbons are produced:
            - If both ribbons are odd, ribbons are aligned on the center
            and then shifted:
              ^ (shift) zig zag atoms if the incoming section is thinner than the
                previous one.
              ^ (2 * shift) zig zag atoms if they differ in a multiple of 4 atoms
                and both configurations are closed or if they differ in (2 + a
                multiple of 4 atoms) and configurations are different (one open
                and one closed.)
              ^ (1 + 2 * shift) zig zag atoms otherwise.
            - If both ribbons are even, they are aligned by the open end of the last
            section, not the incoming one, and then they are shifted:
              ^ (0) zig zag atoms if the shift is 0. In this case both sections are 
              forced to have the same edge open.
              ^ (-1 + 2 * abs(shift)) zig zag atoms in the appropiate direction if abs(shift) is
              bigger than 1. Notice that you can only shift the open edges toward the center.
            - If the ribbons have different parities, they are aligned on the open edge
            of the even ribbon and then shifted:
              ^ (0) zig zag atoms if shift is 0 in the particular cases that (i) the incoming
              odd section is thinner than the last even section or (ii) the last section, 
              being odd, has a border in the open configuration. Only in those cases does it make
              sense to exactly match the edges.
              ^ (bias + 2 * abs(shift)) zig zag atoms in the appropiate direction, where `bias` is
              either -1 or 1 depending on the exact situation and always ensures that the minimum
              shift is always `abs(shift) == 1`.
              
            
        
        If not provided, it defaults to an array full of zeros.
    **kwargs
        Keyword arguments are directly passed to `sisl.geom.graphene_nanoribbon`.
              
    Returns
    ----------
    sisl.Geometry:
        The final structure of the heteroribbon.
    """
    if kwargs.get('kind', 'armchair') == "zigzag":
        raise ValueError("Zigzag heteroribbons are not implemented yet.")
    
    # Initialize all variables that are going to be updated through the
    # iterations
    geom = None
    last_addition = None
    last_open = False
    last_W = -1
    
    # Get the distance of an atom shift.
    atom_shift = kwargs.get('bond', 1.42) * math.sin(math.radians(60))
    
    # If shifts were not provided it means the user wants no shifts,
    # so just initialize the array to all zeros.
    if shifts is None:
        shifts = np.zeros(len(Ws))
    
    # Helper function to manipulate ribbons.
    def _open_start(geometry):
        """Changes the border used for a ribbon.
        
        It does so by shifting half unit cell. This must be done before any
        tiling of the geometry.
        """
        geometry = geometry.move(geometry.cell[0] / 2)
        geometry.xyz = (geometry.fxyz % 1).dot(geometry.cell)
        return geometry
    
    # Loop through all the sections that compose the heteroribbon.
    for i, (W, W_units, shift) in enumerate(zip(Ws, units, shifts)):
        # Check that we are not joining an open odd ribbon with
        # a smaller ribbon, since no matter what the shift is there will
        # always be dangling bonds.
        if last_W % 2 == 1 and W < last_W:
            assert not last_open, (f"At the interface between {last_W} and {W}, {last_W} has an open end."
                    "If the wider ribbon is odd, it must always have a closed end.")
        
        # Get the unit cell of the ribbon that we are going to add to the
        # heteroribbon.
        new_addition = sisl.geom.graphene_nanoribbon(W, **kwargs)
        # Unit cells returned by sisl are always in the "closed" configuration.
        # In even width ribbons, there will always be an open edge and a closed
        # edge. We use "open" in the situation where the top edge is open.
        open_start = False
        
        # If the number of units has been provided as an imaginary number, this means that
        # the user wants the open configuration of the ribbon.
        if W_units.imag > 0:
            # However this is only possible in very specific cases, since in all other
            # cases the border of the incoming section is fully determined by the border
            # of the previous section and the shift.
            if i == 0 or W % 2 == 0 and last_W % 2 == 1:
                new_addition = _open_start(new_addition)
                open_start = not open_start
                W_units = int(W_units.imag)
            else:
                raise ValueError("This section does not allow choosing the configuration, but an imaginary",
                    f" number {W_units} was passed as the number of units."
                )
        
        # If this is not the first ribbon section in the heteroribbon, we are going to
        # calculate the shifts.
        if not i == 0:
            # Get the difference in width between the previous and this ribbon section
            W_diff = W - last_W
            # And also the mod(4) because there are certain differences if the width differs
            # on 1, 2, 3 or 4 atoms. After that, the cycle just repeats (e.g. 5 == 1, etc).
            diff_mod = W_diff % 4
            
            # Now, we need to calculate the offset that we have to apply to the incoming
            # section depending on several factors.
            if diff_mod % 2 == 0 and W % 2 == 1:
                # Both sections are odd
                
                # In this case, ribbons are aligned based on their centers.
                offset = (last_addition.center() - new_addition.center())[1]
                
                different_configs = last_open != open_start
                
                if W < last_W:
                    # The incoming section is thinner than the last one. Note that at this point
                    # we are sure that the last section has a closed border, otherwise we
                    # would have raised an error. At this point, centers can differ by any
                    # integer number of atoms, but we might need to open the start depending
                    # on the shift so that the sections match. 
                    if shift % 2 == 0 and diff_mod == 2 or shift % 2 == 1 and diff_mod == 0:
                        # We must open the start
                        new_addition = _open_start(new_addition)
                        open_start = not open_start
                    
                    # Shift limits are a bit complicated and are different for even and odd shifts.
                    # This is because a closed incoming section can shift until there is no connection
                    # between ribbons, while an open one needs to stop before its edge goes outside
                    # the previous section.
                    shift_lims = {
                        "closed": last_W // 2 + W // 2 - 2,
                        "open": last_W // 2 - W // 2 - 1
                    }
                    
                    shift_pars = {
                        lim % 2: lim for k, lim in shift_lims.items()
                    }
                    
                    assert (shift % 2 == 0 and abs(shift) <= shift_pars[0])\
                        or (shift % 2 == 1 and abs(shift) <= shift_pars[1]),\
                        (f"Shift must be an even number between {-shift_pars[0]} and"
                         f" {shift_pars[0]} or an odd number between {-shift_pars[1]}"
                         f" and {shift_pars[1]}")
                    
                    offset += shift * atom_shift
                else:
                    # At this point, we already know that the incoming section is wider and
                    # therefore it MUST have a closed start, otherwise there will be dangling bonds.
                    if diff_mod == 2 and last_open or diff_mod == 0 and not last_open:
                        # In these cases, centers must match or differ by an even number of atoms.
                        offset += (2*shift) * atom_shift
                        
                        # And these are the limits for the shift
                        if last_open:
                            # To avoid the current open section leaving dangling bonds.
                            shift_lim = (W - last_W) // 4
                        else:
                            # To avoid sections being disconnected.
                            shift_lim = max((last_W - 3) // 2, 0) + (W - last_W) // 4
                        
                        min_shift, max_shift = - shift_lim, shift_lim                    
                    else:
                        # Otherwise, centers must differ by an odd number of atoms.
                        offset += (1 + 2*shift) * atom_shift
                        
                        # And these are the limits for the shift
                        if last_open:
                            # To avoid the current open section leaving dangling bonds.
                            min_shift = - max(0, (W - last_W) // 4)
                            max_shift = max(0, (W - last_W) // 4 - 1)
                        else:
                            # To avoid sections being disconnected.
                            shift_lim = max((last_W - 3) // 2, 0) + (W - last_W) // 4
                            min_shift, max_shift = -shift_lim - 1, shift_lim
                    
                    assert min_shift < shift < max_shift, (f"Shift must be between {min_shift}"
                        f" and {max_shift}, but {shift} was provided."
                    )
                    
            else:
                # There is at least one even section.
                
                # In this case ribbons are aligned based on one of their edges.
                # We choose the edge of the even ribbon that is open (i.e. has
                # dangling bonds). If both ribbons are even, we align it based
                # on the last section, not the incoming one.
                if last_W % 2 == 0 and last_open or W % 2 == 0 and open_start:
                    # Align them on the top edge
                    offset = last_addition.xyz[:, 1].max() - new_addition.xyz[:, 1].max()
                else:
                    # Align them on the bottom edge
                    offset = last_addition.xyz[:, 1].min() - new_addition.xyz[:, 1].min()
                    
                # We have to make sure that the open edge of the even ribbons (however
                # many there are) is always shifted towards the center. Shifting in the
                # opposite direction would result in dangling bonds.
                no_shift = False

                if diff_mod % 2 == 0:
                    min_shift = max(0, (W - last_W) - 1)
                    max_shift = W // 2
                        
                    # Both ribbons are even
                    if (last_open and shift == 0) or (not last_open and shift != 0):
                        # If there is no shift, we must make sure that both ribbons are open or close.
                        # If there is shift, both ribbons must have different configurations to avoid
                        # dangling bonds.
                        new_addition = _open_start(new_addition)
                        open_start = not open_start

                    if shift == 0:
                        no_shift = True
                    else:
                        bias, shift_sign = {
                            True: (-1, 1),
                            False: (1, -1)
                        }[last_open]
                        
                elif W % 2 == 1:
                    # Last section was even, incoming section is odd.
                    min_shift = 0
                    max_shift = last_W // 2

                    if W < last_W and shift == 0:
                        # Incoming odd section is thinner than current section.
                        # We can make the ends match if we ensure that the odd section is open.Make ends match (shift = 0): The odd ribbon must be open.
                        new_addition = _open_start(new_addition)
                        open_start = not open_start

                        no_shift = True
                    else:
                        bias, shift_sign = {
                            (True, True): (-1, 1),
                            (True, False): (1, -1),
                            (False, True): (1, 1),
                            (False, False): (-1, -1)
                        }[(W < last_W, last_open)]

                else:
                    # Last section was odd, incoming section is even.
                    min_shift = 0
                    max_shift = W // 2
                    
                    if last_open:
                        assert shift == 0, ("The shift between an open odd ribbon and an incoming"
                            "even ribbon must be always 0, otherwise the odd section will have a dangling bond"
                        )
                        no_shift = True
                    else:
                        bias, shift_sign = {
                            True: (-1, -1),
                            False: (1, 1)
                        }[open_start]

                if not no_shift:
                    # Check that the shift value is correct. There are two things to check here:
                    # Provided shift is in the right direction
                    right_shift_sign = shift*shift_sign - abs(shift) == 0
                    # Provided shift falls within the bounds.
                    shift_in_bounds = min_shift <= abs(shift) <= max_shift
                    
                    assert (right_shift_sign and shift_in_bounds), (
                        f" Shift must be between {min_shift*shift_sign} and {max_shift*shift_sign}, but {shift} was provided."
                    )

                    offset += (bias + 2*shift) * atom_shift
                
            # Apply the offset that we have calculated.
            new_addition = new_addition.move([0, offset, 0])
        
        # Check how many times we have to tile the unit cell (tiles) and whether
        # we have to cut the last string of atoms (cut_last)
        tiles, res = divmod(W_units + 1, 2)
        cut_last = res == 0
        
        # Tile the current section unit cell
        new_addition = new_addition.tile(tiles, 0)
        # Cut the last string of atoms.
        if cut_last:
            new_addition.cell[0,0] *= W_units / (W_units + 1)
            new_addition = new_addition.remove({"x": (new_addition.cell[0,0] - 0.01, None)})
        
        # Initialize the heterorribon or append to it if it's already initialized.
        if i == 0:
            geom = new_addition
        else:
            geom = geom.append(new_addition, 0)
        
        # Update the state for the next iteration
        last_addition = new_addition
        last_open = open_start != cut_last
        last_W = W
            
    return geom
```
</details>

And some examples of structures you can get:

```python
graphene_heteroribbon([7,9,6], units=[2,4,2]).plot(axes="xy")
```
newplot - 2022-01-12T042342 537

```python
graphene_heteroribbon([7,8,9,8,7], [3,4,2,2,3], shifts=[0,0,-1,1,0]).plot(axes="xy")
```
newplot - 2022-01-12T042418 632

```python
graphene_heteroribbon([7,8,9,8,7], [2j,1,2,1j,2]).plot(axes="xy")
```

newplot - 2022-01-12T042459 816

And this is just trying random numbers, you always get the valid ribbons or an error if that's not possible.

Also, this might be useful to design TranSIESTA setups with graphene electrodes? Something like:

```python
graphene_heteroribbon([31, 9, 7, 5, 7, 9, 31], [4,9,5,3,6,9,4]).plot(axes="xy")
```

newplot - 2022-01-12T043341 061

and then you just cut the cell to get the graphene electrodes, for example.

pfebrer:
Closes #420 

I still need confirmation from @tfrederiksen that there's nothing super wrong with the way it can be used :)

Is the code structure/arguments fine?

### DIFF
--- a/docs/api/default_geom.rst
+++ b/docs/api/default_geom.rst
@@ -14,6 +14,9 @@ geometries via a `pull request <pr>`_.
 
 All methods return a `Geometry` object.
 
+Some of the geometries are created in section based geometries, such as `heteroribbon`.
+This functionality is provided through the `composite_geometry`
+
 
 Bulk
 ====
@@ -51,6 +54,8 @@ Surfaces (slabs)
    zgnr
    graphene_nanoribbon
    nanotube
+   heteroribbon
+   graphene_heteroribbon
 
 
 2D materials
@@ -62,3 +67,12 @@ Surfaces (slabs)
    honeycomb
    bilayer
    graphene
+
+
+Helpers
+=======
+
+.. autosummary::
+   :toctree: generated/
+
+   composite_geometry
--- a/docs/api/default_geom.rst
+++ b/docs/api/default_geom.rst
@@ -76,3 +76,5 @@ Helpers
    :toctree: generated/
 
    composite_geometry
+   CompositeGeometrySection
+
--- a/src/sisl/geom/__init__.py
+++ b/src/sisl/geom/__init__.py
@@ -42,6 +42,7 @@ Surfaces
    graphene
 
 """
+from ._composite import *
 from .basic import *
 from .bilayer import *
 from .category import *
--- a/src/sisl/geom/_composite.py
+++ b/src/sisl/geom/_composite.py
@@ -1,10 +1,22 @@
 from dataclasses import dataclass, copy, fields 
+from abc import abstractmethod
 
-from sisl.messages import warn
+from sisl.messages import warn, SislError
+
+
+__all__ = ["composite_geometry"]
 
 
 @dataclass
 class _geom_section:
+
+    @abstractmethod
+    def build_section(self, geometry):
+        ...
+
+    @abstractmethod
+    def add_section(self, geometry, geometry_addition):
+        ...
 
     def _junction_error(self, prev, msg, what):
         """Helper function to raise an error if the junction is not valid.
@@ -14,7 +26,7 @@ class _geom_section:
         """
         msg = f"Error at junction between sections {prev} and {self}. {msg}"
         if what == "raise":
-            raise ValueError(msg)
+            raise SislError(msg)
         elif what == "warn":
             warn(msg)
 
@@ -29,8 +41,8 @@ def composite_geometry(sections, section_cls=_geom_section, **kwargs):
     sections: array-like of (_geom_section or tuple or dict)
         A list of sections to be added to the ribbon.
 
-        Each section is either a `_geom_section` or something that will
-        be parsed to a `_geom_section`.
+        Each section is either a `composite_geometry.section` or something that will
+        be parsed to a `composite_geometry.section`.
     section_cls: class, optional
         The class to use for parsing sections.
     **kwargs:
@@ -50,12 +62,12 @@ def composite_geometry(sections, section_cls=_geom_section, **kwargs):
         
         return copy.copy(s)
 
-    sections = [conv(section) for section in sections]
-    
     # Then loop through all the sections.
     geom = None
     prev = None
     for i, section in enumerate(sections):
+        section = conv(section)
+
         new_addition = section.build_section(prev)
         
         if i == 0:
@@ -66,3 +78,5 @@ def composite_geometry(sections, section_cls=_geom_section, **kwargs):
         prev = section
     
     return geom
+
+composite_geometry.section = _geom_section
--- a/src/sisl/geom/_composite.py
+++ b/src/sisl/geom/_composite.py
@@ -1,8 +1,7 @@
-from dataclasses import dataclass, copy, fields 
 from abc import abstractmethod
+from dataclasses import copy, dataclass, fields
 
-from sisl.messages import warn, SislError
-
+from sisl.messages import SislError, warn
 
 __all__ = ["composite_geometry"]
 
--- a/src/sisl/geom/_composite.py
+++ b/src/sisl/geom/_composite.py
@@ -3,11 +3,11 @@ from dataclasses import copy, dataclass, fields
 
 from sisl.messages import SislError, warn
 
-__all__ = ["composite_geometry"]
+__all__ = ["composite_geometry", "CompositeGeometrySection"]
 
 
 @dataclass
-class _geom_section:
+class CompositeGeometrySection:
 
     @abstractmethod
     def build_section(self, geometry):
@@ -30,7 +30,7 @@ class _geom_section:
             warn(msg)
 
 
-def composite_geometry(sections, section_cls=_geom_section, **kwargs):
+def composite_geometry(sections, section_cls, **kwargs):
     """Creates a composite geometry from a list of sections.
 
     The sections are added one after another in the provided order.
@@ -78,4 +78,5 @@ def composite_geometry(sections, section_cls=_geom_section, **kwargs):
     
     return geom
 
-composite_geometry.section = _geom_section
+
+composite_geometry.section = CompositeGeometrySection
--- a/src/sisl/geom/composite.py
+++ b/src/sisl/geom/composite.py
@@ -1,0 +1,68 @@
+from dataclasses import dataclass, copy, fields 
+
+from sisl.messages import warn
+
+
+@dataclass
+class _geom_section:
+
+    def _junction_error(self, prev, msg, what):
+        """Helper function to raise an error if the junction is not valid.
+
+        It extends the error by specifying details about the sections that
+        are being joined.
+        """
+        msg = f"Error at junction between sections {prev} and {self}. {msg}"
+        if what == "raise":
+            raise ValueError(msg)
+        elif what == "warn":
+            warn(msg)
+
+
+def composite_geometry(sections, section_cls=_geom_section, **kwargs):
+    """Creates a composite geometry from a list of sections.
+
+    The sections are added one after another in the provided order.
+
+    Parameters
+    ----------
+    sections: array-like of (_geom_section or tuple or dict)
+        A list of sections to be added to the ribbon.
+
+        Each section is either a `_geom_section` or something that will
+        be parsed to a `_geom_section`.
+    section_cls: class, optional
+        The class to use for parsing sections.
+    **kwargs:
+        Keyword arguments used as defaults for the sections when the .
+    """
+    # Parse sections into Section objects
+    def conv(s):
+        # If it is some arbitrary type, convert it to a tuple
+        if not isinstance(s, (section_cls, tuple, dict)):
+            s = (s, )
+        # If we arrived here with a tuple, convert it to a dict
+        if isinstance(s, tuple):
+            s = {field.name: val for field, val in zip(fields(section_cls), s)}
+        # At this point it is either a dict or already a section object.
+        if isinstance(s, dict):
+            return section_cls(**{**kwargs, **s})
+        
+        return copy.copy(s)
+
+    sections = [conv(section) for section in sections]
+    
+    # Then loop through all the sections.
+    geom = None
+    prev = None
+    for i, section in enumerate(sections):
+        new_addition = section.build_section(prev)
+        
+        if i == 0:
+            geom = new_addition
+        else:
+            geom = section.add_section(geom, new_addition)
+        
+        prev = section
+    
+    return geom
--- a/src/sisl/geom/nanoribbon.py
+++ b/src/sisl/geom/nanoribbon.py
@@ -245,7 +245,9 @@ class _heteroribbon_section(_geom_section):
         It does so by shifting half unit cell. This must be done before any
         tiling of the geometry.
         """
-        geometry = geometry.move(geometry.cell[self.long_ax] / 2)
+        move = np.array([0., 0., 0.])
+        move[self.long_ax] = -geometry.xyz[self.W, self.long_ax]
+        geometry = geometry.move(move)
         geometry.xyz = (geometry.fxyz % 1).dot(geometry.cell)
         return geometry
 
--- a/src/sisl/geom/nanoribbon.py
+++ b/src/sisl/geom/nanoribbon.py
@@ -411,7 +411,7 @@ class _heteroribbon_section(_geom_section):
             self._junction_error(prev, "LONE ATOMS: Previous odd section, which has an open end,"
                 " is wider than the incoming one. A wider odd section must always"
                 " have a closed end. You can solve this by making the previous section"
-                " one unit smaller or larger (L = L +- 1).", "raise"
+                " one unit smaller or larger (L = L +- 1).", self.on_lone_atom
             )
         
         # Get the difference in width between the previous and this ribbon section
@@ -420,8 +420,8 @@ class _heteroribbon_section(_geom_section):
         # on 1, 2, 3 or 4 atoms. After that, the cycle just repeats (e.g. 5 == 1, etc).
         diff_mod = W_diff % 4
 
-        # Now, we need to calculate the offset that we have to apply to the incoming
-        # section depending on several factors.
+        # Calculate the shifts that are valid (don't leave atoms with less than 2 bonds)
+        # This depends on several factors.
         if diff_mod % 2 == 0 and W % 2 == 1:
             # Both sections are odd
 
@@ -450,7 +450,7 @@ class _heteroribbon_section(_geom_section):
 
                 # Update the valid shift limits if the sections are aligned on any of the edges.
                 shift_offset = self._offset_from_center(align, prev)
-                valid_shifts += shift_offset
+                valid_shifts -= shift_offset
             elif prev.W == W:
                 valid_shifts = np.array([0])
             else:
@@ -474,7 +474,7 @@ class _heteroribbon_section(_geom_section):
                 shift_offset = self._offset_from_center(align, prev)
 
                 # Apply the offsets and calculate the maximum and minimum shifts.
-                min_shift, max_shift = -shift_lim + shift_offset, shift_lim + shift_offset
+                min_shift, max_shift = -shift_lim - shift_offset, shift_lim - shift_offset
 
                 valid_shifts = np.arange(min_shift, max_shift + 1, 2)
         else:
--- a/src/sisl/geom/nanoribbon.py
+++ b/src/sisl/geom/nanoribbon.py
@@ -7,14 +7,14 @@ from numbers import Integral
 import numpy as np
 
 from sisl import Atom, geom
-from .composite import _geom_section, composite_geometry
+from ._composite import _geom_section, composite_geometry
 from sisl._internal import set_module
 
 from ._common import geometry_define_nsc
 
 __all__ = [
     'nanoribbon', 'graphene_nanoribbon', 'agnr', 'zgnr',
-    'heteroribbon', 'graphene_heteroribbon', '_heteroribbon_section'
+    'heteroribbon', 'graphene_heteroribbon',
 ]
 
 
@@ -164,6 +164,7 @@ def zgnr(width, bond=1.42, atoms=None):
     agnr : armchair graphene nanoribbon
     """
     return graphene_nanoribbon(width, bond, atoms, kind='zigzag')
+
 
 @set_module("sisl.geom")
 @dataclass
@@ -582,14 +583,14 @@ class _heteroribbon_section(_geom_section):
 def heteroribbon(sections, section_cls=_heteroribbon_section, **kwargs):
     """Build a nanoribbon consisting of several nanoribbons of different widths.
 
-    This function basically uses `composite_geometry`, but defaulting to the usage
-    of `_heteroribbon_section` as the section class.
+    This function uses `composite_geometry`, but defaulting to the usage
+    of `heteroribbon.section` as the section class.
 
-    See `heteroribbon_section` and `composite_geometry` for arguments.
+    See `heteroribbon.section` and `composite_geometry` for arguments.
 
     Returns
     -------
-    sisl.Geometry:
+    Geometry:
         The final structure of the heteroribbon.
 
     Notes
@@ -618,6 +619,8 @@ def heteroribbon(sections, section_cls=_heteroribbon_section, **kwargs):
     """
     return composite_geometry(sections, section_cls=section_cls, **kwargs)
 
+heteroribbon.section=_heteroribbon_section
+
 
 @set_module("sisl.geom")
 def graphene_heteroribbon(sections, section_cls=_heteroribbon_section, bond=1.42, atoms=None, **kwargs):
@@ -628,8 +631,10 @@ def graphene_heteroribbon(sections, section_cls=_heteroribbon_section, bond=1.42
 
     See also
     ----------
-    `heteroribbon` : for argument details and how it behaves
+    heteroribbon : for argument details and how it behaves
     """
     if atoms is None:
         atoms = Atom(Z=6, R=bond * 1.01)
     return composite_geometry(sections, section_cls=section_cls, bond=bond, atoms=atoms, **kwargs)
+
+graphene_heteroribbon.section=_heteroribbon_section
--- a/src/sisl/geom/nanoribbon.py
+++ b/src/sisl/geom/nanoribbon.py
@@ -7,10 +7,10 @@ from numbers import Integral
 import numpy as np
 
 from sisl import Atom, geom
-from ._composite import _geom_section, composite_geometry
 from sisl._internal import set_module
 
 from ._common import geometry_define_nsc
+from ._composite import _geom_section, composite_geometry
 
 __all__ = [
     'nanoribbon', 'graphene_nanoribbon', 'agnr', 'zgnr',
--- a/src/sisl/geom/nanoribbon.py
+++ b/src/sisl/geom/nanoribbon.py
@@ -10,7 +10,7 @@ from sisl import Atom, geom
 from sisl._internal import set_module
 
 from ._common import geometry_define_nsc
-from ._composite import _geom_section, composite_geometry
+from ._composite import CompositeGeometrySection, composite_geometry
 
 __all__ = [
     'nanoribbon', 'graphene_nanoribbon', 'agnr', 'zgnr',
@@ -168,7 +168,7 @@ def zgnr(width, bond=1.42, atoms=None):
 
 @set_module("sisl.geom")
 @dataclass
-class _heteroribbon_section(_geom_section):
+class _heteroribbon_section(CompositeGeometrySection):
     """
     Parameters
     ----------
@@ -619,7 +619,7 @@ def heteroribbon(sections, section_cls=_heteroribbon_section, **kwargs):
     """
     return composite_geometry(sections, section_cls=section_cls, **kwargs)
 
-heteroribbon.section=_heteroribbon_section
+heteroribbon.section = _heteroribbon_section
 
 
 @set_module("sisl.geom")
@@ -629,12 +629,12 @@ def graphene_heteroribbon(sections, section_cls=_heteroribbon_section, bond=1.42
     Please see `heteroribbon` for arguments, the only difference is that the `bond` and `atoms`
     arguments default to ``bond=1.42`` and ``Atoms(Z=6, R=bond*1.01)``, respectively.
 
-    See also
-    ----------
+    See Also
+    --------
     heteroribbon : for argument details and how it behaves
     """
     if atoms is None:
         atoms = Atom(Z=6, R=bond * 1.01)
     return composite_geometry(sections, section_cls=section_cls, bond=bond, atoms=atoms, **kwargs)
 
-graphene_heteroribbon.section=_heteroribbon_section
+graphene_heteroribbon.section = _heteroribbon_section
--- a/src/sisl/geom/nanoribbon.py
+++ b/src/sisl/geom/nanoribbon.py
@@ -1,16 +1,21 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
+from dataclasses import dataclass, field
 from numbers import Integral
 
 import numpy as np
 
 from sisl import Atom, geom
+from .composite import _geom_section, composite_geometry
 from sisl._internal import set_module
 
 from ._common import geometry_define_nsc
 
-__all__ = ['nanoribbon', 'graphene_nanoribbon', 'agnr', 'zgnr']
+__all__ = [
+    'nanoribbon', 'graphene_nanoribbon', 'agnr', 'zgnr',
+    'heteroribbon', 'graphene_heteroribbon', '_heteroribbon_section'
+]
 
 
 @set_module("sisl.geom")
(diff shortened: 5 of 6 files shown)


sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+    kind: {'armchair', 'zigzag'}, optional
+        The kind of ribbon to be produced. Defaults to 'armchair'.
+
+    Returns
+    ----------
+    sisl.Geometry:
+        The final structure of the heteroribbon.
+    """
+    if kind == "zigzag":
+        raise ValueError("Zigzag heteroribbons are not implemented yet.")
```
zerothi, line 248: This should be a `NotImplementedError`

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
     agnr : armchair graphene nanoribbon
     """
     return graphene_nanoribbon(width, bond, atoms, kind='zigzag')
+
+
+Heteroribbon_section = namedtuple("Section", ["W", "L", "shift", "atoms"], defaults=(None, 1, 0, None))
```
zerothi, line 164: I would prefer this being a local variable. In that way there is no confusion, and defaults can be changed for future stuff, this would force them to use the function attribute.

Also, the `defaults=(1, 0, None)` suffices, that would mean `W` is required argument.

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+
+
+@set_module("sisl.geom")
+def heteroribbon(sections, bond, atoms=None, kind="armchair"):
+    """Builds a nanoribbon with multiple widths.
```
zerothi, line 169: Build a nanoribbon consisting of several nanoribbons of different widths

More descriptive?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+
+    The function makes sure that only valid ribbons, that is
+    ribbons with no dangling bonds, are being produced.
+    Explanatory errors are raised if that's not possible. Only the 
+    connection at the periodic boundary remains unchecked.
```
zerothi, line 174: The tests do not cover these cases, could we get one that issues the errors?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+
+    It only works for armchair ribbons for now.
```
zerothi, line 176: This should be in a `Notes` section, to highlight it.

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+
+    Parameters
+    ----------
+    sections: array-like of (Heteroribbon_section or array-like)
```
zerothi, line 180: instead of `Heteroribbon_section` write `heteroribbon.Section` or a `tuple` (which is equally retrievable) to backreference the attribute.

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+        A list of sections to be added to the ribbon.
+
+        Each section is either a `Heteroribbon_section` or an array that will
```
zerothi, line 183: also here and the line after

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+        be converted to a `Heteroribbon_section`. Therefore it should contain:
+            W: int
+                The width of the section.
+            L: int, optional
+                The number of units of the section. Note that a "unit" is
+                not a unit cell, but half of it. I.e. a zigzag string.
```
zerothi, line 189: This should be remembered to be changed for the other direction, since then it is a bit different.

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+
+                There are two situations in which you can choose the starting
+                configuration of your border ("open", "closed"). At the beggining of
+                the heteroribbon and on an even section after an odd one. In those
+                cases, an integer imaginary number here will indicate that you want
+                the open configuration (e.g. 2j indicates that you want 2 units, starting
+                with an open configuration). If you pass an imaginary number on an invalid
+                situation, an error will be raised.
```
zerothi, line 197: why not have the `Section` with another field ("open" true/false)? or?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+
+                If not provided, it defaults to 1.
+            shift: int, optional
+                The shift of this section with respect to the previous one.
```
zerothi, line 201: Can the shift be negative?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+                Value to pass to the `atoms` argument of `nanoribbon`. If not provided,
+                it defaults to the `atoms` argument passed to this function.
+    bond: float
+        The bond length of the ribbon.
+    atoms: Atom, optional
+        The default atom to use for the sections. For each section, if it contains an `atoms`
+        value, it will be used instead of this one.
+    kind: {'armchair', 'zigzag'}, optional
+        The kind of ribbon to be produced. Defaults to 'armchair'.
+
```
zerothi, line 241: I would like to see an `Examples` section that shows the kind of structures with the shifts, that would be cool. Perhaps very narrow ones that can be shown easily in an ASCII drawing?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+                        either -1 or 1 depending on the exact situation and always ensures that the minimum
+                        shift is always `abs(shift) == 1`.
+
+                If not provided, it defaults to 0.
+            atoms: Atom, optional
+                Value to pass to the `atoms` argument of `nanoribbon`. If not provided,
+                it defaults to the `atoms` argument passed to this function.
+    bond: float
+        The bond length of the ribbon.
+    atoms: Atom, optional
```
zerothi, line 236: This should not be optional, rather forced argument

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+    geom = None
+    last_addition = None
+    last_open = False
+    last_W = -1
+
+    # Parse sections into Heteroribbon_section objects
+    sections = [Heteroribbon_section(*section) for section in sections]
+
+    # Get the distance of an atom shift.
+    atom_shift = bond * math.sin(math.radians(60))
```
zerothi, line 261: sin(60) == 3**.5/2

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
     agnr : armchair graphene nanoribbon
     """
     return graphene_nanoribbon(width, bond, atoms, kind='zigzag')
+
+
+Heteroribbon_section = namedtuple("Section", ["W", "L", "shift", "atoms"], defaults=(None, 1, 0, None))
+
+
+@set_module("sisl.geom")
+def heteroribbon(sections, bond, atoms=None, kind="armchair"):
```
zerothi, line 168: `atoms` should not be optional, what happens for sections that do not have one?
Hmm...

zerothi (changes requested):
Great job! Minor changes here and there. @tfrederiksen the name of the function, what do you think?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+                If not provided, it defaults to 0.
+            atoms: Atom, optional
+                Value to pass to the `atoms` argument of `nanoribbon`. If not provided,
+                it defaults to the `atoms` argument passed to this function.
+    bond: float
+        The bond length of the ribbon.
+    atoms: Atom, optional
+        The default atom to use for the sections. For each section, if it contains an `atoms`
+        value, it will be used instead of this one.
+    kind: {'armchair', 'zigzag'}, optional
```
zerothi, line 239: when you have a set, you don't need `optional`, it is implicit

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,461 @@ def zgnr(width, bond=1.42, atoms=None):
...
+            else:
+                _junction_error(i, "Incoming section does not allow choosing the configuration,"
+                    f" but an imaginary number ({W_units}) was provided. Choosing configuration"
+                    " is only allowed for the first section or for an incoming even section if the"
+                    " previous section is odd."
+                )
+
+        # If this is not the first ribbon section in the heteroribbon, we are going to
+        # calculate the shifts.
+        if not i == 0:
```
zerothi, line 329: why not just `i > 0` will it ever be negative?

sisl/geom/nanoribbon.py
pfebrer: Because then if you pass lists you would need to pass all parameters (shift, atoms...) to be able to specify if it is open or not. Hmm I'm not sure what's best :thinking:

sisl/geom/nanoribbon.py
pfebrer: Yes, it depends on the situation.

sisl/geom/nanoribbon.py
pfebrer: I will add examples.

Do you know how to draw a ribbon in ASCII without using all your lifetime? I would rather not do it, but if you want to do it feel free :laughing:.

Idea for the future: ASCII backend for `GeometryPlot`.

sisl/geom/nanoribbon.py
pfebrer: True

sisl/geom/nanoribbon.py
pfebrer: Really? Why is that? An argument can be positional but only accept values within a set, doesn't it?

sisl/geom/nanoribbon.py
zerothi: ok, perhaps a short note, "The +- shift of this section" or something like that?

sisl/geom/nanoribbon.py
zerothi: Hehe, yeah, it was just an idea. :)

zerothi: I don't know if there exists an easy ascii molecular drawer

sisl/geom/nanoribbon.py
zerothi: I meant for the documentation:

```
kind: {'armchair', 'zigzag'}
```
nothing more, this is described in the numpy-docs manual.

sisl/geom/nanoribbon.py
zerothi: I would just prefix the variable as `_heteroribbon_section` to ensure it isn't *easily* imported/known.

sisl/geom/nanoribbon.py
tfrederiksen: `jp2a` is a quite nice command line tool to convert from jpg to ascii. Here is an example:
```python
composite_agnr([5, 6, 7, 7, 6, 5]).plot(axes="xy", show_cell=False, show_atoms=True, atoms_scale=0.1)
```
newplot-1
Then I manipulate with `gimp` to remove the frame and to get thick black bonds, and to export in `jpg` format:
structure_fat
Then it can be converted to ascii:
```
$ jp2a --chars=****"  " --width=80 -b structure.jpg
+--------------------------------------------------------------------------------+
|                                   **********                                   |
|                                  **        **                                  |
|                                ***          ***                                |
|                        **********            **********                        |
|                      ***      ***            ***      ***                      |
|                     **          **          **          **                     |
|                   ***            ***      ***            ***                   |
|          ************            ************            ************          |
|         ***        ***          ***        ***          ***        ***         |
|        **            **        ***           **        **            **        |
|**********            ************            ************            **********|
|        ***          ***        ***          ***        ***          ***        |
|         ***        ***          ***        ***          ***        ***         |
|          ************            ************            ************          |
|          ***      ***            ***       **            **        **          |
|         **          **          **          **          **          **         |
|**********            ************            ************            **********|
|**********            ************            ************            **********|
|        ***          ***        ***          ***        ***          ***        |
|          **        **            **        **            **        **          |
|           **********              **********              **********           |
+--------------------------------------------------------------------------------+
```
It requires a bit of playing around to get a reasonably looking result. But maybe this could be automatized?

tfrederiksen:
Great work @pfebrer ! Overall it looks great to me. However, I am not fully convinced about the use of imaginary numbers to specify "open" structures. I think it follows naturally that each consequtive "column" in the structure needs to be of alternating type. Hence, for a given input of widths and repetitions are only two possible structures that can be created. This could be one flag to the function, say `invert=True/False`. What do you think?

tfrederiksen:
> Great job! Minor changes here and there. @tfrederiksen the name of the function, what do you think?

I think it is good (and relatively short) as it is. In the literature I see mostly the formulation "graphene nanoribbon heterojunctions" for these structures. Other possibilities could therefore be the pair `nanoribbon_heterojunction` and `gnr_heterojunction`? (Or `composite_nanoribbon` and `composite_gnr`?)

Could the method eventually be extended to 2D nanomeshes?

zerothi:
> I think it is good (and relatively short) as it is. In the literature I see mostly the formulation "graphene nanoribbon heterojunctions" for these structures. Other possibilities could therefore be the pair `nanoribbon_heterojunction` and `gnr_heterojunction`? (Or `composite_nanoribbon` and `composite_gnr`?)

I like `nanoribbon_hetero`, kind of feel the "junction" is just too long. ;)

Hmm.. On the other hand explicitness is good... hmmm. Since `hetero` is already in the litterature, I like that name.

pfebrer:
> I think it follows naturally that each consequtive "column" in the structure needs to be of alternating type. Hence, for a given input of widths and repetitions are only two possible structures that can be created. This could be one flag to the function, say invert=True/False.

It follows naturally in most cases, but there are two specific cases where it does not follow naturally:
- **At the beginning of the unit cell**: The configuration of the initial string forces the creation of completely different ribbons. One **can't be related** to the other just **by a simple `invert` operation**. As you noted, you probably can always find a string of atoms where you can start with the "sisl provided" configuration and then shift, but I don't believe that this is a good solution :(
- **When an even section comes in and the previous section was odd**: This case is fully defined if you specify the `initial atom/final atom` as in your approach of course, but not in this case where the ribbons are automatically aligned and `0` means "aligned". There are two valid ways to "perfectly align" an odd ribbon with an even ribbon: 
   - If the bottom edge of the even ribbon is open ("closed" configuration), they are aligned at the bottom:
     ```python
     graphene_heteroribbon([(7, 1), (8, 1)])
     ```
     newplot - 2022-01-13T115922 013
     In the case where the previous section is closed, a shift is then allowed, but only upwards, as a downward shift would leave the last atom of the even section dangling:
     ```python
     graphene_heteroribbon([(7, 2), (8, 1, 1)])
     ```
     
     newplot - 2022-01-13T120710 062

   - If the top edge of the even ribbon is open ("open" configuration), they are aligned at the top:
     ```python
     graphene_heteroribbon([(7, 1), (8, 1j)])
     ```
     newplot - 2022-01-13T121351 198
    And then in the case when the previous section is closed, only a downwards shift is allowed.

    If there is only one such choice, the obtained final structures are related by a mirror symmetry, certainly, as you note. But if there is **more than one such choice, then it isn't the case anymore**. Therefore, you need to have the ability to choose.

That's why I only allow imaginary numbers in these two cases. We could provide an extra field to specify this change instead of using this notation, but as I told Nick this has the pitfall that you need to specify all the parameters for the section, because sections are passed as tuples. Whatever you prefer.

pfebrer:
> I like `nanoribbon_hetero`, kind of feel the "junction" is just too long. ;)

But then the graphene version of it would be `graphene_nanoribbon_hetero` ? :sweat_smile:

zerothi:
> That's why I only allow imaginary numbers in these two cases. We could provide an extra field to specify this change instead of using this notation, but as I told Nick this has the pitfall that you need to specify all the parameters for the section, because sections are passed as tuples. Whatever you prefer.

But that is the same as the other arguments in the Section tuple, no?

> > I like `nanoribbon_hetero`, kind of feel the "junction" is just too long. ;)
> 
> But then the graphene version of it would be `graphene_nanoribbon_hetero` ? sweat_smile

Agreed... Hmm....
I also like your original one.. I don't know ;) !

sisl/geom/nanoribbon.py
pfebrer: The question is, would someone use it? Hahah

I've seen that `gnuplot` can draw in ascii and there is an extension for `matplotlib` (https://github.com/gooofy/drawilleplot) that can do it too.

zerothi: Yeah, this is overkill here. :)

sisl/geom/nanoribbon.py
pfebrer: Ok, I've changed it. But I don't understand the logic. I will read the numpy-docs manual :)

pfebrer:
> But that is the same as the other arguments in the Section tuple, no?

Yes, it's just that the more parameters you add, the worse it gets :) Specially if parameters are not easily ordered by importance. E.g. we can all agree that length will be very used so it makes sense that it is the first optional argument.

pfebrer:
Also, for the name: It is not really a `heterojunction` but potentially several `heterojunctions`, so probably calling it `heterojunction` would not picture what the function does exactly.

zerothi:
> > But that is the same as the other arguments in the Section tuple, no?
> 
> Yes, it's just that the more parameters you add, the worse it gets :) Specially if parameters are not easily ordered by importance. E.g. we can all agree that length will be very used so it makes sense that it is the first optional argument.

True... Perhaps you should add the possibility of accepting `dict` as sections.
```python
     sections = [Heteroribbon_section(*section) for section in sections]
# to
def conv(s):
    if isinstance(s):
        return Heteroribbon_section(**s)
    return Heteroribbon_section(*s)
sections = [conv(section) for section in sections]
```

pfebrer:
Well in that case you might as well build the section yourself instead of passing a dict :)

zerothi:
> Well in that case you might as well build the section yourself instead of passing a dict :)

In any case I think it might be useful for some.

pfebrer:
However would you then introduce the `open` parameter? That forces the user to replace a list with a dict just to change the configuration.

pfebrer:
The imaginary number seems natural to me. `2j` is just `2` shifted by a phase :)

zerothi:
I don't know about this, @tfrederiksen what do you say?

pfebrer:
Another option could be using negative numbers (`-2` instead of `2j`)

tfrederiksen:
> * **At the beginning of the unit cell**: The configuration of the initial string forces the creation of completely different ribbons. One **can't be related** to the other just **by a simple `invert` operation**.

I think I didn't express my idea correctly. It is not to apply an invert operation after creating the structure, but to let the user select the "phase" of the first column (open/closed). I fully agree that this choice will create different structures. But the consecutive columns *must* respect phase alternation, no?. Indeed the meaning of shift would have to be redefined (no longer just multiples of the lattice vector in `y`).

My point is just that from the user perspective it may be simpler if one does not have to keep track of the phase, but just specify opening type and then widths/reps/shifts.

pfebrer:
Oooh ok I understand now, so

```python
graphene_nanoribbon([(7, 2), (11, 2)], open_start=True)
```
would be the old
```python
graphene_nanoribbon([(7, 2j), (11, 2)])
```
?

I like it :) But then the second case still remains.

tfrederiksen:
> Oooh ok I understand now, so
> 
> ```python
> graphene_nanoribbon([(7, 2), (11, 2)], open_start=True) # or invert=True
> ```
> 
> would be the old
> 
> ```python
> graphene_nanoribbon([(7, 2j), (11, 2)])
> ```
> 
> ?

Exactly!

> > I like it :) But then the second case still remains.

What do you mean by second case? The meaning of shift would then have to refer to half the lattice vector in `y`.

pfebrer:
> Additionally, in my opinion it is not very clear from the user perspective if a given "column" should be denoted "closed" (real) or "open" (imaginary). It seems ambiguous to me.

For odd columns I think it's very well defined. Why not? You have to consider if you are looking at it from the left or the right, but that seems fine to me. I mean I do not call the column "open" or close, the same column can be an open or closed border. For even columns it is arbitrary I agree, but once you establish which is which it isn't ambiguous either.

tfrederiksen:
> > Additionally, in my opinion it is not very clear from the user perspective if a given "column" should be denoted "closed" (real) or "open" (imaginary). It seems ambiguous to me.
> 
> For odd columns I think it's very well defined. Why not? You have to consider if you are looking at it from the left or the right, but that seems fine to me. I mean I do not call the column "open" or close, the same column can be an open or closed border. For even columns it is arbitrary I agree, but once you establish which is which it isn't ambiguous either.

My point is just that understanding all of this would not be required by the user. One could just choose one of two possible "openings" and then build anything from there.

pfebrer:
> What do you mean by second case? The meaning of shift would then have to refer to half the lattice vector in y.

Hmm ok, then you'd have to, for example always align at the bottom and then shift. I get your point. But I don't think that makes it more intuitive necessarily. That might make it harder to build valid ribbons because you have to know the allowed shift values. The main point of building this function for me was that I didn't have to think about which shifts are valid and which are not. I like the idea of just going by:
- `0`: aligned
- `-1`: first possible downward shift (if available)
- `1`: first possible upwards shift (if available)

If the user needs to specify exactly where to place the ribbon (your proposal) I believe they have to think even more about phases, since otherwise they won't generate valid ribbons.

tfrederiksen:
> > What do you mean by second case? The meaning of shift would then have to refer to half the lattice vector in y.
> 
> Hmm ok, then you'd have to, for example always align at the bottom and then shift. I get your point. But I don't think that makes it more intuitive necessarily. That might make it harder to build valid ribbons because you have to know the allowed shift values. The main point of building this function for me was that I didn't have to think about which shifts are valid and which are not. I like the idea of just going by:
> 
>     * `0`: aligned
> 
>     * `-1`: first possible downward shift (if available)
> 
>     * `1`: first possible upwards shift (if available)
> 
> 
> If the user needs to specify exactly where to place the ribbon (your proposal) I believe they have to think even more about phases, since otherwise they won't generate valid ribbons.

All integer shifts should be valid (in units of half the lattice vector). But clearly the function needs to pick the only meaningful phase for that column not to mess up the bonding pattern.

pfebrer:
No, they are not, because some shifts leave dangling bonds at the edges of the ribbon

pfebrer:
And which shifts are valid depends on the exact characteristics of the junction, it is not straightforward

tfrederiksen:
@pfebrer , what about a quick chat on discord now?

pfebrer:
After talking with @tfrederiksen, I think the best way to go is to let users choose if they want the "supervised" version that makes sure you never leave dangling bonds or a "grown up" version that just lets you do whatever you want by shifting always (shift * half_lattice).

This would be switched on/off with a flag e.g. `graphene_nanoribbon([...], supervised=False)`.

Do you agree @zerothi?

zerothi:
> After talking with @tfrederiksen, I think the best way to go is to let users choose if they want the "supervised" version that makes sure you never leave dangling bonds or a "grown up" version that just lets you do whatever you want by shifting always (shift * half_lattice).
> 
> This would be switched on/off with a flag e.g. `graphene_nanoribbon([...], supervised=False)`.
> 
> Do you agree @zerothi?

Sounds good to me. However, just be reading the name, I can't clarify its intent.
Could we name it:
```python
heteroribbon(..., allow_dangling=False|True)
```
for clarity?

pfebrer:
By the way in the "not supervised" version what would it be better:
- `0` alignment means the bottom atom at the same position as the previous section.
- Alignment is with respect to a fixed reference (as in Thomas implementation).

?

> Could we name it:

Yes, sure!

zerothi:
> (although `allow_dangling` sounds like you raise errors on dangling, but it's more than that. We will find the right name :))

Ok, I see, yes. Just so it has some description of what its doing.

pfebrer:
Ok, after some introspection, I discovered I was doing two things differently from what Thomas was doing:
- **Aligning at the bottom, center or top automatically**. I have introduced an argument `align` that lets you choose where to align the sections (for each section). If you set this to `"auto"`, you get the behavior that I implemented at the beginning.
- **Only allowing certain shift values to avoid dangling bonds**. Now this is an opt-in feature. I called it `quantum_shifts`. You need to set `quantum_shifts=True` (open for suggestions on the name :)).

I've also introduced `on_dangling` so that you can choose to get errors, warnings or nothing when dangling bonds are produced. This is regardless of whether you are using quantum shifts or not.

So, to summarize, by default you get almost Thomas' implementation, you would just need to turn errors off:

```python
graphene_heteroribbon([(7, 2), (11, 2)], on_dangling="ignore").plot(axes="xy")
```

newplot - 2022-01-14T025827 391

```python
graphene_heteroribbon([(7, 2), (11, 2, -1)], on_dangling="ignore").plot(axes="xy")
```
newplot - 2022-01-14T030025 172

But, if you want, you can go "god mode":

```python
graphene_heteroribbon([(7, 2), (11, 2)], align="auto", quantum_shifts=True).plot(axes="xy")
```
newplot - 2022-01-14T030317 327

```python
graphene_heteroribbon([(7, 2), (11, 2, -1)], align="auto", quantum_shifts=True).plot(axes="xy")
```

newplot - 2022-01-14T030202 807

And in between you can switch on/off the individual features as you wish.

Finally, what was previously achieved with imaginary numbers is now achieved by specifying where to align the sections. So:

```python
graphene_heteroribbon([(7, 2), (8, 2j)]) # before
graphene_heteroribbon([(7, 2), (8, 2, 0, "top")], align="auto", quantum_shifts=True) # now
```

Thanks to @tfrederiksen for the discussions, I feel like now it is a much complete tool! :)

(and I hope this is close to the definitive version, because it took me quite a long time hahaha)

zerothi:
@pfebrer please see my commits

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,576 @@ def zgnr(width, bond=1.42, atoms=None):
...
+    >>> # A simple 7-11-AGNR with the sections aligned on the center
+    >>> heteroribbon([(7, 2), (11, 2)], bond=1.42, atoms="C")
+    >>> # The same AGNR but shifted up
+    >>> heteroribbon([(7, 2), (11, 2, 1)], bond=1.42, atoms="C")
+    >>> # And down
+    >>> heteroribbon([(7, 2), (11, 2, -1)], bond=1.42, atoms="C")
+    >>> # The same AGNR but with a bigger 11 section and a 9-atom bridge
+    >>> heteroribbon([(7, 1), (9,1), (11, 4), (9,1), (7,1)], bond=1.42, atoms="C")
+    >>> # One that you have probably never seen before
+    >>> heteroribbon([(7, 1j), (10, 2), (9, 1), (8, 2j, -1)], bond=1.42, atoms="C")
```
zerothi, line 260: So how would this look in the new one? The imaginary number should be -2, no?

zerothi (changes requested):

pfebrer:
Hey, sorry I will get home by 19 aprox and I will then update the examples (the number of units now is always a positive real integer), what you have to do is to change the alignment of the section (align it on top). Regarding setting `correct_shifts=True` as the default, I don't know if that's a good idea. `on_dangling = 'raise'` already makes sure that by default ribbons with dangling bonds are not produced. It might be surprising that the shifts are getting modified by default and users might not understand what's happening. They'll just think the function is not properly implemented. So I'd rather keep it an opt-in feature.

By the way I'm not sure that the function does what one would expect as "correcting shifts". I mean it's not that the shift value is rounded to the closest valid shift 🤔

zerothi:
No worries.

So then I don't fully understand what `correct_shifts` does?  
I thought the idea was that it does something *physical*, i.e. keeping correct graphene structure. Whereas for `False`, it may produce wrong graphene junctions? If the former I think `True` is better, but perhaps I am wrong?

pfebrer:
There are two separate things.

Imagine that for a given junction the valid shifts (as in they don't leave dangling bonds) are:
```
valid_shifts = [0, 2, 4, 6]
```

Then:
- `on_dangling` controls what to do if `shift not in valid_shifts`.The default is to raise an error. So no dangling bonds are allowed by default.
- If `quantum_shifts` is set to `True`, the shift is modified: `shift = valid_shifts[shift]`. That is why I called it "quantum", the shift value is the index of the valid shift you want to use.

zerothi:
Ok, but that is extremely difficult on follow, I. E. Changing quantum shift between true false, forces you to change shift values as well. What about `on_dangling=correct` where it will select the nearest valid shift (always +)?

pfebrer:
> Changing quantum shift between true false, forces you to change shift values as well.

Yes, it's a completely different mode of operation. To me it's useful because it means the same regardless of what the particular widths of the junctions are. And it's also useful to play around if you are not looking for a particular ribbon or you are trying to understand how much shift do you need.

> What about on_dangling=correct where it will select the nearest valid shift (always +)?

I believe it's not a very good idea to allow that. The nearest shift might be really far from the requested, and the obtained junction might have nothing to do with what the user was thinking about. I think it's better to just issue meaningful errors containing the valid shifts and make the user correct the value in their code.

pfebrer:
Well actually after decoupling the functionalities I found that what I needed most was the ability to choose the alignment, but the quantum shifts are also nice to play around. I would not turn them on by default that's it.

zerothi:
Ok, thanks for the clarification. That was not obvious to me :)

What about naming it `shifts_are_indices` instead? The quantum name doesn't really give it meaning?

pfebrer:
Hmm it's not really true either :sweat_smile:

Because if valid shifts is `[-4, -2, 0, 2, 4]` the corresponding "quantum shifts" are `[-2, -1, 0, 1, 2]`. I.e. `0` shift is always the most aligned possibility.

pfebrer:
I think there's no perfect way to describe this in just an argument name. Probably the most important thing is that by reading the docs it can be understood. What would you add there?

tfrederiksen:
Amazing job @pfebrer! I have been playing a bit around with the current version and I think it has become even more intuitive and easy to use! I really like the alignment feature and how shifts are provided relative to the previous block.

I have a few comments:

- I wonder if the meaning of parameter `on_dangling` could be made more precise. I would say that all graphene nanoribbons created in sisl have dangling bonds at the edges (sp2 carbons with just two neighbours have a single dangling bond each, which in reality is typically saturated with hydrogen). Do you rather mean that the function tries to avoid passing geometries where some atoms have *just a single neighbour*?
- The behaviour of `on_dangling` didn't work as expected for me. For instance, I could not force the creation of a disconnected geometry with crazy parameters like:
```python
sisl.geom.graphene_heteroribbon([(7,2), (8,2,-10)], on_dangling='ignore')
```
- On the other hand, the default `on_dangling='raise'` (actually also `'warn'` without any warning) accepted inputs like these:
```python
sisl.geom.graphene_heteroribbon([(7,2), (8,2,-5)]).plot(axes="xy", nsc=[2,1,1])
```
newplot
```python
sisl.geom.graphene_heteroribbon([(7,2), (8,), (9,)], invert_first=True).plot(axes="xy", nsc=[2,1,1])
```
newplot-1

- I do not really understand why you limit the creation of geometries freely according to the given input parameters. Will the users not look at the created geometries and decide if these are what they really want?
- Could it be possible to enable that a `section` could be just an integer for the width (with defaults to `L=1`, `shift=0` and `align='a'`). Say that `graphene_heteroribbon([(7,2), (9,2)]` could also be written as `graphene_heteroribbon([(7,2), 9, 9]`?
- `section` could perhaps also be called `segment`.

In any case, the tool is already super powerful. The above are just some feedback for your consideration.

tfrederiksen:
By the way, I just noticed that while `sisl.geom.graphene_heteroribbon([(7,2)])` works (creating the usual 7-AGNR), `sisl.geom.graphene_heteroribbon([(7,1), (7,1)])` fails with this message:
```
ValueError: Error at junction between sections 0: Section(W=7, L=1, shift=0, align=None, atoms=None) and 1: Section(W=7, L=1, shift=0, align=None, atoms=None).
DANGLING BONDS: Shift must be between 1 and 0 but 0 was provided.
```

pfebrer:
Thanks for the report @tfrederiksen!! Indeed the cases that you mentioned contained some sneaky bugs :) Things should work fine now.

The fact that this one: 
image

didn't report any warning was not a bug though. The notes mentioned that the periodic boundary was not checked. I have however now introduced the `pbc` argument. If it is set to `True`, the connection is checked. 

> Could it be possible to enable that a section could be just an integer for the width (with defaults to L=1, shift=0 and align='a'). Say that graphene_heteroribbon([(7,2), (9,2)]) could also be written as graphene_heteroribbon([(7,2), 9, 9])

This is now possible, good idea! However, the default alignment for the section is `None` (i.e. default global alignment is used) for consistency with the rest of the ways of creating a section. I think otherwise it can be confusing. So `7` is basically the  same as the already valid `(7, )`

**Things left to decide:**

@zerothi I brought back `correct_shifts=False` as the default. Is that fine after understanding what it does?
Any last minute ideas on a better name?

> I do not really understand why you limit the creation of geometries freely according to the given input parameters. Will the users not look at the created geometries and decide if these are what they really want?

Do you mean why errors are raised by default? I preferred to be safe by default. Although all structures might be valid depending on the exact thing that the user is looking for, I think in most cases ribbons with a single neighbor C will not be desired. If a call to this function ends inside some library code, the user won't be able to check the structure, so I prefer to keep it like this. Would you say it makes more sense `on_dangling="ignore"` as the default?

If you meant why the shift is modified to avoid dangling bonds (what I called quantum shifts), in the last commit Nick switched it on by default. I've now switched it back off. Anyway, this is something that can be switched on/off, users are not forced to use it.

> I wonder if the meaning of parameter on_dangling could be made more precise. I would say that all graphene nanoribbons created in sisl have dangling bonds at the edges (sp2 carbons with just two neighbours have a single dangling bond each, which in reality is typically saturated with hydrogen). Do you rather mean that the function tries to avoid passing geometries where some atoms have just a single neighbour?

I agree with Thomas. Proposal for the name?

> section could perhaps also be called segment.

I think Nick also preferred this, I'll change it tomorrow.

I also need to update the examples, once everything is definitive I'll do it.

zerothi:
> > Could it be possible to enable that a section could be just an integer for the width (with defaults to L=1, shift=0 and align='a'). Say that graphene_heteroribbon([(7,2), (9,2)]) could also be written as graphene_heteroribbon([(7,2), 9, 9])
> 
> This is now possible, good idea! However, the default alignment for the section is `None` (i.e. default global alignment is used) for consistency with the rest of the ways of creating a section. I think otherwise it can be confusing. So `7` is basically the same as the already valid `(7, )`

Great!
 
> **Things left to decide:**
> 
> @zerothi I brought back `correct_shifts=False` as the default. Is that fine after understanding what it does? Any last minute ideas on a better name?

Agreed, perhaps `shifts_as_valid_index`? Long, but more descriptive?
But perhaps the `quantum` is not so bad after understanding what it does... `shifts_quantum`? the reason for having `shifts` first is that it is more clear to me that it adheres to the `shifts` argument.

> > I do not really understand why you limit the creation of geometries freely according to the given input parameters. Will the users not look at the created geometries and decide if these are what they really want?
> 
> Do you mean why errors are raised by default? I preferred to be safe by default. Although all structures might be valid depending on the exact thing that the user is looking for, I think in most cases ribbons with a single neighbor C will not be desired. If a call to this function ends inside some library code, the user won't be able to check the structure, so I prefer to keep it like this. Would you say it makes more sense `on_dangling="ignore"` as the default?

I would advocate `on_dangling="raise"`, we should try and return only valid stuff. Alternatively the `on_dangling` could accept a callable with a single argument `msg`? Then if `on_dangling` is `ValueError` or a subclass of `Exception` it will be raised?
Hmm... I don't really know what the best solution is here. But definitely raising an error when they invalid.

> If you meant why the shift is modified to avoid dangling bonds (what I called quantum shifts), in the last commit Nick switched it on by default. I've now switched it back off. Anyway, this is something that can be switched on/off, users are not forced to use it.

I agree my change was wrong.

> > I wonder if the meaning of parameter on_dangling could be made more precise. I would say that all graphene nanoribbons created in sisl have dangling bonds at the edges (sp2 carbons with just two neighbours have a single dangling bond each, which in reality is typically saturated with hydrogen). Do you rather mean that the function tries to avoid passing geometries where some atoms have just a single neighbour?
> 
> I agree with Thomas. Proposal for the name?

`on_lone_atom`? Or something like that?
 
> > section could perhaps also be called segment.
> 
> I think Nick also preferred this, I'll change it tomorrow.

I don't really have a preference here, both seem fine to me. But this naming convention should then be honed in later geometry-creation functions that are based on section/segments. So considering this, which would you prefer?

> I also need to update the examples, once everything is definitive I'll do it.

Great!

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,533 @@ def zgnr(width, bond=1.42, atoms=None):
...
     return graphene_nanoribbon(width, bond, atoms, kind='zigzag')
+
+
+_heteroribbon_section = namedtuple("Section", ["W", "L", "shift", "align", "atoms"], defaults=(1, 0, None, None))
+
+
+@set_module("sisl.geom")
+def heteroribbon(sections, bond, atoms, kind="armchair",
+                 invert_first=False, align="bottom",
+                 correct_shifts=False, on_dangling="raise", pbc=True):
```
tfrederiksen, line 170: What about `low_coordination="warn"` as default (replacing `on_dangling`), referring to creation of structures with neighbours < 2?

Maybe `correct_shifts` could be a function of the above parameter:
- `low_coordination="raise"` => `correct_shifts=True`
- `low_coordination="warn/ignore"`=> `correct_shifts=False` (the user knows what they are doing)?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,533 @@ def zgnr(width, bond=1.42, atoms=None):
...
+    align: {"bottom"/"b", "top"/"t", "center"/"c", "auto"/"a"}
+        The default alignment between sections.
+        Sections `align` argument in their `heteroribbon.Section` will have precedence.
+
+        If automatic alignment is requested, sections are aligned:
+         - If both sections are odd: On their center.
+         - If previous section is even: On its open edge (top or bottom)
+         - If previous section is odd and incoming section is even: On the bottom.
+    correct_shifts: bool, optional
+        Whether or not the implementation will correct faulty shifts.
```
tfrederiksen, line 219: Alternative suggestion: `Whether or not the implementation will correct shifts that would generate low-coordinated atoms  (neighbours < 2)`?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,533 @@ def zgnr(width, bond=1.42, atoms=None):
...
+
+        If ``False``, sections are just shifted (`shift`) number of atoms.
+
+        If ``True``, shifts are quantized in the sense that shifts that produce dangling
```
tfrederiksen, line 223: `...that shifts that produce low-coordinated atoms (neighbours < 2) are ignored`?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,533 @@ def zgnr(width, bond=1.42, atoms=None):
...
+        bonds are ignored. Then:
+            - ``shift = 0`` means aligned.
+            - ``shift = -1`` means first possible downwards shift (if available).
+            - ``shift = 1`` means first possible upwards shift (if available).
+        If this is set to `True`, `on_dangling` is overwritten to `"raise"`.
+    on_dangling: {"raise", "warn", "ignore"}
+        What to do when a junction between sections produces dangling bonds.
```
tfrederiksen, line 230: `...when a junction between sections produce low-coordinated atoms (neighbours < 2)`?

sisl/geom/nanoribbon.py
```diff
@@ -154,3 +159,533 @@ def zgnr(width, bond=1.42, atoms=None):
...
+    >>> # A simple 7-11-AGNR with the sections aligned on the center
+    >>> heteroribbon([(7, 2), (11, 2)], bond=1.42, atoms="C")
+    >>> # The same AGNR but shifted up
+    >>> heteroribbon([(7, 2), (11, 2, 1)], bond=1.42, atoms="C")
+    >>> # And down
+    >>> heteroribbon([(7, 2), (11, 2, -1)], bond=1.42, atoms="C")
+    >>> # The same AGNR but with a bigger 11 section and a 9-atom bridge
+    >>> heteroribbon([(7, 1), (9,1), (11, 4), (9,1), (7,1)], bond=1.42, atoms="C")
+    >>> # One that you have probably never seen before
+    >>> heteroribbon([(7, 1j), (10, 2), (9, 1), (8, 2j, -1)], bond=1.42, atoms="C")
```
tfrederiksen, line 259: Imaginary numbers are now gone, or?

pfebrer:
> So considering this, which would you prefer?

Would that be for any tool that builds a composite geometry or just if the parts are appended one after the other?

zerothi:
> > So considering this, which would you prefer?
> 
> Would that be for any tool that builds a composite geometry or just if the parts are appended one after the other?

In either case I presume... They need to fit together in one way or another. However, as has been discussed an extension that creates graphene-like flakes with *patches* of different materials might be implemented later. There patches are probably more saying since they extend in more than 1 direction.

However, for *composite* or *appended* geometries I think they would fall under the same category?

tfrederiksen:
> we should try and return only valid stuff.

OK, but how do we define *valid*? I think it depends on what the users have in mind. I would say that a geometry defined mathematically by the input parameters cannot be "invalid" (but perhaps look strange or unusual to most people).

zerothi:
> > we should try and return only valid stuff.
> 
> OK, but how do we define _valid_? I think it depends on what the users have in mind. I would say that a geometry defined mathematically by the input parameters cannot be "invalid" (but perhaps look strange or unusual to most people).

True, but to some extend wouldn't physically *valid* geometries never have dangling bonds. Sure users can fix this after (I still need to get that one function to work ;)) but here in this case where Hydrogen are never added and dangling bonds are somewhat unphysical.

Hmm... But again, I see.
From your perspective I can see that `on_dangling` should never be there, not even as a warning since it is the user responsibility to correct the geometries returned. Hmm... Perhaps that is ok, just a strict clarification in the `Notes` section about this function returning geometries with 2 dangling bonds may occur.

tfrederiksen (approved):
Looks great @pfebrer ! I have just some minor suggestions.

By the way, you probably also want to add these new functions to the `sisl/geom/__init__.py` so they show up in the documentation.

zerothi:
Coming to think of this, I think you might be right. For future geometry creation routines it might be better to allow end-users to correct the geometries and just get simple things that can be build upon....

I think I am convinced, vote to remove `on_dangling` all-together :)

tfrederiksen:
> wouldn't physically _valid_ geometries never have dangling bonds

Well, any carbon edge would lead to atoms with dangling bonds, no? I do not see clearly why you would allow atoms with one dangling bond (like any of the other gnr functions do at the edges) but not two or three. This would depend on what chemistry the user have in mind.

A dangling bond is physical. It is just very reactive.

sisl/geom/nanoribbon.py
pfebrer: Yes!

sisl/geom/nanoribbon.py
pfebrer: I think you should be able to provide the exact shift yourself and still get the function to error if you get a single bonded atom. I'm thinking mainly if you do high throughput and you know you don't want single bonded edges, but because of some bug in your code you are producing the wrong structures.

tfrederiksen: One could also have a parameter `min_coordination=2`, i.e., that the user specifies the minimum coordination that can be tolerated?

zerothi:
> > I think I am convinced, vote to remove on_dangling all-together :)
> 
> Why remove it and not just `on_lone_atom="ignore"` as the default? All heteroribbons that I have seen in the literature don't have single bonded carbons, so it is surely something people have in mind when they build ribbons. I don't know if it should be called a "valid" ribbon or not, but I believe it is useful anyway.

Wouldn't it be better to provide methods that can check conditions? That would make it more versatile?

pfebrer:
> Wouldn't it be better to provide methods that can check conditions? That would make it more versatile?

Ok, I will do that. Something like `junction_plugin` that receives all the junction information and can modify the shifts, issue errors or whatever.

zerothi:
hang on ;)

How would you imagine this interface? I am thinking about something like this:

```python
any_geometry_creation_function(..., validator=func)
```
where the `validator` can validate the geometry. That would mean generality for future geometry creation functions?

@tfrederiksen comments?

Also, this may not need to be added now, just to add this later. :)

zerothi:
I.e. lets try and break the PR's down. A validator is quite a task. :)

tfrederiksen:
> I.e. lets try and break the PR's down. A validator is quite a task. :)

I agree. Such a tool seems a bit related to #202 where the idea was to add new atoms depending on the current coordination.

The more I think about it I believe a good solution for *this* PR could be to let the user choose the *minimum coordination* that is considered acceptable. Perhaps `min_coordination=2` could be the default (corresponding to the typical use case). Then all the checks that @pfebrer already implemented could come into use.

pfebrer:
If it's just a generic validator then it will lose the ability to provide meaningful messages about what shifts produce ribbons without hanging carbons. I will be able to just check if there are atoms with one neighbour and that's it.

The reason I wrote this function to begin with was actually that I didn't have to think each time how much do I need to shift my ribbon, because I want to do high-throughput with many different combinations. If the functionality to report why your junction is wrong and provide helpers that would make it right, then I will basically be back to where I started :(. Except for the ability to align, that would be a step forward at least.

zerothi:
> > I.e. lets try and break the PR's down. A validator is quite a task. :)
> 
> I agree. Such a tool seems a bit related to #202 where the idea was to add new atoms depending on the current coordination.
> 
> The more I think about it I believe a good solution for _this_ PR could be to let the user choose the _minimum coordination_ that is considered acceptable. Perhaps `min_coordination=2` could be the default (corresponding to the typical use case). Then all the checks that @pfebrer already implemented could come into use.

I was more thinking about the future. **IF** we are going to add a `validator` argument, then the `min_coordination` should be obsoleted, in which case I think it is not good to add it. So if you think a validator is a good idea, then we shouldn't do anything, just put the argument there and add a function call of the validator in the end.

pfebrer:
> The more I think about it I believe a good solution for this PR could be to let the user choose the minimum coordination that is considered acceptable. Perhaps min_coordination=2 could be the default (corresponding to the typical use case). Then all the checks that @pfebrer already implemented could come into use.

Would `min_coordination` values other than `1` and `2` make sense? If not, that seems like it could be a boolean argument as well.

tfrederiksen:
> Would `min_coordination` values other than `1` and `2` make sense? If not, that seems like it could be a boolean argument as well.

Good point. The advantage of an integer here is perhaps just that it seems very intuitive (`min_coordination` is quite self-explanatory, I think).

zerothi:
> If it's just a generic validator then it will lose the ability to provide meaningful messages about what shifts produce ribbons without hanging carbons. I will be able to just check if there are atoms with one neighbour and that's it.
> 
> The reason I wrote this function to begin with was actually that I didn't have to think each time how much do I need to shift my ribbon, because I want to do high-throughput with many different combinations. If the functionality to report why your junction is wrong and provide helpers that would make it right is gone, then I will basically be back to where I started :(. Except for the ability to align, that would be a step forward at least.

I agree that one looses some functionality. But... A validator can provide more detailed information such as which atoms, their positions and neighbours etc. In that case it should be pretty fast to track down. In either case, a high-throughput solution requires no intervention, and thus you don't really need the exact segment/section information anyway, no?

Another way would be to have 2 validators, one for the full geometry, and one that checks two sections before they are attached? Hmm...

pfebrer:
> Another way would be to have 2 validators, one for the full geometry, and one that checks two sections before they are attached? Hmm...

> A validator can provide more detailed information such as which atoms, their positions and neighbours etc. In that case it should be pretty fast to track down.

I would say it's not so fast to track down. You need to plot the geometry, check which is the atom with that index, check the shift that you provided (if you directly called the function) and then think about which shift would be right.

> In either case, a high-throughput solution requires no intervention, and thus you don't really need the exact segment/section information anyway, no?

Well, it would be nice to debug honestly. In my (small) experience generic errors that pop up from deep within the stack are hard to debug and really annoying, specially if you are not the developer.

> Another way would be to have 2 validators, one for the full geometry, and one that checks two sections before they are attached?

Yes, that's what I thought as well. The latter is a specific thing of this function and could implement the things I'm referring to, the latter is generic.

zerothi:
> > Another way would be to have 2 validators, one for the full geometry, and one that checks two sections before they are attached?
> 
> Yes, that's what I thought as well. The latter is a specific thing of this function and could implement the things I'm referring to, the former is generic.

I think this solution would be acceptable in this and future geometry creation methods.

pfebrer:
As the last commit says, I've changed the paradigm for how it works behind the scenes so that it is hopefully flexible enough to adapt to what we've been talking. The interface remains exactly the same. Let's see if you think it's better or worse, we can always revert.

Basically, **all the work is done by the section objects themselves**, who know how to append to the previous section. Then there's the generic `composite_geometry` that just manages the loop to make the sections adhere one after the other. Benefits of this (as I see it):
- Maybe in the future a heteroribbon section can understand how to merge with another type of section.
- Since you can change the section class that you pass to `composite_geometry`, you can always subclass `heteroribbon_section` to implement whatever you need (e.g. add some checks or try to implement how to manage lattice mismatches).

At this point the discussion may have gotten too deep to rush the PR into the next release, so it might be better to just give it some more time to make sure things are done properly? :thinking: For example it could be useful to try to implement other composite geometries and see if the design holds or it is useless.

tfrederiksen:
Amazing, well done @pfebrer ! Just two really minor comments:
- The new heterostructure functions do not show up in the list of available geometries when calling `help(sisl.geom)`.
- `1j` is still present in the documentation.

pfebrer:
Thanks! Yes I know, I didn't finish the last details since I'm not sure this is actually useful, so I just wanted to have a proof of concept :sweat_smile:

zerothi:
I am playing with this and trying to amend some things to clarify the intent of the code.
Hold on ;)

zerothi:
Sorry, I accidentially pushed to your branch, I'll revert!

zerothi:
After discussing with @pfebrer we have come up with some other ideas that @pfebrer will start exploring. :)

zerothi:
@pfebrer we had a meeting and you were doing some explorations, but now I can't recall those explorations? What is the status here?

zerothi:
Great, thanks for the reminder. 

Ok, so what about I run it through again, and then we just merge this? Then we can always think about the next thing.

After having read your recap of our conversation, perhaps the most ideal thing would be to fix `Geometry.add/append` to accept a class that *fixes* the merging instead of relying on axis fixtures etc.

pfebrer:
> Ok, so what about I run it through again, and then we just merge this? Then we can always think about the next thing.

What do you mean by "run it through"?

> After having read your recap of our conversation, perhaps the most ideal thing would be to fix Geometry.add/append to accept a class that fixes the merging instead of relying on axis fixtures etc.

Yes but that would require the `Geometry` itself to have some knowledge of which type of geometry it is. Otherwise this merging class will have no idea what to do with it.

zerothi:
> > Ok, so what about I run it through again, and then we just merge this? Then we can always think about the next thing.
> 
> What do you mean by "run it through"?

I'll check this PR again.

 
> > After having read your recap of our conversation, perhaps the most ideal thing would be to fix Geometry.add/append to accept a class that fixes the merging instead of relying on axis fixtures etc.
> 
> Yes but that would require the `Geometry` itself to have some knowledge of which type of geometry it is. Otherwise this merging class will have no idea what to do with it.

True, hmm...

zerothi:
I am thinking this could be helpful. So I think we should merge it. :)

zerothi:
Any comments @tfrederiksen @pfebrer

pfebrer:
The only pitfall I see is that this introduces some kind of "framework" for composite structures, which might be a pain in the future if you want to go in another direction. But if you think it's fine, then OK 👍

pfebrer:
Ok, is it properly documented (can't explore it now I'm on the metro)? If it is then for me it is ok to merge it. Otherwise I can try to document it this afternoon or tomorrow.

zerothi:
I'll have a go and ask for another go by you. I think the section classes could be hosted in the function as a property
```python
def heteroribbon(...):


heteroribbon.section = _heteroribbon_section
```
or similar.

zerothi:
@pfebrer feel free to have a look now :)

pfebrer:
Yeah I think it's fine! Thank you for the polishing :)

PR #424, opened 2022-01-17
mnt: bumped Python requirement

zerothi:
This allows namedtuples with defaults and also
aligns with numpy requirements.


zerothi:
@pfebrer @tfrederiksen note that I will bump to 3.7. FYI

Issue #310, opened 2021-03-04
numpy.ndarray binary incompatibility error in environments with `numpy<1.20`


bosonie:
After upgrading sisl to 0.11.0 I can not import the package anymore.

The call:
``from sisl._indices import indices_le``
returns:
``ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject``

OS Fedora 31
Python 3.7.6
numpy 1.17.5 (and others, see below)

Tried to reinstall numpy (same version) as suggested on the web, but couldn't solve the problem.
I am bound to this old numpy version due to other packages requirement, but I also tried in a new environment (same python version) numpy 1.18 and 1.19. Got the same error. Version of numpy 1.20, on the contrary, works.

zerothi:
Thanks, so are you using conda? Because installing sisl should use your local numpy version (there are no hard-coded numpy versions).

If you build `sisl` locally it uses your currently installed numpy version?

bosonie:
I'm using `virtualenv`. The new environment has no `numpy` and therefore it is installed as a dependence of sisl (numpy>=1.13 is the requirement).
In any other environment (for example the local env where I have numpy 1.17.5), it uses the installed numpy.
I tried both installing with PyPI or locally after `git clone`. The result is the same. Unless U=I have numpy 1.20, I get the same error.

zerothi:
It really sounds like the build is using another numpy version from your installed one.

There is _nothing_ hard-coded regarding the things you mention, the pyx files are cythonized at build time against your numpy + cython and compiler.

Please give more details
- build output
- module versions + paths as printed once loaded in python

bosonie:
I'm not sure what you require. But let's start from the very basic. I have a new `virtualenv` called `www`. I first install numpy 1.17.5 (pip install numpy==1.17.5) and then I do `pip install sisl`. The output is:
```
Collecting sisl
  Downloading sisl-0.11.0.tar.gz (4.1 MB)
     |████████████████████████████████| 4.1 MB 1.7 MB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Requirement already satisfied: setuptools in ./lib/python3.7/site-packages (from sisl) (54.0.0)
Collecting netCDF4
  Downloading netCDF4-1.5.6-cp37-cp37m-manylinux2014_x86_64.whl (4.7 MB)
[...]

PR #427, opened 2022-01-28
Github action to build wheels and test them

pfebrer:
This is what I have until now:
- **Linux**: Wheels build succesfully and pass tests for all python versions.
- **MacOS**: Wheels build successfully but don't pass tests.
- **Windows**: For some python versions wheels build successfully, but don't pass tests.

You can see the workflows logs here: https://github.com/pfebrer/sisl/actions


pfebrer:
Yes sure, it was just to do a quick test, mac wheels building and tests passing now :heavy_check_mark: 

Let's see if it works for windows with the last commit :) (the windows thing is an obsession now hahaha I must get it right before I die!)

zerothi:
> Yes sure, it was just to do a quick test, mac wheels building and tests passing now heavy_check_mark
> 
> Let's see if it works for windows with the last commit :) (the windows thing is an obsession now hahaha I must get it right before I die!)

OK, great! :)

pfebrer:
I submitted https://github.com/adang1345/delvewheel/issues/26 as my last attempt to success, otherwise I give up on windows, I don't understand it :)

zerothi:
Ok, my similar experience. Hard to debug when you don't have experience.

I suggest we remove the windows builds and possibly the macos arm64 (although they will soon be required). The error for arm64 seems to be because of wrong os being setup.:
```

cibuildwheel version 2.3.1

Build options:
  platform: 'macos'
  architectures: {<Architecture.x86_64: 'x86_64'>}
  build_selector: BuildSelector(build_config='cp39-macosx_arm64')
  output_dir: PosixPath('wheelhouse')
  package_dir: PosixPath('.')
  test_selector: TestSelector(build_config='*')
  before_all: ''
  before_build: ''
  before_test: ''
  build_frontend: 'pip'
  build_verbosity: 0
  dependency_constraints: DependencyConstraints(PosixPath('/usr/local/Cellar/pipx/1.0.0/.cache/dca626a67d99499/lib/python3.10/site-packages/cibuildwheel/resources/constraints.txt'))
  environment: ParsedEnvironment([])
  manylinux_images: None
  musllinux_images: None
  repair_command: 'delocate-listdeps {wheel} && delocate-wheel --require-archs {delocate_archs} -w {dest_dir} {wheel}'
  test_command: 'python -c "import sisl; print(sisl.__version__)" pytest {package}\n'
  test_extras: '[test]'
  test_requires: ['numpy>=1.13', 'scipy>=0.18', 'netCDF4', 'pyparsing>=1.5.7']

Here we go!

cibuildwheel: No build identifiers selected: BuildSelector(build_config='cp39-macosx_arm64')
```

But again, I have no idea... :(

zerothi:
Note I rebased to remove the faulty commit. :)

pfebrer:
The tests passed for `windows_amd64`!!! :) :) :) :) :) :)

> I suggest we remove the windows builds and possibly the macos arm64 (although they will soon be required). The error for arm64 seems to be because of wrong os being setup.:

Now it shows a different error, the `gcc` compiler does not support `arm64`.

Could you check the `cp-38-win_amd64` wheel? It fails because an invalid option being passed to the compiler:
```
gfortran.exe: error: unrecognized command-line option '-mno-cygwin'; did you mean '-mno-clwb'?
```
A quick search showed me that this flag was removed from gcc long ago. I think this flag is set by `f2py`, do you know how to remove it?

zerothi:
> The tests passed for `windows_amd64`!!! :) :) :) :) :) :)

GREAT!

> > I suggest we remove the windows builds and possibly the macos arm64 (although they will soon be required). The error for arm64 seems to be because of wrong os being setup.:
> 
> Now it shows a different error, the `gcc` compiler does not support `arm64`.
> 
> Could you check the `cp-38-win_amd64` wheel? It fails because an invalid option being passed to the compiler:
> 
> ```
> gfortran.exe: error: unrecognized command-line option '-mno-cygwin'; did you mean '-mno-clwb'?
> ```
> 
> A quick search showed me that this flag was removed from gcc long ago. I think this flag is set by `f2py`, do you know how to remove it?

No, I don't, it is hard-coded in numpy...
I don't know if a `setup.cfg` could be used to bypass this, it must be known to numpy, however, I can't immediately find anything?
https://github.com/numpy/numpy/search?o=desc&q=mno-cygwin&s=created&type=issues

pfebrer:
> Could you check the cp-38-win_amd64 wheel? It fails because an invalid option being passed to the compiler:

I think this might be because the oldest numpy version for python `3.8` (`1.17` I think) adds this flag, and the oldest supported version for python `3.9` doesn't. So maybe just asking for a newer version of numpy in the windows build will solve the problem. I will try.

pfebrer:
Yeah the problem that we are seeing is definitely this: https://github.com/numpy/numpy/pull/16468

pfebrer:
Asking for higher numpy versions seems to work, wheels are building across all python versions for windows. I have two questions:

- Do you know how to specify these numpy versions only for windows and keep `"oldest-supported-numpy"` for linux and mac?
- The optional dependencies in `pyproect.toml` don't seem to work, as this line appears in the logs:
  ```
  WARNING: sisl 0.1.dev1+g76b544f does not provide the extra 'test'
  ```

  If I try locally `pip install .[viz]` it also tells me that it does not provide the extra 'viz'. Can you try to solve this? :)

pfebrer:
**UPDATE**: 

:heavy_check_mark: Tests now are able to read the `sisl-files` test files 
:heavy_check_mark:  Python 3.10 specific bugs are solved 

The only problems left are (you can ignore the previous comment, this includes everything that is wrong now):
- **Windows**:
  - Tests that need to write files fail because of permission errors, most likely a CI specific problem.
  - Tests that make use of `np.float128` or `np.int64` fail. This is because **there are no such data types** in windows' `numpy`. The limits are `np.float64` and `np.int32`. You can see the tests that fail here: https://github.com/pfebrer/sisl/runs/5012095967?check_suite_focus=true, after the permission errors.
- **Mac**:
  - Some tests at `test_selector.py` fail. The tests that fail are not always the same, this doesn't seem like a deterministic thing. See how [here](https://github.com/pfebrer/sisl/actions/runs/1774673988) the python 3.7 run succeeds, but in the [next run](https://github.com/pfebrer/sisl/actions/runs/1774776095) some of these tests fail. The python 3.9 run does the opposite.

zerothi:
Thanks!

> The only problems left are (you can ignore the previous comment, this includes everything that is wrong now):
> 
>     * **Windows**:
>       
>       * Tests that need to write files fail because of permission errors, most likely a CI specific problem.

Yeah, perhaps the tmp directory is not the way to do it in windows...

>       * Tests that make use of `np.float128` or `np.int64` fail. This is because **there are no such data types** in windows' `numpy`. The limits are `np.float64` and `np.int32`. You can see the tests that fail here: https://github.com/pfebrer/sisl/runs/5012095967?check_suite_focus=true, after the permission errors.

Yes, I have read about this elsewhere. I'll try and go through them so we can mark them skipped for os == win.
 
>     * **Mac**:
>       
>       * Some tests at `test_selector.py` fail. The tests that fail are not always the same, this doesn't seem like a deterministic thing. See how [here](https://github.com/pfebrer/sisl/actions/runs/1774673988) the python 3.7 run succeeds, but in the [next run](https://github.com/pfebrer/sisl/actions/runs/1774776095) some of these tests fail. The python 3.9 run does the opposite.

Yes, I recall this problem of mac... I don't know why mac tests are so stupidly slow, this test is *random* in the sense that it checks for correctness of something that is varying in time. And the test is separating them by ~ 0.5 second, which might not be enough on mac.

zerothi:
Could you rebase and try again?

pfebrer:
> Yes, I recall this problem of mac... I don't know why mac tests are so stupidly slow, this test is random in the sense that it checks for correctness of something that is varying in time. And the test is separating them by ~ 0.5 second, which might not be enough on mac.

Wow hahahah. Maybe the hardware for the MAC CI is 20 years old or something. It could be an iPod touch :laughing:

pfebrer:
MAC tests passing now :heavy_check_mark: 

Apart from the permission errors, there's one test failing for sparse matrices in windows, could you check if you know what is going on?

pfebrer:
This might be the reason for the permission errors: https://stackoverflow.com/a/23212515/11991536.

pfebrer:
Ok, now I'm sure that the problem is just that some files remain open and that's the problem since windows doesn't allow to open a file that is already open. I guess this is `netcdf4` keeping files open for some reason??

I tried changing the names of the problematic files, and permission errors were gone: https://github.com/pfebrer/sisl/runs/5014709101?check_suite_focus=true

Then I tried raising an error if `f.unlink()` fails in `sisl_tmp`, and I get lots of errors like this one for netcdf files:
```
 ________________ ERROR at teardown of test_nc_dynamical_matrix ________________
  
  self = <sisl.conftest.sisl_tmp.<locals>.FileFactory object at 0x000001C4E81F7250>
  
      def teardown(self):
          while len(self.files) > 0:
              # Do each removal separately
              f = self.files.pop()
              if f.is_file():
                  try:
                      f.unlink()
                  except Exception as e:
  >                   raise e
  
  Users\runneradmin\AppData\Local\Temp\tmpkbz2h663\lib\site-packages\sisl\conftest.py:90: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  Users\runneradmin\AppData\Local\Temp\tmpkbz2h663\lib\site-packages\sisl\conftest.py:88: in teardown
      f.unlink()
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  
  self = WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/grdyn.nc')
  missing_ok = False
  
      def unlink(self, missing_ok=False):
          """
          Remove this file or link.
          If the path is a directory, use rmdir() instead.
          """
          try:
  >           self._accessor.unlink(self)
  E           PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\grdyn.nc'
```

So that's the problem, any idea how to fix it? :)

zerothi:
Could you try and rebase? I have fixed some of the things you asked for (hopefully all).

pfebrer:
Only two errors to go:

```
____________________________ test_nc_multiple_fail ____________________________
  
  sisl_tmp = <sisl.conftest.sisl_tmp.<locals>.FileFactory object at 0x000001CD0D573A90>
  sisl_system = <sisl.conftest.sisl_system.<locals>.System object at 0x000001CD08D645B0>
  
      def test_nc_multiple_fail(sisl_tmp, sisl_system):
          # writing two different sparse matrices to the same
          # file will fail
          f = sisl_tmp('gr.nc', _dir)
          H = Hamiltonian(sisl_system.gtb)
          DM = DensityMatrix(sisl_system.gtb)
      
  >       with ncSileSiesta(f, 'w') as sile:
  
  Users\runneradmin\AppData\Local\Temp\tmp1nsh8s6w\lib\site-packages\sisl\io\siesta\tests\test_siesta.py:61: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  Users\runneradmin\AppData\Local\Temp\tmp1nsh8s6w\lib\site-packages\sisl\io\sile.py:725: in __init__
      self.__dict__['fh'] = _netCDF4.Dataset(str(self.file), self._mode,
  src\netCDF4\_netCDF4.pyx:2307: in netCDF4._netCDF4.Dataset.__init__
      ???
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  
  >   ???
  E   PermissionError: [Errno 13] Permission denied: b'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\gr.nc'
  
  src\netCDF4\_netCDF4.pyx:1925: PermissionError
  ______________ TestObject.test_read_write_grid[tsvncSileSiesta] _______________
  
  self = <sisl.io.tests.test_object.TestObject object at 0x000001CD13BB41F0>
  sisl_tmp = <sisl.conftest.sisl_tmp.<locals>.FileFactory object at 0x000001CD0E817D60>
  sisl_system = <sisl.conftest.sisl_system.<locals>.System object at 0x000001CD08D645B0>
  sile = <class 'sisl.io.siesta.tsvncSileSiesta'>
  
      @pytest.mark.filterwarnings("ignore", message="*gridncSileSiesta.read_grid cannot determine")
      @pytest.mark.parametrize("sile", _my_intersect(["read_grid"], ["write_grid"]))
      def test_read_write_grid(self, sisl_tmp, sisl_system, sile):
          g = sisl_system.g.rotate(-30, sisl_system.g.cell[2, :])
          G = Grid([10, 11, 12])
          G[:, :, :] = np.random.rand(10, 11, 12)
      
          f = sisl_tmp("test_read_write_grid.win", _dir)
          # Write
          try:
  >           sile(f, mode="w").write_grid(G)
  
  Users\runneradmin\AppData\Local\Temp\tmp1nsh8s6w\lib\site-packages\sisl\io\tests\test_object.py:357: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  Users\runneradmin\AppData\Local\Temp\tmp1nsh8s6w\lib\site-packages\sisl\io\sile.py:725: in __init__
      self.__dict__['fh'] = _netCDF4.Dataset(str(self.file), self._mode,
  src\netCDF4\_netCDF4.pyx:2307: in netCDF4._netCDF4.Dataset.__init__
      ???
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  
  >   ???
  E   PermissionError: [Errno 13] Permission denied: b'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_read_write_grid.win'
  
  src\netCDF4\_netCDF4.pyx:1925: PermissionError
```
The first one I believe it is because `test_nc1` and `test_nc2` have a line like `tb.write(sile(..., "w"))`.

zerothi:
Thanks, try now... :)

pfebrer:
Thank youuu! We are so close :) Only one test failing for `windows` and `python <= 3.8`. Seems like a parsing error:

```
 ______________ TestObject.test_read_write_geometry[chgSileVASP] _______________
  
  self = <sisl.io.tests.test_object.TestObject object at 0x000002251766C708>
  sisl_tmp = <sisl.conftest.sisl_tmp.<locals>.FileFactory object at 0x000002251766CB88>
  sisl_system = <sisl.conftest.sisl_system.<locals>.System object at 0x0000022514F7BE88>
  sile = <class 'sisl.io.vasp.chgSileVASP'>
  
      @pytest.mark.parametrize("sile", _my_intersect(["read_geometry"], ["write_geometry"]))
      def test_read_write_geometry(self, sisl_tmp, sisl_system, sile):
          G = sisl_system.g.rotate(-30, sisl_system.g.cell[2, :])
          G.set_nsc([1, 1, 1])
          f = sisl_tmp("test_read_write_geom.win", _dir)
          # These files does not store the atomic species
          if issubclass(sile, (_ncSileTBtrans, deltancSileTBtrans)):
              return
          # Write
          sile(f, mode="w").write_geometry(G)
          # Read 1
          try:
              with sile(f, mode='r') as s:
                  g = s.read_geometry()
              assert g.equal(G, R=False, tol=1e-3) # pdb files have 8.3 for atomic coordinates
          except UnicodeDecodeError as e:
              pass
          # Read 2
          try:
              with sile(f, mode='r') as s:
  >               g = Geometry.read(s)
  
  Users\runneradmin\AppData\Local\Temp\tmp2kx2lv5b\lib\site-packages\sisl\io\tests\test_object.py:254: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  Users\runneradmin\AppData\Local\Temp\tmp2kx2lv5b\lib\site-packages\sisl\geometry.py:658: in read
      return sile.read_geometry(*args, **kwargs)
  Users\runneradmin\AppData\Local\Temp\tmp2kx2lv5b\lib\site-packages\sisl\io\sile.py:523: in pre_open
      return func(self, *args, **kwargs)
  Users\runneradmin\AppData\Local\Temp\tmp2kx2lv5b\lib\site-packages\sisl\io\vasp\car.py:149: in read_geometry
      sc = self.read_supercell()
  Users\runneradmin\AppData\Local\Temp\tmp2kx2lv5b\lib\site-packages\sisl\io\sile.py:514: in pre_open
      return func(self, *args, **kwargs)
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  
  self = <sisl.io.vasp.chgSileVASP object at 0x000002251766C948>
  
      @sile_fh_open(True)
      def read_supercell(self):
          """ Returns `SuperCell` object from the CONTCAR/POSCAR file """
      
          # read first line
          self.readline()  # LABEL
          # Update scale-factor
  >       self._scale = float(self.readline())
  E       ValueError: could not convert string to float: '\x1a\n'
  
  Users\runneradmin\AppData\Local\Temp\tmp2kx2lv5b\lib\site-packages\sisl\io\vasp\car.py:127: ValueError
```
Could you make it fail for this specific case? That should be it. I'm so looking forward to seeing all green hehe

zerothi:
rebase and try ;)

pfebrer:
The xfail mark doesn't seem to work :(

pfebrer:
I think it needs to be:

```python
if sys.platform.startswith("win") and issubclass(sile, chgSileVASP):
    pytest.xfail("Windows reading/writing supercell fails for some unknown reason")
```

zerothi:
> I think it needs to be:
> 
> ```python
> if sys.platform.startswith("win") and issubclass(sile, chgSileVASP):
>     pytest.xfail("Windows reading/writing supercell fails for some unknown reason")
> ```

Yes, discovered it simultaneously, committed, rebase, 

Could you also name the action "Build wheels" so it is clear what it does?

pfebrer:
Done. By the way, if it works, the only thing left would be this:

> Do you know how to specify these numpy versions only for windows and keep "oldest-supported-numpy" for linux and mac?

zerothi:
> Done. By the way, if it works, the only thing left would be this:
> 
> > Do you know how to specify these numpy versions only for windows and keep "oldest-supported-numpy" for linux and mac?
> 
> Any idea?

Could you remind my why this was the case? If this is generic, we should probably report it upstream with oldest-supported-numpy.

However, I'll add the things to pyproject.toml and setup.cfg ASAP.

pfebrer:
This was because `numpy <= 1.17` had a bug comparing major versions of `GCC`. They compared versions as strings and since `"11"` < `"4"`, they added a flag to `gcc` that is deprecated since `gcc v4`. :sweat_smile:

zerothi:
Note, that I am pretty sure the versions you specify will trigger the problem seen in #310 for users who have older numpy versions installed.

So I am afraid this will cause other problems. In this case we just know the problem, so perhaps ok.

pfebrer:
Not if we bump the numpy requirement for installation. But I was wondering if we could do that only for windows. I've tried to find out how to do it but I don't seem to find a solution.

pfebrer:
I see you have already done that :)

zerothi:
Ah, you want to bump the required numpy version for windows only... I suppose we could do that, for convenience.

Hold on, there were some typos in the latest commit

pfebrer:
> No, the problem arise because one builds sisl against a newer version of numpy than what is currently installed. So if they pip install sisl, it will use numpy >1.19, but they can have numpy 1.16 installed. And then it will break.

That's what I mean, in the installation dependencies we should also have `numpy > 1.19` for windows.

Something like:

```
"numpy >= 1.13; sys_platform != 'win32'",
"numpy >= 1.19.5 ; sys_platform == 'win32' ; python_version <= '3.9'"
"numpy >= 1.21.5 ; sys_platform == 'win32' ; python_version >= '3.10'"
```

zerothi:
try again

pfebrer:
By the way, is ''win32' an alias for all windows? Because the wheels that we have are for the 64bit version

zerothi:
https://stackoverflow.com/questions/2144748/is-it-safe-to-use-sys-platform-win32-check-on-64-bit-python

pfebrer:
Screenshot from 2022-02-01 12-40-42

I'm going to frame this picture and hang it on my bedroom's wall.

sisl/conftest.py
```diff
@@ -5,6 +5,7 @@
 
 import contextlib
 import os
+import sys
```
zerothi, line 8: Please remove this edit, not needed.

zerothi:
great work!!! @pfebrer :rocket:

pfebrer:
Done, could you update the `on` part of the workflow to whatever you want, merge and trigger the workflow from your user? I don't get the `Run workflow` button in my repo but I think it's just because the workflow already existed.

zerothi:
> Done, could you update the `on` part of the workflow to whatever you want, merge and trigger the workflow from your user? I don't get the `Run workflow` button in my repo but I think it's just because the workflow already existed.

Yeah, I need to play with this, and add tokens etc. for it to actually work. 
One thing I can't immediately see here is whether the sources are also uploaded? But I can check this, and probably manually do this.

Once this is completed we can post 0.12.1 release.

pfebrer:
The wheels are uploaded in the `artifacts` zip file. You can just download them and upload them manually along with the source distribution to pypi for this time, and  leave the full CI deployment for the future (?). 

https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/#manual-method

pfebrer:
Well I guess the "automatic method" on that guide is not that difficult, whatever you prefer :)

pfebrer:
On the "automatic method" you can see that they add a step to build the source distribution only on ubuntu.

pfebrer:
I added the steps for deployment. This is how the workflow looks right now:

Screenshot from 2022-02-01 18-27-57

- It builds both wheels (tested) and the source distribution.
- If it succeeds, it then uploads everything to `testpypi`.
- If it succeeds, it installs `sisl` from `testpypi` and runs the tests again (for windows, ubuntu and mac but only with python 3.8).
- If the installation pass the tests, it finally publishes the files to `pypi`.

You only need to set the `TESTPYPI_TOKEN` and `PYPI_TOKEN` secrets, and then trigger the workflow. The only thing I'm not sure if it will work is the 3rd step, since I haven't been able to test it. But it is quite straightforward, I believe there will be no problem.

zerothi:
> How can I force a version name to test uploading to testpypi? I'm trying to test it, but I get `'0.1.dev1+g71ee4bb' is an invalid value for Version`

I don't really know... :( But definitely we should solve this...

zerothi:
> > How can I force a version name to test uploading to testpypi? I'm trying to test it, but I get `'0.1.dev1+g71ee4bb' is an invalid value for Version`
> 
> I don't really know... :( But definitely we should solve this...

I got it, simply do:

```bash
export SETUPTOOLS_SCM_PRETEND_VERSION=0.0.1
```
and it will use that version, regardless!!! :)
I don't have time now, so I'll have to look at this tomorrow.

pfebrer:
That's very nice!

zerothi:
@pfebrer I have some problems with the ci tests:
Please have a look here.
https://github.com/zerothi/sisl/runs/5122380937?check_suite_focus=true

The problem is that release candidates are not defaulted to be downloaded using pip, so one has to extract the version and put it in... So what do we do here? How can I get `sisl[test]==${{ github.ref[1:] }}`?  Do you know?

pfebrer:
You already solved this, right? At least for mac it seems

pfebrer:
Could it be that the problem with ubuntu is just that the wheels weren't available yet (by just some seconds) when the command tries to install them? Maybe just letting the CI sleep for 10 seconds can solve the problem :sweat_smile:

zerothi:
> Could it be that the problem with ubuntu is just that the wheels weren't available yet (by just some seconds) when the command tries to install them? Maybe just letting the CI sleep for 10 seconds can solve the problem 😅

Yeah, that's my current idea as well.. Just seems stupid... ;)

zerothi:
And then I tried adding a sleep, and the results are reversed, now mac is failing... damn this.I think we cannot test them like that... :(

pfebrer:
I tried to use it on my computer and it works :tada: It was quite satisfying to install `sisl` from `pip` in 10s and without dependencies :))

PR #430, opened 2022-01-31
Fixing python 3.10 viz tests

pfebrer:
As mentioned in #427, some viz tests failed on python 3.10. After checking what was the problem, what I don't understand is why didn't they fail with other python versions :sweat_smile:


pfebrer:
Hmm it was just to be safe. I don't think anyone will implement a under method and expect the arguments to be filled with defaults using the current instance settings.

Actually, I don't think this wrapping is a good idea at all. It complicates things and makes stack traces hard to follow. I did it in a time when I liked magic, but I guess I have grown out of that and now I prefer robustness :sweat_smile:. In the future I would like to somehow replace this for some other approach.

PR #439, opened 2022-02-13
Construction of common surface slabs

tfrederiksen:
Although available with other tools (like ASE) I think it would be handy to have native sisl functions to create common surface slabs.

Here I implemented, to begin with, the three simple fcc surfaces: `fcc100`, `fcc110`, and `fcc111`. If my approach is suitable I can extend with the bcc slabs, add tests, etc.


sisl/geom/flat.py
```diff
@@ -70,3 +70,124 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+        g.set_nsc([3, 3, 1])
+    else:
+        g.set_nsc([3, 3, 3])
+    if np.all(g.maxR(True) > 0.):
+        g.optimize_nsc()
+    return g
+
+
+def fcc100(alat, atoms, size=(1, 1, 2), vacuum=None):
+    """ Face-centered cubic (100) surface
```
zerothi, line 88: From this documentation it isn't clear to me along which lattice vector the surface is?

Also, since this is a two-layer structure we are in the situation that one might want to end on an A, or a B?
Instead of using `str` as choice, perhaps integers would be ok? so `end_layer=0|1` or something, the same applies to the other surfaces.

sisl/geom/flat.py
```diff
@@ -70,3 +70,124 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+    # A-layer
+    g = Geometry([0, 0, 0], atoms=atoms, sc=sc).tile(size[2], 2)
+    # B-layer
+    g.xyz[1::2] += (sc.cell[0] + sc.cell[1]) / 2
+    g = _finish_slab(g, size, vacuum)
+    return g
+
+
+def fcc110(alat, atoms, size=(1, 1, 2), vacuum=None):
+    """ Face-centered cubic (110) surface
```
zerothi, line 117: Same as above.

sisl/geom/flat.py
```diff
@@ -70,3 +70,124 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+    vacuum : float, optional
+        distance added to the third lattice vector to separate
+        the slab from its periodic images
+
+    See Also
+    --------
+    fcc110 : construction of fcc(110) surface
+    fcc111 : construction of fcc(111) surface
+    """
+    sc = SuperCell(np.array([np.sqrt(0.5), np.sqrt(0.5), 0.5]) * alat)
```
zerothi, line 107: here I would prefer
`np.array([0.5 ** 0.5, 0.5 ** 0.5, 0.5]) * alat` for just a bit clearer reading.

sisl/geom/flat.py
```diff
@@ -70,3 +70,124 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+        if True returns an orthogonal lattice
+
+    See Also
+    --------
+    fcc100 : construction of fcc(100) surface
+    fcc110 : construction of fcc(110) surface
+    """
+    if orthogonal:
+        # 2-atom basis per layer
+        sc = SuperCell(np.array([np.sqrt(0.5), np.sqrt(0.375) * 2, 1 / np.sqrt(3)]) * alat)
```
zerothi, line 169: in this case one could do
`np.array([0.5, 0.375, 1/3]) ** 0.5`?

sisl/geom/flat.py
```diff
@@ -70,3 +70,124 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+    sc = SuperCell(np.array([1., np.sqrt(0.5), np.sqrt(0.125)]) * alat)
+    # A-layer
+    g = Geometry([0, 0, 0], atoms=atoms, sc=sc).tile(size[2], 2)
+    # B-layer
+    g.xyz[1::2] += (sc.cell[0] + sc.cell[1]) / 2
+    g = _finish_slab(g, size, vacuum)
+    return g
+
+
+def fcc111(alat, atoms, size=(1, 1, 3), vacuum=None, orthogonal=False):
```
zerothi, line 145: Same as above

zerothi (changes requested):
Great addition!

To not confuse with the other things, I would prefer if this was in a separate file, `surfaces`, or something like that?

Also, would it make sense to change the methods to `fcc_slab(..., miller=(1, 1, 0))` ? In this way we could easily add more without causing too many problems?
Or what do you think here?

tfrederiksen:
> Great addition!
>
> To not confuse with the other things, I would prefer if this was in a separate file, `surfaces`, or something like that?

OK, I'll then introduce `surfaces.py`.

> Also, would it make sense to change the methods to `fcc_slab(..., miller=(1, 1, 0))` ? In this way we could easily add more without causing too many problems? Or what do you think here?

I agree that collecting the above in one more general `fcc_slab` function seems attractive. I'll give it a try.

sisl/geom/flat.py
tfrederiksen: We could introduce another optional parameter `axis=2` to select along which lattice vector the layers are grown?

Regarding the layer growth one could also have `layer_offset=0` which simply circles through the possibilities for the first growth layer (A/B for 100 and 110 and A/B/C for 111)?

sisl/geom/flat.py
tfrederiksen: OK

sisl/geom/flat.py
zerothi: Agreed!  

The choice of what to signify as the *important* layer is also not clear. For instance for TS I would mostly like to control the end-surface layer, I don't care too much about the start-surface layer. And when testing different number of layers it seems rather annoying to have to calculate `size` depending on what your initial layer and the number of layers you want. In other cases the initial layer would be important (or just the other electrode).

So perhaps it would be best if one could do something like:
```python
def fcc_slab(alat, atoms, start=None, end=None, ...):
    # possibly layer_start/layer_end
    if start is not None and end is not None:
         raise ValueError("Only one of start or end may be supplied")
```
Then one could easily figure out how to define the full slab?

pfebrer:
Hey, this will be bery nice to have!

What do you think of a `slab` dispatcher in the functions that create bulk lattices? I.e.:

```python
sisl.geom.fcc.slab("111", ...)
```

Or something similar

zerothi:
> Hey, this will be bery nice to have!
> 
> What do you think of a `slab` dispatcher in the functions that create bulk lattices? I.e.:
> 
> ```python
> sisl.geom.fcc.slab("111", ...)
> ```
> 
> Or something similar
> 
> EDIT: Ok, I see that you already proposed it, nice! hahah

But when we discussed about the `heteroribbons` we agreed that dispatchers for methods is too obfuscated for end users? Right? So which way to go?

Should geometry creators really be classes? (Note this is how Pymatgen does it).

pfebrer:
Right, what would be the point of making them classes though? If `fcc` is a class, then methods such as `slab` would all be class methods. And an instance of that class would never be created, since the job of a geometry creator is to return a `Geometry`. So to me it sounds more like `fcc` should then be a module, not a class. Shouldn't it?

tfrederiksen:
> What do you think of a `slab` dispatcher in the functions that create bulk lattices? I.e.:
> 
> ```python
> sisl.geom.fcc.slab("111", ...)
> ```

I think this looks a bit complicated from the user perspective. I would prefer a simple function `fcc_slab(alat, atoms, '111')` that returns a geometry.

sisl/geom/flat.py
tfrederiksen: What do you think about `fcc_slab(alat, atoms, miller, start='A', end=None, ...)` as the default? With the idea that either `start` or `end` can be a string `A`, `B`, or `C`?

sisl/geom/surfaces.py
```diff
@@ -0,0 +1,116 @@
...
+    orthogonal : bool, optional
+        if True returns an orthogonal lattice
+    offset : int, optional
+        index to shuffle the layer sequence, eg ABC to BCA
+    """
+    if isinstance(miller, int):
+        miller = str(miller)
+
+    if isinstance(miller, str):
+        miller = (int(miller[0]), int(miller[1]), int(miller[2]))
```
zerothi, line 40: we should check that `len(miller) == 3`

sisl/geom/surfaces.py
```diff
@@ -0,0 +1,116 @@
...
+            # slide ABC layers relative to each other
+            B = (offset + 2) % 6
+            C = (offset + 4) % 6
+            vec = 1.5 * sc.cell[0] + sc.cell[1] / 2
+            for i in range(2):
+                g.xyz[B+i::6] += vec / 3 - i % 2 * sc.cell[0]
+                g.xyz[C+i::6] += 2 * vec / 3 - sc.cell[0]
+
+    else:
+         raise ValueError("The desired Miller index is not implemented")
```
zerothi: change to `raise NotImplementedError`

zerothi (changes requested):
What about the `offset` value, would it make sense to control start/ending layer?

Otherwise I think it looks more or less ready?

sisl/geom/flat.py
zerothi: I think allowing `str` and `int` are fine. I probably prefer `int` since it can be negative as well, indicating the *final* of the sequence.
So
```python
# ABC stacking
A == 0 == -3 == 3
B == 1 == -2 == 4
C == 2 == -1 == 5
```
Internal conversion is easy:
```python
if isinstance(start, str):
    start = "ABC".index(start)
if isinstance(start, int):
    start = start % 3
```
and it will always be `0, 1, 2`. I am thinking about programatically creating variations of slabs, and here integers are easier to manage?

sisl/geom/surfaces.py
tfrederiksen: Is this necessary? If `miller` does not match the implemented cases an error is raised anyways.

zerothi: Yes, but passing `miller='1111'` should fail I think.

zerothi: > Yes, but passing `miller='1111'` should fail I think.

and `miller='11'` should also fail. That would raise an `IndexError` which is somewhat counter intuitive, I think.

tfrederiksen: Done!

sisl/geom/flat.py
tfrederiksen: Done!

sisl/geom/surfaces.py
tfrederiksen: Done!

zerothi:
I have to say, I am not particularly fond of `surface_slab`, it is very user-specific, and these methods tend to create more questions than they actually solve. In particular because one person may expect one lattice constant, while others may have found other values for slightly different simulations, in particular your pbe values. In this case the reference is very good, however, what happens when other materials or reference values enter? What should we then do?
And what happens when users wants to do it for Siesta PBE, DZ, and DZP and SZP, same XC, but different basis sets with different lattice constants.

As it is, I am probably more in favour of leaving it out to not spread confusion? What do you think about this?

tfrederiksen:
> I have to say, I am not particularly fond of `surface_slab`, it is very user-specific, and these methods tend to create more questions than they actually solve. In particular because one person may expect one lattice constant, while others may have found other values for slightly different simulations, in particular your pbe values. In this case the reference is very good, however, what happens when other materials or reference values enter? What should we then do? And what happens when users wants to do it for Siesta PBE, DZ, and DZP and SZP, same XC, but different basis sets with different lattice constants.
> 
> As it is, I am probably more in favour of leaving it out to not spread confusion? What do you think about this?

OK, I see your point. I'll remove `surface_slab`.

zerothi:
I can merge and squash, removing the surface slabs, if you want?

zerothi:
> > I can merge and squash, removing the surface slabs, if you want?
> 
> Maybe we first want to add a few more structures like BCC and rock-salt (NaCl)?

Yeah, good idea. :)

tfrederiksen:
> > > I can merge and squash, removing the surface slabs, if you want?
> > 
> > 
> > Maybe we first want to add a few more structures like BCC and rock-salt (NaCl)?
> 
> Yeah, good idea. :)

OK, I'll get to it when I find time.

tfrederiksen:
The rock-salt structure could be a simple addition next by combining two intertwined `fcc_slab`s and #444.

zerothi:
I don't know if we should use `halite` instead of `rocksalt`? With clear description of their equivalence? What do you think?

zerothi:
Also the `size` parameter is not well-defined. We should think about whether this refers to number of atoms, or layers. Because currently the `size` is very different depending on `orthogonal`... Is this intuitive.

I can see that size is not per see telling anything about number of atoms. But, especially when we are talking about cutting the surface it seems ligit to say, I want a 5 layer think fcc 111 slab ending in a 'C', but currently this cannot be done as I see it?

tfrederiksen:
> I don't know if we should use `halite` instead of `rocksalt`? With clear description of their equivalence? What do you think?

I do not have a clear preference here.

tfrederiksen:
> Also the `size` parameter is not well-defined. We should think about whether this refers to number of atoms, or layers. Because currently the `size` is very different depending on `orthogonal`... Is this intuitive.
> 
> I can see that size is not per see telling anything about number of atoms. But, especially when we are talking about cutting the surface it seems ligit to say, I want a 5 layer think fcc 111 slab ending in a 'C', but currently this cannot be done as I see it?

`size[:2]` sets the repetition of the basis along the first two lattice vectors (see the `repeat` call in `_ finish_slab`) while `size[2]` is *strictly* the number of layers. Since `orthogonal=True` sometimes involves a basis of two atoms per layer, this means implies a doubling of atoms as controlled by `size[:2]`.

zerothi:
> `size[:2]` sets the repetition of the basis along the first two lattice vectors (see the `repeat` call in `_ finish_slab`) while `size[2]` is _strictly_ the number of layers. Since `orthogonal=True` sometimes involves a basis of two atoms per layer, this means implies a doubling of atoms as controlled by `size[:2]`.

Yes, exactly, but this is not clear for end-users. The question is whether it would be better to allow users to define number of atoms along slab normal direction.
And the ambiguity of getting a twice as big system when changing flag to `orthogonal`.

Perhaps the better choice would be that `size=(x, y), nz=z` or something similar, so that one can easily control number of atoms along z? Or should users always do:
```python
fcc_slab(, size=(1, 1, 4).remove(-1).tile(3, 0).tile(3, 1)
```
?

tfrederiksen:
> Perhaps the better choice would be that `size=(x, y), nz=z` or something similar, so that one can easily control number of atoms along z? Or should users always do:
> 
> ```python
> fcc_slab(, size=(1, 1, 4).remove(-1).tile(3, 0).tile(3, 1)
> ```
> 
> ?

The third element in `size` *is* simply the number of layers. And the first two are just `repeat` calls (to keep atom ordering according to a sort with `lattice=[2, 1, 0]`.

What about keeping `size` as it is, but to add another parameter `layers` that would convert to `size = (1, 1, layers)`?

tfrederiksen:
I replaced `size=None` by `layers=None, rep=(1, 1)`. I think this is indeed more transparent.

zerothi:
Ok, some final notes ;) I am picky, I know ;-)

I am starting to think that perhaps we shouldn't have `repeat` like this?
For transiesta `tile` is much better. 

So here are some suggestions, I would be happy to implement them if you agree?
- remove `rep` argument, doing `fcc_slab(...).repeat(..).repeat(...)` should be easy enough, this is more to not have both `rep` and `tile` arguments. I think the function should preferably do one thing.
- should we default the `vacuum` to `10`? This value is kind of arbitrary, but currently the default returned geometry is not a slab, it is a periodic geometry, this of course requires one to document the `None` value sufficiently.
- allow `layers` to be a str, in this case `layers='ABCABC'` would define both `end` and `layers=len(layers)`, it needs to check for correctness of input string.
- make `orthogonal`, `start` and `end` keyword only arguments. I have a feelling that `vacuum` could be sufficient as an optional argument, but also to allow future API to change a bit once we know what it can be used for, forcing kw-args would allow one to change the API without breaking compatibility.

tfrederiksen:
> Ok, some final notes ;) I am picky, I know ;-)
> 
> I am starting to think that perhaps we shouldn't have `repeat` like this? For transiesta `tile` is much better.
> 
> So here are some suggestions, I would be happy to implement them if you agree?
> 
>     * remove `rep` argument, doing `fcc_slab(...).repeat(..).repeat(...)` should be easy enough, this is more to not have both `rep` and `tile` arguments. I think the function should preferably do one thing.
> 
>     * should we default the `vacuum` to `10`? This value is kind of arbitrary, but currently the default returned geometry is not a slab, it is a periodic geometry, this of course requires one to document the `None` value sufficiently.
> 
>     * allow `layers` to be a str, in this case `layers='ABCABC'` would define both `end` and `layers=len(layers)`, it needs to check for correctness of input string.
> 
>     * make `orthogonal`, `start` and `end` keyword only arguments. I have a feelling that `vacuum` could be sufficient as an optional argument, but also to allow future API to change a bit once we know what it can be used for, forcing kw-args would allow one to change the API without breaking compatibility.

I'm fine with all your comments above. I would perhaps just suggest we change the default vacuum to 20 Å (if one is going to put an adsorbate on the slab then 10 Å starts to be on the low side).

Perfect with me if you want to do these implementations.

One thing I am not totally happy about in my implementation is the lines around `translate2uc` in `_finish_slab`. In my tests I found it necessary to introduce the small shift and to set some small negative values to zero. I feel this is quite a generic problem and could be more naturally addressed internally in the `translate2uc` function. Maybe you see a cleaner way of achieving this?

tfrederiksen:
With `layers` as as `str`, would it make sense to actually allow stacking faults like a sequence `ABCBC` (but not `AA` which would lead to physically incorrect bond lengths)? Not sure if there is any use case though...

One could also consider a more general input structure like `layers=ABC-V-CABC` where the vacuum gap `-V-` is inserted in the middle?

zerothi:
> With `layers` as as `str`, would it make sense to actually allow stacking faults like a sequence `ABCBC` (but not `AA` which would lead to physically incorrect bond lengths)? Not sure if there is any use case though...
> 
> One could also consider a more general input structure like `layers=ABC-V-CABC` where the vacuum gap `-V-` is inserted in the middle?

Cool idea! Lets just get this in as it is, then we can continue along that direction! But definitely useful!

tfrederiksen:
> Cool idea! Lets just get this in as it is, then we can continue along that direction! But definitely useful!

OK, please go ahead!

tfrederiksen:
One more thing: Could you also revise the doc-lines under "See Also"? I do not really follow the description `fcc_slab : routine called to create the slabs, kwargs is passed directly here` that you introduced.

zerothi:
@tfrederiksen could you please have a look, I have done what we agreed + some more.
You can now specify vacuum as you please, i.e.
- `layers='ABC ABC'` will only place vacuum in between the two slabs.
- `layers=' ABC ABC', vacuum=(1, 2, 1)` will place 3 vacuum spaces, before (1 Ang), middle (2 Ang) and last (1 Ang)
- `layers=' ABC` will only place vacuum first (not last)

Note that I have changed the *vacuum* meaning. I.e. the vacuum is the *exact* amount of vacuum in between slabs. This will make distance calculations more precise, if needed.

sisl/geom/surfaces.py
```diff
@@ -0,0 +1,587 @@
...
+    >>> fcc_slab(alat, atoms, "111", layers='BCABCA')
+
+    111 surface, with (1 Ang vacuum)BCA(2 Ang vacuum)ABC(3 Ang vacuum)
+    >>> fcc_slab(alat, atoms, "111", layers=' BCA ABC ', vacuum=(1, 2, 3))
+
+    111 surface, with (20 Ang vacuum)BCA
+    >>> fcc_slab(alat, atoms, "111", layers=' BCA', vacuum=20)
+
+    111 surface, with (2 Ang vacuum)BCA(1 Ang vacuum)ABC(1 Ang vacuum)
+    >>> fcc_slab(alat, atoms, "111", layers=' BCA ABC ', vacuum=(2, 1))
```
tfrederiksen, line 244: Why not `vacuum=(2, 1, 1)`? Can you at least add an explanation how the third vacuum distance is determined? Otherwise this example is a bit confusing, I think.

Actually, wouldn't it not be more precise to circle through the vacuum list, say `layers='A B C A B C ' , vacuum=(1, 2)` would give alternating vacuum distances of 1 and 2 Ang?

sisl/geom/surfaces.py
```diff
@@ -0,0 +1,587 @@
...
+
+    if geoms_none[-1] is None:
+        # ensure that nsc is 1 if there has been inserted vacuum
+        out.set_nsc(c=1)
+
+    return out
+
+
+@set_module("sisl.geom")
+def fcc_slab(alat, atoms, miller, layers=None, vacuum=20., *, orthogonal=False, start=None, end=None):
```
tfrederiksen, line 190: Just out of curiosity: I haven't seen this use of `*` before. What is it doing?

sisl/geom/surfaces.py
```diff
@@ -0,0 +1,587 @@
...
+        sets the last layer in the slab. Only one of `start` or `end` must be specified.
+        If set together with `layers` being a str, then they *must* be conforming.
+
+    Examples
+    --------
+    111 surface, starting with the A layer
+    >>> fcc_slab(alat, atoms, "111", start=0)
+
+    111 surface, starting with the B layer
+    >>> fcc_slab(alat, atoms, "111", start=1)
```
tfrederiksen, line 229: As a remark, I found that while this line is fine
```python
sisl.geom.fcc_slab(4.08, 79, 111, layers="ABC ABC", start="A")
```
the following lines fail:
```python
sisl.geom.fcc_slab(4.08, 79, 111, layers="AB CABC", start="A")
sisl.geom.fcc_slab(4.08, 79, 111, layers="AB CABC", start=("A", "C"))
```
Is this intuitive?

sisl/geom/surfaces.py
```diff
@@ -0,0 +1,587 @@
...
+    orthogonal : bool, optional
+        if True returns an orthogonal lattice
+    start : int or string, optional
+        sets the first layer in the slab. Only one of `start` or `end` must be specified.
+        If set together with `layers` being a str, then they *must* be conforming.
+    end : int or string, optional
+        sets the last layer in the slab. Only one of `start` or `end` must be specified.
+        If set together with `layers` being a str, then they *must* be conforming.
+
+    Examples
```
tfrederiksen, line 369: Maybe we could shorten this and refer to the `fcc_slab` examples (to avoid duplication)?

sisl/geom/surfaces.py
```diff
@@ -0,0 +1,587 @@
...
+    --------
+    NaCl(100) slab, starting with A-layer
+    >>> rocksalt_slab(5.64, ['Na', 'Cl'], 100)
+
+    6-layer NaCl(100) slab, ending with A-layer
+    >>> rocksalt_slab(5.64, ['Na', 'Cl'], 100, layers=6, end='A')
+
+    6-layer NaCl(100) slab, ending with A-layer
+    >>> rocksalt_slab(5.64, ['Na', 'Cl'], 100, layers=6, end='A')
```
tfrederiksen, line 540: This example is repeated two times. Let's change the latter to
```
4-layer NaCl(100) slab, starting with a Cl A-layer, and with
a vacuum gap of 20 Å both below and above it 
>>> rocksalt_slab(5.64, ['Cl', 'Na'], 100, layers=' ABAB ')

File: sisl/geom/surfaces.py [hunk 13]
```diff
@@ -0,0 +1,587 @@
...
+    111 surface, with (20 Ang vacuum)BCA
+    >>> fcc_slab(alat, atoms, "111", layers=' BCA', vacuum=20)
+
+    111 surface, with (2 Ang vacuum)BCA(1 Ang vacuum)ABC(1 Ang vacuum)
+    >>> fcc_slab(alat, atoms, "111", layers=' BCA ABC ', vacuum=(2, 1))
+
+    111 periodic structure with ABC(20 Ang vacuum)BC
+    The unit cell parameters will be periodic in this case, and it will not be
+    a slab.
+    >>> fcc_slab(alat, atoms, "111", layers='ABC BC', vacuum=20.)
```
tfrederiksen, line 249: I think we should add one more example showing the orthogonal cell and use of `repeat`:
```
111 surface in an orthogonal (4x5) cell, maintaining the atom ordering
according to `lattice=[2, 1, 0]`:
>>> fcc_slab(alat, atoms, "111", orthogonal=True).repeat(5, axis=1).repeat(4, axis=0)
```

tfrederiksen (commented):
Looks really good! The vacuum feature is very handy. I have just a few suggestions for the documentation.

sisl/geom/surfaces.py
zerothi: I have described this in the documentation, but yeah, should be emphasized in this example.
>  If shorter the last item will be repeated (like `zip_longest`).

tfrederiksen:
Would it make sense to enable also `layers=(3, 3)` etc to obtain the equivalent of `layers='ABC ABC ' ` without having to specify the string?

sisl/geom/surfaces.py
zerothi: I think this is correct! `start` should be omitted when explicitly defining the `layers`, but it will accept it if `layers[0] == start`, bottom line, if using `layers` as a str, don't supply `start`.

zerothi:
> Would it make sense to enable also `layers=(3, 3)` etc to obtain the equivalent of `layers='ABC ABC ' ` without having to specify the string?

That would be ideal, but as you say it would require restructuring of the `start` argument as well. Perhaps, lets get this in first, then we can work further. It seems a bit harder...

zerothi:
Done @tfrederiksen could you have a look, I think it should cover all cases ;)

zerothi:
> Would it make sense to enable also `layers=(3, 3)` etc to obtain the equivalent of `layers='ABC ABC ' ` without having to specify the string?

Note, that when doing the above `layers=(3, 3)` it will *not* append a vacuum to the end, as end-point vacuums should be specified specifically. Otherwise we would have to use another designator for non-vacuum positions.

So the rule is, if *any* vacuum position is specified, only that one is used, otherwise a default vacuum in the end is added.

sisl/geom/surfaces.py
zerothi: I think this is resolved, otherwise, please open again.

tfrederiksen:
I think there is some issue with the current code. If I do
```python
sisl.geom.fcc_slab(4.08, 79, 111, layers='ABC CABC')
```
the second part is not starting with a C-layer.

Even simpler put, I get a wrong ordering (CAB-order) with:
```python
sisl.geom.fcc_slab(4.08, 79, 111, layers=3, start='B')
```

zerothi:
> ```python
> .08, 79, 111, layers=3, start='B')
> sisl.geom.fcc_slab(4.08, 79, 111, layers='BCA')
> ```

Thanks for asserting I didn't mess things up (which I clearly did), it seems I forgot a sign, could you check, I think it is correct now?

tfrederiksen:
Now it works. All ready, I think.

tfrederiksen:
For completeness I added also the `rocksalt` crystal in analog of `fcc`. I also realized that the basic crystals had `nsc=[1, 1, 1]` as the default and changed it to `[3, 3, 3]`.

zerothi:
This is now merged!!! Thanks @tfrederiksen!

I squashed quite a bit of commits.

pfebrer:
I'm using this, thanks @tfrederiksen :)

zerothi:
> I'm using this, thanks @tfrederiksen :)

Could you comment on the documentation of the vacuum stuff? :-)

zerothi:
Thanks, I have updated the documentation a bit.

PR #442, opened 2022-02-18
bug: fixed behaviour in .outer

zerothi:
It should behave very much like inner.

This breaks outer usage on StateC objects.
But I think it has not really been used, so
this should be fine.

@tfrederiksen could you please check this, if this can replace #441 ? I think I have used `outer` completely wrong, thanks for noticing its weird behaviour. :)


tfrederiksen (approved):
It looks great! Much better solution than #441 .

zerothi:
Great! Thanks! Merged!

PR #444, opened 2022-02-26
Function to translate atoms in geometry into the unit cell

tfrederiksen:
Through my work with the surface slabs I think it would be convenient with a function that translates atoms (if outside) into the unit cell. I think similar algorithms are already in use internally. This PR offers this as a new function `translate2uc`.

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -2352,7 +2352,7 @@ class Geometry(SuperCellChild):
         return g
     translate = move
 
-    def translate2uc(self, atoms=None, axes=(0, 1, 2), tol=1e-5):
+    def translate2uc(self, atoms=None, axes=(0, 1, 2)):
         """Translates atoms in the geometry into the unit cell
 
         One can translate a subset of the atoms by supplying `atoms`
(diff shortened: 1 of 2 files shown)


zerothi:
Great, thanks!

Could you comment on the importance of the `tol` parameter?

I will just commit here an addition to allow doing this for individual lattice vectors.

zerothi:
Regarding the `tol` I also have some tolerances here and there, but I am just afraid they'll be wrongly applied. Consider a system with a lattice vector of 1000 A and a tolerance of 1e-5 means a physical shift of 0.01 Ang...

zerothi:
Hmm, as it stands I am probably more in favour of removing the `tol`, parameter? Would this be ok, or are there issues with this?

tfrederiksen:
> Hmm, as it stands I am probably more in favour of removing the `tol`, parameter? Would this be ok, or are there issues with this?

OK, I agree this will be cleaner. Eventually the users can, if necessary, apply the small shift before and after the function call.

tfrederiksen:
Don't you think we could insert an optional offset to this method?
```python
def translate2uc(..., offset=None):
    if isinstance(offset, float):
        d = np.ones(3) * offset
    elif isinstance(offset, list):
        d = np.array(offset)
    else:
        d = np.zeros(3)
    g = g.move(d)
    # do the existing stuff in this function
    g = g.move(-d)
    return g
```

zerothi:
I think we should first create a few more test cases to see if there are better routes, I have thought about:
- small shifts
- using round with a "number of significant digits"
- instead of using fractional cells, then use a normalized fractional cell such that all numerics are done on the same scale

To solve this I would rather collect the set of difficult problems before trying to attack it without realizing what the root cause is.

tfrederiksen:
OK, what about leaving this as a pending issue then?

PR #445, opened 2022-03-02
Suggestions to docs

tfrederiksen:
Summary:
* I suggest to bring the "publications" page up to a little more prominent location
* I found two places describing how to cite sisl (reduced one of the places and inserted a link) 
* two publications added
* fixed `ase.visualization.view` import problem (at least the original line didn't work for me)
* a few typos corrected

### DIFF
--- a/docs/cite.rst
+++ b/docs/cite.rst
@@ -6,13 +6,13 @@ Citing sisl
 sisl is an open-source software package intended for the scientific community. It is
 released under the MPL-2 license.
 
-You are encouraged to cite sisl you use it to produce scientific contributions.
+You are encouraged to cite sisl when you use it to produce scientific contributions.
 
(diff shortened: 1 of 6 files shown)


tfrederiksen:
By the way, I notice that the [`html-docs`](https://zerothi.github.io/sisl/index.html) show `0.1.dev1+gff3f170` under the "home button". Why not `0.12.1+something`?

zerothi (approved):
Thanks a lot!

zerothi:
> By the way, I notice that the [`html-docs`](https://zerothi.github.io/sisl/index.html) show `0.1.dev1+gff3f170` under the "home button". Why not `0.12.1+something`?

Yeah, I haven't exactly uncovered why this is happening, because when I do `sdata --version` I get a sensible version! So quite annoying!

PR #446, opened 2022-03-03
WIP: Introducing data sources to the viz module

pfebrer:
Ever since I started to implement the visualization module I've wanted to introduce the concept of "data sources". Data sources allow you to get array data in a simple way that can be specified from a GUI. E.g.: If you wanted to colour atoms according to their charges, it would be impossible to pass an array of charges from the GUI. The same thing applies to drawing forces as arrows, for example.

You could implement a particular solution for each of these cases, but that would be dirty and hard to maintain. With data sources (if properly implemented) this is simplified, since the user only needs to tell the interface which data they want at which place. The only thing that matters is that the output of the data source is broadcastable to the particular property.

I have introduced a proof of concept here to use it for atoms and bonds styles, with the goal of finally removing the undocumented `BondLengthMap`. To this end, I have introduced two types of `DataSource`s: `AtomData` and `BondData`, with some examples for each of them.

`DataSource` instances **load data lazily once the source is provided** to it. You can provide parameters on initialization or **perform any kind of operations on and between them with python syntax**. When the source is provided, the data is calculated and all operations are applied.

Here's an example of how one might use data sources to get a map of bond lengths:
```python
import sisl
import sisl.viz
from sisl.viz.data_sources import BondLength

# Define a supercell graphene structure
geom = sisl.geom.graphene().tile(3, 0).tile(3, 1)
# Shift one of the atoms to produce different bond lengths
mod_geom = geom.move([-0.2, 0, 0], atoms=9)

# Plot it, using the BondLength data source to color the bonds.
geom_plot = mod_geom.plot(axes="xy", bonds_style={"color": BondLength(), "width": 5}, points_per_bond=30)

# Update the bonds traces to make it look better (this will be solved)
geom_plot.update_traces(selector={'name': 'bonds'}, marker_line_width=0)
``` 
newplot

The three main points are:
- This can be easily specified, even from the GUI.
- You don't need to know which bonds are being displayed, it is taken care of for you.
- You are not limited to just setting the `color` according to bond length. You can use the `BondLength` data source for any property.

Here is an example of how a `DataSource` might accept initialization parameters to tweak its behavior. In this case we display the strain of each bond compared to a reference structure using `BondStrain`:
```python
import sisl
import sisl.viz
from sisl.viz.data_sources import BondStrain

# Define a supercell graphene structure
geom = sisl.geom.graphene().tile(3, 0).tile(3, 1)
# Shift one of the atoms to produce different bond lengths
mod_geom = geom.move([-0.2, 0, 0], atoms=9)

# Plot it, using the BondLength data source to color the bonds.
geom_plot = mod_geom.plot(axes="xy", bonds_style={"color": BondStrain(geom), "width": 5}, points_per_bond=30)

# Update the bonds traces to make it look better (this will be solved)
geom_plot.update_traces(selector={'name': 'bonds'}, marker_line_width=0)
```

newplot - 2022-03-03T011350 651

Now, a more complex (not necessary useful) example to show how to operate with data sources and use them both in **bonds and atoms**:
```python
import sisl
from sisl.viz.data_sources import BondLength, AtomX, AtomY

# Define a supercell graphene structure
geom = sisl.geom.graphene().tile(3, 0).tile(3, 1)
# Shift one of the atoms to produce different bond lengths
mod_geom = geom.move([-0.2, 0, 0], atoms=9)

# Plot it, using data sources for bonds and atoms
geom_plot = mod_geom.plot(
    axes="xy",
    bonds_style={"color": BondLength(), "width": BondLength() * 5}, 
    points_per_bond=25,
    atoms_style=[
        {"atoms": {"fx": (0.5, None)},"color": AtomX(), "size": abs(AtomY() * AtomX()) ** 1.4}
    ]
)

# Update the bonds traces to make it look better (this will be solved)
geom_plot.update_traces(selector={'name': 'bonds'}, marker_line_width=0)
```
newplot - 2022-03-03T010836 745

This still needs work, but I wanted to know if you had some suggestions/comments about it before continuing. Thanks!


zerothi:
This is great! Let me know when it is ready!

sisl/viz/data_sources/data_source.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _get_composite_source(self, op, other=None):
+        """Given an operator, returns a DataSource that will evaluate to it."""
+        return CompositeDataSource(self, op, other)
+
+    def _get_unary_source(self, op):
+        """Given an operator, returns a DataSource that will evaluate to it."""
+        return UnaryDataSource(self, op)
+
+    # Define all supported operators
```
pfebrer, line 11: In the `DataSource` class I want to implement all operators like:

```python
def __op__(self, other):
    return self._get_composite_source("__op__", other)
```

I've tried setting all attributes dynamically in a loop:
```python
class A:
    pass

ops = ("__mul__", "__add__")

for op in ("__mul__", "__add__"):
    def function(self, other):
        print(op)
    setattr(A, op, function)
```
But then `op` keeps changing so every method gets the last `op`, and
```python
A() * 2
```
prints `__add__`. Do you have any idea how can I acheive this?

pfebrer:
By the way, for atoms this is moreless the same as categories in the sense that they are lazily evaluated. I mean, this:

```python
from sisl.geom.category import AtomX

cat = AtomX(None, 3)
geometry.sub(cat)
```
would be equivalent to this:
```python
from sisl.viz.data_sources import AtomX

sel = AtomX() < 3
# When evaluated, sel will be a boolean array.
geometry.sub(sel)
```

I wonder if the two things could be unified somehow :thinking:

sisl/viz/data_sources/data_source.py
zerothi: I think sadly that in this case it is because the `op` is *global*.
The fix is to hide things in a wrapper function, then the variable becomes a local one:
```python
ops = ("__mul__", "__add__")
for op in ops:
    def wrapper(op):
        def function(self, other):
            r""" {op} """
            print(f"{self.__class__.__name__}", op, other)
        return function
    setattr(A, op, wrapper(op))
```
It ain't pretty, but it works.

zerothi: It seems to be because `op` is the *same* globally defined variable, I tried to change variable to a 2nd one to hide it, to no avail. Agreed this is a bit weird... ;(

pfebrer: > I tried to change variable to a 2nd one to hide it, to no avail.

Yeah tried the same thing :sweat_smile:

It ain't pretty, but much better than defining all of them explicitly! Thanks!

Also, basically I want to support all operations that numpy arrays support, do you know if there's a better way instead of defining a list of operations?

pfebrer: Niceee! This has simplified the code a lot :)

pfebrer: And it is much better, since you can do `np.log( BondLength() )` and it will get evaluated lazily as well. This is great!

zerothi: Exactly ;)

Please note my comment in your commit, `numpy.ufuncs` can do lots of things that might break behaviour if not carefully implemented. :)

pfebrer:
Ok, thanks! Any other argument that I have to rule out? I have no experience with this.

zerothi:
> Ok, thanks! Any other argument that I have to rule out? I have no experience with this.

I think that is the most pressing, especially since this seems to be more used *behind the scenes*, i.e. you control what happens. :)

pfebrer:
This one is pretty cool (I think): A pdos plot being used for sizing atoms through the `AtomPDOS` data source.

Imagine I have created a PDOS plot somehow:
```python
import sisl
import sisl.viz

pdos_plot = sisl.get_sile("mysystem.PDOS").plot()
```
The one I have been testing has this form (if we split the contributions by species):
```python
pdos_plot.split_DOS(on="species")
```
newplot - 2022-03-03T180614 282

Now we can plot the geometry attached to the pdos data, using `AtomPDOS` to color the atoms:

```python
from sisl.viz.data_sources import BondLength, AtomPDOS

# Get the geometry from the PDOS plot
geom = pdos_plot.geometry

# We will use AtomPDOS to get atom-resolved data from our PDOS plot
# We ask for an energy range (from -1 to 0), but we could ask for specific energies as well.
data = AtomPDOS(pdos_plot, E=slice(-1, 0))
# And then just use this data source to size all the atoms
geom.plot(axes="yx", atoms_style={"size": data * 10})
```
newplot - 2022-03-03T181314 392

Since we have asked for the range between -1 and 0, only carbons have significant PDOS. We also can see exactly which ones, which is nice.

To not worry about the multiplying factor, we could `normalize` the data returned by the data source:

```python
geom.plot(axes="yx", atoms_style={"size": data.normalize(5, 30)})
```
newplot - 2022-03-03T181901 282

The best part is that you could have used this same syntax for coloring or opacity, you just need to pass the data source to the property you wish.

Also, now that this is integrated into the framework, it is not that verbose to ask for subplots showing the PDOS-sized geometries for different energy ranges:

```python
# Get the energy range that the PDOS plot contains
Erange = pdos_plot.settings['Erange']
# Create the energy breakpoints for each subplot
breakpoints_E = np.linspace(*Erange, 5)
starts, ends = breakpoints_E[:-1], breakpoints_E[1:]

# And generate the subplots
geom.plot(
    axes="yx", subplots="atoms_style",
    atoms_style=[
        {"size": AtomPDOS(pdos_plot, slice(start, end)).normalize(0, 15)}
        for start, end in zip(starts, ends)
    ]
).update_layout(height=1000)
```
newplot - 2022-03-03T182503 382

And as always if you change the backend (in this case to matplotlib) it works as well:
PDOS_atoms

So you could animate this instead of subplotting it and you'd get a nice animation in blender going through the energies :)

zerothi:
This is pretty rad! :) :1st_place_medal:

pfebrer:
I'm basically trying to refactor the whole `viz` module now :laughing:

The idea of data sources lead to thinking about defining pieces of functionalities as **nodes**. Nodes have input and output ports which can be connected to create kind of a "workflow". The main drive is that this is more functional, since each node is a pure function. This increases modularity and eases reusability and tunability. For example, one node's output can be linked to several other nodes inputs'. This is useful if that one node does a heavy computation, which will only need to be done once now.

For now, I just refactored the PDOS plot as a proof of concept. But I kind of like the way it works, so I think I will continue refactoring everything. Here's an example of what are the steps to plot a PDOS plot following the individual steps:

```python
from sisl.viz.nodes.node import Node
# Imports for the PDOS related nodes.
from sisl.viz.plots.pdos import PDOSDataSIESTA, PDOSDataH, PDOSDataWFSX,\
                                get_PDOS_requests, PlotterNodePDOS, AtomPDOS
# Imports for the plotting backend
from sisl.viz.nodes.canvas import PlotlyCanvas, MatplotlibCanvas

# Set to True to see when a node has recalculated its output.
#Node._debug = True

# The PDOSData* nodes are data sources that standarize any entry point to
# an xarray with a given structure. These are the old "entry_points".
data = PDOSDataSIESTA(pdos_file="/path/to/file.PDOS")

# The "get_PDOS_requests" node processes PDOS data to get the PDOS for
# particular orbital requests. This is the old "set_data". We feed it
# the PDOSData* node.
PDOS_requests = get_PDOS_requests(
    data, requests=[
        {"atoms": {"Z": 7}}, 
        {"atoms": {"Z": 6}, "name": "Carbon"}
    ], Erange=[-2, 2]
)

# A PlotterNode gets the processed data and creates abstract actions (backend agnostic)
# that should be performed on the figure. The output of this node
# must be fed to a canvas (backend specific).
plotter = PlotterNodePDOS(processed_PDOS=PDOS_requests)

# We feed the PlotterNode to a canvas (in this case the plotly canvas), which will perform
# the actions on the figure.
figure = PlotlyCanvas(plot_actions=plotter)
```

And now `figure` is a node that we need to evaluate to get it's result. Evaluating `figure` will evaluate the whole tree. But if at some point a node that is already evaluated and its inputs haven't changed is reached, there's no need to evaluate any deeper. You evaluate a node with `node.get()`:

```python
figure.get()
``` 

newplot

So, you get the same as if you used the current `PdosPlot` (whose API I think I should maintain). But I'm happy because it is now much more modular and one can join the pieces however they want. E.g. one can ditch the plotting part and just get the processed PDOS, or have multiple figures from the same plotter node:

```python
matplotlib_fig = MatplotlibCanvas(plot_actions=plotter)
matplotlib_fig.get()
```
download (1)

or reuse the PDOS data to generate the atom-resolved pdos data to feed to a geometry plot:

```python
atom_pdos = AtomPDOS(data, E=-1)
```

or **whatever they want**, that's the main point. I'm trying to provide the pieces to make building higher level things easier. To make it more visual, this is how the current node tree looks:

sislviz_nodes

Also, connections between nodes are "smart" in the sense that they allow communication in both directions. Nodes have an `automatic_recalc` input that by default is `False`. If set to `True`, whenever there is an update through the path, all the nodes in the path will recalculate. For example, if I set:

```python
figure.update_inputs(automatic_recalc=True)
```

and now I update the processor
```python
PDOS_requests.update_inputs(E=[0, 2])
```

this will be the picture:

sislviz_nodes (1)

where nodes in green have been recalculated, nodes in gray didn't need any recalculation because they are sooner in the path than the update, and nodes in red are outdated. The node in red hasn't been recalculated because it doesn't have `automatic_recalc=True` and it is not in the path between the update and the node that wanted automatic updates.

If I now evaluate the matplotlib figure:
```python
matplotlib_figure.get()
```

this will be the actions performed:
sislviz_nodes (2)

This is just so you know where I'm trying to go :sweat_smile: And if there is something that you want to suggest of course it would be nice to do it now before I refactor everything hahah

By the way I still plan to provide an API very similar to the current one, where `Plot` objects are thin wrappers for the nodes (and are themselves also nodes).

zerothi:
That sounds like a very good idea!

Would it make sense to generalize the data container so the specific plots are limited in implementation details? How much can be done?

I would also really like to enable your plotting functionality in the `sdata` command, I think that would be great since cli for fast plots can be very nice!

pfebrer:
> Would it make sense to generalize the data container so the specific plots are limited in implementation details? How much can be done?

Yes, I am in fact trying to move towards this! I am relying on `xarray.DataArray`/`Dataset`, which I think is quite nice because they are not a `sisl` specific object and people may know them already (if not, they should :)).

For example, all the `PDOSData*` nodes return a `DataArray`/ with exactly the same structure and same naming conventions. The arrays are named by the property they contain and store the units as attributes. In this way, they can be plotted by a generic plotter. The processors know about this conventions and also return things that follow this rules. With this, you can get to the processed data of a PDOS plot like this:

```python
from sisl.viz.plots.pdos import PDOSDataSIESTA, get_PDOS_requests
pdos_data = PDOSDataSIESTA(pdos_file="Myfile.PDOS")

processed_PDOS = get_PDOS_requests(
    pdos_data, requests=[
        {"atoms": {"Z": 7}}, 
        {"atoms": {"Z": 6}, "name": "Carbon"}
    ], Erange=[-2, 2]
)

processed_PDOS.get()
```
Screenshot from 2022-03-07 09-00-20

And the processed bands like this:
```python
from sisl.viz.plots.bands import BandsDataSIESTA, process_bands

bands_data = BandsDataSIESTA(root_fdf=root_fdf)

filtered_bands = process_bands(bands_data, Erange=[-2, 2])

filtered_bands.get()
```
Screenshot from 2022-03-07 09-04-49

Here the `ticks` and `ticklabels` should go to the metadata of the `k` coordinate, instead of the global array, but I haven't done it yet. See e.g. the `band` coordinate, which contains "style" specifications:

```python
filtered_bands.get().band
```
Screenshot from 2022-03-07 09-10-16

It contains only one style specification, but it could contain one for each value of the coordinate, as the `request` coordinate of the processed PDOS:

```python
processed_PDOS.get().request
```
Screenshot from 2022-03-07 09-13-54

Now, both can be plotted with the same plotter node: `PlotterNodeLines`, specifying the x coordinate, the y coordinate and the coordinate to iterate over to plot multiple lines.

PDOS:

```python
from sisl.viz.nodes.plotters import PlotterNodeLines
from sisl.viz.nodes.canvas import PlotlyCanvas, MatplotlibCanvas

plotter = PlotterNodeLines(data=processed_PDOS, x="E")
#iterate_over="request" is implicit since it is the only other coordinate.

figure = PlotlyCanvas(plot_actions=plotter)

figure.get()
```
newplot - 2022-03-07T092016 311

Bands:

```python
plotter = PlotterNodeLines(data=filtered_bands, x="k", y="E", iterate_over="band")

figure = PlotlyCanvas(plot_actions=plotter)

figure.get()
```

newplot - 2022-03-07T092036 328

See how the axes already know about the names of the units of each variable thanks to the standardized data. The best part is that you could use the call to `PlotterNodeLines` in some other way to get a different plot. Or plot something other than lines with a different `PlotterNode`, but still keep the same styles (e.g. color) to easily identify the same PDOS requests across different plots.

> I would also really like to enable your plotting functionality in the sdata command, I think that would be great since cli for fast plots can be very nice!

That would be great!

Now my challenge is how to define the `Plot` objects. Should they be a fixed workflow that is contained inside the `Plot` node, and all the inputs of the nodes are exposed through `Plot` inputs? Or should it be a workflow with node types (e.g. `PDOSData*` -> `PDOSprocessor*` -> `PlotterNode*` -> `CanvasNode`), and ask users to provide the nodes somehow? :thinking:

zerothi:
> > Would it make sense to generalize the data container so the specific plots are limited in implementation details? How much can be done?
> 
> Yes, I am in fact trying to move towards this! I am relying on `xarray.DataArray`/`Dataset`, which I think is quite nice because they are not a `sisl` specific object and people may know them already (if not, they should :)).
> 
> For example, all the `PDOSData*` nodes return a `DataArray`/ with exactly the same structure and same naming conventions. The arrays are named by the property they contain and store the units as attributes. In this way, they can be plotted by a generic plotter. 

Is DataArray the way forward? They will occupy tons of disk space when you have lots of different species? Simply because the quantum numbers will be duplicated for each energy point? Or how do you organize this?

> The processors know about this conventions and also return things that follow this rules. With this, you can get to the processed data of a PDOS plot like this:
> 
> ```python
> from sisl.viz.plots.pdos import PDOSDataSIESTA, get_PDOS_requests
> pdos_data = PDOSDataSIESTA(pdos_file="Myfile.PDOS")
> 
> processed_PDOS = get_PDOS_requests(
>     pdos_data, requests=[
>         {"atoms": {"Z": 7}}, 
>         {"atoms": {"Z": 6}, "name": "Carbon"}
>     ], Erange=[-2, 2]
> )
> 
> processed_PDOS.get()
> ```
> 
> Screenshot from 2022-03-07 09-00-20
> 
> And the processed bands like this:
> 
> ```python
> from sisl.viz.plots.bands import BandsDataSIESTA, process_bands
> 
> bands_data = BandsDataSIESTA(root_fdf=root_fdf)

Why not just `fdf=root_fdf`? `root` does not tell us anything?

> filtered_bands = process_bands(bands_data, Erange=[-2, 2])

I see you use this `Erange` quite a bit of places, couldn't we use `E=np.linspace(-2, 2)` isn't this a bit more intuitive?

> 
> filtered_bands.get()
> ```
> 
> Screenshot from 2022-03-07 09-04-49
> 
> Here the `ticks` and `ticklabels` should go to the metadata of the `k` coordinate, instead of the global array, but I haven't done it yet. See e.g. the `band` coordinate, which contains "style" specifications:
> 
> ```python
> filtered_bands.get().band
> ```
> 
> Screenshot from 2022-03-07 09-10-16
> 
> It contains only one style specification, but it could contain one for each value of the coordinate, as the `request` coordinate of the processed PDOS:
> 
> ```python
> processed_PDOS.get().request
> ```
> 
> Screenshot from 2022-03-07 09-13-54
> 
> Now, both can be plotted with the same plotter node: `PlotterNodeLines`, specifying the x coordinate, the y coordinate and the coordinate to iterate over to plot multiple lines.
> 
> PDOS:
> 
> ```python
> from sisl.viz.nodes.plotters import PlotterNodeLines
> from sisl.viz.nodes.canvas import PlotlyCanvas, MatplotlibCanvas
> 
> plotter = PlotterNodeLines(data=processed_PDOS, x="E")
> #iterate_over="request" is implicit since it is the only other coordinate.
> 
> figure = PlotlyCanvas(plot_actions=plotter)
> 
> figure.get()
> ```
> 
> newplot - 2022-03-07T092016 311
> 
> Bands:
> 
> ```python
> plotter = PlotterNodeLines(data=filtered_bands, x="k", y="E", iterate_over="band")
> 
> figure = PlotlyCanvas(plot_actions=plotter)
> 
> figure.get()
> ```
> 
> newplot - 2022-03-07T092036 328
> 
> See how the axes already know about the names of the units of each variable thanks to the standardized data. The best part is that you could use the call to `PlotterNodeLines` in some other way to get a different plot. Or plot something other than lines with a different `PlotterNode`, but still keep the same styles (e.g. color) to easily identify the same PDOS requests across different plots.
> 
> > I would also really like to enable your plotting functionality in the sdata command, I think that would be great since cli for fast plots can be very nice!
> 
> That would be great!
> 
> Now my challenge is how to define the `Plot` objects. Should they be a fixed workflow that is contained inside the `Plot` node, and all the inputs of the nodes are exposed through `Plot` inputs? Or should it be a workflow with node types (e.g. `PDOSData*` -> `PDOSprocessor*` -> `PlotterNode*` -> `CanvasNode`), and ask users to provide the nodes somehow? thinking
> 
> I think I like the first one better since it is simpler for unexperienced users, but then it can be more confusing to expose nodes for reusability.

Perhaps we should discuss this at some point?  ;)

pfebrer:
> Is DataArray the way forward? They will occupy tons of disk space when you have lots of different species? Simply because the quantum numbers will be duplicated for each energy point? Or how do you organize this?

The PDOS data is stored with an `orb` coordinate, which is the orbital index. During the processing, something like `pdos_data.geometry._sanitize_orbitals(atoms=request['atoms'], n=request['n'], ...)` happens. (not this syntax but this is basically what is happening).

I kind of love `DataArray`/`Dataset`. Once you get the hang of them they are quite easy to use and very powerful. Perhaps if there is some case where you can not scape from sparse data we can use `pandas`, but I think `xarray` should be the priority :)

> Why not just fdf=root_fdf? root does not tell us anything?

You are right, `fdf` is better, `root_fdf` was a name choice that I did when I started :)

> I see you use this Erange quite a bit of places, couldn't we use E=np.linspace(-2, 2) isn't this a bit more intuitive?
The problem is that if I set `E` to accept arrays, it is much harder to make it work with the GUI. Perhaps `slice(-2, 2)` could represent a range, but then I have to do some parsing from/to json so that the GUI doesn't break and everything works fine.

> Perhaps we should discuss this at some point? ;) 

Ok, I'll try to split everything into nodes and then we can discuss how to assemble the nodes back into `Plot` objects, with the perspective of all the different nodes.

zerothi:
> > Is DataArray the way forward? They will occupy tons of disk space when you have lots of different species? Simply because the quantum numbers will be duplicated for each energy point? Or how do you organize this?
> 
> The PDOS data is stored with an `orb` coordinate, which is the orbital index. During the processing, something like `pdos_data.geometry._sanitize_orbitals(atoms=request['atoms'], n=request['n'], ...)` happens. (not this syntax but this is basically what is happening).
> 
> I kind of love `DataArray`/`Dataset`. Once you get the hang of them they are quite easy to use and very powerful. Perhaps if there is some case where you can not scape from sparse data we can use `pandas`, but I think `xarray` should be the priority :)

I agree that `xarray` is great, but one should still have in mind the memory limitations when using it... :)

> 
> > Why not just fdf=root_fdf? root does not tell us anything?
> 
> You are right, `fdf` is better, `root_fdf` was a name choice that I did when I started :)

Great, I think `root_fdf` is used many places, no?

> 
> > I see you use this Erange quite a bit of places, couldn't we use E=np.linspace(-2, 2) isn't this a bit more intuitive?
> The problem is that if I set `E` to accept arrays, it is much harder to make it work with the GUI. Perhaps `slice(-2, 2)` could represent a range, but then I have to do some parsing from/to json so that the GUI doesn't break and everything works fine.

Ok, this isn't totally clear to me, it is just that nearly all other places `E` represents an array. So if this could be done I think it would streamline the API.

> > Perhaps we should discuss this at some point? ;)
> 
> Ok, I'll try to split everything into nodes and then we can discuss how to assemble the nodes back into `Plot` objects, with the perspective of all the different nodes.

Great!

pfebrer:
> I agree that xarray is great, but one should still have in mind the memory limitations when using it... :)

But if the data is stored in a dense compact form (e.g. storing orbital information by orbital index), there's no problem attached to using `xarray` instead of some other tool, right? Otherwise I don't understand what you are saying.

> Great, I think root_fdf is used many places, no?

Yep hahah. In fact it was one of the weak points. The `Plot` object had a `root_fdf` setting, which only makes sense in SIESTA. With the refactoring this should get better.

> Ok, this isn't totally clear to me, it is just that nearly all other places E represents an array. So if this could be done I think it would streamline the API.

Agree. Now that we added nodes, there could probably be a `linspace` node that allows you to easily build an array from a GUI. You still need to think how to manage when the setting is already an array (how to display it in the GUI, if it should be displayed). However: For cases where you already calculated something and you want to filter the data, providing an array of energies doesn't make much sense, since the data is already calculated. What would you do for those cases? `xarray` interprets slices as ranges when filtering data, perhaps we should stick to that? But slices are not JSON serializable so it introduces more complexity to the interface with another language.

zerothi:
> > I agree that xarray is great, but one should still have in mind the memory limitations when using it... :)
> 
> But if the data is stored in a dense compact form (e.g. storing orbital information by orbital index), there's no problem attached to using `xarray` instead of some other tool, right? Otherwise I don't understand what you are saying.

It is only if you are using quantum numbers as columns (and not rows). In that case case one will have duplicated data. But I assume you are not doing this then? If so then all is fine ;)

> 
> > Great, I think root_fdf is used many places, no?
> 
> Yep hahah. In fact it was one of the weak points. The `Plot` object had a `root_fdf` setting, which only makes sense in SIESTA. With the refactoring this should get better.
> 
> > Ok, this isn't totally clear to me, it is just that nearly all other places E represents an array. So if this could be done I think it would streamline the API.
> 
> Agree. Now that we added nodes, there could probably be a `linspace` node that allows you to easily build an array from a GUI. You still need to think how to manage when the setting is already an array (how to display it in the GUI, if it should be displayed). However: For cases where you already calculated something and you want to filter the data, providing an array of energies doesn't make much sense, since the data is already calculated. What would you do for those cases? `xarray` interprets slices as ranges when filtering data, perhaps we should stick to that? But slices are not JSON serializable so it introduces more complexity to the interface with another language.

Yeah, not easy... :(

zerothi:
> By the way, do you think it would be useful to implement a `Geometry._sanitize_orbitals` method accepting `atoms`, quantum numbers...? Then I would get rid of all the filtering machinery in `sisl.viz` and use the standard methods in `sisl`. It would be great.

There is an `_sanitize_orbs`?

pfebrer:
I didn't know it existed :sweat_smile:. Do you think it could make sense to have orbital categories as well? Basically I would like to get orbital indices using orbital properties like quantum numbers

pfebrer:
Can I create a PR to type hint as much code as possible? I'm adding typehints to all nodes to be able to infer an input field from them and it would be useful to have sisl type hinted, otherwise the checker gives me errors. It also would be useful to have some usual types defined in `sisl` to reuse them (e.g. what can go into an `atoms` argument). Two questions:
- Does it make sense to have a `sisl.typing` module where we define types that are used throughout sisl?
- Are you in favour of having overloads to specify the output type based on input type?

PR #447, opened 2022-03-04
Minor things

tfrederiksen:
Found a few things to fix:

- the examples for the new slab functions are not shown correctly in html
- with ` geometry_define_nsc` there is no need to set `nsc` initially
- a link to the "citation" html page didn't show up. I think the paragraph can simply be removed.


tfrederiksen:
Hold on, I have one more thing to add here...

zerothi:
Great! Thanks, I'll hold on!

tfrederiksen:
There you go!

zerothi:
Thanks! :+1:

PR #453, opened 2022-03-18
enh: scaling lattice vectors by different factors

pfebrer:
This allows passing a vector with three scale factors to `Geometry.scale` and `Supercell.scale`, which are understood as the scale factor for each lattice vector.

I needed this because I need to change the surface parameters of a slab to make it fit with another lattice, while keeping the perpendicular (bulk) direction untouched.

For example:

```python
sisl.geom.graphene().scale([1.2, 1.2, 1])
```

makes the honeycomb lattice a bit bigger while keeping the same vacuum.

PS: Maybe we might as well accept a full transformation matrix? (although that should probably be a different `transform` method).

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -7,6 +7,7 @@ from numbers import Integral, Real
 from math import acos
 from itertools import product
 from collections import OrderedDict
+from collections.abc import Iterable
 from pathlib import Path
 import warnings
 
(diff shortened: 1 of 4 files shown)


pfebrer:
Perhaps it would be useful to have a `what="abc" | "xyz"` argument for if people would like to scale the coordinates instead of the lattice vectors?

sisl/geometry.py
```diff
@@ -3005,13 +3006,30 @@ def scale(self, scale) -> Geometry:
...
+           make sense.
         """
-        xyz = self.xyz * scale
-        atoms = self.atoms.scale(scale)
-        sc = self.sc.scale(scale)
+        if isinstance(scale, Iterable):
+            # The scale is a vector
+
+            # Atoms are not rescaled
+            atoms = self.atoms
```
zerothi, line 3020: I would say that atoms should still be scaled according to the maximum scaling factor. If anything the `what` should be able to bypass this (I agree such an argument would make sense here).

sisl/supercell.py
```diff
@@ -7,6 +7,7 @@
 """
 
 import math
+from typing import Iterable
```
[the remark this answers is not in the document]
zerothi: also note in `geometry`

sisl/supercell.py
```diff
@@ -676,9 +677,15 @@ def scale(self, scale):
...
         Parameters
         ----------
-        scale : ``float``
-           the scale factor for the new lattice vectors
+        scale : ``float`` or array-like of floats with shape (3,)
+           the scale factor for the new lattice vectors.
+           If it's an arrray of shape (3,), each lattice vector is scaled
+           by a different factor.
         """
+        if isinstance(scale, Iterable):
```
zerothi, line 685: alternatively, simply do `(self.cell.T * scale).T` this would work regardless of check

zerothi (changes requested):
Great addition!

Please note the wrong usage of `typing` module, I don't think they should be mixed.

Agree with the `what` argument, please use same method as in `center`. :)

pfebrer:
I think I had a different idea of what `what` should do than what you had in mind. I thought that both `"xyz"` and `"cell"` would rescale everything, but the first would mean that scale factors are meant for cartesian coordinates while the second one would mean that scale factors are meant for the lattice vectors.

Do you think someone could want to scale coordinates but not the cell? I think that would be a very specific case.

On another note, how do you know the maximum scaling factor for atoms' radii if you scale lattice vectors?

zerothi:
Ah that makes sense, yeah.

Then instead of cell, use abc (used elsewhere).

For scaling lattice vectors i would calculate the length before and after, and use the largest diff. Possible an argument is needed to not scale atoms?

zerothi:
> Ah that makes sense, yeah.
> 
> Then instead of cell, use abc (used elsewhere).
> 
> For scaling lattice vectors i would calculate the length before and after, and use the largest diff. Possibly an argument is needed to not scale atoms?

Perhaps the argument `atoms_scale` could be true, false or a float (for specific scaling of these?) then false would be equal to 1.

pfebrer:
To know the max scaling factor when `what="abc"` I needed to know the vertices of the cell before and after. At least I think so? Could you check if it is right?

I introduced a method in the supercell called `cell_vertices` to get the vertices:
1. Does it make sense?
2. Should I sum the origin?

pfebrer:
> Perhaps the argument atoms_scale could be true, false or a float (for specific scaling of these?) then false would be equal to 1.

I think accepting floats is not needed, as scaling atoms and coordinates by different factors would be very strange. I don't think it's worth the burden of accepting multiple data types.

You can always scale the atoms afterward if you are in the strange case that you need this.

zerothi:
Check my latest additions, I think it is pretty much good to go?

zerothi:
I can't really figure out if `vertices` should be a property?

sisl/supercell.py
```diff
@@ -670,21 +670,20 @@ def _assert(m, v):
...
-        verts_fxyz[4:, 0] = 1
-        verts_fxyz[[2, 3, 6, 7], 1] = 1
-        verts_fxyz[1::2, 2] = 1
-
-        return (verts_fxyz @ self.cell).reshape(2, 2, 2, 3)
+        verts = np.zeros([2, 2, 2, 3])
+        verts[1, :, :, 0] = 1
+        verts[:, 1, :, 1] = 1
+        verts[:, :, 1, 2] = 1
+        return (verts.reshape(-1, 3) @ self.cell).reshape(2, 2, 2, 3)
```
pfebrer, line 686: I think there's no reason to reshape here, you can just go `return verts @ self.cell`

pfebrer:
Great! Thanks for cleaning it up for me hahaha.

> I can't really figure out if vertices should be a property?

I wrote it as a method because I like to be careful with properties since people will use them as if they were attributes. In this case maybe it makes sense, since the calculation is quite cheap and does not scale with anything (it is always equally cheap). Also in the future it would make sense to make cache it. Yeah I think it is a good candidate for a property.

sisl/supercell.py
zerothi: true!

zerothi:
Ok, .it just came to my mind that you may need to offset the vertices? In this case an argument might be needed. So lets keep it as is.

PR #454, opened 2022-03-23
Allow more flexible MP replace with multiple replacement "levels"

ahkole:
For superconductivity calculations I am doing in `siesta` I am trying to create a MonhorstPack (MP) grid which is zoomed in to a very small region around the Fermi energy where the superconducting gap is. The `mp.replace` function from `sisl` is very useful for this, however, it only allows me to replace the original points once, while I would like to repeat the replacement procedure in an iterative fashion to create multiple "levels" of zoom in (after each replacement I run the calculation again and check which new points actually fall in the small region around the gap and zoom in on these points even more). This isn't possible with the current replace function because it assumes that the volume of each k-point is still equal to the volume in the original grid, so the volume check fails for the second replacement. I tried to fix this in this pull request by adding a new internal variable `self._k_vol` that stores the volume of each k-point in the grid. This seems to work very well. It does have the drawback of having to drag around more data, but it does make the `mp.replace` function a lot more flexible.

 - [ ] Tests added

### DIFF
--- a/sisl/physics/brillouinzone.py
+++ b/sisl/physics/brillouinzone.py
@@ -1401,6 +1401,7 @@ class MonkhorstPack(BrillouinZone):
         self._size = size # vector
         self._centered = centered
         self._trs = i_trs
+        self._k_vol = np.full((self._k.shape[0], 3), self._size / self._diag)
 
     def __str__(self):
         """ String representation of MonkhorstPack """
@@ -1744,18 +1745,6 @@ class MonkhorstPack(BrillouinZone):
         if not isinstance(mp, MonkhorstPack):
             raise ValueError('Object `mp` is not a MonkhorstPack object')
 
-        # We can easily figure out the BZ that each k-point is averaging
-        k_vol = self._size / self._diag
-        # Compare against the size of this one
-        # Since we can remove more than one k-point, we require that the
-        # size of the replacement MP is an integer multiple of the
-        # k-point volumes.
-        k_int = mp._size / k_vol
-        if not np.allclose(np.rint(k_int), k_int):
-            raise SislError(f'{self.__class__.__name__}.reduce could not replace k-point, BZ '
-                            'volume replaced is not equivalent to the inherent k-point volume.')
-        k_int = np.rint(k_int)
-
         # 1. find all k-points
         k = self.in_primitive(k).reshape(1, 3)
         dk = (mp._size / 2).reshape(1, 3)
@@ -1768,6 +1757,18 @@ class MonkhorstPack(BrillouinZone):
         idx = np.logical_and.reduce(np.abs(diff_k) <= dk, axis=1).nonzero()[0]
         if len(idx) == 0:
             raise SislError(f'{self.__class__.__name__}.reduce could not find any points to replace.')
+
+        # Figure out the BZ that each k-point is averaging
+        k_vol = self._k_vol[idx, :]
+        # Compare against the size of this one
+        # Since we can remove more than one k-point, we require that the
+        # size of the replacement MP is an integer multiple of the
+        # k-point volumes.
+        k_int = mp._size / k_vol
+        if not np.allclose(np.rint(k_int), k_int):
+            raise SislError(f'{self.__class__.__name__}.reduce could not replace k-point, BZ '
+                            'volume replaced is not equivalent to the inherent k-point volume.')
+        k_int = np.rint(k_int)
 
         # Now we have the k-points we need to remove
         # Figure out if the total weight is consistent
@@ -1792,10 +1793,12 @@ class MonkhorstPack(BrillouinZone):
 
         self._k = np.delete(self._k, idx, axis=0)
         self._w = np.delete(self._w, idx)
+        self._k_vol = np.delete(self._k_vol, idx, axis=0)
 
         # Append the new k-points and weights
         self._k = np.concatenate((self._k, mp._k), axis=0)
         self._w = np.concatenate((self._w, mp._w * weight_factor))
+        self._k_vol = np.concatenate((self._k_vol, mp._k_vol), axis=0)
 
 
 @set_module("sisl.physics")


ahkole:
The first thing that should probably be tested is that this new version still produces the same k-grids when you just do a single replacement. Is there a test suite available for the `mp.replace` function?

Also, I don't think that there are necessarily any user visible changes, except for the fact that the function now also works to create multiple levels of replacement. And I also don't think that the documentation has to be updated for this change.

zerothi:
Could you give an example where the old implementation fails?  As I recall the old implementation checks the volume and then the size needs to be correct.

ahkole:
> Could you give an example where the old implementation fails? As I recall the old implementation checks the volume and then the size needs to be correct. 

An example would be that you start with a 3x3x3 grid, replace the Gamma-point by a smaller 3x3x3 grid (size = 1./3) and then replace the Gamma-point again by an even smaller 3x3x3 grid (size=1.(3*3)). I think then it should fail.

> Point is that a recursive replacement should be done _in reverse_ since then there is no alteration of the volumes... But yes, perhaps a better check would be appropriate. I wonder whether the weights could be used since they should directly correspond to the volume.

What do you mean exactly with that the recursive replacement should be done in reverse?

> Perhaps the only requirement would be that the sum of the weighs are the same? It would allow obscure grids, but consistent weights... Hmmm...

I also thought about the weights, but if you only use the weights then you are only checking that the volumes of the "boxes" of the replaced point and the replacement grid match, while with this `k_vol` approach you really check that also the shapes of the boxes match (i.e. that the lengths of the boxes in all three directions match).

zerothi:
Reverse, meaning you should start with the smallest.


Yeah, I seem to recall this weight inconsistency when I initially implemented it.

ahkole:
> Reverse, meaning you should start with the smallest.

I am still not sure if I understand how you would do this, since sometimes the points you have to replace by the smallest (most zoomed-in) k-grid are not present yet when you start with your replacements. For example,

> An example would be that you start with a 3x3x3 grid, replace the Gamma-point by a smaller 3x3x3 grid (size = 1./3) and then replace the Gamma-point again by an even smaller 3x3x3 grid (size=1.(3*3)). I think then it should fail.

say that in the above example we don't want to replace the Gamma-point again in the second replacement, but instead want to replace the top left point of the 3x3x3 grid that replaced the original Gamma-point (i.e. one of the points in the corner of the cube if you see this smaller 3x3x3 grid as a cube around the Gamma-point). That point does not exist yet before you do the first replacement, so how would you do that recursive replacement in reverse?

zerothi:
But you can still do it reverse recursively. See this small snippet:
```
def fold_level(level, above, k, increase_1=True):
    if level == 0:
        return above
    dsize = 1. / above._diag[0] * above._size[0]
    if level == 1 and increase_1 and Nk == 9:
        # sample 3 x 3 points
        dsize *= 3
    dNk = Nk + 1 - Nk % 2
    bz = s.MonkhorstPack(gr, [dNk, dNk, 1], size=[dsize, dsize, 1], displacement=k, trs=False)
    print('Equivalent lower level sampling: {}'.format(int(dNk / dsize)))
    bz = fold_level(level - 1, bz, k)
    try:
        above.replace(k, bz)
    except Exception as e:
        print(e)
        print('All k-points')
        print(bz.k)
    return above
```
you'll just keep creating the new points, and first replace the bottom most ones :)
(the example is not functional since it is taken from a bigger context, sorry about that...

ahkole:
Ah, yes, now I see it. So you are basically doing replacements in your replacement grids in a bottom-up fashion until you get to the original grid instead of doing all your replacements directly in the original grid. This could technically work, but would probably get very complicated for my use case, since what I am doing right now is:
1. Generate an initial MP grid
2. Run `siesta` with this grid and get as output a list of all kpoints (with their associated indices in the original grid) that contribute to the PDOS I am trying to calculate.
3. Using this list of indices replace all contributing kpoints by a zoomed in grid.
4. Run `siesta` again with this replaced grid and get a new output list with contributing kpoints
5. Using this new output list do the replacements.
6. etc.

So then I would have to find some mapping between the indices of the k-points in the MP grid to the indices within my replacement grids, which is probably technically possible but which becomes complicated pretty quickly. To me it feels like it would be a lot simpler to work with if I could just recursively replace directly in the original grid.

zerothi:
Why aren't you doing this in sisl? Then an incremental zooming could be done at a much reduced computational cost, and much simpler. I. E. It seems to me that you are calculating the full pdos again and again with a finer and finer grid, instead you could search only the replaced k-points for much higher throughput?

zerothi:
Also, it becomes difficult to retain the volumes when reading external kp files. No? Does your pr solve this?

zerothi:
It is not that I think this is a bad idea (thanks for the pr), but if there are still cases it can't handle then I don't see the immediate benefit. I.e. in this case you still need to do everything in one go since the volumes are not retained between different executions, and then the other proposed scheme is equivalent.

ahkole:
> Why aren't you doing this in sisl? Then an incremental zooming could be done at a much reduced computational cost, and much simpler. I. E. It seems to me that you are calculating the full pdos again and again with a finer and finer grid, instead you could search only the replaced k-points for much higher throughput?

I am still in the proof-of-concept phase to figure out if these zoomed-in grids are useful for my calculations, so there is still a lot of room for improvement yes. The reason I am doing it like this is that I am working with an experimental version of siesta that implements the BdG equations and I don't know if that code can already write the BdG Hamiltonian in a form that can be used by sisl. The quickest way to get something working was to add some code to the associated pdos routine to output the k-points that contribute to the pdos. But it would be much more efficient to not have to calculate the pdos again and again (and even worse I also have to do the SCF again each iteration).

> Also, it becomes difficult to retain the volumes when reading external kp files. No? Does your pr solve this?

There is no way to store the volumes no. But as far as I know there also is no way to store and retrieve the other information needed for MonkhorstPack.replace (i.e. size of the BZ, number of points in each direction) in the current implementation, right?

> It is not that I think this is a bad idea (thanks for the pr), but if there are still cases it can't handle then I don't see the immediate benefit. I.e. in this case you still need to do everything in one go since the volumes are not retained between different executions, and then the other proposed scheme is equivalent.

The current pr wasn't really intended to prevent you having to do everything in one go but more to allow more flexibility to do the replacements in any order you like. There is one other problem I encountered where this approach might be beneficial compared to the other proposed scheme: in the third iteration for a bulk Nb superconductor I had to replace ~100,000 k-points by a zoomed-in grid and for so many points replacing them with a for-loop over the points was extremely slow (hadn't finished after 3 hours) and I suspect that it won't be much better when using recursion as in the other proposed scheme. And when you have these volumes it is easier to implement a bulk replacement function that replaces all the points at once entirely with NumPy operations. I tried to implement something to do that and it was indeed much faster (completed the replacement almost instantaneously). This bulk replacement is not in this pr, but this pr could maybe be the basis for that? Or do you have a better idea how to make it manageable to replace a lot of k-points?

zerothi:
Could you share the code, then I can have a look at proposed solutions?

It sounds like 100000 k-points is not going to help you, I.e. I can't see how much better the descriptions of the electronic structure will be, that is something like 0.001/4 Ang k-point density, which seems ridiculously high! 

As for bdg in sisl, if you could share the code (or I'll ask alberto/gabore bout the branch), then I can implement it, should be trivial I think.

ahkole:
> Could you share the code, then I can have a look at proposed solutions?

Here is a snippet of the changes I made to the `replace` function for the bulk replacements:

        # 1. find all k-points
        k_is_indices = False
        if isinstance(k[0], Integral):
            # k contains indices of k-points to be replaced
            k_is_indices = True
            idx = np.expand_dims(k, axis=0)
        else:
            k = self.in_primitive(k).reshape(1, 3)
            dk = (mp._size / 2).reshape(1, 3)
            # Find all points within [k - dk; k + dk]
            # Since the volume of each k-point is non-zero we know that no k-points will be located
            # on the boundary.
            # This does remove boundary points because we shift everything into the positive
            # plane.
            diff_k = self.in_primitive(self.k % 1. - k % 1.)
            idx = np.logical_and.reduce(np.abs(diff_k) <= dk, axis=1).nonzero()[0]
            if len(idx) == 0:
                raise SislError(f'{self.__class__.__name__}.reduce could not find any points to replace.')

        # Figure out the BZ that each k-point is averaging
        k_vol = self._k_vol[idx, :]
        # Compare against the size of this one
        # Since we can remove more than one k-point, we require that the
        # size of the replacement MP is an integer multiple of the
        # k-point volumes.
        k_int = mp._size / k_vol
        if not np.allclose(np.rint(k_int), k_int):
            raise SislError(f'{self.__class__.__name__}.reduce could not replace k-point, BZ '
                            'volume replaced is not equivalent to the inherent k-point volume.')
        k_int = np.rint(k_int)

        # Now we have the k-points we need to remove
        # Figure out if the total weight is consistent
        total_weight = self.weight[idx].sum(axis=0)
        replace_weight = mp.weight.sum()
        if np.all(np.abs(total_weight - replace_weight) < 1e-8):
            weight_factor = 1.
        elif np.all(np.abs(total_weight - replace_weight * 2) < 1e-8):
            weight_factor = 2.
            if self._trs < 0:
                info(f'{self.__class__.__name__}.reduce assumes that the replaced k-point has double weights.')
        else:
            print('k-point to replace:')
            print(' ', k.ravel())
            print('delta-k:')
            print(' ', dk.ravel())
            print('Found k-indices that will be replaced:')
            print(' ', idx)
            print('k-points replaced:')
            print(self.k[idx, :])
            raise SislError(f'{self.__class__.__name__}.reduce could not assert the weights are consistent during replacement.')

        if k_is_indices:
            mp._k = mp.in_primitive(np.reshape(np.expand_dims(mp._k, axis=0) + np.expand_dims(self._k[k], axis=1), (-1, 3)))
            mp._w = np.tile(mp._w, idx.shape[1])
            mp._k_vol = np.tile(mp._k_vol, (idx.shape[1], 1))
        
        idx = idx.flatten()
        self._k = np.delete(self._k, idx, axis=0)
        self._w = np.delete(self._w, idx)
        self._k_vol = np.delete(self._k_vol, idx, axis=0)

        # Append the new k-points and weights
        self._k = np.concatenate((self._k, mp._k), axis=0)
        self._w = np.concatenate((self._w, mp._w * weight_factor))
        self._k_vol = np.concatenate((self._k_vol, mp._k_vol), axis=0)

The idea was that the `k` variable in `mp.replace(k, mp)` is then a list of indices of k-points instead of the coordinates of the k-points and the replacement mp-grid is shifted so that its origin is placed at the position of the k-point that is being replaced.

And here is a snippet of the code I am using in my script to do the replacements:

    import numpy as np
    import sisl

    # First create MP grid for initial SIESTA run
    # (Assuming you have the geometry of your
    # system in the file STRUCT.fdf)
    fdfSile = sisl.get_sile('STRUCT.fdf')
    geom = fdfSile.read_geometry()
    nk = 50
    mp = sisl.MonkhorstPack(geom, nk, trs=False)  # SOC and superconductivity don't use Time-Reversal Symmetry (trs) I think

    # Output this grid to a file
    rkpSile = sisl.io.siesta.rkpSileSiesta('SLabel.PDOS.RKP', 'w')
    rkpSile.write_data(mp.k, mp.weight)

    # Now run SIESTA with the PDOS calculation that you want,
    # using the kpoints we just created (by using the
    #     PDOS.kgrid.File "SLabel.PDOS.RKP"
    # option for the PDOS kgrid). Also make sure that it
    # is a superconductivity calculation (Spin.Nambu)
    # and that you are using the ParallelOverK option
    # (only implemented things there for now)

    # Read in the kpoints that contribute to the PDOS
    iksContrib = np.loadtxt('/path/to/outputs/SLabel.PDOS.KP.CONTRIB', usecols=1, dtype=int)
    kpointContrib = mp.k[iksContrib - 1, :]  # Note the conversion from 1-based Fortran indexing to 0-based Python/C-indexing

    # NB: You could also directly use the kpoints (usecols=(2,3,4)) from the file,
    # but I find this easier so you don't have to worry about conversion between
    # different units of the kpoints etc.

    # Now replace all the contributing kpoints by a denser grid (i.e. a 3x3x3 grid here)

    # OLD STYLE
    nk1 = 3
    for k in kpointContrib:
        mpZoom = sisl.MonkhorstPack(geom, nk1, displacement=k, trs=False, size=1./nk)
        mp.replace(k, mpZoom)

    # NEW STYLE WITH BULK REPLACE
    nk1 = 3
    mpZoom = sisl.MonkhorstPack(geom, nk1, trs=False, size=1./nk)
    mp.replace(iksContrib-1, mpZoom)

    # Write new kpoints to file
    rkpSile = sisl.io.siesta.rkpSileSiesta('SLabel.PDOS.RKP', 'w')
    rkpSile.write_data(mp.k, mp.weight)

    # Use this new kgrid for another PDOS calculation
    # You could in princple iterate this as often
    # as you'd like. For subsequent iterations, you should use
    # size = 1/(nk*nk1*...*nk(N-1)) for iteration N

> As for bdg in sisl, if you could share the code (or I'll ask alberto/gabore bout the branch), then I can implement it, should be trivial I think.

Nils should have added you to the fork of the code that we are working on now. The code I added to output the k-points that contribute to the PDOS is on the `bdg-future-kp-pdos-k-grid-zoom` branch in the `pdosbkp.F90` file.

> It sounds like 100000 k-points is not going to help you, I.e. I can't see how much better the descriptions of the electronic structure will be, that is something like 0.001/4 Ang k-point density, which seems ridiculously high!

As can be seen here 
image
every iteration I did so far (including the last one where ~100,000 points are replaced) did noticeably change the PDOS (although you could maybe argue that the difference between `iter 2` and `iter 3` is not that significant). This was for bulk Nb with a superconducting intraorbital-pairing of `Delta = 1.525 meV`. Here the starting PDOS grid is `50x50x50` and in every iteration each contributing point is replaced by a `3x3x3` grid. The number of contributing points for each iteration was:

    Original 50x50x50 grid: 1704
    Iter 1: 7982
    Iter 2: 101434
    Iter 3: 2260872

We suspect that we need such dense k-grids because the energy scales are so small that we need very small electronic temperatures and smearing of the PDOS (to not smear states over the superconducting gap).

PR #457, opened 2022-03-28
mnt: changed all root_fdf to fdf

zerothi:
@pfebrer could you please see if everything is ok by doing this?

I reran the tests, and all seems to pass on my comp, but I don't know how you test it?

### DIFF
--- a/sisl/viz/_plotables_register.py
+++ b/sisl/viz/_plotables_register.py
@@ -39,7 +39,7 @@ for HSile in get_siles(attrs=["read_hamiltonian"]):
     register(HSile, FatbandsPlot, "H")
 
 for cls in get_plot_classes():
-    register(siesta.fdfSileSiesta, cls, "root_fdf", overwrite=True)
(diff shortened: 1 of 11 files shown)


pfebrer:
Looks good to me! Although I was already fixing this in https://github.com/zerothi/sisl/pull/446 :sweat_smile:  But it's great that you had the time to look at the code in sisl.viz :)

By the way, the documentation is automatically updated using the parameters of each plot class with `python -m sisl.viz._doc_updater` (as explained in that file). But even that will probably change.

I wouldn't bother too much at modifying the current `sisl.viz` module as most things will change to make it more modular and hopefully everything will be easier to understand :) So I wouldn't push this to `main` as it would probably create unnecessary conflicts with the other PR.

zerothi:
Ok! I'll close!

pfebrer:
Sorry :)

zerothi:
No worries! :)

PR #458, opened 2022-04-05
enh: added multiple replacements for k-points

zerothi:
This is required for replacing large quantities of
k-points. The idea was initiated by @ahkole and polished by @zerothi.

A displacement argument and as_index argument ensures that one
can fine-tune the placement. However, in these cases it is not
strictly checking whether the indices are actually *inside* the
replaced k-points. Probably this should be checked but users
may want to do crazy stuff.

Co-authored-by: A.H. Kole <a.h.kole@uu.nl>
Co-authored-by: Nick Papior <nickpapior@gmail.com>

@ahkole, could you please have a look here? I think this will fix #454 since there were some corner cases that could not be handled in your situation (for instance the detection of indices was not clear in all cases, users could pass `k=[0, 0, 0]` and it would be understood as an index).
I have benchmarked this and it is much much faster (see `benchmarks/bz_replace.py`).

Would this suffice?

### DIFF
--- a/benchmarks/bz_replace.py
+++ b/benchmarks/bz_replace.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+# This benchmark creates a very large graphene flake and uses construct
+# to create it.
+
+# This benchmark may be called using:
+#
+#  python $0
+#
+# and it may be post-processed using
+#
+#  python stats.py $0.profile
+#
+import cProfile
+import pstats
+import sys
+import sisl
+import numpy as np
+from tqdm import tqdm
+
+if len(sys.argv) > 1:
+    nlvls = int(sys.argv[1])
+else:
+    nlvls = 2
+ndim = 2
+print(f"dimensions = {ndim}")
+print(f"levels = {nlvls}")
+
+
+# this will create 10x10 points
+if nlvls > 2:
+    nks = [40 for _ in range(nlvls)]
+else:
+    nks = [50 for _ in range(nlvls)]
+
+if ndim == 1:
+    def get_nk(nk):
+        return [nk, 1, 1]
+    ns = [2 for _ in range(nlvls)]
+else:
+    def get_nk(nk):
+        return [nk, nk, 1]
+    if nlvls > 2:
+        ns = [50 for _ in range(nlvls)]
+    else:
+        ns = [100 for _ in range(nlvls)]
+
+
+# Replace many k-points
+def yield_kpoint(bz, n):
+    yield from np.unique(np.random.randint(len(bz), size=n))[::-1]
+
+# Replacement function
+
+
+def add_levels(bz, nks, ns, fast=False, as_index=False, debug=False):
+    """ Add different levels according to the length of `ns` """
+    global nlvls
+
+    lvl = nlvls - len(nks)
+    nreps = 0
+
+    if fast:
+        # we need to copy the bz since for each ik, the new_bz gets
+        # changed in add_levels.
+        # If there was only 1 ik per level, then all would work fine.
+        bz = bz.copy()
+
+    from io import StringIO
+    s = StringIO()
+
+    def print_s(force=True):
+        nonlocal s
+        out = s.getvalue()
+        spaces = " " * (lvl * 2)
+        out = spaces + out.replace("\n", f"\n{spaces}")
+        if force:
+            print(out)
+
+        # reset s
+        s = StringIO()
+    if debug:
+        print(f"lvl = {lvl}", file=s)
+        print_s()
+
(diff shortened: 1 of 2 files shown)


zerothi:
@ahkole do you mind having a look at this one? :-D

ahkole:
@zerothi sorry I haven't replied yet. I looked at the code and it looks good, definitely more robust than my implementation of the bulk replacement! I also wanted to test it but I got busy with other projects and didn't get around to doing that yet. I am currently at a Spring School until Friday but as soon as I get back from that I will run some tests on my system to see if it works for my case.

ahkole:
@zerothi I tested the `mp.replace` on my bulk Nb superconductor system and the bulk replacement seems to work perfectly. There is one small difference with my implementation of the bulk replacement: sometimes this new version produces k-points that are outside the first Brillouin Zone, while my implementation moved all k-points inside the first BZ with the `mp.in_primitive` function. As an example, this new version returned a k-point `[ 0.50666667,  0.10666667, -0.18666667]`, while my version returned the k-point `[-0.49333333,  0.10666667, -0.18666667]`. This might not be that important because the reciprocal space is periodic anyway, but I thought it was worth to mention.

I would however say that it does not really fix https://github.com/zerothi/sisl/pull/454 because that PR was about a separate issue, i.e. allowing more flexible MP replacements by carrying around a `k_vol` attribute for each k-point. We talked about the bulk replacements in that PR's discussion thread, but the main issue that PR was trying to address is this replacement flexibility, not the replacement speed.

zerothi:
> @zerothi I tested the `mp.replace` on my bulk Nb superconductor system and the bulk replacement seems to work perfectly. There is one small difference with my implementation of the bulk replacement: sometimes this new version produces k-points that are outside the first Brillouin Zone, while my implementation moved all k-points inside the first BZ with the `mp.in_primitive` function. As an example, this new version returned a k-point `[ 0.50666667, 0.10666667, -0.18666667]`, while my version returned the k-point `[-0.49333333, 0.10666667, -0.18666667]`. This might not be that important because the reciprocal space is periodic anyway, but I thought it was worth to mention.

Thanks, I'll try and fix this.

> I would however say that it does not really fix #454 because that PR was about a separate issue, i.e. allowing more flexible MP replacements by carrying around a `k_vol` attribute for each k-point. We talked about the bulk replacements in that PR's discussion thread, but the main issue that PR was trying to address is this replacement flexibility, not the replacement speed.

As I wrote there, the problem is when reading from other files the volumes get lost, and then it all breaks down... So I would rather not add an attribute that only works *sometimes*. It still checks that the *total* weight/volume is equivalent, so if users do weird stuff, then all should be fine, no?

ahkole:
> As I wrote there, the problem is when reading from other files the volumes get lost, and then it all breaks down... So I would rather not add an attribute that only works sometimes. It still checks that the total weight/volume is equivalent, so if users do weird stuff, then all should be fine, no?

Ah okay. A check that only checks that the total volume of the replaced points is consistent would also be fine yes, but as far as I can see this version still does exactly the same as the old where it checks the volume of a single k-point, right? Because I still got the error that the volume was not equivalent when I tested my second "level" of replacements (which does work with the version where you have the additional `k_vol` attribute).

ahkole:
> Ah okay. A check that only checks that the total volume of the replaced points is consistent would also be fine yes, but as far as I can see this version still does exactly the same as the old where it checks the volume of a single k-point, right? Because I still got the error that the volume was not equivalent when I tested my second "level" of replacements (which does work with the version where you have the additional k_vol attribute).

Or more specifically, it requires that the volume of the replacement MP grid is consistent with the volume of one of the k-points from the original grid (before any replacements). And there is no way of guaranteeing this consistency in the second "level" of replacement.

ahkole:
> As I wrote there, the problem is when reading from other files the volumes get lost, and then it all breaks down...

I also personally do not see why this is a problem because as far as I can see there is no way of creating a `MonkhorstPack` object from a file at the moment (you can only read k-points and weights from a file, not the total `MonkhorstPack` object), so there is no situation where this would break down since you have to create the object from scratch all the time anyways.

zerothi:
Thanks for your comments!

I will think some more about this. I am a bit reluctant to ship another *large* array together which seems to complicate things for only one specific thing. 

So many thanks for bringing this to my attention!

zerothi:
@ahkole what about now? I have added an argument to disable volume checking. I think this is safer in the long run, also given the batch replacement.

What is important is that it still checks that the k-point weights are equivalent, and hence the total volume is protected.
I would rather this, than having an extra array... What do you say to this approach?

ahkole:
> @ahkole what about now? I have added an argument to disable volume checking. I think this is safer in the long run, also given the batch replacement.

I tested it and it works perfectly, giving the exact same k-points as my old version!

> What is important is that it still checks that the k-point weights are equivalent, and hence the total volume is protected.
I would rather this, than having an extra array... What do you say to this approach?

Yes, this approach is fine for me. It's also probably marginally faster when skipping the volume check :)

zerothi:
Great, I'll merge this! Thanks for testing this!

PR #462, opened 2022-05-03
Fixed reading of n, l, zeta in hsx_read_specie1

ahkole:
@juijan and I noticed that the reading of the quantum numbers of the orbitals (n, l, zeta) from the HSX file does not work correctly for HSX file version 1. There is an inconsistency between the ordering of the number during reading and writing. The writing routine (`write_hsx` in `io_hsx.F90`) writes them in the order `n(1), l(1), zeta(1), n(2), l(2), zeta(2), ...`, while the reading routine expects it in the order `n(1), n(2), ..., l(1), l(2), ..., zeta(1), zeta(2), ...`. This PR fixes the reading routine to use the same ordering.

### DIFF
--- a/sisl/io/siesta/_src/hsx_read.f90
+++ b/sisl/io/siesta/_src/hsx_read.f90
@@ -1354,7 +1354,7 @@ subroutine read_hsx_specie1(fname, ispecie, no_specie, n_specie, l_specie, zeta_
 
 ! Internal variables and arrays
   integer :: iu, ierr, version
-  integer :: is
+  integer :: is, io
 
   ! Local readables
   integer :: nspecies, lna_u, lno_u, lnspin, nsc(3)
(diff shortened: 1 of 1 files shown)


zerothi:
@ahkole something unrelated, I think your email address on your local machine does not match your github accounts email address.

The point is that github cannot recognize your commits and thus you will not be attributed in the github history (I would prefer to acknowledge your work, but I can't do anything here...)
See e.g. here https://github.com/zerothi/sisl/pull/462/commits where the commit author says A. H. Kole and not your username as is customary.

FYI!

ahkole:
@zerothi yes, that may be correct. I usually commit using my university email address but I created this GitHub account a long time ago with my personal email address so that's probably responsible for the mismatch. But I wasn't aware that GitHub then cannot attribute commits to my account, thanks for letting me know!

I will try to add my university address as a secondary email to my account and hopefully GitHub will then be able to properly recognize the commits.

zerothi:
Yeah, it isn't exactly clear to me how this works, but I think you could check it by making a fake branch and commit some things there, then see if your user-name pops up.

ahkole:
I think it even works retroactively because I just added my university email to my account and now this commit https://github.com/zerothi/sisl/pull/462/commits does show a proper link to my account :)

zerothi:
Even better! Great!

PR #464, opened 2022-05-10
Add option to sdata to write DOS to stdout or file

juijan:
As the title suggests: adds a new option to `sdata` (`--dos-write`) , which calculates the density of states and either writes it to standard output or to a file (if provided). Additional arguments are the same as for `--dos`. 

Not sure if you want to add it, but I found it useful. 🤷

### DIFF
--- a/sisl/io/siesta/eig.py
+++ b/sisl/io/siesta/eig.py
@@ -32,13 +32,17 @@ class eigSileSiesta(SileSiesta):
         sdata siesta.EIG --plot eig_spread.png
 
 
-    One may also plot the DOS using the eigenvalues and the k-point weights:
+    One may also extract/plot the DOS using the eigenvalues and the k-point weights:
 
     .. code:: bash
 
         sdata siesta.EIG --dos
         # or to save to png file
         sdata siesta.EIG --dos dos.png
+	# or to write DOS to stdout
+        sdata siesta.EIG --dos-write
+	# or to write DOS to file
+        sdata siesta.EIG --dos-write dos.dat
 
     This will default to plot the DOS using these parameters:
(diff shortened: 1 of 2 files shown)


zerothi:
Great addition, perhaps this should be streamlined like what I have for the `TBT.nc` files? I.e. that on can get multiple data sets and save them via `--out`?

Idea:
- change `sdata *.EIG --dos [dos-args]` to `sdata *.EIG --dos [dos-args] --plot`
- allow `sdata *.EIG --out`
This would make the two commands more coherent, I think?
And allow one to collect multiple DOS arguments for txt files/plots?
```

**Comment by juijan** [comment]
That does sound more coherent. 

I gave it a show. Let me know what you think.

**Comment by zerothi** [comment]
could you have a look at my edits? I think it is much cleaner and easier now?

**Comment by zerothi** [comment]
Ok, @juijan, now it is definitely there. Do you want to have a look, or should I merge?

**Comment by juijan** [comment]
Looks good to me!

File: sisl/io/siesta/eig.py [hunk 1]
```diff
@@ -140,29 +144,177 @@ def opts(*args):
...
+                                same_E = np.allclose(E, ns._data[-i])
+                            except Exception:
+                                same_E = False
+
+                if not same_E:
+                    ns._data_header.append("Energy")
+                    ns._data.append(E)
+
+                def calc_dos(E, eig, w):
+                    nonlocal distribution
```
juijan, line 260: Is there any specific reason, why it is beneficial to choose `nonlocal` over passing as a argumetn?

zerothi: No, none at all in this case... :)

PR #466, opened 2022-05-24
Enable dtype change in sisl.Grid.copy

tfrederiksen:
The `sisl.Grid.copy` function did not actually allow for a change of the array dtype (as suggested in the documentation). This branch fixes this.


zerothi (approved):

zerothi:
Great! Thanks for finding this!!!

PR #476, opened 2022-07-18
Refactoring the viz module to a more modular design.

pfebrer:
The discussion for this PR started in #446, but the final aim is very different from what I intended at the beggining of that PR, so I prefer to open a new one.

Basically, I refactored the module to have a more modular and functional design while keeping the functionalities that I thought were important. The refactoring is still not complete. I have refactored PDOS and bands as a proof of concept and wanted to know your opinion before carrying on.

As a first introduction and to display the differences, this is now the code of `BandsPlot`, which at a high level is much easier to understand from scratch than the previous design:

```python
@Plot.from_func
def BandsPlot(bands_data: BandsData = None, 
    Erange=[-2, 2], E0=0, E_axis: Literal["x", "y"] = "y", bands_range=None, spin=None, 
    bands_style={'color': 'black', 'width': 1, "opacity": 1}, spindown_style={"color": "blue", "width": 1},
    gap=False, gap_tol=0.01, gap_color="red", gap_marker={"size": 7}, direct_gaps_only=False, custom_gaps=[],
    bands_mode: Literal["line", "scatter", "area_line"] = "area_line"
):
    if bands_data is None:
        raise ValueError("You need to provide a bands data source in `bands_data`")

    # Filter the bands
    filtered_bands = filter_bands(bands_data, Erange=Erange, E0=E0, bands_range=bands_range, spin=spin)

    # Add the styles
    styled_bands = style_bands(filtered_bands, bands_style=bands_style, spindown_style=spindown_style)

    # Determine what goes on each axis
    x = get_axis_var(axis="x", var="E", var_axis=E_axis, other_var="k")
    y = get_axis_var(axis="y", var="E", var_axis=E_axis, other_var="k")

    # Get the actions to plot lines
    bands_plottings = PlotterNodeXY(data=styled_bands, x=x, y=y, set_axrange=True, what=bands_mode)

    # Gap calculation
    gap_info = calculate_gap(filtered_bands)
    # Plot it if the user has asked for it.
    gaps_plottings = draw_gaps(bands_data, gap, gap_info, gap_tol, gap_color, gap_marker, direct_gaps_only, custom_gaps, E_axis=E_axis)

    return CompositePlotterNode(bands_plottings, gaps_plottings, composite_method=None)
```

The main ideas are explained in [this notebook](https://github.com/zerothi/sisl/files/9130706/Sisl.viz.refactor.display.zip) (I include the html output so that you don't need to run it), because I thought it was much easier to explain in a notebook. I tried to not go into the details and explain the framework at a high level, so I expect it will be a fast read even if it seems a bit long.

Looking forward to hear your thoughts whenever you have time!


pfebrer:
Hmm I think it's better not to rush it, so you can publish the release without it. Actually I was waiting for your opinion on the core ideas, because maybe there are some things that can be simplified/made more functional.

Whenever you have time if you could look at the html file that I provided [here](https://github.com/zerothi/sisl/files/9130706/Sisl.viz.refactor.display.zip) with a rendered notebook that would be nice :)

zerothi:
Ok, I'll have a look (hopefully this week or next)! Thanks!

zerothi:
Ok, I have had a look. And all-in-all I really like your ideas!

I have uploaded a new notebook with comments in them,
[Sislvizrefactordisplay.txt](https://github.com/zerothi/sisl/files/10569010/Sislvizrefactordisplay.txt)

(change extension to `ipynb` :)

I say we go! :)

pfebrer:
Thanks for looking at it!

I will go through some of your comments here:

1. Will this be problematic for nodes storing large data-sets? Imagine nodes with grids and doing a workflow: read -> do function -> reduce grid. This would result in 2 full grids, and one at reduced size? Not only that, also the arguments to the functions (which needs to be in memory for recalculations).

Yes, I also thought that this would be problematic with grids. But then my thought was that if it does not make sense to store its inputs and outputs it shouldn't be a node. E.g. if you have a plot that does `... -> interpolate_grid -> smooth_grid -> reduce_grid -> ...` you should probably pack all that functionality into a single node: `... -> process_grid -> ...`.

2. Will a static code analyzer figure out that a node with a return value is valid input? (not necessary, just out of interest)

No, it doesn't. This is one of the reasons I want to get rid of defining all functions as nodes (see design tweak below).

3. Perhaps there is something I am missing, but couldn't you just have done @Node.from_funct as well? What is the difference from a Workflow?

Yeah this is a bit tricky, but I think `Workflow` deserves a class of its own. The difference between a simple `Node` and a `Workflow` is that a node can contain any computation inside it, while a workflow is just a way of organizing nodes. A workflow just links nodes with each other, there is nothing computed outside the nodes it contains.

Of course you can also consider that you can split all the functionality in a node into individual pieces that are linked together, but you wouldn't want to store checkpoints for all of them :sweat_smile:

## My thoughts for a design tweak

I didn't like the fact that there was `@Node.from_func` on top of every function of `sisl.viz` (I had to introduce the default lazy_context=False to make them usable as normal functions). Also I didn't like that to use any other function in a workflow you need to define it explicitly as a node. I.e.:

```python
from sisl import some_func
from sisl.viz import Node, Workflow

node_func = Node.from_func(some_func)

@Workflow.from_func
def my_workflow(...):
    ...
    a = node_func()
    ...
```

I have now an implementation (not merged here) that on `Workflow` definition builds replaces all function calls with Node calls. I.e. it allows you to do:

```python
from sisl import some_func
from sisl.viz import Workflow

@Workflow.from_func
def my_workflow(...):
    ...
    a = some_func()
    ...
```

and it will be exactly the same as the previous code.

To me this is a great improvement because:
- You don't need to redefine functions as nodes.
- `sisl.viz` functionality doesn't need to be defined as nodes.
- It makes going from workflow to function very simple (just remove the decorator). It is also much easier the other way.
- Because of the previous to points, there is no need now for the `lazy_context`. If you use a node it will be lazy. Otherwise just use the normal functions.

I do the change from function calls to node calls getting the Abstract Syntax Tree, modifying it and recompiling. I don't know if it is too "hacky", but it seems very robust to me, and cleaner than any other implementation. Do you think it is fine or should I stick to "more normal" python trickery?

Also, if you don't need to define all functionalities as nodes I thought that this might be useful not only for `sisl.viz`. You could create a workflow with any functions in `sisl` (in fact with any function in any module). So do you think it makes sense to have something like `sisl.nodes` to keep this functionality?

Cheers!

zerothi:
I'll have to wrap my head a bit more around the node stuff, I am a bit worried it might make the code more complicated and not as intuitive. 

I have however thought about the dispatch methods since there are some short-comings, here the nodes might be interesting.

A note on the workflows, it seems to me that if you do it via `ast` you'll be forcing all things in the node tree and thus you'll be surprised about a workflow that does the `process_grid` thing, no?

pfebrer:
> I'll have to wrap my head a bit more around the node stuff, I am a bit worried it might make the code more complicated and not as intuitive.

Which part will be more complicated? I don't mean to define sisl functions as nodes, I just mean to store the `Node` and `Workflow` class in `sisl.nodes` instead of `sisl.viz.nodes`. Although since the functionality has nothing to do with sisl or science I also thought of creating a separate python package that `sisl.viz` would depend on. I don't know 😅

> A note on the workflows, it seems to me that if you do it via ast you'll be forcing all things in the node tree and thus you'll be surprised about a workflow that does the process_grid thing, no?

What do you mean by doing the `process_grid` thing? Inside the workflow you would just call `process_grid()`, so `process_grid` is just converted to a node without caring about what's inside. All you know is that the developer wants to checkpoint the `process_grid` function.

zerothi:
1. Ok, sure lets do that! Could you make a separate PR, then we can get it in immediately, I'll promise to just accept any PR's in that sub-module so you can quickly go about it ;)
2. It is more that end-users needs to be aware of the shortcomings of the workflow. Meaning they have to consider that every function call will be a node (static memory consumption). And also, what happens for `x = y + z` code lines? Will this be turned into a lambda and then a node? I found this article which arguably has some points to the matter https://www.georgeho.org/manipulating-python-asts/ It might be dangerous to change the code `in-place`. But if you like it the way it is, then fine!

pfebrer:
2. I want to clarify that I'm not replacing the code in place. When you create a workflow, you get a new object that is the workflow, but the original function remains untouched (it's the same as for a normal node).

Regarding `x = y + z`, I don't know what to do with it. The thing is that if `y` or `z` are a node, then the sum produces a node as well. So the only problem is if `y` and `z` are both direct inputs of the workflow. Then you would need to do what you are saying. I don't know if it is worth it to try and catch these things. You need to determine all the inputs 

My first idea was to restrict the usage of workflows to very simple things. At the beggining I thought I would only allow assignments from function calls. E.g.:

```python
@Workflow.from_func
def my_workflow(a, b):
      val = calc_val(a, b)
      shift = calc_shift(a)
      return shift_val(val, shift)
```

I think if you do this you are forcing devs to create easy to understand workflows and it is very clear where the checkpoints are. 

Then I thought I could just replace function calls by node calls and everything would work the same. So in my current implementation it is now valid to do:

```python
@Workflow.from_func
def my_workflow(a, b):
      return shift_val(calc_val(a, b), calc_shift(a))
```

and it produces the same workflow with the same checkpoints. My idea in doing this was that it would be much simpler to convert any given function into a workflow, which is nice. 

However, I don't know if I should go back to forcing the first syntax. I like it because it explicitly declares the checkpoints. And you can access checkpoints of the workflow from variable names (e.g. `workflow.val` or `workflow.shift`). In some sense these are not just checkpoints, it is as if you had a state, only that you know what are the minimal functions to run to keep it up to date.

What I'm thinking now is that I could support the second syntax, but a variable declaration could mean that we should create a breakpoint. This is a bit different because checkpoints would be a thing external to nodes instead of using their intrinsic functionality. One could allow:

```python
def my_workflow(a, b):
      # Private variables indicate that these are just temporal values and should not be checkpointed
      _val_left = calc_val(a, b)
      _val_right = calc_val(b, a)
     
      # I want to create a checkpoint here
      val = _val_left + _val_right 

      # Another checkpoint is implicitly created for the output.
      return shift_val(val, calc_shift(a))
```

Sorry for the long text :) Any ideas on what you think is best?

pfebrer:
In the last idea, you wouldn't necessarily need to use `_*` for temporal variables, it could also be:

```python
@Workflow.from_func(store=["val"])
def my_workflow(a, b):
      # Temporal values that are not checkpointed
      val_left = calc_val(a, b)
      val_right = calc_val(b, a)
     
      # I want to create a checkpoint here
      val = val_left + val_right 

      # Another checkpoint is implicitly created for the output.
      return shift_val(val, calc_shift(a))
```

zerothi:
Would all `nodes` (`calc_val`, `calc_shift`) be present in the workflow node list? I think one shouldn't be able to access variables, rather they should access nodes and use `get`:

```python
val = my_workflow.nodes.calc_val.get()
```
should be the way to get it, no? Since each node is accessible and the values them selves are `.get()`'ed then you wont have this problem.  
In this way changing of variable names is not a problem! Consider taking a workflow that does `A`, but now it should also do `B`, but you accidentally used a variable name in the first workflow that clashes with the understanding of the 2nd workflow...

zerothi:
To add to that, if a user wants a check-point, they should use a node! :)
So wanting to checkpoint your sum, you would need to `nodify` it ;)

pfebrer:
> So wanting to checkpoint your sum, you would need to nodify it ;)

Yes, but that's what happens automatically, thanks to your advice on `Ufuncs`. Any operation that you do with nodes results in a node. E.g. if you do `y = node_1 + node_2`, then `y` is a "`SumNode`" with it's inputs linked to `node_1` and `node_2` outputs. That is as if you did: `y = SumNode(node_1, node_2)`.

> This of course gets problematic when you have multiple calls to the same node function... hmm... How to do that!

Yes, exactly. Naming checkpoints by their node name has the downside that if you use the same node more than once you need to modify the names. What this PR implements right now is to add a number to the name if that's the case. I.e. there is `calc_val` and `calc_val_2`.

I thought that giving names to checkpoints was much nicer for this and because a normal user would understand it much better without knowing the inner workings I believe. `calc_val` is the generic name for the function that computes the value, but if you give a name to its output it can be much more meaningful in the context of the workflow. 

The naming for the checkpoints would just be a local table to access the nodes outputs, it doesn't impose anything on the node itself. So a separate workflow would be able to use the same node and map its output to another name. (this is just the normal way variables work in python, nothing particularly strange, you can have multiple variables pointing to the same object and each named differently).

pfebrer:
One could even argue if the variable name is more stable than the node name. You might change the way you compute a given variable of your workflow, or the function name from a third party package might simply change.

zerothi:
Well, everything is up for changes ;)
The problem is sharing workflows. It requires documentation.

With this I think your idea of *notifying* the workflow what to expose is the best idea.
```python
@Workflow.from_func(store=["val"])
```
You could imagine a `Workflow.add_store("variable")` in case users wants to examine a temporary variable.

The above approach ensures that the implementors of the workflow will document the stored variables for consistency and ease of use.

pfebrer:
> You could imagine a Workflow.add_store("variable") in case users wants to examine a temporary variable.

Yes, my idea was exactly that! It would generate a copy of the workflow with this extra checkpoint.

> The problem is sharing workflows. It requires documentation.

I would say it requires the same amount of documentation as any other object, no? You need to document the variables that you expose and try not to change them, private variables need less documentation since they are not expected to be used by users, etc...

zerothi:
> > You could imagine a Workflow.add_store("variable") in case users wants to examine a temporary variable.
> 
> Yes, my idea was exactly that! It would generate a copy of the workflow with this extra checkpoint.

Good.

> > The problem is sharing workflows. It requires documentation.
> 
> I would say it requires the same amount of documentation as any other object, no? You need to document the variables that you expose and try not to change them, private variables need less documentation since they are not expected to be used by users, etc...

Yes, but here it is a bit more difficult, right. 
The workflow returns `something`, but you want users to be able to access temporary variables defined (in the workflow)!
This is not the same as documenting a function! Variables in a function is by definition `private`, what you are doing is making them public in some way.

Ideally the `add_store` should change the documentation and insert the variable name and some documentation on the variable, in this way it becomes automatic, and users will also more easily document their changes.

zerothi:
Yes, just so we agree that the documentation needs to follow the store values ;)

pfebrer:
Ok, so it is now clean enough for you to check (not finished yet). I feel it's better to check at the code directly instead of looking at the changes.

The new `viz` module has the following submodules:
- `data`: Contains the data structures with multiple entry points, to make sure the data looks the same regardless of where it came from.
- `processors`: Contains many functions to process data in a functional way.
- `plotters`: Functions that determine the actions to perform on the plot (backend agnostic).
- `figure`: Functions that actually draw the plot. Each backend has their own figure implemented, implementing the drawing actions that the `sisl.viz` uses (`draw_line`, `draw_scatter`...)
- `plots`: Combining all of the above, the implementation of specific plots (bands, pdos, geometry) as functions. We also define the workflow (`sisl.node.Workflow`) version of them. When registering plotables, the workflow version is used, so that the plot can be modified (with `update_inputs`, `update_settings` has been deprecated).
- `data_sources`: This is a bit experimental. The functionality here could be in `processors`. It contains functions that process data to generate usually styles for atoms, bonds, etc... The idea is that they can help link different plots. E.g. an `AtomPDOS` data source will help styling atoms in a geometry plot from pdos data, etc... I still need to think about the best way to organize this.

This modular and functional design has allowed the tests to be much more "normal" now. See for example `processors/tests/*`, where I am able to test the functionality of each function separately, as it should be :) It also makes it easier to have a high coverage (> 90 % for processors).

pfebrer:
Now it's almost finished (I'd say it is finished but I would like to give it a final look after I sleep). 

All the documentation has been updated (it is much simpler :partying_face: ).

zerothi:
Thanks for this.

Some comments:
- your naming choices seem very blended, I have a hard time distinguishing what `data` vs `data_sources` exactly means, are they related, if so, could `data_sources` be moved into `data`?
- same thing with `plots` and `plotters`, could we name them something that defines them more appropriately? 
- `processors` is probably just me, but these looks like tools to manipulate sisl objects, or results thereof? `processors` makes me think of CPU's...
Otherwise, looks great! It is big, so will take me time to get used to this :)

zerothi:
Also, could you run `isort` on the code?

pfebrer:
Thanks! 

Do you have naming suggestions? I'm happy to change the names.

I think it would be useful for you to go through the `docs/visualization/viz_module/diy/Building a new plot class.ipynb` notebook. It goes through the main concepts and it is now very simple! It shouldn't take more than 5min

zerothi:
Hmm, I don't have a good suggestion now, are they something the end user will play with? Or is it an internal thing (then we can just do it later when we have better ideas?)

zerothi:
Just an idea, would it make sense to do something like this:
```python

def inject_plot(yfunction, base=sisl.viz.Plot):
    def dec(cls):
        d = dict(cls.__dict__)
        d["function"] = staticmethod(function) # or possibly check if it needs to be a class call?
        return type(cls.__name__,cls.__bases__ + (base,), d)
    return dec

@inject_plot(a_cool_plot)
class CoolPlot:
    def color_like(self, color): ...
```
Or is this annoying, another complexity level?

Otherwise, it looks good! :)

pfebrer:
I think the `class` definition not inheriting from `Plot` is too obscure, isn't it?

But I like the idea of the decorator as an alternative to defining the `function` variable.  Or would it be too little benefit to have the decorator and explicitly inherit from `Plot`.

By the way is it possible to support

```python
class CoolPlot(Plot, a_cool_plot):
    ...
```

? (I know, a bit crazy haha)

Or maybe

```python
class CoolPlot(Plot[a_cool_plot]):
    ...
```

This one is possible for sure, but I don't know if it is sane, you tell me 😅

zerothi:
But my decorator is having `Plot` as the base? Check `type` arguments. The `base` argument allows to change the default base. It might check that base is a subclass of `Plot`, just to be sure! 

Your first is not viable, and the 2nd is too obscure IMHO...

pfebrer:
Yes I know that it has the `base` argument, but the class definition without inheritance looks very weird 😅

The `Plot[function]` syntax would just be a shorthand for `Plot.from_func(function)`, which already exists (this is `Node` functionality).

Maybe if we don't agree we could just leave it as it is for now. And then if someone uses it (which would already be very nice) and complains about it being too verbose we can think about it.

zerothi:
> Yes I know that it has the `base` argument, but the class definition without inheritance looks very weird 😅
Ah, like that. Yeah, true.

> The `Plot[function]` syntax would just be a shorthand for `Plot.from_func(function)`, which already exists (this is `Node` functionality).

I think this is too weird. I would think an index of a plot could possibly be hooked only to the backend, that wouldn't be so weird. But an indexer that subclasses with a function handle... hmm... not something I would expect.

> 
> Maybe if we don't agree we could just leave it as it is for now. And then if someone uses it (which would already be very nice) and complains about it being too verbose we can think about it.

Yes, lets do that. :)
Could you go through the tests that fail, i.e. there is one place where you have:

```
def test_calculate_gap(bands_data, gap):
    
        spin = bands_data.attrs["spin"]
    
        gap_info = calculate_gap(bands_data)
    
        # Check that the gap value is correct
>       assert gap_info['gap'] == gap
E       assert 2.4695733283635555 == 2.5
```
I think here it would be better to use `gap_info['gap'] == pytest.approx(2.5)` with appropriate range.

Could you go through them and make the tests succeed, then I'll merge!

pfebrer:
Fixed the bug :)

But don't merge yet, I realised I still need to add some docstrings.

pfebrer:
Hmm this is very strange, the test is failing with exactly the same exact gap value and only for spin orbit.

And I can't reproduce the failing test locally. Can you? The test data is generated randomly so I wonder if the CI sets a value for the random seed and that is what explains the exact same value error.

I don't think it is just a numerical issue, because from 2.5 to 2.46 it's too much error. That's why I don't want to add `pytest.approx()`.

zerothi:
I get these errors:
```
    @pytest.mark.parametrize("nodes_lazy", [True, False])
    def test_workflow(nodes_lazy):
    
        def sum_node(a, b):
            return a + b
    
        @Workflow.from_func
        def my_workflow(a, b, c):
            first_sum = sum_node(a, b)
            return sum_node(first_sum, c)
    
        with temporal_context(context=Node.context, lazy=nodes_lazy):
            #It shouldn't matter whether nodes have lazy computation on or off for the working of the workflow
            with temporal_context(context=Workflow.context, lazy=True):
>               val = my_workflow(a=2, b=3, c=4)

/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/tests/test_context.py:59: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:103: in __init__
    self.setup(*args, **kwargs)
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/workflow.py:740: in setup
    super().setup(*args, **kwargs)
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:131: in setup
    self._output = self._blank
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/workflow.py:846: in _set_output
    self.nodes.output._output = value
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:537: in __getattr__
    return GetAttrNode(obj=self, key=key)
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:110: in __init__
    self.get()
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:355: in get
    evaluated_inputs = self.map_inputs(
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:314: in map_inputs
    input_val = func(input_val)
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:340: in evaluate_input_node
    return node.get()
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/workflow.py:825: in get
    return self.nodes.output.get()
/opt/gnu/12.3.0/python/packages/3.11.4/sisl-dev/0/lib/python3.11/site-packages/sisl/nodes/node.py:537: in __getattr__
    return GetAttrNode(obj=self, key=key)
E   RecursionError: maximum recursion depth exceeded
!!! Recursion detected (same locals & position)
```
```
bands_data = <sisl.viz.data.bands.BandsData object at 0x14fbed324410>, gap = 2.5

    def test_calculate_gap(bands_data, gap):
    
        spin = bands_data.attrs["spin"]
    
        gap_info = calculate_gap(bands_data)
    
        # Check that the gap value is correct
>       assert gap_info['gap'] == gap
E       assert 2.4695733283635555 == 2.5
```

I did use the updated one...

zerothi:
But is the gap you have entered to the test the correct one? I can not imagine anything where the gap is exactly 2.5...

pfebrer:
Those errors are locally in your computer?

In this case it is exactly 2.5 (to machine precision, which I don't think will introduce an error of 0.03) because the data is sintetically generated with that gap.

pfebrer:
I thought the error could be related to python 3.11 but even with 3.11 I could not reproduce it 😪

zerothi:
Yes, locally on my machine... That is weird...

pfebrer:
And to add more mistery it only fails for spin orbit, not even non-colinear spin. Hmm I think I'll mark it as skip for now.

The workflow one I also get it so I'll try to fix it.

pfebrer:
Ok, interesting, I can reproduce if I run the full test suite (previously I was only running the `viz` tests)

zerothi:
> Ok, interesting, I can reproduce if I run the full test suite (previously I was only running the `viz` tests)

Sounds like a state?

pfebrer:
It's a bug in the generation of the synthetic data. The bands are random polynomials, and if there were band crossings in the valence band the gap was not properly set. At some point of running the tests numpy's random seed is set and that's why it's generating always exactly the same gap.

zerothi:
Could you rebase, then the tests should run through

pfebrer:
I still haven't pushed the fix, give me 30 min or something like that hehe

pfebrer:
Unbelievably, this is finally ready for merge :)

zerothi:
Some things that needs to be cleaned in the commits (probably I'll do a squash to remove everything)
- changes in `.ipynb` related to the version and nb_format should be reverted, it doesn't do anything
- changes in the `docs/*/generated` should be deleted, they should be auto-generated. So remove them.

I think that is more or less it.

pfebrer:
Done, sorry.

Shouldn't the `docs/**/generated` directories be in `.gitignore`?

Issue #478, opened 2022-07-26
Rename bond current to transmission


zerothi:
Should we rename `bond_current` to `bond_transmission` and let `bond_current` be the bias window integrated quantity?

tfrederiksen:
Such a change sounds sensible to me.

anshugaur:
Dear Nick,
 1.  Does sisl (ver. 0.12.2) calculate bond_current (bias window integrated quantity)? 
 2.  In sisl, do the extracted properties (orbital_current, bond_current, atom_current, and vector_current) have any dependence on the applied bias?
 3. The transmission (vs. energy) plots calculated from TBT at different biases (V) are shifted in energy by +V/2. (image attached). While comparing energy-resolved quantities (from sisl) at different applied biases, should we compare these values at the same energy or energies corresponding to the same transmission at applied bias?
 4.  Is it meaningful to compare these properties at a non-zero bias (point 3)?

We have a large multi-component system and we are interested to check which atoms (bonds_current, vector_current) contribute towards the current pathway leading to total current at non-zero bias.  The system contains ionic bonds and we see large bond_current (or bond_transmission) between atoms to different ionicity. However, our hypothesis is that at an applied bias (larger than the band gap), the current should be primarily through single atoms (or orbitals, of the same ionicity) which form the edge of the conduction band.

Thanks,
Anshu
![Transmission](https://user-images.githubusercontent.com/39057814/181221768-adb1b50e-2ebf-4a99-936d-236c1a68f3e6.png)

zerothi:
1. No it is not an integrated quantity.
2. Why would they not be dependent on bias? And why would they be dependent on the bias? After that line of thought, what would you think is the right answer?
3. I don't understand what you mean? My response in the prior thread makes it clear that the transmission and bond transmissions are related.
4. Sure, why not? But you have to be more careful at what you compare. Interpreting this is your responsibility as the researcher.

zerothi:
This was fixed in #479

PR #479, opened 2022-07-29
mnt: fixed #477 and #478, changed *_current to *_transmission

### LINKED ISSUE #477: Energy in Bond currents using SISL
Whenever I try to extract bond currents from .nc file, A energy is required to fill in "bond_current(elec, E, kavg=True, isc=None, only='out', orbitals=None, uc=False)"  I don't understand what this Energy means? Please help me resolve it. 
Additional information about the system - The system has a 0.65 eV band gap. When plotting transmission, the first peak on the right side of transmission lies at E = 0.075 eV and the second peak at E = 0.405 eV.
Thanks in advance

zerothi:
The orbital_current was confusing for end-users as it
signalled it was a bias-window integrated quantity.

Now the routines have changed names and additional ones
have been added.

*_transmission: are the orbital/atom/... resolved
  transmissions (they have `E` as argument)
*_current: are the orbital/atom/... bias window
  integrated quantities that reflects the total
  current (they don't have `E` as argument)

This will break old users scripts as the interfaces
has also changed.
Now all electrode arguments defaults to the 1st electrode
and energy arguments are generally the first ones (except DOS).

This *should* be more what users want as for 2 electrode systems
they'll generally only calculate one orbital_current for the first
electrode.

For old users one can accommodate these settings by using this
small snippet:
```python 
if sisl.__version_tuple__[:3] >= (0, 13, 0):
   tbt.orbital_transmission(0.1, "Left")
else:
   tbt.orbital_current("Left", 0.1)
```
Same goes for the other equivalent routines.

I am not sure whether the interface changes are a good idea, I would be grateful for any comments! @tfrederiksen @sofiasanz (others!)

 - [X] Closes #477 and #478
 - [X] Tests added
 - [X] Documentation for functionality
 - [X] User visible changes (including notable bug fixes) are documented in `CHANGELOG`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once
 we hit release version 1.0.0.
 
-## [UNRELEASED] - YYYY-MM-DD
+## [0.13.0] - YYYY-MM-DD
 
 ### Added
 - enabled unit specification for lengths in cube-files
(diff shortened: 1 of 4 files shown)


zerothi:
@harshiiiiiii and @anshugaur comments on these changes are most welcome!

harshiiiiiii:
Thanks for the clarification and updates @zerothi .
Nothing more for now.

zerothi:
I'll merge this and see whether comments arrive ;)

PR #482, opened 2022-08-17
Too large energies printed in siesta made sisl crash

pfebrer:
When SIESTA has to print an energy value that it is too large to be formatted, it prints `**********************`. This was making `outSileSiesta.read_energy()` fail in those cases, since it was trying to convert the string to a float. In my case the very large energy was the ion-electron energy, and although I only cared about the total energy it reads all of them and crashes because it can not read a particular one.

With this commit when the energy can not be parsed to float `np.nan` is asigned. We could check for asterisks in the string, but I think that would be a waste of time. Also, I didn't know whether to choose `np.inf` over `np.nan`. The problem is that the sign of the "infinite" value is unknown and it is not really infinite, so I think `np.nan` is better.

### DIFF
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -15,6 +15,7 @@ from sisl import Geometry, Atom, SuperCell
 from sisl.utils import PropertyDict
 from sisl.utils.cmd import *
 from sisl.unit.siesta import unit_convert
+from sisl.messages import warn
 
 __all__ = ['outSileSiesta']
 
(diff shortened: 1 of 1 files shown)


zerothi:
Great addition!

One thing, I would like a warning to be raised (so users are aware of possible output-problems). We already have the key, so it would be specific! Could you please add this in the catch clause?

pfebrer:
Done! You can change the message to whatever you prefer :sweat_smile:

Issue #486, opened 2022-09-19
Sign of the y-component of spin moments may be wrong


juijan:
Hey Nick,

is it possible that the y-component of the spin moments is incorrectly calculated?

It looks to me like we should add a minus, because:

$$\eqalign{
\mathbf{S}_i^y 
   &= Re( \langle \psi_i | \boldsymbol\sigma_y \mathbf S | \psi_i \rangle ) \\
   &= Re( \left(\matrix{\langle \psi_i^{\uparrow} | \cr \langle \psi_i^{\downarrow}| } \right)
             \left(\matrix{ 0 & -i \mathbf S \cr i \mathbf S & 0} \right)
             \left(\matrix{| \psi_i^{\uparrow}\rangle \cr | \psi_i^{\downarrow}\rangle } \right)\\
   &= Re(- i \langle \psi_i^{\uparrow} |  S | \psi_i^{\downarrow} \rangle + i \langle \psi_i^{\downarrow} |  S | \psi_i^{\uparrow} \rangle ) \\
   &= Im(  \langle \psi_i^{\uparrow} |  S | \psi_i^{\downarrow} \rangle - \langle \psi_i^{\downarrow} |  S | \psi_i^{\uparrow} \rangle) \\
   &= D2.imag - D1.imag
}$$

https://github.com/zerothi/sisl/blob/3f11290ec7dd11e87c4b04f6eca24e568b11ed4d/sisl/physics/electron.py#L410-L417
https://github.com/zerothi/sisl/blob/3f11290ec7dd11e87c4b04f6eca24e568b11ed4d/sisl/physics/electron.py#L427-L432

zerothi:
Did you check with siesta? Thanks!

juijan:
I did, and now I have yet another suggestion.

If I understand it correctly I believe this is implemented in the [spin_texture](https://gitlab.com/siesta-project/siesta/-/blob/master/Util/COOP/spin_texture.f90) utility:

$$\eqalign{
\mathbf{S}_i^y= - Im(\langle \psi_i^{\uparrow} |  S | \psi_i^{\downarrow} \rangle +  \langle \psi_i^{\downarrow} |  S | \psi_i^{\uparrow}\rangle) = - D2.imag - D1.iamg
}$$

zerothi:
Your equation in the initial question looks correct, i believe that the single minus sign in spin texture is due to siestas convention on the off-diagonal imaginary part in the spin box. But I have to double check, now we are at it.

[...]

PR #487, opened 2022-09-19
Fix issue 486 - in correct sign of the y component of the spin moments

juijan:
Fix issue #486 - in correct sign of the y component of the spin moments


zerothi:
Did you check with siesta values?

zerothi:
I added a new commit which also fixes PDOS calculation. :)

Could you have a look?

juijan:
The commit for the PDOS calculations looks good to me.

zerothi:
Thanks!!

Next time, please ensure you have a clean *main*, i.e. your main was several commits behind :)

PR #489, opened 2022-09-23
bug: fixed #488 with the cell order

### LINKED ISSUE #488: Reorder honeycomb lattice vectors?
I used `sisl.geom.graphene` to write a simple POSCAR file of graphene. However, VASP complains about this POSCAR (and quits!) because the lattice vectors do not form a right-handed system:
```
ERROR: the triple product of the basis vectors 
 is negative exchange two basis vectors
```
Therefore, what do you think about permanently interchanging the first two lattice vectors of `honeycomb` in sisl? Perhaps other codes will also make this complaint?

zerothi:
Also added tests. @tfrederiksen 

 - [x] Closes #488
 - [x] Tests added

### DIFF
--- a/sisl/geom/tests/test_geom.py
+++ b/sisl/geom/tests/test_geom.py
@@ -3,8 +3,9 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 import pytest
 
-from sisl import Atom
+from sisl import Atom, SuperCell
 from sisl.geom import *
+from sisl._math_small import cross3, dot3
 
 import math as m
 import numpy as np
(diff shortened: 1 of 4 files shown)


sisl/geom/tests/test_geom.py
```diff
@@ -13,36 +14,59 @@
...
 
 
+class CellDirect(SuperCell):
+
+    @property
+    def volume(self):
+        return dot3(self.cell[0, :], cross3(self.cell[1, :], self.cell[2, :]))
+
+
+def is_positive(geometry):
```
tfrederiksen, line 24: Maybe this function could be named `is_righthanded` to be more explicit about what is tested?

zerothi: done! Thanks!

PR #494, opened 2022-10-10
enh: enabling reading of file-handles and StringIO

### LINKED ISSUE #484: Using sisl to parse aiida-siesta output
**Describe the feature**
I am trying out AiiDA to manage my workflow with siesta. I really like what the aiida-siesta (https://docs.siesta-project.org/projects/aiida-siesta/en/latest/) has already implemented in terms of calculations and workflows. This plugin also has a simple parser for parsing some of the basic siesta output files.I would like to extend the functionality to also be able to parse information from other siesta output files. Sisl has a large collection of parsers available so it would be great if I could use these to parse the output files produced by siesta. However, I haven't found a good robust way yet to use sisl for this. AiiDA stores retrieved output files from simulations in a file repository and gives you access to these files through a `FolderData` API (https://aiida.readthedocs.io/projects/aiida-core/en/stable/topics/data_types.html#topics-data-types-core-folder). This API is able to provide a file handle object for reading or can simply read and return the entire content of the file. On the other hand, to use sisl for parsing a file you need to supply the filename of the file and sisl handles the rest. Unfortunately I don't have access to the filename, just to either a file handle or the complete file content. Is there a way to make sisl work with AiiDA retrieved files? So far I have come up with a couple of potential solutions:
1. Write the contents of the AiiDA retrieved file to a temporary file in the current working directory with the right extension, use `get_sile` to parse this temporary file and then remove it. This should work out of the box but involves a lot of unnecessary IO. This also probably becomes unnecessarily slow if you have large output files.
2. Modify sisl to not only accept filenames but also the entire file contents as a string (or something equivalent for binary files). You could then maybe use the `StringIO` and `BytesIO` classes to provide a file like interface to the other methods in the sile. The big disadvantage from this is that you cannot derive the filetype from the extension so you would have to supply it some other way. And this may become slow and require tons of memory (because the entire file needs to be loaded into memory) if you have large files.
3. Modify sisl to accept a file handle instead of a filepath. This to me seems like the nicest option. You still need some way to determine the file type (i.e. you could give a filepath with extension in addition to the file handle (the name of the original file is still available from within AiiDA so this would be easy) or there could be some other input parameter for this).

What do you think? Is there a way that sisl could be used for parsing files retrieved by AiiDA? And if so, do any of the above solutions look like a good one? Or do you have a better proposal? I would also be happy to help with any implementation and/or testing once we have settled on an approach.

zerothi:
This patches the Sile (ascii) formats to enable
reading directly from a TextIOBase, anything else will be parsed as a filename to be used normally.

This is done using the __new__ method to patch the class creation and omit the file's opening etc.
Only 2 tests have been added for now, but more should be added to assert it works on a multitude of things.

Probably the fdf sile will be problematic due to the include statements that may interfere.

The dynamic class creation will append a "Buffer" suffix so the class nature is clear. However there will only be some internal methods which will be overwritten.

This addition was requested and great inspiration from @pfebrer and @ahkole, thanks!

Also enabled users to overwrite the dynamic creation of the buffer-class for user-defined classes.
This may come in handy in certain situations where the buffersile is not enough.

 - [x] Closes #484
 - [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ we hit release version 1.0.0.
 ## [0.13.0] - YYYY-MM-DD
 
 ### Added
+- enabled ASCII siles to read from file-handles and buffers, #484
 - enabled unit specification for lengths in cube-files
 - added `kwargs` passed to eigenstate functions in `berry_phase`
 	and `conductivity`
--- a/sisl/io/siesta/tests/test_basis.py
+++ b/sisl/io/siesta/tests/test_basis.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 import pytest
 import os.path as osp
+from io import StringIO
 from sisl.io.siesta.basis import *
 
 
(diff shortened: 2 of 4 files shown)


zerothi:
@ahkole, please check-out this PR, it should have everything you need?

Once you clear it, it will go in!

zerothi:
Note that the iter problem was fixed upstream in the main branch. So you won't see it here

zerothi:
@ahkole, I believe I can just merge this, please open a new issue if needed!

PR #499, opened 2022-11-01
Easy construction of hydrogenic orbitals

tfrederiksen:
Suggestion for a new class that provides the normalized radial function for hydrogenic orbitals.

### DIFF
--- a/sisl/orbital.py
+++ b/sisl/orbital.py
@@ -1140,7 +1140,7 @@ class HydrogenicOrbital(AtomicOrbital):
 
     Parameters
     ----------
-    Z : float
+    Zeff : float
         effective atomic number
     n : int
         principal quantum number
@@ -1157,14 +1157,21 @@ class HydrogenicOrbital(AtomicOrbital):
 
     """
 
-    def __init__(self, Z, n, l, m, *args, **kwargs):
+    def __init__(self, Zeff, n, l, m, *args, **kwargs):
+
+        self._Zeff = Zeff
 
         R = kwargs.get("R", 10.)
         r = np.linspace(0, R, 1000)
         a0 = 0.529177 # Bohr radius
-        z = 2 * Z / (n * a0)
+        z = 2 * Zeff / (n * a0)
         pref = (z ** 3 * factorial(n - l - 1) / (2 * n * factorial(n + l))) ** 0.5
         L = eval_genlaguerre(n - l - 1, 2 * l + 1, z * r)
         Rnl = pref * np.exp(-z * r / 2) * (z * r) ** l * L
 
         super().__init__(n, l, m, (r, Rnl), q0=kwargs.get("q0", 0.), tag=kwargs.get("tag", ""))
+
+
+    def copy(self):
+        """ Create an exact copy of this object """
+        return self.__class__(self._Zeff, self.n, self.l, self.m, self.q0, self.tag)
--- a/sisl/orbital.py
+++ b/sisl/orbital.py
@@ -10,7 +10,7 @@ from math import factorial as fact
 import numpy as np
 from numpy import cos, sin
 from numpy import take, sqrt, square
-from scipy.special import lpmv
+from scipy.special import lpmv, factorial, eval_genlaguerre
 from scipy.interpolate import UnivariateSpline
 
 from ._internal import set_module
@@ -21,7 +21,7 @@ from .shape import Sphere
 from .utils.mathematics import cart2spher
 
 
-__all__ = ["Orbital", "SphericalOrbital", "AtomicOrbital"]
+__all__ = ["Orbital", "SphericalOrbital", "AtomicOrbital", "HydrogenicOrbital"]
 
 
 # Create the factor table for the real spherical harmonics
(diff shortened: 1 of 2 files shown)


sisl/orbital.py
```diff
@@ -1130,3 +1130,50 @@ def __setstate__(self, d):
             self.__init__(d["name"], q0=d["q0"], tag=d["tag"])
         else:
             self.__init__(d["name"], (d["r"], d["f"]), q0=d["q0"], tag=d["tag"])
+
+
+@set_module("sisl")
+class HydrogenicOrbital(AtomicOrbital):
+    r""" A hydrogen-like atomic orbital defined by an effective atomic number Z
```
zerothi, line 1137: The first line should be a single line (shorthand in documentation).

Then additional descriptions should follow.

I think it would be nice to add the following information:
```
A hydrogenic atom (Hydrogen-like) is an atom with a single valence electron.
```

Could we also put:

```
See [1]_ for details about the hydrogenic atom.

References
---------------
.. [1] : https://en.wikipedia.org/wiki/Hydrogen-like_atom
```
this will put the link in a reference section.

sisl/orbital.py
```diff
@@ -1130,3 +1130,50 @@ def __setstate__(self, d):
...
+    in addition to the usual quantum numbers (n, l, m).
+    The returned orbital is properly normalized.
+
+    See https://en.wikipedia.org/wiki/Hydrogen-like_atom
+
+    Parameters
+    ----------
+    Zeff : float
```
zerothi, line 1145: Could we perhaps just use `Z`? And emphasize `effective`? More if future orbitals also makes use of Z?

sisl/orbital.py
```diff
@@ -1130,3 +1130,50 @@ def __setstate__(self, d):
...
+
+        R = kwargs.get("R", 10.)
+        r = np.linspace(0, R, 1000)
+        a0 = 0.529177 # Bohr radius
+        z = 2 * Zeff / (n * a0)
+        pref = (z ** 3 * factorial(n - l - 1) / (2 * n * factorial(n + l))) ** 0.5
+        L = eval_genlaguerre(n - l - 1, 2 * l + 1, z * r)
+        Rnl = pref * np.exp(-z * r / 2) * (z * r) ** l * L
+
+        super().__init__(n, l, m, (r, Rnl), q0=kwargs.get("q0", 0.), tag=kwargs.get("tag", ""))
```
zerothi, line 1174: here just do:
```
super().__init__(n, l, m, (r, Rnl), **kwargs)
```

sisl/orbital.py
```diff
@@ -1130,3 +1130,50 @@ def __setstate__(self, d):
...
+
+    """
+
+    def __init__(self, Zeff, n, l, m, *args, **kwargs):
+
+        self._Zeff = Zeff
+
+        R = kwargs.get("R", 10.)
+        r = np.linspace(0, R, 1000)
+        a0 = 0.529177 # Bohr radius
```
zerothi, line 1168: I have added the Bohr radius in the constant module, could you use `a0("Ang")`? This should give the same value (but with more significant digits).

zerothi (changes requested):
Looks great!

Just some minor details and questions.

sisl/orbital.py
tfrederiksen: Better now?

sisl/orbital.py
tfrederiksen: Done

sisl/orbital.py
```diff
@@ -1130,3 +1131,54 @@ def __setstate__(self, d):
...
+        pref = (z ** 3 * factorial(n - l - 1) / (2 * n * factorial(n + l))) ** 0.5
+        L = eval_genlaguerre(n - l - 1, 2 * l + 1, z * r)
+        Rnl = pref * np.exp(-z * r / 2) * (z * r) ** l * L
+
+        super().__init__(n, l, m, (r, Rnl), **kwargs)
+
+
+    def copy(self):
+        """ Create an exact copy of this object """
+        return self.__class__(self._Z, self.n, self.l, self.m, self.q0, self.tag)
```
zerothi, line 1184: this won't work, just saw... You are discarding *args* in the above, so you should pass q0 and tag as keyword arguments.

Actually you should remove `*args` from `__init__` to ensure that users don't pass something that is lost.

tfrederiksen: I see. Better now?

zerothi: Yes, great!

zerothi:
Ok, one last thing, would you mind me moving the Z to the end of arguments? So the order would be: `n, l, m, Z`? In this way it looks the most like the AtomicOrbital?
I can do the change if you don't mind.

Once I have merged I will:
- amend the documentation (put the orbital in the correct place)
- Add a test for the copy
- Add `__getstate__` routines (for pickling)

tfrederiksen:
I've pushed some further changes. Feel free to take it further from here.

zerothi:
Now merged! Thanks!

PR #503, opened 2022-11-06
Plotly is incompatible with ipywidgets 8.x

pfebrer:
Apparently plotly is incompatible with ipywidgets 8.x for now (https://github.com/plotly/plotly.py/issues/3686) and this is making the plots in the docs not display properly:

Screenshot from 2022-11-06 13-53-55


zerothi:
Great! Thanks!

PR #504, opened 2022-11-07
Allow reading forces from transiesta in output file

pfebrer:
In transiesta runs, both siesta and transiesta forces are written to the output file. I don't know if the siesta forces make any sense, but those are what the `outSile.read_force` function read.

For transiesta runs, it should read the ts forces. I just added an argument to specify that you want `"ts"` forces and it works, but perhaps `outSile` should be able to determine if it is a siesta or transiesta run?

### DIFF
--- a/sisl/io/siesta/out.py
+++ b/sisl/io/siesta/out.py
@@ -282,7 +282,7 @@ class outSileSiesta(SileSiesta):
         return next_geom()[1]
 
     @sile_fh_open()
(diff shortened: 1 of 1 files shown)


pfebrer:
I could not add tests because there are no TS runs with forces in `sisl-files`.

zerothi:
Great! In the other read routines we use `key` for specification on which run type it is. I think it is best if all are the same.

Regarding automatically figuring out the run, hmm... That should also affect all the other routines no?

PR #505, opened 2022-11-07
Sile function `step_to(..., reread=True)`

tfrederiksen:
While working on #500 I found (sometimes) an unexpected behaviour of the `step_to(..., reread=True)` function: It would not always read from the very beginning of the sile.

Consider a simple text file `inp.txt` with the following lines:
```
1 a
1 b
2 a
2 b
...
```
and this test script
```python
import sisl
from sisl.io import Sile, add_sile, sile_fh_open

class mySile(Sile):

    @sile_fh_open()
    def read(self):
        self.step_to('b')
        l = self.step_to('a')[1]
        print(l)

add_sile('txt', mySile)

f = sisl.get_sile('inp.txt{mySile}')
f.read()
```
With sisl version 55250091a584f1faa8b1835ba9973d721d3b08c8 the printed line is `2 a`. I would have expected the default value `reread=True` to find the first appearance of "a" in the file, i.e., `1 a`.

This branch forces the file to be reopened from the very beginning when `reread=True`.

### DIFF
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -738,8 +738,7 @@ class Sile(BaseSile):
             # Force close and reopen from the beginning
             self.close()
             self._open()
-        # The previously read line...
-        line = self._line
+
         if isinstance(keywords, str):
             # convert to list
             keywords = [keywords]
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -734,6 +734,10 @@ class Sile(BaseSile):
         r""" Steps the file-handle until the keyword(s) is found in the input """
         # If keyword is a list, it just matches one of the inputs
         found = False
+        if reread:
+            # Force close and reopen from the beginning
+            self.close()
+            self._open()
         # The previously read line...
         line = self._line
         if isinstance(keywords, str):
@@ -746,21 +750,6 @@ class Sile(BaseSile):
             if l == '':
                 break
             found = self.line_has_keys(l, keys, case)
-
-        if not found and (l == '' and line > 0) and reread:
-            # We may be in the case where the user request
-            # reading the same twice...
-            # So we need to re-read the file...
-            self.close()
-            # Re-open the file...
-            self._open()
-
-            # Try and read again
-            while not found and self._line <= line:
-                l = self.readline()
-                if l == '':
-                    break
-                found = self.line_has_keys(l, keys, case)
 
         if ret_index:
             idx = -1
--- a/sisl/io/vasp/out.py
+++ b/sisl/io/vasp/out.py
@@ -100,10 +100,10 @@ class outSileVASP(SileVASP):
             "Solvation": "solvation",
         }
 
-        def readE(itt):
+        def readE(itt, reread=True):
             nonlocal name_conv
             # read the energy tables
-            f = self.step_to("Free energy of the ion-electron system", reread=False)[0]
+            f = self.step_to("Free energy of the ion-electron system", reread=reread)[0]
             if not f:
                 return None
             next(itt) # -----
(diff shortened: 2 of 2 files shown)


zerothi:
No, that is exactly the behaviour intended. For instance in an md simulation you would want to step to the next entry with the same key. Reread simply allows it to re-open it in case it was at a position in the middle of the file.

But maybe the documentation is lacking or not explicit enough.

In any case I don't think this pr should go in.

tfrederiksen:
> No, that is exactly the behaviour intended. For instance in an md simulation you would want to step to the next entry with the same key. Reread simply allows it to re-open it in case it was at a position in the middle of the file.

I think the natural choice is to use `reread=False` in the second call to `step_to` (if one wishes to continue from the current line).

But maybe I do not understand what is intended here.

tfrederiksen:
Put differently, how would you change my script above to get the code to return `1 a` *after* `step_to("b")` has been executed?

zerothi:
The point of `step_to` is to step to the next occurrence of a specific key in the file.

In your example one should manually close and re-open the file:

```python
self.close()
with self:
    self.step_to(...)
```
I know it is not elegant, perhaps `_open` should be made public to make this more clean...

However, the idea with `reread` is that:
```python
self.step_to('b')
l = self.step_to('1 a', reread=True)
```
would give you the line requested, whereas
```python
self.step_to('b')
l = self.step_to('1 a', reread=False)
```
would never find it because it had already stepped beyond it.

The idea would be that you don't have to re-read the content of a big file again and again if the key is located far from the beginning of the file, say you know the entries are *supercell*, *geometry*, *orbitals*, then you should just step in this order and you'll have minimal IO.

Perhaps it should have a new flag called `reopen`, and `reread` should be renamed to `allow_reread` to clarify intent.

tfrederiksen:
To me `reread=True` implies "reading the file again from the beginning" (i.e., potentially rereading some lines).

Therefore, with my example, I would say
```python
line = self.step_to('b') # first call, with both reread True or False, it brings us to
# line is "1 b"
line = self.step_to('a', reread=False) # with False we continue from "1 b" and get
# line is "2 a"
line = self.step_to('a', reread=False)
# line is "3 a", the MD-trajectory case, etc
```
On the contrary, one may want another behaviour:
```python
line = self.step_to('b')
# line is "1 b"
line = self.step_to('a', reread=True)  # now we force to start from the beginning,
                                       # independent of the line above, so
#line is "1 a"
line = self.step_to('a', reread=True) # we keep going from the beginning each time
line == "1 a"
```
Why is this not the intended behaviour?

tfrederiksen:
> Perhaps it should have a new flag called `reopen`, and `reread` should be renamed to `allow_reread` to clarify intent.

Sorry, with this I start to understand the intent (and thus the "unexpected" behaviour from my perspective mentioned initially). These two flags would make things much clearer, I think.

To clear things up, the current (and thus originally intended and correct) behaviour of `step_to` in sisl is this, right?
```python
f, line = self.step_to('b')
# line is "1 b"
f, line = self.step_to('a', reread=False) # continue from "1 b"
# line is "2 a"
while f:
    f, line = self.step_to('a', reread=False)
# The sequence will be 1b, 2a, 3a, ..., None
# Note that we here miss the first occurrence of 1a
```
and
```python
f, line = self.step_to('b')
# line is "1 b"
f, line = self.step_to('a', reread=True) # continue from "1 b"
# line is "2 a"
while f:
    f, line = self.step_to('a', reread=True)
# The sequence will be 1b, 2a, 3a, ...., 1a, 2a, 3a, ... infinite loop
```

zerothi:
> I can try to prepare a new PR with `reopen=False` and `allow_reread=True` as keywords to `step_to`, OK?

Sounds like a good idea! Thanks!

PR #511, opened 2022-11-13
Possible bug: within_inf with periodic=False returns sc atoms for nsc > 1

pfebrer:
I think I found a bug. If you call `Geometry.within_inf` with `periodic=False` it would work properly only if the geometry had `nsc=[1,1,1]`, otherwise it would basically not care if periodic was set to False and return atoms in neighboring cells.

Or was this intended behavior?


zerothi:
I believe you are correct, thanks for discovering this bug!

PR #514, opened 2022-11-17
Surface slab geometries: clarify `vacuum=None` as a possibility for building electrodes

tfrederiksen:
For building an electrode geometry, consistent with a given slab geometry, it may be convenient to use `vacuum=None` to get a crystal structure with the correct orientation, e.g., 
```python
elec = sisl.geom.fcc_slab(alat, atoms, 111, layers=3, vacuum=None)
device = sisl.geom.fcc_slab(alat, atoms, 111, layers=9, vacuum=20) * (3, 3, 1)
device = device.sort(axis=[2, 1, 0])
```
I tried to make this possibility more clear in the documentation.


zerothi:
Great, thanks!

PR #515, opened 2022-11-17
doc: fixed indices in sums

tfrederiksen:


zerothi:
Thanks for noticing and fixing!

PR #519, opened 2022-11-29
Fix reading of basis

juijan:
This partially reverts a change that breaks the reading of basis orbitals from 'ORB_INDEX' file when multiple species are present.

The current version works fine for the first species. However, not for the next species.

This happens because, as soon as the loop reaches a new species it will try to create a new atom. Instead of first reading all orbitals for the new species. When checking whether the current `i_s` is in the list of `species` we really want to check it for the previous atom, not for the current atom. 

Hence, moving the definition of `i_s`  below is necessary.

I assume this change was intended to avoid, accessing `i_s` or `spec`  before they are first assigned.

Instead, I suggest assigning an arbitrary value before the loop. The initial value will never to used, because `i_a != ia` can not happen before the second iteration of the loop.

### DIFF
--- a/sisl/io/siesta/orb_indx.py
+++ b/sisl/io/siesta/orb_indx.py
@@ -83,11 +83,11 @@ class orbindxSileSiesta(SileSiesta):
         orbs = []
         specs = []
         ia = 1
+        i_s = 0
         for _ in range(no):
             line = self.readline().split()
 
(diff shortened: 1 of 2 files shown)


zerothi:
Thanks for this! :)

Could you please have a look #516 ?

PR #521, opened 2022-12-06
Minor fixes for IO-API

tfrederiksen:
- removed TBtrans double-entry
- PHtrans as separate toc-entry (same spirit as the siesta/transiesta division)


zerothi:
Thanks!

PR #523, opened 2023-01-07
mnt: make viz work with new spin conventions

pfebrer:
All viz tests pass again.

I added a check for PDOS arrays that have `ndim == 2` because `pdosSileSiesta.read_data` returns an array without spin dimension in unpolarized cases, but the return should probably be changed to 3 dimensions for consistency with the rest of sisl.


zerothi:
you have accidentally changed the `files` submodule and pushed the changes here.

True about pdosSile problem, will fix ASAP!

pfebrer:
Sorry, fixed!

zerothi (approved):

PR #526, opened 2023-01-17
bug: reading the EDM from fdf did not read geometry automatically

pfebrer:
`fdf` was not an option for reading the geometry when reading the energy density matrix.

### DIFF
--- a/sisl/io/siesta/fdf.py
+++ b/sisl/io/siesta/fdf.py
@@ -1976,7 +1976,7 @@ class fdfSileSiesta(SileSiesta):
         if f.is_file():
             if "geometry" not in kwargs:
                 # to ensure we get the correct orbital count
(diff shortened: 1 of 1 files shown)


zerothi:
I have also added this possibility for the TSHS and overlap matrix. It shouldn't be necessary since the TSHS file always contain the geometry, however, it seems weird to not have it when the others are present.

Thanks!

pfebrer:
Does the TSHS contain all the basis information?

zerothi:
No, I plan on deprecating the TSHS file since the newest hsx file contains everything the tshs file contained, plus basis information. So it should be superfluous, I just need to change the electrode blocks to accept hsx files.

pfebrer:
Then shouldn't `fdf` come before `TSHS` in the order of preference for reading the geometry? (in the DM/EDM I mean)

PR #531, opened 2023-02-04
enh: Added Nodes and Workflows.

pfebrer:
This is the result of the discussion that started here: https://github.com/zerothi/sisl/pull/476#issuecomment-1375561025

A new submodule is introduced, `sisl.nodes`, that implements functionality to create checkpointable workflows based on nodes.

This framework will be used by the visualization module to make plots much more modular. It also allows more flexibility and the GUI might be able to automatically use nodes and workflows (would be great).

A notebook to introduce the framework has been added to the documentation. But very briefly, one can create a workflow like:

```python
from sisl.nodes import Workflow
import numpy as np

# Some helper functions that will be used inside the workflow
def normalize_data(data, vmin=0, vmax=1, normalize=True):
    if not normalize:
        print("NOT NORMALIZING")
        return data
    
    print("NORMALIZING")
    return (data - vmin) / max(data) * (vmax - vmin)

def mix(x, y, coef):
    print("MIXING")
    return coef * x + (1 - coef) * y

# The function that computes
def compute(start=0, stop=2, linear_shift=1, power=2, linear_coef=0.5, normalize=True):
    linear = np.linspace(start, stop) + linear_shift
    non_linear = np.linspace(start, stop) ** power
    
    mixed = mix(linear, non_linear, linear_coef)
    
    normalized = normalize_data(mixed, normalize=normalize)
    
    return normalized

# We convert it to a workflow
compute_flow = Workflow.from_func(compute)
```

All the computations done directly inside the workflow are then converted into nodes. You can visualize it (which I think helps a lot):

```python
compute_flow.network.visualize(notebook=True)
```

Screenshot from 2023-02-04 20-16-32

And then you can use it to get results:

```python
# Instantiate the workflow with all the default arguments except for power.
flow = compute_flow(power=3)

# Run it and get the result
flow.get()
```
```
MIXING
NORMALIZING
array([0.09090909, 0.09462585, 0.09837969, 0.10220772, 0.10614702,
       0.11023468, 0.11450779, 0.11900345, 0.12375873, 0.12881074,
       0.13419656, 0.13995328, 0.146118  , 0.1527278 , 0.15981977,
       0.16743101, 0.1755986 , 0.18435964, 0.19375121, 0.20381041,
       0.21457432, 0.22608004, 0.23836466, 0.25146526, 0.26541894,
       0.28026278, 0.29603389, 0.31276934, 0.33050623, 0.34928165,
       0.36913268, 0.39009643, 0.41220997, 0.4355104 , 0.46003482,
       0.4858203 , 0.51290395, 0.54132284, 0.57111408, 0.60231474,
       0.63496193, 0.66909273, 0.70474424, 0.74195353, 0.78075771,
       0.82119386, 0.86329907, 0.90711044, 0.95266505, 1.        ])
```

Update its inputs and recompute, then only the necessary things are recomputed again.
```python
flow.update_inputs(normalize=False)
flow.get()
```

```
NOT NORMALIZING
array([0.5       , 0.52044216, 0.54108832, 0.56214247, 0.58380862,
       0.60629075, 0.62979286, 0.65451895, 0.68067302, 0.70845906,
       0.73808107, 0.76974305, 0.80364899, 0.84000289, 0.87900875,
       0.92087056, 0.96579231, 1.01397802, 1.06563167, 1.12095725,
       1.18015878, 1.24344023, 1.31100562, 1.38305893, 1.45980416,
       1.54144532, 1.62818638, 1.72023137, 1.81778426, 1.92104905,
       2.03022975, 2.14553035, 2.26715484, 2.39530723, 2.5301915 ,
       2.67201166, 2.8209717 , 2.97727562, 3.14112742, 3.31273109,
       3.49229063, 3.68001003, 3.87609329, 4.08074442, 4.2941674 ,
       4.51656623, 4.74814491, 4.98910743, 5.2396578 , 5.5       ])
```

### DIFF
--- a/docs/build_docs.sh
+++ b/docs/build_docs.sh
@@ -6,6 +6,9 @@ which sphinx-build
 
 # Ensure single-core
 export SISL_NUM_PROCS=1
+# Inform to the workflow visualization function that the 
+# notebooks are to be exported to html
+export SISL_NODES_EXPORT_VIS=1
 
 # First make clean
 make clean
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -107,6 +107,7 @@ of places to search/ask for answers:
    installation
    tutorials.rst
    scripts/scripts
+   nodes/nodes_intro
 
 .. toctree::
    :maxdepth: 2
--- a/docs/nodes/nodes_intro.ipynb
+++ b/docs/nodes/nodes_intro.ipynb
@@ -1,0 +1,694 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "ab557674",
+   "metadata": {},
+   "source": [
+    "Nodes and workflows\n",
+    "--------\n",
+    "\n",
+    "Nodes and workflows are a functionality within `sisl` that allows you to define functional workflows that are lazily computed in a very easy way. Some of the reasons that you might want to use the `sisl.nodes` framework are:\n",
+    "\n",
+    "- **Cleaner and maintainable code**: The framework forces to write your workflows in a functional style. That is, **each piece of functionality** must be packed into **a node, which must be a pure function**, i.e. a function whose result depends only on its inputs. If you manage to do this, your code will be much more reusable and reproduceable by other people. It also helps a lot in testing your code, since you can very easily test each piece individually.\n",
+    "- **Easier to use by an external interface**: Graphical interfaces (GUI) can have a hard time interacting with code if there is no clear division of functionalities or the inputs to provide are very complex. Nodes are pieces of functionality with simple input fields. Whenever an input to a node is too complex, you can always generate it from another node, simplifying the input that the user needs to provide. Linking a node's output to another node's input is therefore a very efficient way of creating \"workchains\" from a GUI by providing only simple inputs.\n",
+    "\n",
+    "*Note*: The plan is to convert [`sisl-gui`](https://pypi.org/project/sisl-gui/) to use these nodes and workflows, so whatever that you develope within the framework will be usable automatically by the GUI.\n",
+    "\n",
+    "### Nodes\n",
+    "\n",
+    "In `sisl.nodes`, you have the `Node` class:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bc48e2c4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sisl\n",
+    "from sisl.nodes import Node"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1883c9ef",
+   "metadata": {},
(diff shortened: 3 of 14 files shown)


zerothi:
Great!

But, why are all the input_fields stuff here? No usage of nodes etc. I think we should leave them out of this PR...
Also, an input field requires interaction, so per see not applicable in the `nodes` sub-module, if I am not mistaken?

pfebrer:
Yes, I didn't know what to do with the input fields. They were a bit integrated into the `Node` class so I decided to keep the basic ones (probably an act of lazyness). But I agree that I should remove them for now, and they could be determined afterwards if needed. I will try to remove them tomorrow.

By the way for now I went with the approach of checkpointing everything that is done in the workflow, and variable assignments just create aliases for easier access to nodes. I think this is the less complex approach. Perhaps we could keep it like this in beta, ask people to play with it and then have a more definitive API before the next release (?).

For example one thing that I thought could be useful is to store some intermediate value without necessarily checkpointing it (i.e. not storing the inputs and therefore still recomputing each time). But I'm not sure about the implications and if it is useful.

zerothi:
> Yes, I didn't know what to do with the input fields. They were a bit integrated into the `Node` class so I decided to keep the basic ones (probably an act of lazyness). But I agree that I should remove them for now, and they could be determined afterwards if needed. I will try to remove them tomorrow.

Ok.
 
> By the way for now I went with the approach of checkpointing everything that is done in the workflow, and variable assignments just create aliases for easier access to nodes. I think this is the less complex approach. Perhaps we could keep it like this in beta, ask people to play with it and then have a more definitive API before the next release (?).

Yeah, I would still be more keen on the *at creation specify what is exposed* which should be more stable...

> For example one thing that I thought could be useful is to store some intermediate value without necessarily checkpointing it (i.e. not storing the inputs and therefore still recomputing each time). But I'm not sure about the implications and if it is useful.

Hmm... Maybe...

pfebrer:
> Yeah, I would still be more keen on the at creation specify what is exposed which should be more stable...

Yes, but that is an extra layer on top I would say. For now these aliases just allow you to get the nodes by variable name from the `nodes` attribute, like:

```python
# Retreive the node using the name I used for the variable.
my_flow.nodes.some_alias
```

Then one can add attributes to the workflow that map directly to the output of the node.
```
# Access the current output of the node
my_flow.some_alias
```

Is that what you mean by exposed, or you have in mind that "exposed = checkpointed"?

zerothi:
Hmm, I'll have to check- what I mean ;)

pfebrer:
I removed the input fields :+1: 

Doing that I realized that `inspect.Signature` is more useful than I thought, because it can parse *args and **kwargs into a dictionary of arguments.

zerothi:
Great! thanks!

PR #540, opened 2023-02-11
bug: fixed swapaxes handling

### LINKED ISSUE #539: What should `Geometry.swapaxes` do if it includes `xyz` in `what`?
From the documentation I would expect that calling `Geometry.swapaxes` with `xyz` in the `what` argument would fully swap two Cartesian axes. So essentially you are not changing your geometry but just rotating it space by rotating the Cartesian axes. I.e. if you have a slab with vacuum in the z-direction you could transform it to a slab with vacuum in the x-direction by swapping the Cartesian axes. However, the way it is implemented right now the call only swaps the Cartesian components of the atom positions but leaves the Cartesian components of the cell vectors intact which completely changes your geometry. Is this the intended behaviour or is it a bug?

zerothi:
Now the swapaxes is correct for Cartesian coordinate swaps.

Also extended it to be much more versatile.
One can do collapsed swappings, which might not be that useful. But rather easy to follow given that I wanted the str arguments.

Thanks to @ahkole for discovering the bug leading to this.

Signed-off-by: Nick Papior <nickpapior@gmail.com>

 - [x] Closes #539 
 - [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,13 @@ we hit release version 1.0.0.
   ince the scattering matrix is an S-matrix usage.
   Also changed `se2scat` to `se2broadening` #529
 - allow `BrillouinZone` initialization with scalar weights for all k-points #537
+- Geometry.swapaxes and SuperCell.swapaxes, these are now more versatile by
+	allowing multiple swaps in a single run, #539
+
 
 ### Fixed
 - Documentation now uses global references
+- Geometry.swapaxes would not swap latticevector cartesian coordinates, #539
 
 
 ### toolbox.btd
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -69,6 +69,7 @@ OrbitalsArg = Union[
     ArrayLike,
     str, int, dict,
 ]
+TGeometry = "Geometry"
 
 
 @set_module("sisl")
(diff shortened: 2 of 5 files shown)


zerothi:
@ahkole I don't know if the message pinged you or not, but here it is again :)

zerothi:
@ahkole I will merge this as I think it is the correct thing, thanks for bringing this to my attention!

ahkole:
@zerothi Hi, sorry. I recently switched phones and haven't setup everything on my new phone yet (including the mail account where I receive GitHub notifications) so I completely missed this, sorry.

I had a look at the new implementation and as far as I can see all the swaps now work as I would expect. Maybe the only thing that could still be improved is the docstrings for the `swapaxes` method. First of all, inside the `Geometry.swapaxes` docstring there is this example,

    g_s = g.swapaxes("abx", "bcz", 1, "xyz")

But this call should not be possible because `swapaxes` can get at most 3 arguments. Furthermore, what wasn't immediately clear to me from the docstring is that the `what` parameter is only used if you use integer arguments for `axis_a` and `axis_b` and also that using a string of labels instead of a single label for `axis_a` and `axis_b` means that multiple swaps will be executed in serial one after the other.

zerothi:
@ahkole thanks for taking a look! The recent commit should be more verbose! Thanks for finding the doc bug ;)

PR #542, opened 2023-02-27
mnt: changed only= keyword arguments to what=

### LINKED ISSUE #541: `only` or `what`?
#539 made me notice that the keyword in `Geometry.swapaxes` is `what`, while in [`Geometry.rotate`](https://zerothi.github.io/sisl/api/generated/sisl.Geometry.html#sisl.Geometry.rotate) it is `only`. Would it make sense to unify? I like `what` better.

zerothi:
This is to unify interfaces that has these kinds of arguments. It affects Geometry|SuperCell.rotate and some tbtrans siles.

To accommodate backwards compatibility for some time I have added a function to decorate methods enabling a keyword argument and replacing the *new* keyword argument with the old keyword argument. I believe users who use the old argument are not aware of the new one and hence it makes more sense.
A deprecation warning is also issued.

The default for Geometry.rotate is now also changed depending on whether atoms argument is supplied.
- atoms is None: what="xyz+abc"
- else what="xyz" when doing sub-rotations it makes more sense to only rotate the atomic coordinates, while rotations of everything makes more sense to also rotate the lattice vectors.

 - [x] Closes #541
 - [x] Tests added
@tfrederiksen could you please have a look whether this is ok? :)

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -2299,13 +2299,12 @@ class Geometry(SuperCellChild):
              atoms will be rotated.
         rad : bool, optional
              if ``True`` the angle is provided in radians (rather than degrees)
-        what : {'xyz', 'abc', 'abc+xyz'}
+        what : {'xyz', 'abc', 'abc+xyz', <or combinations of "xyzabc">}
             which coordinate subject should be rotated,
-            if ``abc`` is in this string the cell will be rotated
-            if ``xyz`` is in this string the coordinates will be rotated
-            Each of the coordinates may be individually rotated.
+            if any of ``abc`` is in this string the corresponding cell vector will be rotated
+            if any of ``xyz`` is in this string the corresponding coordinates will be rotated
             If `atoms` is None, this defaults to "abc+xyz", otherwise it defaults
-            to "xyz".
+            to "xyz". See Examples.
 
         Examples
         --------
@@ -2330,6 +2329,7 @@ class Geometry(SuperCellChild):
         if atoms is None:
             if what is None:
                 what = "abc+xyz"
+            # important to not add a new dimension to xyz
             atoms = slice(None)
         else:
             if what is None:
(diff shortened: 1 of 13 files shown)


sisl/geometry.py
```diff
@@ -2280,61 +2287,87 @@ def rotate(self, angle, v, origin=None, atoms: Optional[AtomsArg]=None, only='ab
...
         atoms : int or array_like, optional
              only rotate the given atomic indices, if not specified, all
              atoms will be rotated.
-        only : {'abc+xyz', 'xyz', 'abc'}
-             which coordinate subject should be rotated,
-             if ``abc`` is in this string the cell will be rotated
-             if ``xyz`` is in this string the coordinates will be rotated
         rad : bool, optional
              if ``True`` the angle is provided in radians (rather than degrees)
+        what : {'xyz', 'abc', 'abc+xyz'}
```
tfrederiksen, line 2302: Is it clear from this line that one can also specify `what="x"` as in your example below? Or combinations  like `what="xy+b"`?

sisl/geometry.py
```diff
@@ -2264,13 +2265,19 @@ def angle(self, atoms, dir=(1., 0, 0), ref=None, rad=False):
             return ang
         return np.degrees(ang)
 
-    def rotate(self, angle, v, origin=None, atoms: Optional[AtomsArg]=None, only='abc+xyz', rad=False) -> Geometry:
-        """ Rotate geometry around vector and return a new geometry
+    @deprecate_argument("only", "what",
```
tfrederiksen, line 2268: With this branch installed I didn't get the deprecation message when calling with `only`. Shouldn't I?

tfrederiksen (commented):
Looks good, thanks!

sisl/geometry.py
zerothi: thanks, I have amended, could you have a look?

sisl/geometry.py
zerothi: Thanks! Fixed!

sisl/geometry.py
tfrederiksen: Looks fine

zerothi:
Thanks @tfrederiksen

PR #544, opened 2023-03-02
aniSileSiesta

tfrederiksen:
This PR enables reading of siesta ANI files.

### DIFF
--- a/sisl/io/siesta/ani.py
+++ b/sisl/io/siesta/ani.py
@@ -1,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+"""
+Sile object for reading ANI files
+"""
+from ..xyz import xyzSile
+from ..sile import add_sile
+from sisl._internal import set_module
+
+__all__ = ["aniSileSiesta"]
+
+@set_module("sisl.io.siesta")
+class aniSileSiesta(xyzSile):
+
+    pass
+
+add_sile('ani', aniSileSiesta, case=False, gzip=True)
--- a/sisl/io/siesta/ani.py
+++ b/sisl/io/siesta/ani.py
@@ -8,11 +8,15 @@ from ..xyz import xyzSile
 from ..sile import add_sile
 from sisl._internal import set_module
 
+
 __all__ = ["aniSileSiesta"]
+
 
 @set_module("sisl.io.siesta")
 class aniSileSiesta(xyzSile):
 
-    pass
+    def read_geometry(*args, all=True, **kwargs):
+        return super().read_geometry(*args, all=all, **kwargs)
 
-add_sile('ani', aniSileSiesta, case=False, gzip=True)
+
+add_sile('ANI', aniSileSiesta, gzip=True)
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -548,7 +548,8 @@ def sile_fh_open(from_closed=False):
 
 
 def sile_read_multiple(start=0, stop=1000000000, step=None, all=False,
-                       skip_call=None, pre_call=None, post_call=None):
+                       skip_call=None, pre_call=None, post_call=None,
+                       postprocess=None):
     """ Method decorator for doing multiple reads
 
     Parameters
@@ -562,68 +563,90 @@ def sile_read_multiple(start=0, stop=1000000000, step=None, all=False,
     all : bool, optional
        read all entries (takes precedence over start/step)
     skip_call : function, optional
-       read method without actual data processing
+       read method without actual data processing.
+       The skip call must something different from ``None`` if it succeeds.
     pre_call : function, optional
-       function to be applied before each read call
+       function to be applied before each read call.
+       Does *not* apply to the `skip_call`.
     post_call : function, optional
        function to be applied after each read call
+       Does *not* apply to the `skip_call`.
+    postprocess : function, optional
+       function to be applied on the returned data before it gets returned.
 
     Returns
     -------
     single entry or list from multiple reads
     """
     def decorator(func):
+        # ensure we can access the callables
+        nonlocal skip_call, pre_call, post_call, postprocess
+
+        if pre_call is None:
+            def pre_call(): pass
+
+        if post_call is None:
+            def post_call(): pass
+
+        if postprocess is None:
+            def postprocess(x): return x
+
+        def wrap_func(*args, **kwargs):
+            pre_call()
+            r = func(*args, **kwargs)
+            post_call()
+            return r
+
+        # if skip_call is not set, we'll simply use func.
+        # Only in this case will pre and post be called
+        # as well.
+        # they should be equivalent but skip can be used to
+        # skip some processing of the read data.
+        if skip_call is None:
+            skip_call = wrap_func
+
         @wraps(func)
-        def multiple(self, *args, start=start, stop=stop, step=step, all=all,
-                     skip_call=skip_call, pre_call=pre_call, post_call=post_call, **kwargs):
-
-            if skip_call is None:
-                skip_call = func
-            if pre_call is None:
-                def pre_call(): pass
-            if post_call is None:
-                def post_call(): pass
-
-            def wrap_func(self, *args, **kwargs):
-                pre_call()
-                r = func(self, *args, **kwargs)
-                post_call()
-                return r
-
-            def wrap_skip_call(self, *args, **kwargs):
-                pre_call()
-                r = skip_call(self, *args, **kwargs)
-                post_call()
-                return r
+        def multiple(*args, start=start, stop=stop, step=step, all=all, **kwargs):
+            self = args[0]
 
             if all:
-                start, stop, step = 0, 1000000000, 1
+                start, stop, step = 0, 100000000000, 1
+            if stop < 0:
+                # TODO this will probably fail if users do stop=-4 (to not have the last 3 items)
+                # It should be done in a different manner
+                stop = 100000000000
+
+            # start by reading past start
+            for _ in range(start):
+                r = skip_call(*args, **kwargs)
 
             if step is None:
-                for _ in range(start):
-                    wrap_skip_call(self, *args, **kwargs)
-                if hasattr(self, "fh"):
-                    return wrap_func(self, *args, **kwargs)
-                with self:
-                    return wrap_func(self, *args, **kwargs)
+                # read a single element and return
+                def ret_func():
+                    return wrap_func(*args, **kwargs)
 
-            def loop(self):
-                rng = range(start, stop, step)
-                R, r, ir = [], True, 0
-                while r is not None:
-                    if ir in rng:
-                        r = wrap_func(self, *args, **kwargs)
-                        if r is not None:
-                            R.append(r)
-                    else:
-                        r = wrap_skip_call(self, *args, **kwargs)
-                    ir += 1
-                return R
+            else:
+                # read a set of geometries and return
+                def ret_func():
+                    R = []
+                    for ir in range(start, stop, step):
+                        r = wrap_func(*args, **kwargs)
+                        if r is None:
+                            return R
+                        R.append(r)
+
+                        # skip the middle steps
+                        for _ in range(step-1):
+                            r = skip_call(*args, **kwargs)
+                            if r is None:
+                                return R
+                    return R
 
             if hasattr(self, "fh"):
-                return loop(self)
-            with self:
-                return loop(self)
+                return postprocess(ret_func())
+            else:
+                with self:
+                    return postprocess(ret_func())
 
         return multiple
     return decorator
--- a/sisl/io/sile.py
+++ b/sisl/io/sile.py
@@ -36,6 +36,7 @@ __all__ += [
 # Decorators or sile-specific functions
 __all__ += [
     'sile_fh_open',
+    'sile_read_multiple',
     'sile_raise_write',
     'sile_raise_read'
 ]
@@ -544,6 +545,88 @@ def sile_fh_open(from_closed=False):
                     return func(self, *args, **kwargs)
             return pre_open
     return _wrapper
+
+
+def sile_read_multiple(start=0, stop=1000000000, step=None, all=False,
+                       skip_call=None, pre_call=None, post_call=None):
+    """ Method decorator for doing multiple reads
+
+    Parameters
+    ----------
+    start : int, optional
+       position of first read
+    stop : int, optional
+       position of last read
+    step : int, optional
+       number of steps to take (if any at all)
+    all : bool, optional
+       read all entries (takes precedence over start/step)
+    skip_call : function, optional
+       read method without actual data processing
+    pre_call : function, optional
+       function to be applied before each read call
+    post_call : function, optional
+       function to be applied after each read call
+
+    Returns
+    -------
+    single entry or list from multiple reads
+    """
+    def decorator(func):
+        @wraps(func)
+        def multiple(self, *args, start=start, stop=stop, step=step, all=all,
+                     skip_call=skip_call, pre_call=pre_call, post_call=post_call, **kwargs):
+
+            if skip_call is None:
+                skip_call = func
+            if pre_call is None:
+                def pre_call(): pass
+            if post_call is None:
+                def post_call(): pass
+
+            def wrap_func(self, *args, **kwargs):
+                pre_call()
+                r = func(self, *args, **kwargs)
+                post_call()
+                return r
+
+            def wrap_skip_call(self, *args, **kwargs):
+                pre_call()
+                r = skip_call(self, *args, **kwargs)
+                post_call()
+                return r
+
+            if all:
+                start, stop, step = 0, 1000000000, 1
+
+            if step is None:
+                for _ in range(start):
+                    wrap_skip_call(self, *args, **kwargs)
+                if hasattr(self, "fh"):
+                    return wrap_func(self, *args, **kwargs)
+                with self:
+                    return wrap_func(self, *args, **kwargs)
+
+            def loop(self):
+                rng = range(start, stop, step)
+                R, r, ir = [], True, 0
+                while r is not None:
+                    if ir in rng:
+                        r = wrap_func(self, *args, **kwargs)
+                        if r is not None:
+                            R.append(r)
+                    else:
+                        r = wrap_skip_call(self, *args, **kwargs)
+                    ir += 1
+                return R
+
+            if hasattr(self, "fh"):
+                return loop(self)
+            with self:
+                return loop(self)
+
+        return multiple
+    return decorator
 
 
 @set_module("sisl.io")
--- a/sisl/io/xyz.py
+++ b/sisl/io/xyz.py
@@ -98,8 +98,11 @@ class xyzSile(Sile):
         sc : SuperCell, optional
             the supercell to be associated with the geometry
         """
+        line = self.readline()
+        if line == '':
+            return None
         # Read number of atoms
-        na = int(self.readline())
+        na = int(line)
 
         # Read header, and try and convert to dictionary
         header = self.readline()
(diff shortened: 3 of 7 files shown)


sisl/io/siesta/ani.py
```diff
@@ -0,0 +1,61 @@
...
+from sisl import Geometry, SuperCell
+
+__all__ = ["aniSileSiesta"]
+
+@set_module("sisl.io.siesta")
+class aniSileSiesta(SileSiesta):
+    """ ANI file object """
+
+    @sile_fh_open(True)
+    def read_geometry(self, step=1):
```
zerothi, line 21: Could we perhaps have a `start=0, step=1` argument to allow one to start arbitrarily?
Could some value (`none` or negative) be specified to only retrieve 1 of the geometries?

`read_geometry(start=200, step=None) == Geometry` (not a list)

sisl/io/siesta/ani.py
```diff
@@ -0,0 +1,61 @@
...
+
+from .sile import SileSiesta
+from ..sile import add_sile, sile_fh_open
+from sisl._internal import set_module
+from sisl import Geometry, SuperCell
+
+__all__ = ["aniSileSiesta"]
+
+@set_module("sisl.io.siesta")
+class aniSileSiesta(SileSiesta):
```
zerothi, line 17: One could inherit from the `xyzSile` object since they are the same format. In this way you could use `super().read_geometry()`

sisl/io/siesta/ani.py
```diff
@@ -0,0 +1,61 @@
...
+
+        def readG(itt):
+            atoms = []
+            xyz = []
+            # Read the Geometry block
+            try:
+                na = int(next(itt))
+            except:
+                return None
+            next(itt) # blank line
```
zerothi, line 42: The ANI file format sometimes allows the blank line to contain a single number, I know that molden can use this to plot, say the total energy. Would it make sense to return this upon request?

zerothi (changes requested):
Great addition!

It is however a generic file format and should perhaps be contained in the `xyzSile` then the `aniSileSiesta` could simply be a subclass of the xyz sile.

sisl/io/siesta/ani.py
zerothi: but we should consider adding this functionality to the `xyzSile` instead.

tfrederiksen: > but we should consider adding this functionality to the `xyzSile` instead.

by "this functionality" you mean the possibility of returning the header?

sisl/io/siesta/ani.py
tfrederiksen: Yes, good point.

sisl/io/siesta/ani.py
zerothi: basically everything, an `xyz` file may be any number of entries.

zerothi: see my top comment: https://github.com/zerothi/sisl/pull/544#pullrequestreview-1321713959

tfrederiksen:
OK, do you then see the `.ANI` format as basically a `xyzSile` (but one with several entries)?

zerothi:
Yes, that (I think) is the actual `xyz` file format. So for siesta we would simply do `add_sile("ani", xyzSile, ...)` and then don't do anything else.

zerothi:
> Yes, that (I think) is the actual `xyz` file format. So for siesta we would simply do `add_sile("ani", xyzSile, ...)` and then don't do anything else.

Well, perhaps we should do:
```python
aniSileSiesta(xyzSile): pass

add_sile("ani", aniSileSiesta, ...)
```

tfrederiksen:
What about handling the sile as something that can be iterated:
```python 
s = sisl.get_sile(anifile)
g0 = s.read_geometry()
g1 = s.read_geometry()
```
Right now the `xyzSile` forces a start from the beginning, but this could be changed.

zerothi:
> What about handling the sile as something that can be iterated:
> 
> ```python
> s = sisl.get_sile(anifile)
> g0 = s.read_geometry()
> g1 = s.read_geometry()
> ```
> 
> Right now the `xyzSile` forces a start from the beginning, but this could be changed.

This should work already in the `xyzSile`
```python
s = sisl.get_sile(anifile)
with s:
    g0 = s.read_geometry()
    g1 = s.read_geometry()
```

tfrederiksen:
Am I on the right track with wrapping the single-time `read_geometry` function for multiple read attempts as done here?

zerothi:
> Am I on the right track with wrapping the single-time `read_geometry` function for multiple read attempts as done here?

Yes, I think something like this.

My idea would be to unify MD output files with a decorator, so essentially something like this:
```python
@sile_fh_open
@read_multiple(start=0, step=None)
def read_geometry(self, atoms=None, sc=None)
...
```
where the decorator ensures correct calls and adds documentation etc.
```python
def read_multiple(start=0, step=None):
    "decorate a function for doing multiple reads "
    def decorator(func):
        @wraps
        def f(*args, *, start=start, stop=stop, **kwargs):
                 if stop is None:
 return func(*args, **kwargs)
...
```
something like the above could work and make interfaces the same, what do you think?

sisl/io/siesta/ani.py
tfrederiksen: I think this is now enabled.

tfrederiksen:
What do you think of the current layout?

tfrederiksen:
How should `write` be handled for lists of geometries? Would it make sense that a multiple-read call returns an instance from a new class like `GeometryCollection` (instead of a simple list)?

sisl/io/sile.py
```diff
@@ -546,6 +547,50 @@ def pre_open(self, *args, **kwargs):
     return _wrapper
 
 
+def sile_read_multiple(all=False, start=0, step=None):
```
zerothi, line 550: instead of `all` perhaps `start, stop, step`. This goes in line with slice arguments.

Come to think of it, would it make sense to have a single argument named `range` or `rng` this could then also accept slices? Hmm... Might just be annoying when you only need a single element.

Which would you prefer:
```python
# end is not inclusive
read_geometry(range=5) # same as start=0, step=1, stop=5
read_geometry(range=(2, 5)) # same as start=2, step=1, stop=5
read_geometry(range=(2, 5, 2)) # same as start=2, step=2, stop=5
read_geometry(range=slice(2, None)) # same as start=2, stop=-1

# or the more explicit
read_geometry(stop=5)
read_geometry(start=2, stop=6, step=2)
```

sisl/io/siesta/ani.py
```diff
@@ -0,0 +1,21 @@
...
+from ..xyz import xyzSile
+from ..sile import add_sile
+from sisl._internal import set_module
+
+__all__ = ["aniSileSiesta"]
+
+@set_module("sisl.io.siesta")
+class aniSileSiesta(xyzSile):
+
+    def read_geometry(self, *args, **kwargs):
```
zerothi, line 16: I would just change default argument in the argument list here, no need to check in the code.

sisl/io/sile.py
```diff
@@ -546,6 +547,50 @@ def pre_open(self, *args, **kwargs):
...
+            R, r = [], True
+            if hasattr(self, "fh"):
+                if step is None:
+                    for s in range(start):
+                        func(self, *args, **kwargs)
+                    return func(self, *args, **kwargs)
+                while r is not None:
+                    R.append(r)
+                    r = func(self, *args, **kwargs)
+                return R[start+1::step]
```
zerothi, line 580: shouldn't this be `R[start::step]`? otherwise you'll skip `start` element?

sisl/io/sile.py
```diff
@@ -546,6 +547,50 @@ def pre_open(self, *args, **kwargs):
...
+            with self:
+                if step is None:
+                    for s in range(start):
+                        func(self, *args, **kwargs)
+                    return func(self, *args, **kwargs)
+                while r is not None:
+                    R.append(r)
+                    r = func(self, *args, **kwargs)
+                return R[start+1::step]
```
zerothi, line 589: same here?

sisl/io/xyz.py
```diff
@@ -98,8 +99,11 @@ def read_geometry(self, atoms=None, sc=None):
         sc : SuperCell, optional
             the supercell to be associated with the geometry
         """
+        line = self.readline()
+        if line == '':
+            return None
         # Read number of atoms
-        na = int(self.readline())
+        na = int(line)
```
zerothi, line 106: wow, I am surprised this wasn't caught in the tests... this looks like a severe bug...

sisl/io/sile.py
tfrederiksen: It's because the first entry is `True` because I append to `R` before reading the first `r`. Maybe a sloppy way to do it.

sisl/io/sile.py
tfrederiksen: I see your point, but I thought `all` is more in line with other sile functions, like [`outSileSiesta.read_geometry`](https://zerothi.github.io/sisl/api/io/generated/sisl.io.siesta.outSileSiesta.html#sisl.io.siesta.outSileSiesta.read_geometry).

sisl/io/sile.py
zerothi: ah, I see... That didn't catch my eye at first

sisl/io/sile.py
zerothi: Thanks! Yes, but come to think of it, I think the other way would be more pythonic, then we could change the other ones to follow suit on this convention. What do you think?

sisl/io/xyz.py
tfrederiksen: There is also another issue. What if some blank lines are present before or after a xyz-geometry block? Should the code not be able to handle such situations?

zerothi: Preferentially yes. This could easily be solved with
```python
def sile_read_multiple(..., pre_call=None, post_call=None):
    ...
   pre_call()
   r = func(...)
   post_call()
```
or similar. I think this could cover all use cases.

sisl/io/sile.py
tfrederiksen: One could also just stick with `all` (simple and intuitive keyword) and let the user process the returned list if they want only a subset. The whole file will be processed anyway, so I am not sure there is much of an advantage of leaving out data.

sisl/io/xyz.py
tfrederiksen: Then the user would need to specify `pre_call=self.readline()` if there is a blank line before the atom number?

Could `read_geometry()` not simply read blank lines until it finds an integer?

sisl/io/sile.py
zerothi: In some cases reading very large files could be very slow if one is only interested in reading every 5th value. In this case having a `skip` argument to the decorator could faster skip the geometry reading. For instance converting coordinates and atoms into the `Atoms` class + `Geometry` class turns out to be quite heavy when reading, say 1000, geometries.

What about we do this:
```python
def sile_read_multiple(start=0, end=-1, step=None, all=False):
    if all:
        start, end, step = 0, -1, 1
```
this would allow fast argument, and enable the other one as well?

zerothi: Oh, and when not reading everything, it would be wise to discard them on the fly due to the memory overhead, which again isn't big for small MD collections, but for big it could be.

sisl/io/xyz.py
zerothi: > Then the user would need to specify `pre_call=self.readline()` if there is a blank line before the atom number?
> 
> Could `read_geometry()` not simply read blank lines until it finds an integer?

but for more complex files which contains `xyz` you wouldn't know if that integer starts the geometry, or some other segment. It is a rabbit hole ;)

I think the above is the most stable one, after all it is not something the user sees, it is the file implementor who knows how its handled.

sisl/io/sile.py
tfrederiksen: OK, good points. Is there a fast way to step forward a number of lines in a file without invoking `readline()` in a `for` loop?

zerothi: No, not really.

The simplest would be to do `for _ in range(N): self.fh.readline()`, that bypasses all parsers etc., i.e. you don't want to call `self.readline()` as that also does some comment parsing etc.

tfrederiksen: One problem is that the number of atoms in each geometry in a series may not need to be the same. To find out one needs the first line.

Would it be a good idea to implement a keyword in `xyzSile.read_geometry(skip=True)` which steps over the geometry data and return, say, only the number of atoms and the comment line?

zerothi: So you mean the skip function needs to adapt, but that skip function will still read `na` and then parse the correct number of lines. Perhaps I am missing something, but I don't see a problem since `read_geometry` is self-contained and will only ever read the number of lines matching the number of atoms + 2. The `skip` should be a callable that simple reads the file, for an xyz sile it could be:
```python
def skip(self):
    na = int(self.readline())
    self.readline()
    for _ in range(na): self.readline()
```

zerothi: > Would it be a good idea to implement a keyword in `xyzSile.read_geometry(skip=True)` which steps over the geometry data and return, say, only the number of atoms and the comment line?

If anything this should be a separate method, but I don't like it ;) I don't think it is necessary?

zerothi: Do you have an example?

tfrederiksen: I don't see how `sile_read_multiple` can know how many lines to skip without actually calling the implemented read function. One only knows this after entering the data block.

zerothi: Ok, perhaps I was not clear. The function should be something like this:
```python
def sile_read_multiple(all=False, start=0, step=None, skip_call=None, pre_call=None, post_call=None):
    """ Method decorator for doing multiple reads
    Parameters
    ----------
    all : bool, optional
       read all entries (takes precedence over start/step)
    start : int, optional
       position of first read
    step : int, optional
       number of steps to take (if any at all)
   *_call : callable, optional
        callables called before/after read and for fast skipping an element

    Returns
    -------
    single entry or list from multiple reads
    """
    def decorator(func):
         if skip_call is None:
              skip_call = func # perhaps this should be partial'ized with arguments... hmm
        @wraps(func)
        def multiple(*args, all=all, start=start, step=step, end=end, **kwargs):
            if all is True:
                start, end, step = 0, 1e100, 1
            rng = range(start, end, step)
            R, r = [], True
            def read_next(ir):
                   if ir in rng:
                       pre_call()
                       res = func(*args, **kwargs)
                       post_call()
                       return res
                   pre_call() ; res = skip_call(); post_call()
                   return res
                   
            if hasattr(self, "fh"):
                ir = 0
                r = read_next(ir)
                while r is not None:
                    R.append(r)
                    ir += 1
                    r = read_next(ir)
                
                return R[rng]

       ...

        return multiple
    return decorator

Comment (tfrederiksen) [reply sisl/io/sile.py]: I adopted `stop` instead of `end` as this seems to be the custom in `slice`. Otherwise tried to follow your layout.

Comment (zerothi) [reply sisl/io/sile.py]: agreed, thanks for fixing my mistake!

File: sisl/io/siesta/ani.py (hunk 5 above)
Comment (tfrederiksen) [reply sisl/io/siesta/ani.py:16]: I made another suggestion. Is this what you had in mind?

**Comment by tfrederiksen** [comment]
> How should `write` be handled for lists of geometries? Would it make sense that a multiple-read call returns an instance from a new class like `GeometryCollection` (instead of a simple list)?

Any thoughts on this?

File: sisl/io/siesta/ani.py (hunk 5 above)
Comment (zerothi) [reply sisl/io/siesta/ani.py:16]: It could be simpler, simply do:
```python
read_geometry(self, *args, *, all=True, **kwargs)
```

sisl/io/sile.py
```diff
@@ -546,6 +547,89 @@ def pre_open(self, *args, **kwargs):
...
+    post_call : function, optional
+       function to be applied after each read call
+
+    Returns
+    -------
+    single entry or list from multiple reads
+    """
+    def decorator(func):
+        @wraps(func)
+        def multiple(self, *args, start=start, stop=stop, step=step, all=all,
```
zerothi, line 577: as per your animation file, we should probably force the extra arguments to be forced keys:
```python
def multiple(self, *args, *, start=start, stop=stop, step=step, all=all)
```

Secondly, the `*_call` methods should not be added to the method it self as far as I can see, they could just be used without allowing users to change the behaviour. I don't know if they could be useful for users. Hmm... Perhaps they could... What do you think?

zerothi (changes requested):
I would also like to remove the unnecessary commits, could you rebase and squash the commits that shouldn't be used? Then I can have a final go and get this in?

sisl/io/siesta/ani.py
tfrederiksen: Sorry, I tried different variations of this but it failed (I'm afraid I miss something with the syntax).

sisl/io/sile.py
tfrederiksen: I'm probably missing something with the syntax so I'm not sure how you suggest to force the keys here.

Also, regarding the `*_call` methods I get `UnboundLocalError: local variable '*_call' referenced before assignment` if I move it outside `multiple`.

pfebrer:
> How should write be handled for lists of geometries? Would it make sense that a multiple-read call returns an instance from a new class like GeometryCollection (instead of a simple list)?

As I have said already on several other issues, I think sisl should have a well defined way of working with trajectories :) Otherwise, just reading trajectories from files as lists of `Geometry` objects and afterwards concatenating their coordinates in a numpy array is far from optimal. It would also be cool to add MD analysis tools into sisl, but since there's no defined way to treat trajectories, there is no space right now in the package for them, which is sad :(

sisl/io/siesta/ani.py
zerothi: Yeah, my suggestion was wrong, I'll make a PR against your branch, then you can see.

pfebrer:
Hi! I'm checking the `sile_read_multiple` function and the `all` argument seems odd to me. `start`, `stop`, `step` are already enough, no?

Why not go with the conventions of slicing in python? I.e. behaving exactly as `slice`.

Additionally, I know it looks a bit ugly at first sight but `numpy.mgrid` does something like:

```python
gc = xyzSile(f).read_geometry[4::10]
```

Just something to think about, not that I personally think it's a good idea :)

zerothi:
> Hi! I'm checking the `sile_read_multiple` function and the `all` argument seems odd to me. `start`, `stop`, `step` are already enough, no?

See discussion here https://github.com/zerothi/sisl/pull/544#discussion_r1126045934

sisl/io/sile.py
pfebrer: How about leaving the function arguments untouched but allowing slicing to create the loop?

```python
# The original function call
geom = xyzSile(f).read_geometry(atoms=...)

# How it works now if you want to read multiple geometries
geoms = xyzSile(f).read_geometry(atoms=..., start=0, stop=30)

# How I propose to separate the loop from the function call.
# Prepare the loop before calling the function
geoms = xyzSile(f).read_geometry[:30](atoms=...)

# "All" would simply be
geoms = xyzSile(f).read_geometry[:](atoms=...)
```

This is similar to dispatchers, where you would do something like `xyzSile(f).apply(0, 30).read_geometry(atoms=...)`.

pfebrer: I know preparing the loop with slicing is a bit restrictive (it doesn't allow a `pre_call` for example), but you could go for the `.apply(0, 30, pre_call=...)` version of it if you want to further customize it (?)

zerothi: that seems like a good addition, so I imagine `read_geometry[:30](end=15)` would use `end=15` rather than `30` so it _can_ be overwritten in argument list, no? Once https://github.com/tfrederiksen/sisl/pull/2 is in (or a variant) then it would be great with a PR! :)

zerothi: no the idea is that the `pre_call` is only known at decoration time, a user cannot do a `pre_call` overwrite. So that would be fine.

zerothi: I think we shouldn't complicate things unnecessarily, the `[...]` notation would basically just set the defaults in the call.

zerothi (approved):

PR #546, opened 2023-03-07
mnt: added geometry collection with some more features

### LINKED ISSUE #545: Geometry collections -- trajectories
**Describe the feature**
A common requested feature is to enable trajectory classes to handle MD simulations + analysis of such.

I have been a little reluctant to put this into sisl, since sisl was filling a gap (I think) about tight-binding + DFT interoperability.
Although much of sisl has diverged a bit, I would be hesitant about implementing a very well-known and well-solved problem in other packages. I.e. there is no need to re-invent the wheel for doing analysis that could easily be done with other packages.

Here is a list of other packages:
- https://www.mdanalysis.org/
- https://www.mdtraj.org/1.4.2/index.html
- list of tools https://md-how-to.readthedocs.io/en/latest/tools.html
- probably many more (ASE?)

The way to mitigate this could be:
- create a datatype that can easily convert to other packages (without much copying)
- ensure that a `Geometry` class uses references to the trajectory xyz (reduces memory, but causes side-effects)

See also @pfebrer PR #167.

This is also a question for sisl going forward, maintaining lots of different algorithms is hard, and requires time and knowledge. So implementing this in sisl requires a consistent support of that functionality.

@tfrederiksen @pfebrer comments?

zerothi:
Here a first go at the `GeometryCollection`.

@pfebrer sorry to overwrite #167 , I think we shouldn't add any `tile` + friends to the class as it will be a clutter to maintain. Using the nomenclature as `pandas` does it seems appropriate here. What do you think?

@tfrederiksen @pfebrer could you please both have a look here?

 - [x] Closes #545 
 - [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ we hit release version 1.0.0.
 ## [X.Y.Z] - YYYY-MM-DD
 
 ### Added
+- `GeometryCollection` enabling multiple geometries contained in a single collection
 - `Atoms.formula` to get a chemical formula, currently only Hill notation
 - unified the index argument for reading Grids, `read_grid`, this influences
   Siesta and VASP grid reads.
--- a/docs/api/basic.rst
+++ b/docs/api/basic.rst
@@ -29,6 +29,7 @@ Simple objects
    Atom
    Atoms
    Geometry
+   GeometryCollection
    SuperCell
    Grid
 
--- a/sisl/_collection.py
+++ b/sisl/_collection.py
@@ -3,7 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 # To check for integers
 from __future__ import annotations
-from typing import List, Callable
+from typing import Callable
 from collections import UserList
 
 
--- a/sisl/_collection.py
+++ b/sisl/_collection.py
@@ -1,0 +1,39 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+# To check for integers
+from __future__ import annotations
+from typing import List, Callable
+from collections import UserList
+
+
+__all__ = ["Collection"]
+
+
+class Collection(UserList):
+    """ Container for multiple objects in a single object
+
+    This is primarily intended for geometry-collections and friends.
+    For instance one could host a GeometryCollection + Collection(of velocities)
+    associated with the geometries.
+    """
+
+    def applymap(self, func: Callable, **kwargs) -> Collection:
+        """ Apply a function to all elementwise
+
+        Applies the function `func` to each of the contained objects
+        and returns a new collection with the applied function to them.
+
+        Parameters
+        ----------
+        func : callable
+            the function to be called on each object contained
+        **kwargs : optional
+            keyword arguments passed directly to `func`
+
+        Returns
+        -------
+        Collection : a new collection with each geometry transformed by `func`
+        """
+        return self.__class__(func(g, **kwargs) for g in self)
+
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -3,7 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 # To check for integers
 from __future__ import annotations
-from typing import List, Union, Iterator, Optional, TYPE_CHECKING
+from typing import Callable, List, Union, Iterator, Optional, TYPE_CHECKING
 from numbers import Integral, Real
 from math import acos
 from itertools import product
@@ -45,7 +45,7 @@ from ._dispatcher import AbstractDispatch
 from ._dispatcher import ClassDispatcher, TypeDispatcher
 
 
-__all__ = ['Geometry', 'sgeom']
+__all__ = ['Geometry', "GeometryCollection", 'sgeom']
 
 
 # It needs to be here otherwise we can't use it in these routines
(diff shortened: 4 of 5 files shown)


pfebrer:
> sorry to overwrite https://github.com/zerothi/sisl/pull/167 , I think we shouldn't add any tile + friends to the class as it will be a clutter to maintain.

No problem, that was just a first suggestion so that you implemented it in some way :)

> Using the nomenclature as pandas does it seems appropriate here. What do you think?

It's a bit strange because the functionality for modifying geometries is defined as methods, so you would often pass lambda functions to `applymap`: `gc.applymap(lambda g: g.tile(...))`. Perhaps this indicates that the functionality should be moved into functions :eyes: 

The other thing is that to be consistent with the `BrillouinZone` objects, the functionality could be implemented with dispatchers? I.e. `gc.apply.tile(...)`.

zerothi:
> > sorry to overwrite #167 , I think we shouldn't add any tile + friends to the class as it will be a clutter to maintain.
> 
> No problem, that was just a first suggestion so that you implemented it in some way :)
> 
> > Using the nomenclature as pandas does it seems appropriate here. What do you think?
> 
> It's a bit strange because the functionality for modifying geometries is defined as methods, so you would often pass lambda functions to `applymap`: `gc.applymap(lambda g: g.tile(...))`. Perhaps this indicates that the functionality should be moved into functions eyes

But that is also how one does it in pandas, no?

My idea about `applymap` would be:
```python
gcout = gc.applymap(Geometry.tile, reps=2, axis=2)
```
I am not sure I fully understand your last paragraph? Would you advocate `GC.tile` + friends?   
Or how exactly?

> The other thing is that to be consistent with the BrillouinZone objects, the functionality could be implemented with dispatchers? I.e. gc.apply.tile(...).objects, the functionality could be implemented with dispatchers? I.e. gc.apply.tile(...).

That could be nice, but still it would require you to create lots of stuff, and right now I have a problem with the dispatchers, what I really want is some way to do nested calls, i.e. `gc.apply.tile(...).translate(...)` which simply does not work in the dispatchers (a bit side-tracked, but...)

zerothi:
Now I think I understand. But the `GeometryCollection` is not a *single* entity like the `Geometry` object, it really is a set of *data points* and hence I think it could be argued that one shouldn't do it as functions. Consider if you only want to tile every 2nd object. How should the interface for `tile` look?
Agreed the current one does not account for it either, but it could via an argument... hmm...

pfebrer:
I'm now thinking that you are basically implementing a list with some extra functionality (for now, the write method). So it could be:

```python
from collections import UserList

class GeometryCollection(UserList):
    def write(...):
        ...
```

I guess you would still need an `applymap` method if you want a mapped `GeometryCollection`, because python lists don't seem to implement it.

zerothi:
> I'm now thinking that you are basically implementing a list with some extra functionality (for now, the write method). So it could be:
> 
> ```python
> from collections import UserList
> 
> class GeometryCollection(UserList):
>     def write(...):
>         ...
> ```
> 
> I guess you would still need an `applymap` method if you want a mapped `GeometryCollection`, because python lists don't seem to implement it.

Yeah, I thought about that, but I didn't want count... Hmm. Perhaps you are right, I should just do it like that! ;)

pfebrer:
> I am not sure I fully understand your last paragraph? Would you advocate GC.tile + friends?

Sorry, no, what I meant was something like:

```python
from sisl.geometry_ops import tile

gc.applymap(tile)
```

By the way, I still think that the benefit of this with respect to a normal list is very small (although I agree that the `write` method is convenient). For a trajectory, `applymap` will probably be orders of magnitude slower than having an `xyz` array with an extra dimension and applying operations on that. Do you think `Geometry` methods will work with an extra coordinate dimension as they are now? Although a variable cell would make it a bit harder...

zerothi:
> > I am not sure I fully understand your last paragraph? Would you advocate GC.tile + friends?
> 
> Sorry, no, what I meant was something like:
> 
> ```python
> from sisl.geometry_ops import tile
> 
> gc.applymap(tile)
> ```

Hmm... so you want to double everything... Hmm... I am not sure this is a good way of doing it, I think the above is quite clear, although a bit verbose, but one could do:
```python
G = sisl.Geometry
gc.applymap(G.tile)
```
which isn't too hard. :)


> By the way, I still think that the benefit of this with respect to a normal list is very small (although I agree that the `write` method is convenient). For a trajectory, `applymap` will probably be orders of magnitude slower than having an `xyz` array with an extra dimension and applying operations on that. Do you think `Geometry` methods will work with an extra coordinate dimension as they are now? Although a variable cell would make it a bit harder...

Agreed, and that would be the ideal case, but for now I think this is *ok*. Consider that we need to handle differently the `tile` vs. `translate` methods, tile *may* change `nsc`, while `translate` does not.

Actually, come to think of it, `applymap` shouldn't by default return a new GC, consider you want: `gc.applymap(lambda g: g.center())` then it makes no sense...

And no, `Geometry` methods would not work with the extra dimension. I have thought that it _could_ be beneficial if the `GC` class has the full xyz, and all contained geometries merely have a view on the subset coordinates. A different way of doing it, and this would require explicit handling of the methods to actually get any benefit from it...

tfrederiksen:
I think the current proposal looks good, clean and intuitive.

> Actually, come to think of it, `applymap` shouldn't by default return a new GC, consider you want: `gc.applymap(lambda g: g.center())` then it makes no sense...

For such situations, why would one not simply do `[g.center() for g in gc]`?

zerothi:
> For such situations, why would one not simply do `[g.center() for g in gc]`?

I can go with that. :)

Anything else before I merge? Then #544 could be rebased onto main, and return `GeometryCollection`?

zerothi:
> ... `applymap` will probably be orders of magnitude slower than having an `xyz` array with an extra dimension and applying operations on that. ...

The problem with that approach is that it enforces all contained geometries to be the same.

zerothi:
@pfebrer anything else required before merging?

zerothi:
> > The problem with that approach is that it enforces all contained geometries to be the same.
> 
> Yes, I know. But I think it would be worth it to have a special class for that case. It is the most relevant use case I would say.
> 
> Just reading and writing to a list of geometries I think it's a quick fix for the problem of the ANI sile but it's not going to be used in practice for moderately long trajectories (as it is now at least). But anyway you can merge it and later there could be the addition of trajectories if needed.

But the point was that that information should be offloaded into a proper MD analysis library? No? Then we won't have a problem.

In case the full trajectory should be build, it would also mean that the reading utilities should be reworked so data can be concatenated efficiently, rather than doing it at the end.

PR #547, opened 2023-03-08
mnt: nodes context settings are no longer a regular input

### LINKED ISSUE #538: sisl.nodes and automatic_recalc
**Describe the feature**

Instead of passing the automatic_recalc as a function argument, wouldn't it be more explicit to use a separate method for the class?

I.e. a user might find it weird that an _extra_ argument to the function call will change the behaviour.
Since the `update_inputs` is already documented, it shouldn't be too much of a hazzle to make a `Node.set_autocalc(True|False)` or similar? Or perhaps, since the documentation emphasises `lazy` behaviour, we might name it `Node.set_lazy(True|False)`?

Is there a special point on adding extra arguments to the function?

@pfebrer

pfebrer:
Closes #538 

Previously, settings that controlled the behavior of nodes where passed as function arguments, which was confusing:

```python
from sisl.nodes import Node

@Node.from_func
def sum_node(a, b):
    return a + b

my_node = sum_node(1, 2, automatic_recalc=True)
```

Now it is instead controlled with contexts. A node's behavior is defined by its context, which is a chainmap from sisl's main context (`sisl.nodes.context.MAIN_CONTEXT`), the node class context (`Node.context`), the instance context and a global temporal context.

Therefore, the same behavior can be obtained in different ways, which act at different levels:

```python
from sisl.nodes import Node, temporal_context

@Node.from_func
def sum_node(a, b):
    return a + b

my_node = sum_node(1, 2)

# Option 1: Modify the global sisl context, which affects all nodes
sisl.nodes.MAIN_CONTEXT.update(lazy=False)

# Option 2: Set the laziness for all instances of a Node
Node.context.update(lazy=False)

# Option 3: Set the laziness for all instances of a sum_node
sum_node.context.update(lazy=False)

# Option 4: Set the laziness only for this instance
my_node.context.update(lazy=False)

# Option 5: Set a temporal context, which overrides all contexts.
with temporal_context(lazy=False):
    ... do whatever
```

I also changed what `lazy` means. Before this PR it defined what was returned when calling a node class. If lazy, a node instance was returned and if not, the result was directly returned. Since it is now very easy to go from function to node and viceversa I thought that calling a node class should always return a node instance to avoid confusion. Now `lazy` means just whether the node should be recomputed on change of inputs, which was the old `automatic_recalc`, now removed.


sisl/nodes/context.py
```diff
@@ -1,36 +1,103 @@
 import contextlib
+from collections import ChainMap
+from typing import Any, Union
 
-def set_lazy_computation(nodes: bool = True, workflows: bool = True):
-    """Set the lazy computation mode for nodes and workflows.
+# The main sisl nodes context that all nodes will use by default as their base.
+MAIN_CONTEXT = dict(
```
zerothi, line 6: should we call this `SISL_NODE_CONTEXT` for clarity? Since this is importable the `MAIN` seems redundant?

pfebrer: Done!

zerothi:
Great!!!

I'll just merge!

PR #548, opened 2023-03-08
aniSileSiesta part 2

tfrederiksen:
Some further changes seem needed:
- bugfix for single-entry `GeometryCollection`
- reading ANI file should return a `GeometryCollection`
- more tests

Note that in `ani.py` I think the `all=True` keyword should only be passed to `read_geometry` if the user is not already trying to set any of the keywords `start, stop, step, all`.

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -4891,6 +4891,11 @@ class Geometry(SuperCellChild):
 class GeometryCollection(Collection):
     """ Container for multiple geometries in a single object """
 
+    def __init__(self, geometries):
+        if isinstance(geometries, Geometry):
+            geometries = [geometries]
+        super().__init__(geometries)
+
     @property
     def geometries(self) -> List[Geometry]:
         return self.data
--- a/sisl/io/siesta/ani.py
+++ b/sisl/io/siesta/ani.py
@@ -4,6 +4,7 @@
 """
 Sile object for reading ANI files
 """
+from sisl import GeometryCollection
 from ..xyz import xyzSile
 from ..sile import add_sile
 from sisl._internal import set_module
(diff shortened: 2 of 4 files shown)


tfrederiksen:
Btw, I think that reading multiple geometries from an `xyzSile` should also return a `GeometryCollection`. Currently it is just a list.

zerothi:
> Btw, I think that reading multiple geometries from an `xyzSile` should also return a `GeometryCollection`. Currently it is just a list.

Yes, I am finishing some things there. There are quite a bit of logic that is just wrong as it is. :)
So I'll amend some things, and then merge your stuff.

zerothi:
Could you try the latest commit? It should have fixed the inconsistencies present, but it was required to change the `all` behaviour, i.e. the explicitly given arguments are retained.

I would however really like the tests you've added. So 1) does the current main fix the problems, and if so, 2) could you rebase, and remove the changes made, keeping the tests?

zerothi:
I will merge your changes with some changes. 

The `all` flag is just too messy to deal with. The way I could make it work intuitively was that `all` _only_ sets defaults for `start`, `stop`, `step` if they are not set. I.e.
```python
start=1, stop=34, step=1, all=True
```
then `all` has no importance. It also makes more sense when you start doing:
```python
start=2, all=True
```
in which case it will read everything from the 3rd geometry and onwards.
```python
step=3, all=True
```
will then read all with a step size of 3.

zerothi:
> OK, great, please go ahead. Regarding `all` my thought was that this should be a simple keyword that overrules eventual `start/stop/step` instructions. I think the wording is pretty clear what it means.

I partially agree. When testing a few things it just felt weird that if I put `start=X, all=True` it would do `start=0`. In the end I think if you want `all` then no other arguments should be present. And this is most likely how people will use it anyway.

zerothi:
This is now merged! Thanks!

tfrederiksen:
Is it worth to mention ANI files support in the CHANGELOG?

zerothi:
Done!

PR #549, opened 2023-03-12
Bugfix for rotate(..., atoms=list) + additional tests

tfrederiksen:
I believe to have found a bug related to specifying a subset of atoms for rotation. For instance, before this PR:
```
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sisl
>>> g = sisl.geom.graphene() * (2, 2, 1)
>>> g.rotate(30, 'z', atoms=[1, 2])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".local/lib/python3.10/site-packages/sisl/messages.py", line 96, in wrapped
    return func(*args, **kwargs)
  File ".local/lib/python3.10/site-packages/sisl/geometry.py", line 2355, in rotate
    xyz[atoms, idx] = (q.rotate(xyz[atoms] - origin) + origin)[:, idx]
IndexError: shape mismatch: indexing arrays could not be broadcast together with shapes (2,) (3,) 
```
I added a new test function to expose the above (as well as also testing the `origin` keyword).

### DIFF
--- a/sisl/geometry.py
+++ b/sisl/geometry.py
@@ -2346,13 +2346,15 @@ class Geometry(SuperCellChild):
             if d in what:
                 idx.append(i)
 
-        # get which coordinates to rotate
         if idx:
             # Prepare quaternion...
             q = Quaternion(angle, vn, rad=rad)
             q /= q.norm()
             # subtract and add origin, before and after rotation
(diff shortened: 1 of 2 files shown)


zerothi:
Thanks @tfrederiksen

PR #550, opened 2023-03-14
changed SuperCell to Lattice, long overdue

### LINKED ISSUE #95: sc arguments should be named cell

zerothi:
The ambiguity of referencing to the lattice/unit-cell as a supercell is confusing when we are also dealing with the description of the *number of supercells*.
Therefore we have changed all SuperCell references and sc arguments to Lattice and lattice, respectively.

This surely comes at a cost for end-users, since this is a major transition. I will keep backwards compatibility with the older implementations for at least a year (perhaps 2). But they should eventually be gone to prefer the lattice.

@tfrederiksen @pfebrer please chip in here, the discussion about this entered in #95. I have long wanted to change this since it is a bit ambiguous in its meaning.
This change will allow it to differ from "supercells" which I think is much clearer.  
The `Lattice` name was basically chosen to continue the `.cell` access of the lattice vectors.
I have thought about this for some time, but if you have suggestions, please do let me know!
The initial discussion also considered:
- class: `Cell`, lattice vectors `abc`
- class: `Cell`, lattice vectors `cell` (a little problematic IMHO)

 - [x] Closes #95 
 - [x] Tests added (I think I covered everything)


pfebrer:
My first thought is that lattice is nicer than supercell, it feels more natural :+1: 

Then there could also be a `Sites` class and a `Geometry` would be the combination of `Lattice`, `Sites` and `Atoms`. A combination of only `Sites` and `Lattice` could implement the purely geometrical methods that are now implemented in `Geometry`, which maybe would be useful for other things. The first thing that comes to mind is k points in the reciprocal lattice, but I'm sure it could have other applications e.g. in situations where you have a sparse sampling of space :)

tfrederiksen:
I like `Lattice` too, as well as @pfebrer's suggestion of `Sites`.

Note that in 3D there exists only 14 different [Bravais lattices](https://en.wikipedia.org/wiki/Bravais_lattice). Maybe they could be introduced for some symmetry-related functionality?

zerothi:
Ok, so let's merge this and be gone with the madness. @tfrederiksen once merged, it would be nice to see if hubbard works with no changes, I have tried to catch tests as much as possible, but with this major change I have the feeling that I forgot something ;)

tfrederiksen:
With inspiration from spglib, one could rename `.cell` to `.basis` (basis vectors) and reserve `cell` for information about the connectivity etc.

zerothi:
Hmm, I am not too fond of that convention. Kind of like `cell` like unit-cell. I don't fully get your last paragraph?

zerothi:
I was considering `abc` or `cell`, I think `basis` is a little more "mathy", and not so physics oriented. Could be wrong though?

tfrederiksen:
I am just wondering if the notion of  `nsc` is clear for the user, and if `basis` would be a more natural property of `Lattice`.

zerothi:
agreed, `nsc` could probably be spelled out `supercells`. I just hate long words... ;)

tfrederiksen:
Isn't it rather that a supercell means the collection of the unit cell + its neighbours?

zerothi:
yes, but that is essentially what `nsc` means, no? Or should we do `nsc = cell_neighbours`, or something similar? The supercell phrase is just used extensively (in the code and in litterature).

tfrederiksen:
Another word could be `connections`.

zerothi:
That's not entirely fitting either, is it orbital connections, or ... 

Hmm... in the sparse structures we use `Sparse.edges` but neither are referring to the clarity of supercells, I think. The fact that this is an auxiliary supercell I think is quite vital in the understanding of its use, and if that gets lots in the name, I am afraid it will confuse?

pfebrer:
It is weird because `nsc` is not really a property of the lattice, no? It only makes sense in a combination of a lattice, some sites, and something that defines the range of the sites. Perhaps the lattice information should be separated from the auxiliary cell information, which could be an attribute -`aux_cell`- of `Geometry` or whatever other class where it makes sense.

So `lattice` would hold `cell` and `pbc`, and `aux_cell` would hold `nsc`.

tfrederiksen:
Supercells in plural is what I find a bit confusing terminology.

zerothi:
> It is weird because `nsc` is not really a property of the lattice, no? It only makes sense in a combination of a lattice, some sites, and something that defines the range of the sites. Perhaps the lattice information should be separated from the auxiliary cell information, which could be an attribute -`aux_cell`- of `Geometry` or whatever other class where it makes sense.
> 
> So `lattice` would hold `cell` and `pbc`, and `aux_cell` would hold `nsc`.

True... The plot thickens...

PR #560, opened 2023-04-11
first try at moving to scikit-build + cmake

### LINKED ISSUE #517: Move to scikit-build+cmake build for extensions
Title says it all

zerothi:
The compilation will take longer due to default flags in scikit-build which uses lto + more advanced compilation flags.

Perhaps we shuold reduce the flags in the tests since it takes too long.

The nice thing is that we can now build in parallel.

export SKBUILD_BUILD_OPTIONS="-j6"
and
python setup.py ... --global-option=build_ext --global-option=-j6

 - [x] Closes #517 
 - [x] Tests added


zerothi:
This is breaking the build-directories which is quite annoying, but I just couldn't get scikit-build to function with the namespace packages and sisl...

Since this is also the recommended way that pypa assumes projects I think this is ok. I don't like the extra dir, but...

pfebrer:
This seems more complicated than the old `setup.py` no? :sweat_smile: Now for if I add a cython file on a new directory I need to create a CMakeLists.txt file?

pfebrer:
By the way it would be awesome if sisl could compile to Web Assembly. There's a very big project pyodide which has achieved the compilation of CPython and many packages, so they can now be directly ran on the browser.

See for example the console in https://numpy.org/.

This would be specially awesome to deploy a GUI in javascript, since it wouldn't need a server. The GUI would just be a webpage that runs sisl on the browser. It would also be very nice for tutorials in general. This means that **everyone** could run `sisl` with no requirements, except for having a browser (fair enough I would say).

There are some instructions on how to build packages with C extensions:

https://pyodide.org/en/stable/development/new-packages.html#building-a-python-package-in-tree

It is basically all specified in a config file called `meta.yaml`, and `scipy` has already been succesfully compiled (you can import it in numpy's website), so probably it would be fairly easy to make it work by looking at their `meta.yaml`:

https://github.com/pyodide/pyodide/tree/main/packages/scipy

The biggest issue is with fortran sources, but they explain the approach they took in the `info.md` file.

Just in case you want to give it a try, now that you are into the building stuff :)

zerothi:
> 

Hmm, I don't have any plans to do this... :)

pfebrer:
Ok, when you are done with this I will try to do it. I think it is worth it.

zerothi:
> Ok, when you are done with this I will try to do it. I think it is worth it.

It's because it is required from 3.12, otherwise I probably wouldn't have bothered. :)

pfebrer:
By the way, do you plan to get rid of the fortran sources? Is it even possible to read the fortran binary files in some other way?

zerothi:
> By the way, do you plan to get rid of the fortran sources? Is it even possible to read the fortran binary files in some other way?

No, I don't plan on that. I pretty much want them to be present. It is easier to work with, and should be the fastest possible performance. And it will also allow external collaborators to more easily add fortran codes (in case their DFT codes are in fortran). So I think there is little gain (probably only problems) to move to other parsers.

pfebrer:
Ok, I was asking because at some point you asked me to change a fortran extension to Cython and I understood that support for fortran was going to get worse with time. So, what's the "policy"? Use fortran only to interact with fortran codes?

zerothi:
> Ok, I was asking because at some point you asked me to change a fortran extension to Cython and I understood that support for fortran was going to get worse with time. So, what's the "policy"? Use fortran only to interact with fortran codes?

I recall.
- if the code is going to read data-files generated from fortran sources, then fortran files should be used.
- if the code is just doing manipulations on stuff from Python/Cython, then fortran should not be used.

pfebrer:
The README of `scikit-build` states these two issues:
- Editable installs do not work with the latest versions of Setuptools (and had issues with older versions, too).
- The cache directory (_skbuild) may need to be deleted between builds in some cases (like rebuilding with a different Python interpreter).

I would say having editable install is very important, or you develop in some other way? Have you tried if it works?
I don't understand if the second issue implies that you can not build for multiple python versions from the same source directory.

zerothi:
> The README of `scikit-build` states these two issues:
> 
>     * Editable installs do not work with the latest versions of Setuptools (and had issues with older versions, too).
> 
>     * The cache directory (_skbuild) may need to be deleted between builds in some cases (like rebuilding with a different Python interpreter).
> 
> 
> I would say having editable install is very important, or you develop in some other way? Have you tried if it works? I don't understand if the second issue implies that you can not build for multiple python versions from the same source directory.

I agree that having editable installs would be nice, I have just never done it...

But we have to move to a different build-system. And scikit-build is suggested by scipy, and seems to work *OK*. I would rather not go the meson way, it is harder to customize...

Yes, the 2nd point is due to the issue you mention.

pfebrer:
How do you develop sisl if it's not with editable install?

zerothi:
> How do you develop sisl if it's not with editable install?

I install it to a location, then run tests from another directory. It is a bit cumbersome, but it failsafes some problems of the installation process that I have experienced. Now I just got used to it

pfebrer:
Do you need to install `sisl` each time you make a modification? I don't understand

zerothi:
> Do you need to install `sisl` each time you make a modification? I don't understand

Yeah, basically. :) But I of course have scripts that can copy changed files, if needed.

rgommers:
@zerothi I just noticed this, so I thought I'd comment (please ignore if you already considered this): `scikit-build` should be considered legacy at this point, I wouldn't recommend migrating to it now. If you prefer CMake over Meson, then try https://github.com/scikit-build/scikit-build-core; that the funded, newer effort and it's already significantly better than `scikit-build` I believe. They're receptive to feedback or requests for help too.

zerothi:
> @zerothi I just noticed this, so I thought I'd comment (please ignore if you already considered this): `scikit-build` should be considered legacy at this point, I wouldn't recommend migrating to it now. If you prefer CMake over Meson, then try https://github.com/scikit-build/scikit-build-core; that the funded, newer effort and it's already significantly better than `scikit-build` I believe. They're receptive to feedback or requests for help too.

Thanks for the note!  
It wasn't completely clear that it was deprecated. Although I did see that they recommended the scikit-build-core direction.

It just requires additional boiler plate cmake-stuff that I wasn't really prepared to do at the moment. :)

Looks like I'll have to move before merging this one.

I have already opened a few tickets related to its interaction with `pyproject.toml` :)
I'll move and adjust and open tickets. Again thanks!

zerothi:
This is now superseeded by #563 which actually works on CI (the reason is the sometimes creation of `*wrappers*.f90` files.

PR #562, opened 2023-04-14
Wrap fortran files import in try/except block.

pfebrer:
The rest of files that import fortran modules do this, so I think in this case it should also be done for consistency.

### DIFF
--- a/sisl/io/tbtrans/delta.py
+++ b/sisl/io/tbtrans/delta.py
@@ -17,7 +17,10 @@ from sisl.sparse import _ncol_to_indptr
 from sisl.messages import warn, deprecate_argument
 from sisl.unit.siesta import unit_convert
 from ..siesta._help import _csr_to_siesta, _csr_from_sc_off, _mat_spin_convert
-from ..siesta._siesta import siesta_sc_off
+try:
+    from ..siesta._siesta import siesta_sc_off
+except:
+    pass
 
 
 __all__ = ['deltancSileTBtrans']


zerothi:
You are right. But perhaps we should just force the need for the fortran modules? Wouldn't this be acceptable? I don't even know how to install sisl without these modules...

pfebrer:
In `v0.13.0` I removed the fortran extensions from `setup.py` and with this extra patch I could compile sisl on web assembly and use it without problems.

I don't know, it might be useful to allow `sisl` to be used without the fortran modules, just because the fortran extensions are not so easy to compile as the cython ones, at least for now. Compiling to web assembly is an example, but Windows is also another example (perhaps conda could compile easily for windows without the fortran functionality). Since the fortran functionality is "secondary", moreso now that SIESTA starts to write outputs in netcdf format, I would say it's reasonable.

Perhaps the try/except blocks could be removed from the main sisl source but maintain a patch file that when applied allows sisl to be compiled and ran without fortran.

zerothi:
> In `v0.13.0` I removed the fortran extensions from `setup.py` and with this extra patch I could compile sisl to web assembly and use it without problems.
> 
> I don't know, it might be useful to allow `sisl` to be used without the fortran modules, just because the fortran extensions are not so easy to compile as the cython ones, at least for now. Compiling to web assembly is an example, but Windows is also another example (perhaps conda could compile easily for windows without the fortran functionality). Since the fortran functionality is "secondary", moreso now that SIESTA starts to write outputs in netcdf format, I would say it's reasonable.
> 
> Perhaps the try/except blocks could be removed from the main sisl source but maintain a patch file that when applied allows sisl to be compiled and ran without fortran.

Ok, fine. Lets continue to have this possibility. This should be easier with the cmake stuff. Later on we may add this functionality to disable fortran sources.

pfebrer:
What might be better than the `try/except` would be a placeholder module that can be imported and tells the user that there's no fortran functionality when sisl tries to use something from that module.

zerothi:
With the current `scikit-build-core` transitioning you can now do:
```shell
python -m pip install --config-settings=cmake.define.WITH_FORTRAN=NO .
```
and it will install everything, without the fortran sources compiled! :)

pfebrer:
That is very cool!  `CMAKE_ARGS="-DWITH_FORTRAN" python -m pip install` should also work, right?

I've seen other packages prefix their specific flags, do you think it could make sense to name it `SISL_WITH_FORTRAN` to avoid interfering with other flags?

PR #563, opened 2023-04-17
Scikit build core

### LINKED ISSUE #517: Move to scikit-build+cmake build for extensions

zerothi:
Changing from `scikit-build` to `scikit-build-core`.

This also superseeds #560.

 - [x] Closes #517
 - [ ] Tests added


zerothi:
@pfebrer could you perhaps see if you can install this one, if you can get it in, I'll merge ASAP to get on!

pfebrer:
Yes, I'll try this afternoon!

pfebrer:
What do I need to do? `pip install .`?

zerothi:
Yes, or whatever you can do, editable as well... I'll also test a bit more... But if you can surely test editable installs that would be great!

pfebrer:
I did `pip install .` and it seemed to work, but it is very strange because I don't have `scikit-build` installed.

I also tried the editable install and it didn't work.

pfebrer:
Or the build environment that is specified in `build-system` is automatically installed only for the build?

zerothi:
> Or the build environment that is specified in `build-system` is automatically installed only for the build?

Yes, that is the idea! :)

zerothi:
Ok, thanks. I can reproduce. :)

zerothi:
I know that `scikit-build-core` still marks it as experimental. When looking at the `_sisl_editable.pth/py` files, I can see that the paths to the `*.so` files are relative, rather than absolute.

I *guess* one can adapt these to have absolute paths, and then it should work, to test this I would simply edit the `sisl._indices` module and see if one can get by 1 of the libraries, if this gets fixed then we *could* ship a tool in sisl that fixes the editable paths.

zerothi:
For completeness sake, I can get the editable installs to work by using `tools/fixeditable.bash` script

PR #566, opened 2023-04-19
Make documentation tabs more visual

pfebrer:
It was very difficult to understand where the content of the tabs ended.

By adding some custom CSS, we have made it more intuitive.

Previous layout:

Screenshot from 2023-04-19 11-09-49

Current layout:

Screenshot from 2023-04-19 11-21-53

### DIFF
--- a/docs/_static/css/custom_styles.css
+++ b/docs/_static/css/custom_styles.css
@@ -1,0 +1,24 @@
+/*STYLES TO MAKE TABS MORE VISUAL*/
+
+/*The parent container of tabs*/
+.tab-set {
+  border: 1px solid #ccc
+}
+
+/*Individual box of each tab*/
+.tab-set > label {
+  margin-bottom: 0
+}
+
+/*Container appearing below the tabs, with the contents of the active tab*/
+.tab-content {
+  background-color: aliceblue;
+  padding: 20px
+}
+
+/*Initial paragraph of the content. By default there's a top margin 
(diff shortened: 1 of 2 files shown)


docs/requirements.txt
```diff
@@ -15,6 +15,7 @@ sphinx-rtd-theme
 sphinx-autodoc-typehints
 sphinx-inline-tabs
 sphinxcontrib-bibtex
+sphinxcontrib.jquery
```
zerothi, line 18: please remove this, it is implicitly required by rtd theme :)

zerothi: See here: https://github.com/readthedocs/sphinx_rtd_theme/issues/1452#issuecomment-1490504991

pfebrer: Hmm I ran `pip install -r requirements.txt` and it still told me that this package was missing. Perhaps it is only from a certain version of the RTD theme?

zerothi: Hmm, lets remove it as that is recommended by the rtd theme. Hmm.. Not that many are going to do it anyways, it is more if rtd decides to use another backend for search. gh-actions worked without problems, so perhaps something on your end?

pfebrer: Surely, I think it is just that I had an old version of the RTD theme already installed, that's why probably in the requirements the version of the RTD theme should be specified

zerothi:
It is live!

PR #573, opened 2023-05-25
VASP forces from OUTCAR 

tfrederiksen:
Nothing fancy, but this PR enables users to read atomic positions and forces from VASP (eg., in a relaxation run) from the OUTCAR file

 - [ ] Closes #xxxx
 - [x] Tests added

### DIFF
--- a/src/sisl/io/vasp/stdout.py
+++ b/src/sisl/io/vasp/stdout.py
@@ -87,7 +87,7 @@ class stdoutSileVASP(SileVASP):
         -------
         PropertyDict : all energies from the "Free energy of the ion-electron system" segment of VASP output
         """
-        warn(f"{self!s} no longer returns the last entry (but the next in file) as default!")
+        warn(f"{self!s}.read_energy no longer returns the last entry (but the next in file) as default!")
         name_conv = {
             "alpha": "Z",
             "Ewald": "Ewald",
--- a/src/sisl/io/vasp/stdout.py
+++ b/src/sisl/io/vasp/stdout.py
@@ -7,7 +7,7 @@ from .sile import SileVASP
 from ..sile import add_sile, sile_fh_open
 from .._multiple import SileBinder
 
-from sisl.messages import deprecation, warn
+from sisl.messages import deprecation, deprecate_argument
 from sisl.utils import PropertyDict
 from sisl._internal import set_module
 
@@ -59,9 +59,12 @@ class stdoutSileVASP(SileVASP):
 
     @SileBinder()
     @sile_fh_open()
-    @deprecation("all keyword is deprecated, use read_energy[:]() instead", from_version="0.14")
+    @deprecate_argument("all", None, "use read_energy[:]() instead to get all entries", from_version="0.14")
+    @deprecation("WARNING: direct calls to stdoutSileVASP.read_energy() no longer returns the last entry! Now the next block on file is returned.", from_version="0.14")
     def read_energy(self):
-        """ Reads the energy specification from OUTCAR and returns energy dictionary in units of eV
+        """ Reads an energy specification block from OUTCAR
+
+        The function steps to the next occurrence of the "Free energy of the ion-electron system" segment
 
         Notes
         -----
@@ -85,9 +88,9 @@ class stdoutSileVASP(SileVASP):
 
         Returns
         -------
-        PropertyDict : all energies from the "Free energy of the ion-electron system" segment of VASP output
+        PropertyDict : all energies from a single "Free energy of the ion-electron system" segment
         """
-        warn(f"{self!s}.read_energy no longer returns the last entry (but the next in file) as default!")
+
         name_conv = {
             "alpha": "Z",
             "Ewald": "Ewald",
@@ -123,7 +126,6 @@ class stdoutSileVASP(SileVASP):
         v = line.split()
         E.total = float(v[4])
         E.sigma0 = float(v[-1])
-        print(E)
         return E
 
     @SileBinder()
--- a/src/sisl/io/vasp/stdout.py
+++ b/src/sisl/io/vasp/stdout.py
@@ -1,13 +1,15 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
+import numpy as np
+
 from .sile import SileVASP
 from ..sile import add_sile, sile_fh_open
+from .._multiple import SileBinder
 
 from sisl.messages import deprecation
 from sisl.utils import PropertyDict
 from sisl._internal import set_module
-
 
 __all__ = ["stdoutSileVASP", "outSileVASP"]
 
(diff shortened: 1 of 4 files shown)


zerothi:
Thanks for this @tfrederiksen!

I think we need to conform to the idea that `read_forces` should only return forces, not the geometry as well.

Perhaps we should have a method `read_md|steps|trajectory|path|?` to return MD stuff (more than geometry|forces), this would fall in line with the `read_scf` method which also contains such information.

tfrederiksen:
OK, for me it would be fine to rename the current function to `read_trajectory`. And we could also have a `read_forces` (relying on the current function). Is this an acceptable approach?

Note that the relevant block(s) in the OUTCAR contains both positions and forces, so it is basically for free to return both.

zerothi:
> OK, for me it would be fine to rename the current function to `read_trajectory`.

Yes, lets do something like this. What should we name it?
`trajectory` I now think is advocating only coordinates.

Ideas:
- `steps`
- `time_steps`

In any case I think we should return a `PropertyDict` with fields like:
```python
steps = PropertyDict()
steps.geometries = GeometryCollection(geometries)
steps.forces = Collection(forces)
return steps
```
The advantage is that future additions can be done without disrupting code, i.e. some codes can return
energy information as well as forces etc.

> And we could also have a `read_forces` (relying on the current function). Is this an acceptable approach?
Agreed, that would be ok, I think :)
It would be great if they could use the `@sile_read_multiple` decorator for automatic adding arguments.

> Note that the relevant block(s) in the OUTCAR contains both positions and forces, so it is basically for free to return both.
Ok.

zerothi:
Ok, lets do that then. Later when/if we can parse the data for the chemical species, we can add a `geometry` field.

tfrederiksen:
It would of course be easy with the OUTCAR basepath to also read a POSCAR for the chemical species. But opening another file internally in this function seems a bit messy, or?

zerothi:
> It would of course be easy with the OUTCAR basepath to also read a POSCAR for the chemical species. But opening another file internally in this function seems a bit messy, or?

I have done this for e.g. the fdf file. If they are present, then it would be ok. Better to have consistent data. But then we shouldn't read the geometry, only the atomic information.

tfrederiksen:
OK, I'll give it a try when I find time.

tfrederiksen:
Maybe this has been discussed elsewhere, but couldn't the geometry class not optionally hold info about forces, strain, velocities, constraints, etc?

zerothi:
> Maybe this has been discussed elsewhere, but couldn't the geometry class not optionally hold info about forces, strain, velocities, constraints, etc?

This issue is tracked in #11 I have yet to find a stable solution that also works when manipulating the geometry. For instance, what would happen if one translates a subset of the coordinates, does it make sense to remove the forces (etc) or what should it do. But we should continue the discussion in #11

src/sisl/io/vasp/out.py
```diff
@@ -144,6 +145,73 @@ def readE(itt):
             return E[-1]
         return None
 
+
+    @sile_fh_open()
+    def read_trajectory(self, all=True):
```
zerothi, line 150: I think it would be great to use the `@sile_read_multiple` to automatically handle things.

src/sisl/io/vasp/out.py
```diff
@@ -144,6 +145,73 @@ def readE(itt):
...
+            if not f:
+                return None
+            next(itt) # -----
+            P, F = [], []
+            line = next(itt)
+            while "----" not in line:
+                v = line.split()
+                # positions and forces
+                P.append([float(v[0]), float(v[1]), float(v[2])])
+                F.append([float(v[3]), float(v[4]), float(v[5])])
```
zerothi, line 187: It'll probably be faster to do the final conversion in `np.array(C,dtype=np.float64)` or similar, then the code becomes cleaner and faster.

src/sisl/io/vasp/out.py
```diff
@@ -144,6 +145,73 @@ def readE(itt):
...
+            step.force = np.array(F)
+            return step
+
+        itt = iter(self)
+        trajectory = []
+        step = read_step(itt)
+        while step is not None:
+            trajectory.append(step)
+            step = read_step(itt)
+
```
zerothi, line 201: my idea was a little different.
```python
steps = PropertyDict()
steps.geometries = GeometryCollection(geometries)
steps.forces = Collection(forces)
return steps
```
We can consider a PropertyDict wrapper that returns tuples when requesting indices, but that might be another round.

zerothi (changes requested):
Thanks @tfrederiksen! Some comments and improvements are needed.

src/sisl/io/vasp/out.py
tfrederiksen: Yes, but I did this to keep the same spirit as in `read_energy` (one PropertyDict per step).

src/sisl/io/vasp/out.py
tfrederiksen: I agree. (The only thing is that now it is not possible to get just the last entry with `start=-1`).

src/sisl/io/vasp/out.py
tfrederiksen: Done

src/sisl/io/vasp/out.py
zerothi: What do you mean? Does it error out? Or?

src/sisl/io/vasp/out.py
zerothi: Thanks!

src/sisl/io/vasp/out.py
zerothi: Ah, I see. These should be changed for consistency, I'll open a new issue.

src/sisl/io/vasp/out.py
tfrederiksen: The first entry in the file is returned with `read_trajectory(start=-1)[0]`.

tfrederiksen:
@zerothi , the errors exposed above in the test runs seem unrelated to this branch.

src/sisl/io/vasp/stdout.py
```diff
@@ -55,8 +57,10 @@ def accuracy_reached(self):
         """ True if the line "reached required accuracy" was found. """
         return self.step_to("reached required accuracy")[0]
 
+    @SileBinder()
     @sile_fh_open()
-    def read_energy(self, all=False):
+    @deprecation("all keyword is deprecated, use read_energy[:]() instead", from_version="0.14")
```
zerothi, line 62: I would propose:
```python
@deprecate_argument("all", None, <message>, from_version="0.14")
@deprecate("direct calls will no longer return the last entry!", from_version="0.14")
```
or similar, that should produce the same behaviour, and no `warn` in the method would be nice.

src/sisl/io/vasp/stdout.py
```diff
@@ -101,49 +101,62 @@ def read_energy(self, all=False):
...
+                E[name_conv[key]] = float(v[-1])
+            line = self.readline()
+        line = self.readline()
+        E.free = float(line.split()[-2])
+        self.readline()
+        line = self.readline()
+        v = line.split()
+        E.total = float(v[4])
+        E.sigma0 = float(v[-1])
+        print(E)
```
zerothi, line 126: debug statement should be removed

src/sisl/io/vasp/stdout.py
```diff
@@ -101,49 +101,62 @@ def read_energy(self, all=False):
...
+        return E
+
+    @SileBinder()
+    @sile_fh_open()
+    def read_trajectory(self):
+        f = self.step_to("BASIS-vectors are now", allow_reread=False)[0]
```
zerothi, line 132: documentation

src/sisl/io/vasp/stdout.py
```diff
@@ -101,49 +101,62 @@ def read_energy(self, all=False):
...
-        return None
+            # positions and forces
+            P.append(v[:3])
+            F.append(v[3:6])
+            line = self.readline()
+        step = PropertyDict()
+        step.cell = np.array(C, dtype=np.float64)
+        step.xyz = np.array(P, dtype=np.float64)
+        step.force = np.array(F, dtype=np.float64)
+        return step
```
zerothi, line 159: perhaps a `postprocess` argument to `SileBinder` to unpack the Property dict?
```python
def unpack2collection(steps):
    ret = dict()
    for step in steps:
        for key, val in step.items()):
            ret[key].append(val) # won't work as is not list-initialized, but...
    prop = PropertyDict()
    for key, steps in ret.items():
        prop[key] = Collection(steps)
    return prop
```
we really need to discuss how we should deal with multiple return values. and arguments for such

zerothi (changes requested):
minor details

src/sisl/io/vasp/stdout.py
tfrederiksen: I've addressed all comments except this one as I am not sure we want/need to unpack. For instance, the current behaviour seems logical to me:
```python
traj = stdoutSileVASP(f).read_trajectory[:3]()
for i in range(3):
    print(traj[i].force)
# or
for step in traj:
    print(step.force)
```
You would prefer this?
```python
traj = stdoutSileVASP(f).read_trajectory[:3]()
for i in range(3):
    print(traj.force[i])
# or
for F in traj.force:
    print(F)
```

zerothi (approved):

PR #575, opened 2023-05-31
cleaned the procedure for locating maxR

### LINKED ISSUE #574: Orbital set_radial and friends -- allow custom maximum R finding
**Describe the feature**

Currently the search method for the maximum R of an radial function is based on checking when the radial function is strictly 0.

However, this is pretty bad for exponentially decaying functions. So better to retain a "contained" fraction.

I would propose we do something like this:

```python
*Orbital(R = -0.9999) # contains locates minimum r where radial(r) ** 2  >= max(radial(r)**2) * 0.9999
*Orbital(R=10) # retains a maximum radius of 10 Ang
```
I think this would be a better approach and should be simpler for end-users to detect problems.

zerothi:
Instead of searching for a 0 in the radial function we now search for an integration that contians 99.99% of the radial function.
The integration uses the trapezoid integration mechanism with f**2*r**2 as that should be normalized for a
radial function.

This should make it much more stable against different radial function arguments and produce more accurate ranges for the orbitals.

Now a warning will be issued when the routine cannot determine a correct R (based on inputs).

Ensured that all orbitals with radial functions
can pre-calculate the R in case user did not supply it. One can always forcefully set it by supplying it.

@sofiasanz @tfrederiksen could you please have a look here, this should better adapt the R discovery of your `HydrogenicOrbital`

 - [X] Closes #574
 - [x] Tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,10 @@ we hit release version 1.0.0.
 
 ### Changed
 - bumped Python requirement to >=3.8
+- orbitals `R` arguments will now by default determine the minimal radii
+  that contains 99.99% of the function integrand. The argument now
+  accepts values -1:0 which is a fraction of the integrand that the function
+  should contain, a positive value will explicitly set the range #574
 - Added printout of the removed couplings in the `RecursiveSI`
 - `SuperCell` class is officially deprecated in favor of `Lattice`, see #95 for details
   The old class will still be accessible and usable for some time (at least a year)
--- a/src/sisl/orbital.py
+++ b/src/sisl/orbital.py
@@ -770,7 +770,7 @@ class SphericalOrbital(Orbital):
 
     def copy(self):
         """ Create an exact copy of this object """
-        return self.__class__(self.l, self._radial, self.q0, self.tag)
+        return self.__class__(self.l, self._radial, R=self.R, q0=self.q0, tag=self.tag)
 
     def equal(self, other, psi=False, radial=False):
         """ Compare two orbitals by comparing their radius, and possibly the radial and psi functions
@@ -1328,7 +1328,7 @@ class HydrogenicOrbital(AtomicOrbital):
 
     def copy(self):
         """ Create an exact copy of this object """
-        return self.__class__(self.n, self.l, self.m, self._Z, q0=self.q0, tag=self.tag)
+        return self.__class__(self.n, self.l, self.m, self._Z, R=self.R, q0=self.q0, tag=self.tag)
 
     def _radial(self, r):
         r""" Radial functional for the Hydrogenic orbital """
@@ -1375,7 +1375,7 @@ class _ExponentialOrbital(Orbital):
         args = list(args)
 
         # Extract shell information
-        n = kwargs.get("n", None)
+        n = kwargs.pop("n", None)
         l = kwargs.pop("l", None)
         m = kwargs.pop("m", None)
         alpha = kwargs.pop("alpha", None)
@@ -1440,7 +1440,7 @@ class _ExponentialOrbital(Orbital):
 
     def copy(self):
         """ Create an exact copy of this object """
-        return self.__class__(n=self.n, l=self.l, m=self.m, R=self.R, alpha=self.alpha, coeff=self.coeff, q0=self.q0, tag=self.tag)
+        return self.__class__(n=self.n, l=self.l, m=self.m, alpha=self.alpha, coeff=self.coeff, R=self.R, q0=self.q0, tag=self.tag)
 
     def __str__(self):
         """ A string representation of the object """
--- a/src/sisl/orbital.py
+++ b/src/sisl/orbital.py
@@ -8,6 +8,8 @@ from numbers import Integral, Real
 from math import pi
 from math import sqrt as msqrt
 from math import factorial as fact
+from typing import Callable, Optional, Tuple
+import numpy.typing as npt
 
 import numpy as np
 from numpy import cos, sin
@@ -368,7 +370,13 @@ class Orbital:
         self.__init__(d["R"], q0=d["q0"], tag=d["tag"])
 
 
-def radial_minimize_range(radial_func, contains, maxR=100, func=None):
+
+RadialFuncT = Callable[[npt.ArrayLike], npt.NDArray]
+def radial_minimize_range(radial_func: Callable[[RadialFuncT], npt.NDArray],
+                          contains: float,
+                          dr: Tuple[float, float]=(0.01, 0.0001),
+                          maxR: float=100,
+                          func: Optional[Callable[[RadialFuncT, npt.ArrayLike], npt.NDArray]]=None) -> float:
     """ Minimize the maximum radius such that the integrated function `radial_func**2*r**3` contains `contains` of the integrand
 
     Parameters
@@ -377,6 +385,10 @@ def radial_minimize_range(radial_func, contains, maxR=100, func=None):
        the function that returns the radial part
     contains : float
        how much of a percentage the squared function should contain @ R
+    dr : tuple of float, optional
+       the precision of the integral. First number is the coarse integral.
+       The second number determines the fine-integral to exactly determine R between
+       coarser points.
     maxR : float, optional
        maximally searched ``R``, in case there is no cross-over of the integrand
        containing `contains` in this range a ``-contains`` will be returned to
@@ -389,10 +401,7 @@ def radial_minimize_range(radial_func, contains, maxR=100, func=None):
     # We should never expect a radial components above
     assert maxR > 0.05, "maxR too small (> 0.05)"
     assert contains > 0, "contains too small (> 0)"
-
-    # Precision of 0.01 A @ 99.99% (default) of the integrand
-    # The subsequent integral will have a precision of 0.0001 A
-    dr = (0.01, 0.0001)
+    assert len(dr) == 2, "number of sub-divisions is not 2: dr argument"
 
     def func_base(func, r):
         # finding the best integral function for locating max
--- a/src/sisl/orbital.py
+++ b/src/sisl/orbital.py
@@ -29,9 +29,11 @@ from .utils.mathematics import cart2spher
 from sisl.constant import a0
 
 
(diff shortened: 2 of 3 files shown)


sofiasanz:
Nice!

For the hyrogenic orbital I was using for a carbon `pz` orbital I get the following numbers using different `Zeff`:

```python
pz = sisl.HydrogenicOrbital(2, 1, 0, Zeff, q0=1.0, R=-0.999999)
```

Output (maxR and the radial function evaluated at that point):
```
****** Zeff = 3.2 ******
pz.R =  3.8761000000000556
pz.radial([pz.R]) =  [0.00057862]
****** Zeff = 4 ******
pz.R =  3.101200000000003
pz.radial([pz.R]) =  [0.00080775]
****** Zeff = 5 ******
pz.R =  2.4820000000000677
pz.radial([pz.R]) =  [0.0011238]
```

If I don't set the `R` in `pz` (i.e. the default value set in `sisl`), I get:
```
****** Zeff = 3.2 ******
pz.R =  2.9418000000000886
pz.radial([pz.R]) =  [0.00740379]
****** Zeff = 4 ******
pz.R =  2.3537000000000083
pz.radial([pz.R]) =  [0.0103381]
****** Zeff = 5 ******
pz.R =  1.8838999999999964
pz.radial([pz.R]) =  [0.01439109]
```

zerothi:
Ah, now I see what you did. :)
Great! Hope the finetuning is *OK*. :)

Any other suggestions here?

sofiasanz:
> Ah, now I see what you did. :) Great! Hope the finetuning is _OK_. :)
> 
> Any other suggestions here?

Not for the moment! For me it looks fine! Let's see if @tfrederiksen has any other suggestions.

sofiasanz:
Hi Nick, I think there is something weird now in this branch.

I was plotting the LDOS maps for benzene using this branch (commit f89b533db31e71afee465f4e4a95abfbe7baa26) and I got these following figures which look very weird to me (the top row are the wavefunctions, just to compare with the obtained LDOS):

fig-574-orbital-1

I am using this orbital

```python
pz = sisl.HydrogenicOrbital(2, 1, 0, Zeff, q0=1.0, R=10)
```
to expand the lattice in real space. 

I compared to what I get with the main branch (using the same grid shape and everything), using version with commit 9e4307bf9c644bf338a3a1b325e6bd8e24c89e27, and I get the following images:

fig-main-1

It looks like the LDOS changed a lot and also the resolution, which I don't understand.

zerothi:
Hmm... Perhaps wait a bit, I need to check some things.

tfrederiksen:
I like the idea that the user can set a cutoff radius through the specification of a negative `R` (which is then understood as the target volume). My suggestion would be that any `R<0` could in principle be acceptable, for instance if the orbital is not normalized.

By the way, one could also have a function that normalizes the orbital on `r<=R`.

tfrederiksen:
I don't understand the rationale behind `radial**2 * r**3`. The normalisation condition is `int dr radial**2 * r**2 = 1`, no?

zerothi:
> I don't understand the rationale behind `radial**2 * r**3`. The normalisation condition is `int dr radial**2 * r**2 = 1`, no?

Yes, this is where I am currently investigating the problems.
I'll soon have another commit to this branch which will highlight the problem of doing the *actual radial integral*. Hold tight.

tfrederiksen:
> > I don't understand the rationale behind `radial**2 * r**3`. The normalisation condition is `int dr radial**2 * r**2 = 1`, no?
> 
> Yes, this is where I am currently investigating the problems. I'll soon have another commit to this branch which will highlight the problem of doing the _actual radial integral_. Hold tight.

Isn't it not just to solve for $R$ such that $\int_0^R dr f(r) ^2 r^2 =$ `contained`?

zerothi:
As @sofiasanz pointed out, the actual function looked a bit weird in the real-space pictures.

This I *think* can be traced to the short `R` returned from the `radial(r)**2 * r**2` integral. It simply yields a too short `R` to make sense for a proper real space visualization. This is because the integral converges much faster than the radial function does. Hence some other integral might be more appropriate.

I have tried various things, but I think the *simplest* and acceptable version would be the `abs(radial(r))` function. It is a bit weird, but should suffice.

I have now opted for a different path where users can pass arguments to control the minimization procedure.

This can effectively be done like this:
```python
R = {
"contains": 0.9999, # contain 99.99% of func integrated
"func": lambda radial, r: radial(r)**2 * r**2, # function to integrate
"maxR": 100, # max searched R distance (integral *must* converge within this range).
}
HydrogenicOrbital(..., R=R)
```
One can still supply a negative number which then translates to `{"contains": - R}` internally.
The function used is `radial_minimize_range` which is now also exposed through the API. It will allow users to test and play around with things. Comments on the function name and interface would be appreciated.

Please see the latest commit.

zerothi:
@sofiasanz could you try with the latest commit to see if the new default is more appropriate?

sofiasanz:
I like it :) I think also is much clearer and flexible now.

I'll try this commit in the afternoon and plot again the benzene LDOS to compare to what we get with the main branch. What is the default value for `R` now?

zerothi:
> I like it :) I think also is much clearer and flexible now.
> 
> I'll try this commit in the afternoon and plot again the benzene LDOS to compare to what we get with the main branch. What is the default value for `R` now?

```python
for Z in [3.2, 4, 5]:
    orb = HydrogenicOrbital(2, 1, 0, Z)#, R=-0.999999)
    R = np.arange(0, orb.R, 0.001)
    r = orb.radial(R)
    print("99.99", orb.R, r[-1] / r.max()*100)

    orb = HydrogenicOrbital(2, 1, 0, Z, R=-0.999999)
    R = np.arange(0, orb.R, 0.001)
    r = orb.radial(R)
    print("99.9999", orb.R, r[-1] / r.max()*100)
```
yields:
```
99.99   3.888400000000018 0.025069398892111586
99.9999 5.519599999999978 0.0002567943876730368
99.99   3.110700000000002 0.025104120475539306
99.9999 4.415699999999987 0.0002569771567260434
99.99   2.4886000000000186 0.025104120475539272
99.9999 3.5326000000000057 0.00025697715672604347
```
Hope this should be fine enough.

sofiasanz:
I produced this image (using `pz = sisl.HydrogenicOrbital(2, 1, 0, Zeff, q0=1.0, R=10)`) with this branch (f62c12dcc0959f81e86ab53dcd557a5fe4412607) which looks like what I got with the main branch yesterday, so it seems to be working just fine:

fig-574-orbital-1

sofiasanz:
Hi Nick, so I was trying to find the `R` at which I stop getting acceptable images for the LDOS, and I found something a little strange. The thing is that I kept changing `R` when building the `pz` orbital and I always get the same images, which doesn't make sense because I was changing the basis.
Then I found some mismatch between the `Grid` object and the `Geometry` object. For instance, if I fix the `R`, e.g. `R=10.0` I get the following outputs when printing the `Geometry` and `Grid` objects: 

```
---Zeff = 3.2---
Geometry{na: 6, no: 6,
 Atoms{species: 1,
  Atom{C, Z: 6, mass(au): 12.01070, maxR: 10.00000,
   HydrogenicOrbital{2pzZ1, q0: 1.0, SphericalOrbital{l: 1, R: 10.0, q0: 1.0}}
  }: 6,
 },
 maxR: 10.00000,
 Lattice{nsc: [1 1 1],
  A=[4.260, 0.000, 0.000],
  B=[0.000, 22.460, 0.000],
  C=[0.000, 0.000, 14.200],
 }
}

Grid{kind: f, shape: [100 100 1],
 bc: [periodic, periodic,
      periodic, periodic,
      periodic, periodic],
 Geometry{na: 6, no: 6,
  Atoms{species: 1,
   Atom{C, Z: 6, mass(au): 12.01070, maxR: 3.88840,
    HydrogenicOrbital{2pzZ1, q0: 1.0, SphericalOrbital{l: 1, R: 3.888400000000018, q0: 1.0}}
   }: 6,
  },
  maxR: 3.88840,
  SuperCell{nsc: [1 1 1],
   A=[6.840, 0.000, 0.000],
   B=[0.000, 6.460, 0.000],
   C=[0.000, 0.000, 1000.000],
  }
 }
}
```

The `maxR` is different for the `Geometry` and `Grid` objects, and the `maxR` is always the same (for a fixed `Zeff`) in the `Grid` object, regardless the fixed `R` in the `Geometry` object. It looks like is not keeping the information from the `Geometry` of the parent?

zerothi:
@sofiasanz thanks for your check, I have now added a new commit which should fix the problem, now `R` is copied explicitly for all orbitals.

sofiasanz:
Yes, I wrote a message before saying that when printing the `Grid` and `Geometry` objects the `maxR` didn't match, but I realized that the problem was because I was copying the geometry and then the information about the `maxR` was somehow lost. I'll check again with the new commits :)

sofiasanz:
Here I have the images I get with this last commit for `Zeff=3.2` and different `R` values:

fig-574-orbital-1

I start to get a little weird images in the vicinity of R~2.5 Ang for this effective nuclear charge.
I think now is working fine :)

zerothi:
Thanks for checking.

Just for completeness:
```python
import numpy as np
from sisl import HydrogenicOrbital

r = np.arange(0, 10, 0.001)
for Z in [3.2, 4, 5]:
    orb = HydrogenicOrbital(2, 1, 0, Z)
    R = np.arange(0, orb.R, 0.001)
    r = orb.radial(R)
    print("99.99  ", Z, orb.R, r[-1] / r.max()*100)

    orb = HydrogenicOrbital(2, 1, 0, Z, R=-0.999999)
    R = np.arange(0, orb.R, 0.001)
    r = orb.radial(R)
    print("99.9999", Z, orb.R, r[-1] / r.max()*100)
```
which produces
```shell
99.99   3.2 3.888400000000018 0.025069398892111586
99.9999 3.2 5.519599999999978 0.0002567943876730368
99.99   4 3.110700000000002 0.025104120475539306
99.9999 4 4.415699999999987 0.0002569771567260434
99.99   5 2.4886000000000186 0.025104120475539272
99.9999 5 3.5326000000000057 0.00025697715672604347
```
meaning that the default value produces a `R=3.888` which should be fine then. I'll merge this then. :)

Issue #577, opened 2023-06-09
Interfaces for multiple returns


zerothi:
**Describe the bug**
All `read_` routines which may return multiple entries, such as MD, or SCF data should utilize `sile_read_multiple` as a decorator.

However, there are quite a bit of files that implement the wrong interfaces:
```shell
~/codes/sisl/src/sisl/io % grep "all=" **/*.py | grep def                                                                                                                                
openmx/md.py:    def read_geometry(self, *args, all=True, **kwargs):
orca/out.py:    def read_electrons(self, all=False):
orca/out.py:    def read_energy(self, all=False):
orca/out.py:    def read_orbital_energies(self, all=False):
orca/txt.py:    def read_electrons(self, all=False):
orca/txt.py:    def read_energy(self, all=False):
orca/txt.py:    def read_geometry(self, all=False):
siesta/out.py:    def read_geometry(self, last=True, all=False):
siesta/out.py:    def read_force(self, last=True, all=False, total=False, max=False, key="siesta"):
siesta/out.py:    def read_stress(self, key="static", last=True, all=False):
siesta/out.py:    def read_moment(self, orbitals=False, quantity='S', last=True, all=False):
vasp/out.py:    def read_energy(self, all=False):
```
these should all be changed to utilise the decorator.

tfrederiksen:
This would be nice, but how do we get the `sile_read_multiple` to return optionally only the *last* entry?

zerothi:
> This would be nice, but how do we get the `sile_read_multiple` to return optionally only the _last_ entry?

we should probably extend that decorator to return a single element. For instance that can be done via the `postprocess` argument, checking for the size or such. Or simply by checking arguments internally.
The main problems is that:
1. Users may programmatically do something that coincides with a trajectory of size 1, in which case it shouldn't return a single element.
2. I am not sure if we need to add yet another flag to allow this. The thing is that probably the decorator could fix this...

zerothi:
It might also be a `single` flag or `indices=[...]|-1` where if `indices` is not a list, it unwraps the returned argument.

tfrederiksen:
Maybe we should just fix things such that negative values are not simply passed to `range(start, stop, step)`, but instead understood as first reading all entries to build the complete list, and *then* making a selection with counts from the end?

zerothi:
[...]

PR #578, opened 2023-06-11
Negative start/stop/step in sile_read_multiple

tfrederiksen:
Here an attempt to handle negative `start/stop/step` in `sile_read_multiple` as discussed in #577 . If one specifies only `start=-1` one now gets the last entry.

If a negative `start/stop/step` is provided, the code reads all entries and returns `slice(start, stop, step)`.

NB I am unsure if all corner cases are appropriately handled.

### DIFF
--- a/src/sisl/io/sile.py
+++ b/src/sisl/io/sile.py
@@ -660,7 +660,6 @@ def sile_fh_open(from_closed=False, reset=None):
 def sile_read_multiple(start: Optional[int]=None,
                        stop: Optional[int]=None,
                        step: Optional[int]=None,
-                       all: bool=False,
                        skip_call: Optional[Callable[..., Optional[Any]]]=None,
                        pre_call: Optional[Callable[..., Optional[Any]]]=None,
                        post_call: Optional[Callable[..., Optional[Any]]]=None,
@@ -675,9 +674,6 @@ def sile_read_multiple(start: Optional[int]=None,
        position of last read
     step : int, optional
        number of steps to take (if any at all).
-    all : bool, optional
-       read all entries, `start`, `stop` and `step` will not be altered
-       unless they are set to None.
     skip_call : function, optional
        read method without actual data processing.
        The skip call must something different from ``None`` if it succeeds.
@@ -722,60 +718,24 @@ def sile_read_multiple(start: Optional[int]=None,
             skip_call = wrap_func
 
         @wraps(func)
-        def multiple(*args, start=start, stop=stop, step=step, all=all, **kwargs):
+        def multiple(*args, start=start, stop=stop, step=step, **kwargs):
             self = args[0]
             inf = 100000000000
-
-            if all:
-                # parse everything, but still obey inputs
-                if start is None:
-                    start = 0
-                if stop is None:
-                    stop = inf
-                if step is None:
-                    step = 1
-            else:
-                # behave like range(stop) | range(start, stop, step)
-                # almost.
-                #  step only => start, step = 0, inf
-                #  start only => stop = start + 1
-                #  stop only => start = stop - 1
-                if step is not None:
-                    if start is None:
-                        start = 0
-                    if stop is None:
-                        stop = inf
-                elif start is None and stop is None:
-                    start = 0
-                    stop = 1
-
-                if start is None:
-                    if stop is not None:
-                        start = stop - 1
-                    else:
-                        start = 0
-                if stop is None and start > 0:
-                    # this will only occur if start is given, but not stop
-                    stop = start + 1
-                if step is None and stop is not None:
-                    # only handle cases where the requested number of
-                    # elements is longer than 1, then step should be defined,
-                    # else we resort to returning a single entity (not postprocessed)
-                    if stop - start > 1:
-                        step = 1
 
             def check_none(r):
                 if isinstance(r, tuple):
                     return reduce(lambda x, y: x and y is None, r, True)
                 return r is None
 
-            slc = None
-            if start < 0:
-                slc = slice(start, stop, step)
-            elif stop < 0:
-                slc = slice(start, stop, step)
-            if slc:
-                # read all entries and build slice at the end
+            if start is None:
+                start = 0
+            if stop is None:
+                stop = inf
+            if step is None:
+                step = 1
+
+            if start < 0 or stop < 0 or step < 0:
+                # we need to read everything and take a slice afterwards
                 def ret_func():
                     R = []
                     for _ in range(inf):
@@ -783,11 +743,11 @@ def sile_read_multiple(start: Optional[int]=None,
                         if check_none(r):
                             break
                         R.append(r)
+                    slc = slice(start, stop, step)
                     R = R[slc]
                     if len(R) == 1:
                         return R[0]
                     return postprocess(R)
-
             else:
                 # start by reading past start, this is regardless of what we
                 # are about to do
@@ -796,27 +756,22 @@ def sile_read_multiple(start: Optional[int]=None,
                     if check_none(r):
                         return r
 
-                if step is None:
-                    # read a single element and return
-                    def ret_func():
-                        return wrap_func(*args, **kwargs)
-
-                else:
-                    # read a set of geometries and return
-                    def ret_func():
-                        R = []
-                        for _ in range(start, stop, step):
-                            r = wrap_func(*args, **kwargs)
+                # read a set of geometries and return
+                def ret_func():
+                    R = []
+                    for _ in range(start, stop, step):
+                        r = wrap_func(*args, **kwargs)
+                        if check_none(r):
+                            break
+                        R.append(r)
+                        # skip the middle steps
+                        for _ in range(step-1):
+                            r = skip_call(*args, **kwargs)
                             if check_none(r):
                                 break
-                            R.append(r)
-
-                            # skip the middle steps
-                            for _ in range(step-1):
-                                r = skip_call(*args, **kwargs)
-                                if check_none(r):
-                                    return postprocess(R)
-                        return postprocess(R)
+                    if len(R) == 1:
+                        return R[0]
+                    return postprocess(R)
 
             if hasattr(self, "fh"):
                 return ret_func()
--- a/src/sisl/io/sile.py
+++ b/src/sisl/io/sile.py
@@ -724,13 +724,14 @@ def sile_read_multiple(start: Optional[int]=None,
         @wraps(func)
         def multiple(*args, start=start, stop=stop, step=step, all=all, **kwargs):
             self = args[0]
+            inf = 100000000000
 
             if all:
                 # parse everything, but still obey inputs
                 if start is None:
                     start = 0
                 if stop is None:
-                    stop = 100000000000
+                    stop = inf
                 if step is None:
                     step = 1
             else:
@@ -743,7 +744,7 @@ def sile_read_multiple(start: Optional[int]=None,
                     if start is None:
                         start = 0
                     if stop is None:
-                        stop = -1
+                        stop = inf
                 elif start is None and stop is None:
                     start = 0
                     stop = 1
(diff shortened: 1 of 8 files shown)


tfrederiksen:
A major step with the last commit is the proposal to **completely remove the `all`-keyword**. Instead reading would be solely controlled by (the equivalent of) `slice(start, stop, step)`.

For example:
* `start=-1` returns only the last entry (after reading the whole file),
* `stop=1` reads and returns only the first entry,
* `start=10, stop=11` skips the first 9 entries (using `skip_call`), reads the 10th entry, and stops reading further,
* `start=-2, step=-1` returns two entries `(step-1, step-2)` i.e., in reversed order,
* etc.

src/sisl/io/xyz.py
```diff
@@ -100,10 +100,10 @@ def _r_geometry_skip(self, *args, **kwargs):
         return na
 
     @sile_fh_open()
-    @sile_read_multiple(skip_call=_r_geometry_skip, postprocess=GeometryCollection)
+    @sile_read_multiple(stop=1, skip_call=_r_geometry_skip, postprocess=GeometryCollection)
```
tfrederiksen, line 103: It would somehow be cleaner if the default was always to return everything, unless the user provides some selection with `start/stop/step`.

src/sisl/io/sile.py
```diff
@@ -722,72 +718,44 @@ def wrap_func(*args, **kwargs):
...
+                    for _ in range(inf):
+                        r = wrap_func(*args, **kwargs)
+                        if check_none(r):
+                            break
+                        R.append(r)
+                    slc = slice(start, stop, step)
+                    R = R[slc]
+                    if len(R) == 1:
+                        return R[0]
+                    return postprocess(R)
```
zerothi, line 750: I am not sure I like this.

Consider users doing something that results in only returning a single entry at a single iteration. Now all other iterations will use the `GeometryCollection` (or postprocess call) and just one of them will use the raw output.

Wouldn't it be better to force users to do `read_()[0]` in that case? I know it is a bit annoying, but...

Otherwise we could add a `single` argument that replaces `start/stop` with the one that is given.
I.e.
```python
read_geometry(start=3, single=True) == read_geometry(stop=3, single=True)
```

src/sisl/io/sile.py
```diff
@@ -796,12 +764,13 @@ def ret_func():
...
                         R.append(r)
-
                         # skip the middle steps
                         for _ in range(step-1):
                             r = skip_call(*args, **kwargs)
                             if check_none(r):
-                                return postprocess(R)
+                                break
+                    if len(R) == 1:
+                        return R[0]
```
zerothi, line 773: same here.

zerothi (changes requested):
I largely agree on everything, just some minor comments.

It would be great if @pfebrer could have a word here, he was interested in some MD functionality. It is a bit hard this one, since some file-formats are generically used for single geometries (xyz) while others are primarily MD files (ANI). It would feel weird for end-users that `xyz`-files returns a GeometryCollection. But I think the current approach of returning the raw output could be annoying for those doing stuff *on the fly*.

We could accept it, but then it needs clear documentation.

src/sisl/io/sile.py
tfrederiksen: I kind of like the idea of `single` (opposite effect from the original `all`), but it introduces an extra keyword that can be avoided if we can tolerate single-entry cases are handled differently than multi-entry cases.

Anyways,
> ```python
> read_geometry(start=3, single=True) == read_geometry(stop=3, single=True)
> ```
seems a bit confusing to me since with `stop=3` in a `slice` would not include the fourth element. Why not

```python
read_geometry(start=3, single=True) == read_geometry(stop=4, single=True)
```
?
How should eg. `read_geometry(start=2, stop=4, single=True)` be handled?

src/sisl/io/tests/test_xsf.py
```diff
@@ -67,18 +67,18 @@ def test_axsf_geoms(sisl_tmp):
...
 
     with xsfSile(f) as s:
-        rgeoms = s.read_geometry(start=1)
+        rgeoms = s.read_geometry(start=1, stop=2)
         assert isinstance(rgeoms, Geometry)
         assert geoms[1].equal(rgeoms)
 
     with xsfSile(f) as s:
-        rgeoms = s.read_geometry(all=True)
+        rgeoms = s.read_geometry(stop=None)
```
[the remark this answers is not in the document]
tfrederiksen: No, I inserted `stop=1` as the default for `xsfSile`.

src/sisl/io/sile.py
tfrederiksen: Another possibility could be to let `step=0` mean first entry with `start` and last entry with `stop`.

src/sisl/io/sile.py
```diff
@@ -722,72 +718,44 @@
...
+                # we need to read everything and take a slice afterwards
+                def ret_func():
+                    R = []
+                    for _ in range(inf):
+                        r = wrap_func(*args, **kwargs)
+                        if check_none(r):
+                            break
+                        R.append(r)
+                    slc = slice(start, stop, step)
+                    R = R[slc]
```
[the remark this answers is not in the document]
tfrederiksen: I don't understand the error indicated here. Slicing a list should be perfectly fine, no?

pfebrer:
So I would try to make arguments mean exactly the same and have exactly the same defaults as `slice`.

As you discuss, this might be an issue because for some files it feels natural to read only one geometry, so the default should be `stop=1`. I think it's a bad idea to have different defaults for different siles because it's confusing. Maybe it is not a good idea that the function named `read_geometry` can read more than one geometry, and the user could be required to ask explicitly for the iterator. Would it be that bad? Something like:

```python

sile = sisl.get_sile("file.xyz")

# Read the first geometry (or wherever the file handle is at)
sile.read_geometry() 

# Reads a slice of geometries by first creating the iterator and then calling it.
sile.read_geometry[20:2:-1]()
# or
sile.read_geometry.multi(20, 2, -1)()

# In fact, you could use it as an iterator, which feels more natural than passing some `wrap` argument
reader = sile.read_geometry[20:2:-1]

for i, geom in reader:
    ...
```

To me it feels natural and it would avoid these problems of "arbitrary" defaults for each sile.

tfrederiksen:
@pfebrer , thanks for your inputs. For `read_geometry` I agree that it would be intuitive that the default would be to return one entry (the next), but what about `read_energy`? Here one would typically be most interested in the *last* entry.

pfebrer:
Hmm in fact I found it confusing at the beggining that there was a `read_energy` method in the `out` sile with preference for either the last/first, because it's not obvious to me that one is better than the other. In a relaxation, it might make sense to have last as the default, but if it's an MD, is there really an energy that makes more sense reading than all the others in the file?

Since it's not very well defined, I don't see it as a big problem to require the user to ask explicitly for something. I.e. `read_energy[-1]()` or something similar. 

By the way, `last`, `all` and `single` arguments are probably not necessary if we fully support slicing. For example `read_energy[-1]()` should give you just one value while `read_energy[-1:]()` should give you a list of length one. I think it would be easier to make all the reading functions in `sisl` have a consistent interface if we remove the extra keywords.

src/sisl/io/sile.py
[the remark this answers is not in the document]
pfebrer: Maybe `R = R[...]`, i.e. naming the result with the same name as the list, messes with the way the algorithm determines types, and it assumes that `R` can be anything, not just a list (although I agree it seems it would be a misjudgement :smile: ).

src/sisl/io/sile.py
zerothi: If `single` is true, then invalid input should result in an exception.
Yeah, probably only `start` should be valid input.

> Another possibility could be to let `step=0` mean first entry with `start` and last entry with `stop`.

I don't think so, that is somewhat counterintuitive since `stop` is generally not included in `range` expressions. If users want that, they should do `start=, stop=+1, step=stop-start`

src/sisl/io/tests/test_xsf.py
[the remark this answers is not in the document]
zerothi: Ah, thanks! Sorry!

zerothi:
I think @pfebrer is on to something here.

I like both the slicing approach and the `.multi` approach. Likely the `.multi` function is easier in the long run since it can allow additional expressions. But both could be supported for the temporary class.

- allow slicing
- allow `.multi` or perhaps just `.m/M/s`?
- what should we do to improve performance, perhaps have a `same` argument that will allow the first call to pass the returned value to all subsequent methods, that should remove the `Geometry` creation heavy part?
- should we allow `as_xyz` to circumvent the `Geometry` class, I kind of don't like this since the return type will then heavily vary depending on the arguments. Perhaps this is ok, but the documentation of each method will be heavy... However, when returning multiple things, forces, coordinates etc. it might be fine to return in a `PropertyDict` (or some variant of the collection to perform manipulations of the internal things.

zerothi:
I think the convention for returning single elements should be the last, for all methods. Then there will be no confusion.

src/sisl/io/sile.py
tfrederiksen: I meant `step=0` to behave as `single=True`, because that value is not permitted in a slice. We would avoid a new keyword.

zerothi: Ah, I see. Yeah, that could probably also work. I can agree to get this in, but perhaps we should then hold on the other things before we have discussed the commented things (see further down)

pfebrer:
> what should we do to improve performance, perhaps have a same argument that will allow the first call to pass the returned value to all subsequent methods, that should remove the Geometry creation heavy part?

I think it would be enough if you can use the `atoms` argument of `read_geometry`. You can read the basis first and then read all the coordinates. In fact this is how I do it when using `ASE` and it's efficient enough. But I tried to do it in the current implementation of `read_multiple` and I think it didn't work. Perhaps I was doing it wrong.

> I think the convention for returning single elements should be the last, for all methods. Then there will be no confusion.

But then you can't do what you usually suggest:

```
with sile as open_sile:
    for i in range(3):
        open_sile.read_geometry()
```

By the way I was using the word slicing but using `read_energy[...]` is in general the get item operation so `...` could be a single integer, an array of indices or whatever other more complex selection. Just as it is done with atoms selection.

zerothi:
> > what should we do to improve performance, perhaps have a same argument that will allow the first call to pass the returned value to all subsequent methods, that should remove the Geometry creation heavy part?
> 
> I think it would be enough if you can use the `atoms` argument of `read_geometry`. You can read the basis first and then read all the coordinates using the same basis for all geometries. In fact this is how I do it when using `ASE` and it's efficient enough. But I tried to do it in the current implementation of `read_multiple` and I think it didn't work. Perhaps I was doing it wrong.
> 
> > I think the convention for returning single elements should be the last, for all methods. Then there will be no confusion.
> 
> But then you can't do what you usually suggest:
> 
> ```python
> with sile as open_sile:
>     for i in range(3):
>         open_sile.read_geometry()
> ```
Ah, true, so actually it should default to only the first one!

 
> By the way I was using the word "slicing" but using `read_energy[...]` is in general the get item operation so `...` could be a single integer, an array of indices or whatever other more complex selection. Just as it is done with atoms selection.
Agreed.
I

tfrederiksen:
> Ah, true, so actually it should default to only the first one!

I think this highlights a need for different defaults for different `read_*`.

zerothi:
> > Ah, true, so actually it should default to only the first one!
> 
> I think this highlights a need for different defaults for different `read_*`.

True... But this is complicated for end-users... Hmm... We need to think carefully about the impacts in this regard.

tfrederiksen:
> > > Ah, true, so actually it should default to only the first one!
> > 
> > 
> > I think this highlights a need for different defaults for different `read_*`.
> 
> True... But this is complicated for end-users... Hmm... We need to think carefully about the impacts in this regard.

The simplest seem to be defaulting to read all entries (ie `start=stop=step=None`) and let the user select from a list/collection or actively specify `start/stop/step`.

In the example by @pfebrer , one could do
```python
with sile as open_sile:
    for i in range(3):
        open_sile.read_geometry(step=0)
```

zerothi:
Otherwise the defaults would depend on whether the file is open or not?
I.e. if the file is opened in a `with ... as sile:` then we only read the next entry, otherwise everything/last/first...

tfrederiksen:
Now I've implemented `step=0` as the default for `xyzSile`/`xsfSile`/`mdSileOpenMX`(single entry) while we keep `step=1` for `aniSileSiesta` (collection).

pfebrer:
In my opinion, `read_*` routines -where `*` is a word in singular- should not worry about skipping and they should just focus on reading the next thing available. Then we could get rid of all these extra arguments, which will be very hard to document in a way that doesn't confuse users, specially the different defaults for different methods/siles.

It seems clean and clear enough for me to specify the iterator externally (slicing, `multi`, `slice`,`apply`...), just as it is done with the BZ methods for example. If some method is very common to use with `all=True`, there could be an extra method with the plural word, i.e. `read_energy` -> `read_energies` as an alternative for read_energy[:].

zerothi:
I think @pfebrer is on to something simple and effective. I'll have a look at this, as it is vital to get this in ASAP for API stability

zerothi:
I have a somewhat working solution, see in the `sile-slicer` branch, a test is here:
```python
open("/tmp/test.file", 'w').write("""1

C   0.00000000  0.00000000  0.00000000
2

C   0.00000000  0.00000000  0.00000000
C   1.000000  0.00000000  0.00000000
3

C   0.00000000  0.00000000  0.00000000
C   1.000000  0.00000000  0.00000000
C   2.00000  0.00000000  0.00000000
""")

from sisl.io import xyzSile

sile = xyzSile("/tmp/test.file")
print(sile.read_geometry(), help(sile.read_geometry))
print(sile.read_geometry[0](), help(sile.read_geometry[0]))
print(sile.read_geometry[1]())
print(sile.read_geometry[1:]())
```
It isn't complete, but just to get an idea?
I quickly ran into problems when decoration the function, since that happens at class-creation, the function won't bind `self` and thus they won't be passed, hence the `__get__` work-around... A bit ugly IMHO.  
Other suggestions on the implementation side?

pfebrer:
I think the `__get__` implementation is fine, but probably it's better to create a new (bound) instance of `SileContainer` instead of assigning the object instance (`self`) to the shared `SileContainer`. 

After all, this is how functions turn into methods so there's nothing very strange going on. Unless you can modify the `__getitem__` method of the function directly to implement a subscriptable method, I'd say this would be the best way to do it.

pfebrer:
By the way, what is the reason behind the name `SileContainer` ? :sweat_smile:

zerothi:
Lets continue the discussion in #584.

Perhaps this MR should be closed, and then focus should be on the other branch?

PR #581, opened 2023-06-12
undo change of filenames

tfrederiksen:
Filenames in the orca tests were (accidentally?) changed from `.output` to `.out` in da3c29b422164a9aaf8fa3aba7a91ff29a9f0e41, not consistent with the `sisl-files` repo. This branch reverts the change.

### DIFF
--- a/src/sisl/io/orca/tests/test_out.py
+++ b/src/sisl/io/orca/tests/test_out.py
@@ -14,14 +14,14 @@ _dir = osp.join('sisl', 'io', 'orca')
 
 
 def test_tags(sisl_files):
-    f = sisl_files(_dir, 'molecule.out')
+    f = sisl_files(_dir, 'molecule.output')
     out = outSileORCA(f)
     assert out.na == 2
     assert out.no == 62
     assert out.completed()
 
 def test_read_electrons(sisl_files):
-    f = sisl_files(_dir, 'molecule.out')
+    f = sisl_files(_dir, 'molecule.output')
     out = outSileORCA(f)
     N = out.read_electrons(all=True)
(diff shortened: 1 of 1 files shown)


zerothi:
Ah, I should have caught that.
We should probably retain one of them since the code should be able to handle various codes output by testing their `read_` routines.

My main motivation is that future codes will in most cases have an `out` file type. And in that case sisl should be able to handle them by figuring out which one is the file-type. Currently it does this by running some of the `read_*` methods of the sile and if they fail, so will the `Sile` type, removing it from the eligible siles.

I.e. here it is needed to ensure that orca siles are in fact returned. And we should strive to _not_ use specific file extensions unless the code it-self explicitly does so. Will `orca` always write these `.output` files, or is it the user who pipes to `output` files?

tfrederiksen:
Wouldn't be more conventional to call these files `stdout`?

zerothi (approved):

tfrederiksen:
Note that the vasp OUTCAR is *not* stdout, but a fixed file it always produces. Additionally, VASP also generates stdout.

zerothi:
Yes, thanks for that. From my understanding the outcar and stdout is somewhat similar, no?

tfrederiksen:
> Yes, thanks for that. From my understanding the outcar and stdout is somewhat similar, no?

Not really, stdout only gives some brief info on the iterations (scf energies, optimization steps, etc).

tfrederiksen:
I have added some stdout examples in https://github.com/zerothi/sisl-files/pull/7.

zerothi:
Great, probably some cleanup in that file then ;)

Issue #584, opened 2023-06-15
Slicing read_* methods


zerothi:
In my opinion, `read_*` routines -where `*` is a word in singular- should not worry about skipping and they should just focus on reading the next thing available. Then we could get rid of all these extra arguments, which will be very hard to document in a way that doesn't confuse users, specially the different defaults for different methods/siles.

It seems clean and clear enough for me to specify the iterator externally (slicing, `multi`, `slice`,`apply`...), just as it is done with the BZ methods for example. If some method is very common to use with `all=True`, there could be an extra method with the plural word, i.e. `read_energy` -> `read_energies` as an alternative for `read_energy[:]`.

_Originally posted by @pfebrer in https://github.com/zerothi/sisl/issues/578#issuecomment-1591358548_

zerothi:
I have a somewhat working solution, see in the `sile-slicer` branch, a test is here:
```python
open("/tmp/test.file", 'w').write("""1

C   0.00000000  0.00000000  0.00000000
2

C   0.00000000  0.00000000  0.00000000
C   1.000000  0.00000000  0.00000000
3

C   0.00000000  0.00000000  0.00000000
C   1.000000  0.00000000  0.00000000
C   2.00000  0.00000000  0.00000000
""")

from sisl.io import xyzSile

sile = xyzSile("/tmp/test.file")
print(sile.read_geometry(), help(sile.read_geometry))
print(sile.read_geometry[0](), help(sile.read_geometry[0]))
print(sile.read_geometry[1]())
print(sile.read_geometry[1:]())
```
It isn't complete, but just to get an idea?
I quickly ran into problems when decoration the function, since that happens at class-creation, the function won't bind `self` and thus they won't be passed, hence the `__get__` work-around... A bit ugly IMHO.  
Other suggestions on the implementation side?

_Originally posted by @zerothi in https://github.com/zerothi/sisl/issues/578#issuecomment-1591943394_

zerothi:
@tfrederiksen @pfebrer lets continue the discussion here.

Please see the updated branch 584-sile-slicer, it contains more functionality and patches the documentation.

@pfebrer did the changes correspond to your remarks?

pfebrer:
It is a bit strange that `SileSlicer.__getitem__` stores the key and returns itself, no? This allows things like:

```python
sile = sisl.get_sile("file.xyz")

sile.read_geometry[2][:3][2:4:5][-1] # Will only care about the last selector.
```

[...]

PR #585, opened 2023-06-15
WIP: `translate2uc` for sparse matrices with associated geometries.

pfebrer:
As discussed in #584, this PR creates a function that translates all atoms of a sparse matrix to the unit cell, updating both the coordinates and the indices of the matrix elements.

The current state of the code contains only the simplest possible behavior, implemented in the most straightforward way I could think of. And only for `SparseOrbital`, not yet for `SparseAtom`. This is so that you can easily give it a look and see if there is some conceptual mistake. Also, I am aware that this is not the most time and memory efficient implementation possible, but I can optimize it later (suggestions accepted :)). Also I don't know what is the approppiate ratio `optimization/readability` for this.

I thought about it, and I see no way of using the `translate_columns` without making the code more complex and unreadable, because the translation of each matrix element depends on both the columns and rows. So, a column translates differently depending on which row are we.

Demo
-----
I did a very simple test with a hydrogen molecule, which I generated like:

```python
import sisl
from ase.build import molecule
import numpy as np
import plotly.express as px

geom = sisl.Geometry.new(molecule("H2"))

geom.cell[0,0] = 10
geom.cell[1,1] = 10
geom.cell[2,2] = 10

geom.xyz += 5
```

Then I ran a SIESTA calculation for this molecule as is and the same molecule with the first atom displaced one cell in X. I used `ForceAuxCell t` and defaults for the rest of inputs.

I read the DMs and plot them:

```python
uc_dm = sisl.get_sile("original/RUN.fdf").read_density_matrix()
sc_dm = sisl.get_sile("displaced/RUN.fdf").read_density_matrix()

px.imshow(uc_dm.tocsr().toarray()).show()
px.imshow(sc_dm.tocsr().toarray()).show()
```
newplot - 2023-06-16T004432 859

newplot - 2023-06-16T004447 048

Then, I translate the second DM to the unit cell:

```python
moved_dm = sc_dm.translate2uc()

px.imshow(moved_dm.tocsr().toarray()).show()
```
newplot - 2023-06-16T004909 077

All the heavy lifting is done by the `translate_atoms_sc` method, which receives the cell translations of all atoms. So you can displace the atoms differently and it also works. Here I displace the atom that was in the unit cell to the same cell as the displaced atom:

```python
moved_dm = sc_dm.translate_atoms_sc([[0, 0, 0], [1,0,0]])

moved_dm.plot(axes="xz").show()
px.imshow(moved_dm.tocsr().toarray()).show()
```
newplot - 2023-06-16T005212 285
newplot - 2023-06-16T005224 075


pfebrer:
One thing that I'm not sure how to do is to re-sort the column indices (that is if there is already a method that does that). Because I modified them but I left them unsorted. `scipy` doesn't have a problem with it, but probably is better to sort them?

src/sisl/sparse_geometry.py
```diff
@@ -3006,6 +3006,108 @@ def assert_unique(old, new):
         out._csr = csr
         out._geometry = geom
         return out
+    
+    def translate2uc(self):
```
zerothi, line 3010: Probably the interface should be equivalent to the `Geometry.translate2uc`, it would also allow selection of `atoms`.

src/sisl/sparse_geometry.py
```diff
@@ -3006,6 +3006,108 @@ def assert_unique(old, new):
...
+
+        # Get the quotient and remainder of dividing by 1.
+        # The quotient is the supercell index, the remainder is the coordinates in the unit cell.
+        current_sc, uc_fcoords = np.divmod(fxyz, 1)
+        
+        # Simply translate the atoms to move all atoms to the unit cell. That is, all atoms
+        # should be moved to supercell (0,0,0).
+        return self.translate_atoms_sc(-current_sc, geometry_coords=uc_fcoords @ self.cell)
+
+    def translate_atoms_sc(self, sc_translations, geometry_coords=None):
```
zerothi, line 3033: Hmm... I don't particularly like this method. Couldn't we start with the `translate2uc` with the atoms specification, and if there is a need for the `translate_atoms_sc` then reconsider?

I see this method as being the equivalent of a `translate` method, since the `geometry_coords` can change the coords at will.

zerothi (changes requested):
Lets start with the first method, and work from there.

src/sisl/sparse_geometry.py
pfebrer: Yes :+1: This was just to make it as simple as possible for now so that the code was easier to review

src/sisl/sparse_geometry.py
pfebrer: I will remove the `geometry_coords` argument, it was just to avoid recomputing the coordinates in the particular case of `translate2uc` but now I see that it's exactly the same amount of computation whether I precompute or compute inside the method.

I can also make the method private for now. But I thought it was a nice separation of functionality since the only specificity of `translate2uc` is to know how much unit cells is each atom translated. And since the translation of the matrix elements is not trivial to implement I figured it would be nice to keep it like this.

pfebrer:
See now, you still don't like `translate_atoms_sc`? Not the name, but the concept :sweat_smile:

pfebrer:
This is now fully functional, with tests that cover everything :)

After you review I can squash everything into a single commit.

zerothi:
@pfebrer could you have a look? I would think this is fine (made the `translate_atoms_sc` a private function).

zerothi:
Also, once in you should check that this works faster due to cb7e3cb9a664e2c37d224471b1181069f577bbff

zerothi:
Hold on, something is missing, a test for sparse orbital

pfebrer:
Looks good to me :+1:

PR #586, opened 2023-06-16
Slicing IO read_ routines

### LINKED ISSUE #584: Slicing read_* methods

zerothi:
- [x] Closes #584
 - [x] Tests added
     somewhat, more documentation would be great.

### DIFF
--- a/src/sisl/io/_multiple.py
+++ b/src/sisl/io/_multiple.py
@@ -107,7 +107,8 @@ class SileSlicer:
                 retval = func(obj, *args, **kwargs)
 
             if len(retvals) == start:
-                raise RuntimeError(f"{obj.__class__.__name__}.{func.__name__} could not read any entries?")
+                # none has been found
+                return None
 
         # ensure the next call won't use this key
         # This will prohibit the use
--- a/src/sisl/io/_multiple.py
+++ b/src/sisl/io/_multiple.py
@@ -152,17 +152,32 @@ class SileBound:
 
     def _update_doc(self):
         # Override name to display slice handling in help
-        self.__name__ = f"{self.__name__}[...]"
+        default_slice = self.default_slice
+        if self.default_slice is None:
+            default_slice = 0
+
+        self.__name__ = f"{self.__name__}[...|{default_slice!r}]"
         name = self.__func__.__name__
         try:
             doc = self.__doc__
         except AttributeError:
             doc = ""
 
+        if default_slice == 0:
+            default_slice = "the first"
+        elif default_slice == -1:
+            default_slice = "the last"
+        elif default_slice == slice(None):
+            default_slice = "all"
+        else:
+            default_slice = self.default_slice
+
         doc = dedent(f"""{doc}
         Notes
         -----
-        This method enables slicing for handling multiple values (see [...])
+        This method defaults to return {default_slice} item(s).
+
+        This method enables slicing for handling multiple values (see [...|default]).
 
         This is an optional handler enabling returning multiple elements if {name}
         allows this.
--- a/src/sisl/io/_multiple.py
+++ b/src/sisl/io/_multiple.py
@@ -214,6 +214,14 @@ class SileBound:
                 **self.kwargs
         )
 
+    @property
+    def next(self):
+        return self[0]
+
+    @property
+    def last(self):
+        return self[-1]
+
     def __get__(self, obj, objtype=None):
         # I have no idea why this is needed, if I don't have this, then
         # the help(read_geometry) returns without the correct interface,
--- a/src/sisl/io/_multiple.py
+++ b/src/sisl/io/_multiple.py
@@ -38,6 +38,9 @@ class SileSlicer:
             self.skip_func = func
         else:
             self.skip_func = skip_func
+        if postprocess is None:
+            def postprocess(ret):
+                return ret
         self.postprocess = postprocess
         # this is already sliced, sub-slicing shouldn't work (at least for now)
         update_wrapper(self, func)
--- a/src/sisl/io/_multiple.py
+++ b/src/sisl/io/_multiple.py
@@ -1,0 +1,239 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+from functools import (
+        update_wrapper,
+        reduce
+)
+from textwrap import dedent
+from typing import Any, Callable, Optional, Type
+from numbers import Integral
+
+Func = Callable[..., Optional[Any]]
+
+
+class SileSlicer:
+    """ Handling io-methods in sliced behaviour for multiple returns
+
+    This class handler can expose a slicing behavior of the function
+    that it applies to.
+
+    The idea is to attach a function/method to this class and
+    let this perform the function at hand for slicing behaviour
+    etc.
+    """
+    def __init__(self,
+                 obj: Type[Any],
+                 func: Func,
+                 key: Type[Any],
+                 *,
+                 skip_func: Optional[Func]=None,
+                 postprocess: Optional[Callable[..., Any]]=None):
+        # this makes it work like a function bound to an instance (func.__self__
+        # works for instances)
+        self.__self__ = obj
+        self.__func__ = func
+        self.key = key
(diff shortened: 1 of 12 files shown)


src/sisl/io/_multiple.py
```diff
@@ -0,0 +1,236 @@
...
+        # this makes it work like a function bound to an instance (func.__self__
+        # works for instances)
+        self.__self__ = obj
+        self.__func__ = func
+        self.key = key
+        if skip_func is None:
+            self.skip_func = func
+        else:
+            self.skip_func = skip_func
+        self.postprocess = postprocess
```
tfrederiksen, line 41: What happens if `postprocess is None`? Could `list` be a default?

src/sisl/io/siesta/tests/test_ani.py
```diff
@@ -32,41 +32,41 @@ def test_ani(sisl_tmp):
...
-    g = a.read_geometry(start=1, all=False)
+    g = a.read_geometry[1]()
     assert g.na == 2
 
-    g = a.read_geometry(start=1, stop=3, step=1)
+    g = a.read_geometry[1:3:1]()
     assert g[1].na == 3 and len(g) == 2
 
-    g = a.read_geometry(start=1, stop=3, step=2, all=True)
+    g = a.read_geometry[1:3:2]()
```
tfrederiksen, line 69: What about adding some checks with negative indices?

tfrederiksen (commented):
I played a bit around with this and it looks really cool. Well done!

src/sisl/io/_multiple.py
zerothi: It then returns as the decorated routine returns the values (in this case a list of the return values).
I see there was no default, now it is `lambda x: x` (or the equivalent).

So yes, the default is a list.

src/sisl/io/siesta/tests/test_ani.py
zerothi: Added a new test. (I'll keep from pushing, I am reaching the bandwidth limit for the lfs repo)

zerothi:
I don't see a reason to not use this everywhere, i.e. read_energy etc?

Would there be objections to this?

tfrederiksen:
> I don't see a reason to not use this everywhere, i.e. read_energy etc?
> 
> Would there be objections to this?

I agree, no objections. I already tested this out on `stdoutSileVASP.read_energy()`. However, I tend to think it may cause confusion if we change the behaviour everywhere to return the *first* entry instead of the *last* (default up to now).

Maybe raise a warning for a while to make users aware that `read_energy[-1]()` is needed to get the last entry (that I suppose is the typical case)?

zerothi:
Where should we default to return the last? Everywhere, some of @pfebrer comments related to MD are valid.
However, I assume that those doing MD are aware of their needs and thus does require to return all.
I don't know if *all* should default to return the last item?

pfebrer:
Wouldn't the default slice return the confusion of the arguments we have removed. I advocate for no defaults on this, then "legacy" code will be more likely to keep doing the same thing in the future. If in the future we decide that another default is better, it will be very difficult to debug changes in the defaults in already written code.

I would just issue a warning for some time as @tfrederiksen was suggesting (perhaps keep the current API with a deprecation warning).

Also in my case I would spend more time rechecking the documentation for the defaults of each sile than simply specifying what I want 😅

pfebrer:
By the way I don't know if it would be a good idea to use slices for cases where there is more than one dimension. E.g. in `outSileSiesta.read_scf` one could do `read_scf[:, -1]` meaning all MD steps, but just the last scf step. 

It looks nice but it's a bit hard to know if it makes sense implementationwise because it would no longer be just a simple external iterator on the reading function.

pfebrer:
> Yeah, you are probably right. But if we then just are choosing a sane default that applies everywhere? There are pros and cons on choosing first/last, which I think are the only viable ones.

😂
My point is no default, and let the read function read at whatever position the file handle is.

zerothi:
> > Yeah, you are probably right. But if we then just are choosing a sane default that applies everywhere? There are pros and cons on choosing first/last, which I think are the only viable ones.
> 
> joy My point is no default, and let the read function read at whatever position the file handle is.

But a no default is still a default (`[0]`). So I think that is still up for discussion. ;)

I am probably pro the last, since anything but MD are generally most interested in the last (say converged system).

zerothi:
> By the way I don't know if it would be a good idea to use slices for cases where there is more than one dimension. E.g. in `outSileSiesta.read_scf` one could do `read_scf[:, -1]` meaning all MD steps, but just the last scf step.
> 
> It looks nice but it's a bit hard to know if it makes sense implementationwise because it would no longer be just a simple external iterator on the reading function.

This is a fine idea, I think it could be implemented easily as the subsequent tuple keys are simply applied to the returned arrays. Lets see...

pfebrer:
> But a no default is still a default ([0]). So I think that is still up for discussion. ;)

Then what would

```python
with sile as f:
    for i in range(3):
        f.read_geometry()
```

do? Read three times the first or the last geometry?

zerothi:
> > But a no default is still a default ([0]). So I think that is still up for discussion. ;)
> 
> Then what would
> 
> ```python
> with sile as f:
>     for i in range(3):
>         f.read_geometry()
> ```
> 
> do? Read three times the first or the last geometry?

ah yeah, there was this annoyance... damn... why is this so annoying...

Would it be too problematic to check if the file was open? Would that be annoying for end-users?

.e.g.
```python
last = f.read_geometry()
# whereas
with f:
     next = f.read_geometry()
```
?

pfebrer:
Would it be so bad to have a `read_last_geometry()` method if it makes sense for some `Sile`?

tfrederiksen:
> Would it be so bad to have a `read_last_geometry()` method if it makes sense for some `Sile`?

I think this is a good suggestion, making it explicit for the user what is looked up. In particular the warning could suggest users to adapt `read_last_energy()` in their scripts.

pfebrer:
> I am pretty sure that would make sense for all siles that uses the SileBinder

Yeah maybe 😅

I guess if `read_*` functions only parse lines it will be also easier to combine them. For example you could implement something to read geometries and energies in the same pass through the file.

zerothi:
I have thought about it, and I have now added this, what do you think?
```python
read_geometry.last(*args, **kwargs)
read_geometry.next(*args, **kwargs)
```
to explicitly request the next item, and the last item. Remark that `first` would have been misleading as in the case of @pfebrer's case above.

pfebrer:
Sounds good!

But what does `read_geometry()` do?

zerothi:
> Sounds good!
> 
> But what does `read_geometry()` do?

same as `next` so I don't know if that is needed? but more for clarity, as you suggested.

pfebrer:
Hmm it is very difficult to know what users would think about it. But it doesn't seem it could be harmful in the future to maintain this two properties (fingers crossed haha) so we could have them both.

zerothi:
@tfrederiksen you might be interested in the latest commit which changes the `orca` output to follow this method. I couldn't really safe the `all` argument, but I hope this is fine. In the end when we get to a release it will be 0.14. Let me know your thoughts.

Probably this should go in ASAP, and then the other siles will come as required.

pfebrer:
Could `read_energy` and similars return a numpy array by default instead of a `Collection`?

zerothi:
> Could `read_energy` and similars return a numpy array by default instead of a `Collection`?

That would be hard for end users, then the users need to read which indices corresponds to which energies, that was why the property dict was used.  
It would be nice if they all allow collection or dataframes.

tfrederiksen (approved):
I think this looks great. After merge I will revisit #573.

Issue #509, opened 2022-11-09
Group common data in Siles into a property


zerothi:
**Describe the feature**

In #500 and in the tbtncSileTBtrans we added the properties `no` and `na` etc. I think these should be removed since they are not sile specific but should be queried in some other way.

Some ideas:
- some property that returns a dictionary or namedtuple
    - this property could be named `info`
    - `settings` (so other things could also be added, like options etc.)
    - ?

The main reason is that the complexity and handling of many properties that are more geometry specific, or setting specific seems to bloat the classes.

Comments?

tfrederiksen:
How do you imagine that one should access the `no`/`na` quantities within a sile? With calls to `self.info["no"]`? And how/where should the dictionary be created?

tfrederiksen:
OK, I see. So a generic `PropertyDict` to all siles where quantities/settings/info should be registered?

zerothi:
That would be my idea, it reduces the amount of methods, and the documentation for that entry should just be sufficient :)

PR #591, opened 2023-06-21
WIP: Grouping data into Sile.info

tfrederiksen:
Is this a way towards addressing #509?


zerothi:
Thanks for this!

I have been trying to wrap by head around this problem (basically since you created the PR).
I think that something a bit more dynamic could be used. I have been playing with some things, I'll make a PR and request your feedback. It is more complex, but will also allow a more dynamic approach for end-users.

tfrederiksen:
Superseded by #631

PR #594, opened 2023-07-07
Fix missing import in siesta_grid.

juijan:
This import is needed when the `read_grid` routine receives an array of floats for the kwarg `spin`.

`grid_reduce_indices` was called in line 132 (now 133), but never imported causing a NameError.

This should fix it.


zerothi (approved):
Thanks!!!

PR #596, opened 2023-07-10
netCDF4 is now an optional dependency

### LINKED ISSUE #595: Remove required dependency of netCDF4
**Describe the issue**
Should we remove required dependency of netCDF4?

Currently, it isn't forced in the code, so it *could* be disabled.

The latest commit has added it to all optional dependency additions.

Would there be any objections to this?
The thing is that sisl without io usage does not require netCDF, so it would still be useful for users to test things with sisl like this.

zerothi:
An import error will be raised with explanation
on how to mitigate the problem.

 - [x] Closes #595 
 - [x] Tests added

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,7 @@
 numpy>=1.13
 scipy>=1.5.0
-netCDF4
 pyparsing>=1.5.7
+netCDF4
 matplotlib
 ipykernel
 nbsphinx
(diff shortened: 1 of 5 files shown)


docs/requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy>=1.13
 scipy>=1.5.0
-netCDF4
 pyparsing>=1.5.7
+netCDF4
```
tfrederiksen, line 4: Why is `netCDF4` listed here? I also notice that it is still present in `../requirements.txt`.

zerothi: `requirements.txt` was a missing delete. But docs/ is necessary for building the documentation. There I just use everything.

PR #598, opened 2023-07-13
maint: Adapt to breaking changes in scipy.sparse.isspmatrix

pfebrer:
In this PR: https://github.com/scipy/scipy/pull/18528 scipy introduced a breaking change which has already been released in `1.11`. 

`isspmatrix` no longer returns `True` if you pass a sparse array (instead of a sparse matrix).

To keep the old behavior, we should use `scipy.sparse.issparse` instead. `isspmatrix_fmt` (e.g. `isspmatrix_csr`) becomes more cumbersome because there is no equivalent function in scipy to check for sparse arrays as well, and you have to go `issparse(matrix) and matrix.format == "fmt"`.

These changes keep the old behavior of sisl even with `scipy >= 1.11` and don't introduce backwards incompatibilities, so no need to bump the `scipy` dependency (I think).

### DIFF
--- a/src/sisl/io/tbtrans/tbt.py
+++ b/src/sisl/io/tbtrans/tbt.py
@@ -16,7 +16,7 @@ import numpy as np
 ndarray = np.ndarray
 
 # The sparse matrix for the orbital/bond currents
-from scipy.sparse import SparseEfficiencyWarning, csr_matrix, isspmatrix_csr
+from scipy.sparse import SparseEfficiencyWarning, csr_matrix, issparse
 
 import sisl._array as _a
 from sisl import Atoms, Geometry, constant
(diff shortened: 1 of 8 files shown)


zerothi:
this won't work for older scipy's, you need to check scipy version before doing this, preferably importing into a unified name depending on version.

zerothi:
Also, all of sisl still uses sparse matrices, but I agree it would be nice if users could do sparse arrays.

zerothi:
Note this could be collected into a `sisl._scipy_imports` package to make it simpler.

pfebrer:
I have an application that uses sisl with arrays, that's how I noticed the bug.

I checked up to the minimum version that sisl supports now (1.5.0), and `issparse` was already there.

zerothi:
Sorry, I missed it!

zerothi (approved):
Thanks!

pfebrer:
Perfect :)

pfebrer:
By the way I think now the `isspmatrix_csr`, coo, etc... in `sisl.sparse` are unused. Just in case you want to remove them

zerothi:
yeah, they should just be removed.

zerothi:
These got corrected, now a little more to go and we are done! :)

PR #620, opened 2023-09-30
bug: register data classes as plotables

pfebrer:
This one is not that important, you could not `.plot()` on the data classes.

Now this:

```python
from sisl.viz import BandsData

data = BandsData(...)

data.plot.bands()
```

although it is not shown in the tutorials.

I have other little tweak ideas for improvement of `sisl.viz`, but they can wait, no need to rush :)

### DIFF
--- a/src/sisl/viz/_plotables.py
+++ b/src/sisl/viz/_plotables.py
@@ -187,7 +187,15 @@ def register_data_source(
     data_source_init_kwargs: dict = {},
     **kwargs
 ):
+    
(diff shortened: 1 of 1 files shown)


zerothi:
I won't do anything before late tonight, or perhaps even first tomorrow evening. Feel free to make any prs!

pfebrer:
It's better not to change anything now, I might introduce some bugs or take wrong decisions if I do it fast :) It is good enough as it is for now :+1:

PR #621, opened 2023-09-30
Cleaning tutorial notebooks.

pfebrer:
This is the last one hehe

I added a `normalize` function and used it to normalize sizes of atoms so that we don't need to multiply the PDOS by some strange factor to set it as atoms sizes (which is less prone to visualization bugs).

The second commit changes the siesta tutorial to use sisl.viz to plot PDOS and fatbands. I don't know if you want it or you prefer to show the technical parts of doing the analysis in this tutorial. Feel free to discard that commit :)


zerothi:
Missing a clean of the workbook

zerothi:
I think it makes sense to transition to the viz module. I think everything should be in that format. :)

Issue #625, opened 2023-10-09
Default `Lattice.nsc` and relation to periodic/nonperiodic directions


tfrederiksen:
I've noticed that the default `Lattice.nsc = [1, 1, 1]` is obtained when reading a geometry from file (XV, CONTCAR, fdf, etc). However, this prevents a function like `.translate2uc()` to take effect unless one first changes `nsc[i] > 1` or explicitly sets `axes=True` (how I came across this issue).

_Suggestion_: Perhaps `nsc[i] = 0` should be allowed to denote a nonperiodic direction, while a value of `1` would mean periodicity (but without explicit reference to neighbour cells)? This would be consistent with the `pbc`-parameter written to the `xyz`-header (obtained from `pbc = ['T' if n else 'F' for n in nsc]`).

With this definition, we could simply redefine the default `axes=None` in `translate2uc` to: 
```python
-            axes = (self.lattice.nsc > 1).nonzero()[0]
+            axes = (self.lattice.nsc >= 1).nonzero()[0]
```

tfrederiksen:
Thanks for reminding me of #553, I agree these issues are related. I guess the concept of `nsc` is only strictly meaningful for LCAO-type calculations, whereas a geometry is a more general concept.

zerothi:
> Thanks for reminding me of #553, I agree these issues are related. I guess the concept of `nsc` is only strictly meaningful for LCAO-type calculations, whereas a geometry is a more general concept.

Indeed, and that turns out to be difficult to generalize...

tfrederiksen:
Would it be possible to associate `nsc` only with `sparse_geometry`?

zerothi:
But I think this is more a property of the geometry and interaction ranges. For instance it may also be used for neighbour finding (which need not be used in the context of a sparse matrix).

It seems to me to belong to the geometry in conjunction with the lattice.

zerothi:
So the question is whether we need some kind of `Coordinate` class #258 (somewhat related).

Issue #553, opened 2023-03-15
nsc should not belong to Lattice


zerothi:
It is weird because `nsc` is not really a property of the lattice, no? It only makes sense in a combination of a lattice, some sites, and something that defines the range of the sites. Perhaps the lattice information should be separated from the auxiliary cell information, which could be an attribute -`aux_cell`- of `Geometry` or whatever other class where it makes sense.

So `lattice` would hold `cell` and `pbc`, and `aux_cell` would hold `nsc`.

_Originally posted by @pfebrer in https://github.com/zerothi/sisl/issues/550#issuecomment-1469775585_

zerothi:
I agree, it would be nice to put the supercells into the placeholder that really uses them.

In this case it would be the `Geometry` class.

Considerations for alternate names,
- `neighbour_cells`
- `supercells`
- `aux_cells`
- others?
I kind of have a hard time going around the plural `s`, `neighbour_cell` only signifies one direction in this case. @tfrederiksen please chip in here as well.

pfebrer:
I think of it as just a bigger cell, so I don't have a problem with removing the plural 😅

tfrederiksen:
What about `supercell_shape`? Or variants hereof:  `supercell_size`, `supercell_range`...

zerothi:
> What about `supercell_shape`? Or variants hereof: `supercell_size`, `supercell_range`...

Kind of like `supercell_size/range/extent`, has meaning and clarity.... But still a bit long... Hmm...

zerothi:
So should this be a class in it-self, probably that would be the best thing. Since `Geometry`, `Grid` should use this feature.

tfrederiksen:
Could `Supercell` be a child of `Lattice` with the extra info about a `supercell_size` etc?

zerothi:
As @pfebrer pointed out this would still be a bit ambiguous. The supercell is intrinsically connected to some ranged property (orbital in this case) + the lattice vectors. So it is not only due to the Lattice, it is some combination of the Atom + coordinates + Lattice that results in the supercell. At least conceptually. Users may then overwrite the intrinsic supercells.

The idea that `Lattice` has `pbc` would be useful, but it might also be a bit problematic downstream if `pbc[i] == True` but `nsc[i] == 1`, what to do then?

pfebrer:
[...]

PR #629, opened 2023-10-20
Lattice is now owning the boundary conditions

### LINKED ISSUE #626: `Lattice.boundary_cond` boundary conditions for the lattice
**Describe the feature**

this follows the discussion of #553 #625  which discusses the move of `nsc` to another object.

Lets here focus on a more simple thing. Namely how to handle boundary conditions.

It seems to me that the boundary conditions are a more fundamental property of the lattice, and as such belongs in that object.

- currently the boundary conditions are only located in the `Grid` class, but I think we can argue that the `Lattice` is the more natural
  place. Whether or not a grid is used, one can still associate a boundary condition with the `Lattice`
- according to the above I think we have agreed that `nsc` should not be used to predict `bc`. In fact, one might argue that non-periodic boundary conditions when `nsc[axis]>1` is wrong. And this is something that could be checked otherwise.

My proposal:
- move the `bc` arguments of `Grid` into `Lattice` so that the lattice is responsible for the boundary conditions.
- instead of only having `pbc`, we should be more generic, and name it `boundary_cond`/`boundary_condition`. This is very easy to read and shouldn't pose big problems. The current allowed `PERIODIC`, `NEUMANN`, `DIRICHLET`, `OPEN` which should be re-used here as well.
- `pbc` can be a shorthand for checking `boundary_cond == PERIODIC` at its simplest

I think this goes nicely in line with the above points, any comments? @tfrederiksen @pfebrer

zerothi:
This should be a first step towards moving the
nsc attribute from Lattice to somewhere more
appropriate.

This also makes the boundary conditions in the
Grid more manageable since it is the lattice which keeps the information.

Simplified reading xyz files since the boundary
conditions are now explicit.

@tfrederiksen and @pfebrer a review please. :)

 - [x] Closes #626 #625 
 - [x] Tests added

A small snippet showing its use:
```python
from sisl import BoundaryCondition as BC, Lattice

print("Available boundary conditions:")
for b in BC:
    print("  ", b)

# Define a lattice
print(Lattice(1, boundary_condition=BC.PERIODIC))
print(Lattice(1, boundary_condition=[BC.DIRICHLET, BC.NEUMANN, BC.PERIODIC]))
print(Lattice(1, boundary_condition=[BC.DIRICHLET, [BC.NEUMANN, BC.DIRICHLET], BC.PERIODIC]))


print("Setting new boundary conditions:")
latt = Lattice(1, boundary_condition=BC.PERIODIC)
print(latt)
latt.set_boundary_condition(c=[BC.NEUMANN, BC.DIRICHLET])
print(latt)
latt.set_boundary_condition(BC.PERIODIC)

print("Printing stuff")
print(latt.boundary_condition)
print(latt.pbc)

# doing
latt.boundary_condition = ...
# is equivalent to latt.set_boundary_condition
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ we hit release version 1.0.0.
   debugging.
 - marked all `toSphere|toEllipsoid|...` as deprecated
 
+### Changed
+- `Lattice` now holds the boundary conditions (not `Grid`), see #626
+
 
 ## [0.14.2] - 2023-10-04
 
--- a/docs/api/basic.rst
+++ b/docs/api/basic.rst
@@ -30,6 +30,7 @@ Simple objects
    Atoms
    Geometry
    Lattice
+   BoundaryCondition
    Grid
 
 
--- a/src/sisl/geom/_common.py
+++ b/src/sisl/geom/_common.py
@@ -10,15 +10,21 @@ def geometry_define_nsc(geometry, periodic=(True, True, True)):
     """Define the number of supercells for a geometry based on the periodicity """
     if np.all(geometry.maxR(True) > 0.):
         geometry.optimize_nsc()
-        if not periodic[0]:
-            geometry.set_nsc(a=1)
-        if not periodic[1]:
-            geometry.set_nsc(b=1)
-        if not periodic[2]:
-            geometry.set_nsc(c=1)
+        for i, d, per in zip(range(3), "abc", periodic):
+            if per:
+                geometry.lattice.set_boundary_condition(**{d: "Periodic"})
+            else:
+                geometry.set_nsc(**{d: 1})
     else:
         nsc = [3 if p else 1 for p in periodic]
         geometry.set_nsc(nsc)
+
+    for name, per in zip("abc", periodic):
+        if per:
+            per = "Periodic"
+        else:
+            per = "Unknown"
+        geometry.lattice.set_boundary_condition(**{name: per})
 
 
 def geometry2uc(geometry, dx=1e-8):
--- a/src/sisl/grid.py
+++ b/src/sisl/grid.py
@@ -1,6 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
+import logging
 from math import pi
 from numbers import Integral, Real
 
@@ -14,8 +15,8 @@ from . import _array as _a
 from ._help import dtype_complex_to_real, wrap_filterwarnings
 from ._internal import set_module
 from .geometry import Geometry
-from .lattice import LatticeChild
-from .messages import SislError, deprecate_argument
+from .lattice import BoundaryCondition, LatticeChild
+from .messages import SislError, deprecate_argument, deprecation
 from .shape import Shape
 from .utils import (
     cmd,
@@ -29,6 +30,10 @@ from .utils import (
 from .utils.mathematics import fnorm
 
 __all__ = ['Grid', 'sgrid']
+
+_log = logging.getLogger("sisl")
+_log.info(f"adding logger: {__name__}")
+_log = logging.getLogger(__name__)
 
 
 @set_module("sisl")
@@ -76,26 +81,29 @@ class Grid(LatticeChild):
     """
 
     #: Constant for defining a periodic boundary condition
-    PERIODIC = 1
+    PERIODIC = BoundaryCondition.PERIODIC
     #: Constant for defining a Neumann boundary condition
-    NEUMANN = 2
+    NEUMANN = BoundaryCondition.NEUMANN
     #: Constant for defining a Dirichlet boundary condition
-    DIRICHLET = 3
+    DIRICHLET = BoundaryCondition.DIRICHLET
     #: Constant for defining an open boundary condition
-    OPEN = 4
+    OPEN = BoundaryCondition.OPEN
 
     @deprecate_argument("sc", "lattice",
                         "argument sc has been deprecated in favor of lattice, please update your code.",
                         "0.15.0")
+    @deprecate_argument("bc", None,
+                        "argument bc has been deprecated (removed) in favor of the boundary conditions in Lattice, please update your code.",
+                        "0.15.0")
     def __init__(self, shape, bc=None, lattice=None, dtype=None, geometry=None):
-        if bc is None:
-            bc = [[self.PERIODIC] * 2] * 3
 
         self.set_lattice(None)
 
         # Create the atomic structure in the grid, if possible
         self.set_geometry(geometry)
         if lattice is not None:
+            if bc is None:
+                bc = [[self.PERIODIC] * 2] * 3
             self.set_lattice(lattice)
 
         if isinstance(shape, Real):
@@ -106,7 +114,20 @@ class Grid(LatticeChild):
         self.set_grid(shape, dtype=dtype)
 
         # Create the grid boundary conditions
-        self.set_bc(bc)
+        if bc is not None:
+            self.lattice.set_boundary_condition(bc)
+
+    @deprecation("Grid.set_bc is deprecated since boundary conditions are moved to Lattice (see github issue #626", "0.15.0")
+    def set_bc(self, bc):
+        self.lattice.set_boundary_condition(bc)
+
+    @deprecation("Grid.set_boundary is deprecated since boundary conditions are moved to Lattice (see github issue #626", "0.15.0")
+    def set_boundary(self, bc):
+        self.lattice.set_boundary_condition(bc)
+
+    @deprecation("Grid.set_boundary_condition is deprecated since boundary conditions are moved to Lattice (see github issue #626", "0.15.0")
+    def set_boundary_condition(self, bc):
+        self.lattice.set_boundary_condition(bc)
 
     def __getitem__(self, key):
         """ Grid value at `key` """
@@ -131,20 +152,31 @@ class Grid(LatticeChild):
                 return False
         return np.all(abs(reps - np.round(reps)) < 1e-5)
 
-    def set_geometry(self, geometry):
+    def set_geometry(self, geometry, also_lattice: bool=True):
         """ Sets the `Geometry` for the grid.
 
         Setting the `Geometry` for the grid is a possibility
         to attach atoms to the grid.
 
         It is not a necessary entity, so passing `None` is a viable option.
+
+        Parameters
+        ----------
+        geometry : Geometry or None
+            specify the new geometry in the `Grid`. If ``None`` will
+            remove the geometry (but not the lattice)
+        also_lattice : bool, optional
+            whether to also set the lattice for the grid according to the
+            lattice of the `geometry`, if ``False`` it will keep the lattice
+            as it was.
         """
         if geometry is None:
             # Fake geometry
             self.geometry = None
         else:
             self.geometry = geometry
-            self.set_lattice(geometry.lattice)
+            if also_lattice:
+                self.set_lattice(geometry.lattice)
 
     def fill(self, val):
         """ Fill the grid with this value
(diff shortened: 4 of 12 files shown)


zerothi:
Added a small code-snippet in the top post.

pfebrer:
Great!

Some comments:

1. Don't you think that `bc` has a clear enough meaning for a lattice? `boundary_condition` seems to long to me.
2. Could the argument also support strings? In general I feel like enums are not very user friendly because you need to import them, which is tedious.
3. Could it make sense that there was also a `set_pbc` method? With an argument that could be `True` or `False` (False could mean to set open boundary conditions, I think this is what most users would expect).

zerothi:
1. I partially agree. What does Thomas say?
2. Good idea. I'll amend
3. I am not particularly fond of that doing the other would still be good. I'll see if bool can be translated to periodic.
4. It is because it is contained in the numpy array, perhaps this is overkill, so we might just change it. Any pros/cons on str vs int?

zerothi:
@pfebrer oh now I see your point in 4. I'll amend.

tfrederiksen:
I think this looks good. Just some simple observations:

1. `boundary_condition=True` is accepted, but is this intended?
```python
>>> sisl.Lattice(1, boundary_condition=True)
<sisl.Lattice a=1.0, b=1.0, c=1.0, α=90.0, β=90.0, γ=90.0, bc=[U, U, U], nsc=[1 1 1]>
```
2. Would it make sense to set the z-direction to be nonperiodic in the 2D geometries?
```python
>>> sisl.geom.graphene().lattice
<sisl.Lattice a=2.4595, b=2.4595, c=14.2, α=90.0, β=90.0, γ=60.0, bc=[P, P, P], nsc=[3 3 1]>
```
3. Could one enable simple string specifications, like `boundary_condition="ppu"` etc?

pfebrer:
I like Thomas' third point :) But I don't know if it deviates too much from the approach you were taking.

zerothi:
> I think this looks good. Just some simple observations:
> 
>     1. `boundary_condition=True` is accepted, but is this intended?
> 
> 
> ```python
> >>> sisl.Lattice(1, boundary_condition=True)
> <sisl.Lattice a=1.0, b=1.0, c=1.0, α=90.0, β=90.0, γ=90.0, bc=[U, U, U], nsc=[1 1 1]>
> ```

Thanks for discovering this, I'll amend some checks.
>
>     2. Would it make sense to set the z-direction to be nonperiodic in the 2D geometries?
> 
> 
> ```python
> >>> sisl.geom.graphene().lattice
> <sisl.Lattice a=2.4595, b=2.4595, c=14.2, α=90.0, β=90.0, γ=60.0, bc=[P, P, P], nsc=[3 3 1]>
> ```

Yes, this should be incorporated into the `geom` creation routines. Agreed!
I'll amend.
 
>     3. Could one enable simple string specifications, like `boundary_condition="ppu"` etc?

I don't like that... With @pfebrer addition one can do strings, but they should be part of a list. right now (not committed) this would work:
```python
lattice.set_boundary_condition(a="per", b="un") # uses `BC.name.startswith(key.upper())`
# your proposal would be:
boundary_condition=list("ppu")
```
Basically:
- a single string would be broadcasted to all directions
- a list of strings would be broadcasted to each direction
- a list of list of strings would be individual boundaries
Hope this would satisfy the usage.

pfebrer:
You made the same point to me when `axes` in `sisl.viz` only accepted a list, pointing that it would be better to pass `"xy"` instead of `list("xy")` or `["x", "y"]`. I think that it has showed to be the most convenient way and probably it would also be very convenient in this case.

zerothi:
> You made the same point to me when `axes` in `sisl.viz` only accepted a list, pointing that it would be better to pass `"xy"` instead of `list("xy")` or `["x", "y"]`. I think that it has showed to be the most convenient way and probably it would also be very convenient in this case.

Here it is a bit more generic... In `viz` there are only ever 3 axes, `"xyz"`, and `"abc"`, while for boundary conditions we don't really know if there later will be a new condition called "Poisson_*", in which case `"P"` is a bad specifier.  
The point is, boundary condition is not necessarily a static entity, in which case single letter specification is a bad notation. I agree with the current names they are unique, but, I am more concerned that this will be the preferred way of assigning BC, but that would limit future implementations.

zerothi:
Actually, startswith is equally bad... But it would encourage more than 1 letter words. I'll add so 1 letter words are not accepted.

pfebrer:
There are also more complex axes like a specific direction `[1, 1, 0]`. They are simply not allowed in the string syntax. I think it could be possible here as well that the string syntax is a shorthand that only supports the most common cases and if you have something more complex then you go with the "normal" syntax.

pfebrer:
> Actually, startswith is equally bad... But it would encourage more than 1 letter words. I'll add so 1 letter words are not accepted.

Yeah I would just specify explicitly a shorthand string for each case, as you do with spins (I think).

zerothi:
> There are also more complex axes like a specific direction `[1, 1, 0]`. They are simply not allowed in the string syntax. I think it could be possible here as well that the string syntax is a shorthand that only supports the most common cases and if you have something more complex then you go with the "normal" syntax.

What about: `"per,un,per"`? Then a comma/colon/whatever separator would allow this more easily?

pfebrer:
I think a list is better than commas in a string, so as a first thought I wouldn't say it would help with usage :)

zerothi:
> There are also more complex axes like a specific direction `[1, 1, 0]`. They are simply not allowed in the string syntax. I think it could be possible here as well that the string syntax is a shorthand that only supports the most common cases and if you have something more complex then you go with the "normal" syntax.

But what is the most common cases? The only non-common thing is "UNKNOWN"...

> I think a list is better than commas in a string, so as a first thought I wouldn't say it would help with usage :)

Ok, I think I can accept that `updno` always refer to the currently known boundary conditions. I'll amend.

pfebrer:
> But what is the most common cases? The only non-common thing is "UNKNOWN"...

I think more than 90 % of the time people will use periodic or open. But maybe I'm wrong.

pfebrer:
I didn't realise there was `unknown`, how should it be treated? LShould it raise errors when processing is attempted?

zerothi:
> I didn't realise there was `unknown`, how should it be treated? Should it raise errors when processing is attempted?

No, it is a placeholder for something the user doesn't know. It could be unspecified, or simply "unknown".

> > But what is the most common cases? The only non-common thing is "UNKNOWN"...
> 
> I think more than 90 % of the time people will use periodic or open. But maybe I'm wrong.

Hmm, I think the others could be equally nice. :)

zerothi:
I have amended your requests, could you have a look.

pfebrer:
But for example, if the user asks to compute neighbours, what should happen?

zerothi:
> But for example, if the user asks to compute neighbours, what should happen?

I agree, that is one of the points that needs resolution, but perhaps this is more appropriately solved with #553?

The whole point is that `nsc` should control the search space...
While one could argue that `nsc` is not something the user should control, in has proven to be very educational. Users are required to think about the concept of ranges, and can physically see them through `nsc`.

pfebrer:
> The whole point is that nsc should control the search space...

I don't agree with this. Boundary conditions should control the search space. Otherwise it's very confusing because the boundary conditions can be meaningless... (e.g. open boundary conditions with nsc=[3, 3, 3]).

pfebrer:
Or periodic boundary conditions with `nsc=[3, 3, 3]` while in reality there are interactions with second neighbour cells. If you follow `nsc` you are not using periodic boundary conditions then.

zerothi:
> > The whole point is that nsc should control the search space...
> 
> I don't agree with this. Boundary conditions should control the search space. Otherwise it's very confusing because the boundary conditions can be meaningless... (e.g. open boundary conditions with nsc=[3, 3, 3]).

One could do a transiesta calculation with `nsc=3` but you would have open boundary conditions.
It ain't pretty, it ain't right. But what would you do? When specifying anything but periodic it also sets `nsc` to 1?

pfebrer:
> When specifying anything but periodic it also sets nsc to 1?

If it is not periodic, then `nsc` doesn't make any sense, no? So there simply shouldn't be a `nsc` attribute in that case.

zerothi:
> Or periodic boundary conditions with `nsc=[3, 3, 3]` while in reality there are interactions with second neighbour cells. If you follow `nsc` you are not using periodic boundary conditions then.

no, but that could be used to limit the interaction range (intentionally). That is currently a flexibility in `sisl` where one can create a graphene flake, but set `nsc = 1 ` to create a GNR.
I agree it is confusing, but...

I am very much pro making this intuitive, and sustainable, but the way things are moving, #553 in particular, makes me wonder how we actually should implement these constraints, because users could do:
```python
H.geometry.lattice.set_boundary_condition(...)
```
and the Hamiltonian wouldn't know anything about it...



> > When specifying anything but periodic it also sets nsc to 1?
> 
> If it is not periodic, then `nsc` doesn't make any sense, no? So there simply shouldn't be a `nsc` attribute in that case.

I agree. But... The way the objects are interacting is probably not the best, I just can't see how we should move things around. See e.g. the above example which disrupts the full `H`

pfebrer:
So in practical terms maybe the method that sets boundary conditions should create a new `Lattice` object? Just as the methods that return modified geometries do?

zerothi:
> > no, but that could be used to limit the interaction range (intentionally). That is currently a flexibility in sisl where one can create a graphene flake, but set nsc = 1  to create a GNR.
> 
> I agree the flexibility is nice. But it should be an opt-in thing. I remember being very confused at the beggining when I created a geometry and `close` didn't find the periodic neighbours, and it was not very nice to have to set `nsc` manually because I didn't know if it was `3`, `5`, `7`... So yeah in the routines that use an auxiliary cell you could have some option to limit interactions, but by default the boundary conditions should be sufficient to specify what you want to happen.

That's what `optimize_nsc` is for. ;)
But I see your point. How should we go about this. `nsc` is an *advanced* thing, but also ideal for users to know about.

> > and the Hamiltonian wouldn't know anything about it...
> 
> I think you shouldn't be able to modify a `Hamiltonian`'s geometry. So once it is associated, a frozen copy should be made. You could change the boundaries from `periodic` to `open` and the `Hamiltonian` should then discard all supercell interactions. Or you could translate atoms from one cell to another, and the `Hamiltonian` should adapt. As in `translate2uc`. But I don't know if there are any other operations that could make sense.

I agree, but I don't know how to do this simply. Many of the routines does things in-place, so essentially all *changing* things should return a new object.

tfrederiksen:
I played a bit around with this and it looks good to me.

zerothi:
> I played a bit around with this and it looks good to me.

@tfrederiksen what do you think about the names? Is it too long? Should we do with `bc`, or `bound_cond`, comments?

tfrederiksen:
I don't like `bound_cond` too much (neither a short abbreviation nor completely explicit). `bc` could be OK and suitable like with `pbc`. On the other hand, I am more in favour of the explicit form `boundary_condition` which makes the code easier to read. It would be aligned with the property `.lattice` and the function `set_boundary_condition(...)`, etc. `bc` has also the drawback that it resembles two of the directional arguments like in `latt.set_boundary_condition(b=`per`, c='un`), but this may be a minor point. Could one accept both, something like `bc = boundary_condition` for the lazy user?

zerothi:
> I don't like `bound_cond` too much (neither a short abbreviation nor completely explicit). `bc` could be OK and suitable like with `pbc`. On the other hand, I am more in favour of the explicit form `boundary_condition` which makes the code easier to read. It would be aligned with the property `.lattice` and the function `set_boundary_condition(...)`, etc. `bc` has also the drawback that it resembles two of the directional arguments like in `latt.set_boundary_condition(b=`per`, c='un`), but this may be a minor point. Could one accept both, something like `bc = boundary_condition` for the lazy user?

Thanks, lets keep it for now! :)

zerothi:
@pfebrer the latest commit adds an informational warning that is shown when a boundary condition is changed from `PERIODIC` to non-periodic *and* `nsc>1`. It will only be shown if the `bc` is changed.

This is then a place-holder to inform users of inconsistencies, but users are expected to correct it them-selves.
I would be careful about changing `nsc` since it can be attached to a `Hamiltonion`. So this should be a warning for now.

pfebrer:
Yes I also agree it should be either `boundary_condition` or `bc`. I prefer the second, but both can be fine.

> @pfebrer the latest commit adds an informational warning that is shown when a boundary condition is changed from PERIODIC to non-periodic and nsc>1. It will only be shown if the bc is changed.

Ok, let's see how you make `nsc` play with the neighbour finder, because there currently I just don't care about it and calculate the `nsc` needed for periodic boundary conditions :sweat_smile:

zerothi:
> Yes I also agree it should be either `boundary_condition` or `bc`. I prefer the second, but both can be fine.
> 
> > @pfebrer the latest commit adds an informational warning that is shown when a boundary condition is changed from PERIODIC to non-periodic and nsc>1. It will only be shown if the bc is changed.
> 
> Ok, let's see how you make `nsc` play with the neighbour finder, because there currently I just don't care about it and calculate the `nsc` needed for periodic boundary conditions 😅

Hehe, ok, yes, lets play with that as well. The main constraint is that the offsets in the sparse matrix is a 1-1 correspondence with the offsets.

zerothi:
I'll keep the current state and merge, thanks a lot for your feedback!

PR #630, opened 2023-10-23
Included missing notebook in the viz docs

pfebrer:
I don't know why, but there was a notebook that was not included in the `.rst` file (probably I forgot to include it).

### DIFF
--- a/docs/visualization/viz_module/index.rst
+++ b/docs/visualization/viz_module/index.rst
@@ -73,6 +73,7 @@ developer or you are building a framework (e.g. GUI) around `sisl.viz`, this is 
 .. nbgallery::
     :name: viz-plotly-diy-gallery
 
+    diy/Building a new plot.ipynb
     diy/Adding new backends.ipynb
 
 


pfebrer:
Thanks! Could you refresh the documentation? I'm writing a message on discord to advertise the new `sisl.viz` :)

zerothi:
Is it important to get out now? It could be great to get #629 in as well? Shouldn't take too much

pfebrer:
Hmm ok, I'll just save the message

zerothi:
I am quite worried about my limits on the CI's, not infrequently I get emails on having runned too much :)
That is why I am trying to limit *unnecessary* runs. Hope you understand!
(you can print it here for safe keeping? )

pfebrer:
In the case of documentation I think it would be worth to have some caching, specially of the executed notebooks. (the hardest point would be to keep track of which parts of sisl the notebook depends on)

zerothi:
Agreed, I haven't had time to look into GA caching... But that would indeed be great!

PR #631, opened 2023-10-26
Added Info attributes to ascii class files

### LINKED ISSUE #509: Group common data in Siles into a property

zerothi:
Now all ascii files will contain the info attributes.

This enables the class to dynamically add attributes based on the content of Sile._info_attributes_.

Each of those entries will be converted to a regex Pattern and will process on each read line by patching the instance readline function.

This is very crude, and might be slow for very big files with lots of attributes, perhaps it can be streamlined a bit by removing those attributes which are static and already found (or perhaps all non-updatable attributes will be removed for searching once the full file has been processed).

It proves quite simple to use and makes the code look prettier and more readable. But the internals are a bit messy.

 - [x] Closes #509
 - [x] Tests added
 - [ ] Documentation for functionality, `docs/`

@tfrederiksen might you have a look, I know this is a drastically different approach than yours, but I think this could be sustainable for other things.

Note that `value` can be anything, meaning that the attribute could be a function that does one thing or the other...

### DIFF
--- a/src/sisl/io/orca/stdout.py
+++ b/src/sisl/io/orca/stdout.py
@@ -27,7 +27,7 @@ class stdoutSileORCA(SileORCA):
            lambda attr, match: int(match.string.split()[-1])),
         _A("no", r".*Number of basis functions",
            lambda attr, match: int(match.string.split()[-1])),
-        _A("_vdw_", r".*DFT DISPERSION CORRECTION",
+        _A("vdw_correction", r".*DFT DISPERSION CORRECTION",
            lambda attr, match: True, default=False),
         _A("completed", r".*ORCA TERMINATED NORMALLY",
            lambda attr, match: True, default=False),
@@ -120,8 +120,8 @@ class stdoutSileORCA(SileORCA):
                     spin_block = False
 
                 self.readline() # skip ---
-                A = np.empty(self.na, np.float64)
-                for ia in range(self.na):
+                A = np.empty(self.info.na, np.float64)
+                for ia in range(self.info.na):
                     line = self.readline()
                     v = line.split()
                     if spin_block and not spin:
@@ -177,7 +177,7 @@ class stdoutSileORCA(SileORCA):
                 if orbitals is None:
                     return D
                 else:
-                    Da = np.zeros(self.na, np.float64)
+                    Da = np.zeros(self.info.na, np.float64)
                     for (ia, orb), d in D.items():
                         if orb == orbitals:
                             Da[ia] = d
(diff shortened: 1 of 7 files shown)


src/sisl/io/sile.py
```diff
@@ -716,7 +718,188 @@ def close(self):
...
+    _info_attributes_ = []
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.info = _Info(self)
+
+    class _Info:
+        """ The actual .info object that will attached to the instance.
+
+        As of know this is problematic to document.
```
tfrederiksen, line 737: typo -> "now"

src/sisl/io/orca/stdout.py
```diff
@@ -15,80 +15,39 @@
...
-            self._vdw = True
-        return line
-
-    readline.__doc__ = SileORCA.readline.__doc__
+    _info_attributes_ = [
+        _A("na", r".*Number of atoms",
+           lambda attr, match: int(match.string.split()[-1])),
+        _A("no", r".*Number of basis functions",
+           lambda attr, match: int(match.string.split()[-1])),
+        _A("_vdw_", r".*DFT DISPERSION CORRECTION",
```
tfrederiksen, line 30: Why underscore here instead of just `info.vdw`?

tfrederiksen (changes requested):
Overall this looks very good to me; more elegant (but also more complex) than my attempt in #591. I have just some minor comments:
- could you add some critical tests of the new functionality?
- the code still references `self.attribute` instead of `self.info.attribute` in several places, for instance: 
   - https://github.com/zerothi/sisl/blob/d76ae4255c0d4bd773395abd7d4b0ea93c309008/src/sisl/io/orca/stdout.py#L123-L124
   - https://github.com/zerothi/sisl/blob/d76ae4255c0d4bd773395abd7d4b0ea93c309008/src/sisl/io/orca/tests/test_stdout.py#L20-L21
- What about adding the `_info_attributes_` with key quantities also for `siesta`, `vasp` etc?

src/sisl/io/orca/stdout.py
zerothi: Done, changed to `vdw_correction` it was based on the old implementation, but I have now changed to a more meaningful nomenclature. It is not `vdw`, but a `vdw_correction`. Hope you agree.

PR #634, opened 2023-10-30
fixed #633 due to weird combination of boundary lookups

### LINKED ISSUE #633: Bug in `construct`?
**Describe the bug**
I discovered a worrysome behaviour of the `construct` function for simple (but large) systems. It appears that some matrix elements are not always correctly set, at least on my machines.

The following script collects the number of nonzero matrix elements in 10 identical Hamiltonian constructions. The list `nnz` should thus contain identical numbers, but I observe some variation from run to run (by a few matrix elements) if I use `construct`. With a "manual" `close`loop things appear to be OK.

```python
import sisl

g = sisl.geom.graphene() * (60, 6, 1)

# using construct:
nnz = []
for i in range(10):
    H = sisl.Hamiltonian(g)
    H.construct([(0.1, 1.6), (1, -2.7)])
    nnz.append(H.nnz)
print("CONSTRUCT:", nnz)

# manual close-loop
nnz = []
for i in range(10):
    H = sisl.Hamiltonian(g)
    for ia in g:
        i0, ia = g.close(ia, R=(0.1, 1.6))
        H[i0, i0] = 1
        H[i0, ia] = -2.7
    nnz.append(H.nnz)
print("MANUAL LOOP:", nnz)

# version info
import sys, numpy, scipy
print("sys", sys.version)
print("numpy", numpy.__version__)
print("scipy", scipy.__version__)
print("sisl", sisl.__version__)
```
For instance, with the current development version I get
```python
CONSTRUCT: [2880, 2879, 2878, 2879, 2878, 2880, 2880, 2878, 2879, 2880]
MANUAL LOOP: [2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880]
sys 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
numpy 1.21.5
scipy 1.8.0
sisl 0.14.3.dev27+g849437cf
```
I can reproduce this with older versions on a different machine, for instance
- `sisl-0.11.0`:
```python
CONSTRUCT: [2880, 2880, 2878, 2879, 2878, 2878, 2880, 2880, 2879, 2879]
MANUAL LOOP: [2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880]
sys 3.7.6 (default, Jan  8 2020, 19:59:22) 
[GCC 7.3.0]
numpy 1.19.4
scipy 1.6.0
sisl 0.11.0
```
- `sisl-0.13.0`
```python
CONSTRUCT: [2880, 2879, 2880, 2880, 2880, 2880, 2879, 2880, 2878, 2879]
MANUAL LOOP: [2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880, 2880]
sys 3.7.6 (default, Jan  8 2020, 19:59:22) 
[GCC 7.3.0]
numpy 1.19.4
scipy 1.6.0
sisl 0.0.0.dev+851d78c55d40d1f7aa1f5d8320aae1d3364064b8
```

Any ideas where the problem lies?

**Some observations**
The problem seems to be related to the system size/shape, for instance I find
```python
g = sisl.geom.graphene() * (30, 29, 1)
....
CONSTRUCT: [6960, 6960, 6960, 6960, 6960, 6960, 6960, 6960, 6960, 6960]
MANUAL LOOP: [6960, 6960, 6960, 6960, 6960, 6960, 6960, 6960, 6960, 6960]
```
is OK, but
```python
g = sisl.geom.graphene() * (80, 2, 1)
...
CONSTRUCT: [1279, 1280, 1280, 1279, 1279, 1280, 1279, 1280, 1280, 1280]
MANUAL LOOP: [1280, 1280, 1280, 1280, 1280, 1280, 1280, 1280, 1280, 1280]
```
also fails.

zerothi:
The problem of close arose when the searched sphere was just crossing a border of the unit-cell (or supercell). In those cases could the translated to unit-cell indices miss a neighbour.

The fix was rather simple; once the atoms closests has been found, only loop over those in the primary unit-cell.

There are many other small changes here, but they
are more cosmetic and should provide more stability when using orbital distances very close to atomic coordinates.

 - [ ] Closes #633 
 - [x] Tests added

@tfrederiksen might you have a 2nd look here, I tested on some geometries, all seems fine, but this would warrant a 2nd look

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ we hit release version 1.0.0.
   The _info_attributes_ contains a list of attributes that can be
   discovered while reading ascii files see #509
 
+### Fixed
+- fixed cases where `Geometry.close` would not catch all neighbours, #633
+
 ### Changed
 - `Lattice` now holds the boundary conditions (not `Grid`), see #626
 - Some siles exposed certain properties containing basic information
--- a/src/sisl/geom/_common.py
+++ b/src/sisl/geom/_common.py
@@ -10,7 +10,7 @@ def geometry_define_nsc(geometry, periodic=(True, True, True)):
     """Define the number of supercells for a geometry based on the periodicity """
     if np.all(geometry.maxR(True) > 0.):
         geometry.optimize_nsc()
-        for i, d, per in zip(range(3), "abc", periodic):
+        for d, per in zip("abc", periodic):
             if per:
                 geometry.lattice.set_boundary_condition(**{d: "Periodic"})
             else:
--- a/src/sisl/geometry.py
+++ b/src/sisl/geometry.py
@@ -1749,7 +1749,7 @@ class Geometry(LatticeChild, _Dispatchs,
             R = self.maxR() + 0.001
         if R < 0:
             R = 0.00001
-            warn(self.__class__.__name__ +
+            warn(f"{self.__class__.__name__}"
                  ".optimize_nsc could not determine the radius from the "
                  "internal atoms (defaulting to zero radius).")
 
--- a/src/sisl/geometry.py
+++ b/src/sisl/geometry.py
@@ -892,7 +892,7 @@ class Geometry(LatticeChild, _Dispatchs,
 
         # default block iterator
         if R is None:
-            R = self.maxR() * 1.01
+            R = self.maxR() + 0.001
         if R < 0:
             raise ValueError(f"{self.__class__.__name__}.iR unable to determine a number of atoms within a sphere with negative radius, is maxR() defined?")
 
@@ -924,7 +924,7 @@ class Geometry(LatticeChild, _Dispatchs,
             raise SislError(f'{self.__class__.__name__}.iter_block_rand too small iR!')
 
         if R is None:
-            R = self.maxR() + 0.01
+            R = self.maxR() + 0.001
         # The boundaries (ensure complete overlap)
         R = np.array([iR - 0.5, iR + 0.501]) * R
 
@@ -997,7 +997,7 @@ class Geometry(LatticeChild, _Dispatchs,
         if iR < 2:
             raise SislError(f'{self.__class__.__name__}.iter_block_shape too small iR!')
 
-        R = self.maxR() + 0.01
+        R = self.maxR() + 0.001
         if shape is None:
             # we default to the Cube shapes
             dS = (Cube((iR - 0.5) * R),
@@ -1124,7 +1124,7 @@ class Geometry(LatticeChild, _Dispatchs,
         iR :
             the number of `R` ranges taken into account when doing the iterator
         R :
-            enables overwriting the local R quantity. Defaults to ``self.maxR() + 0.01``
+            enables overwriting the local R quantity. Defaults to ``self.maxR() + 0.001``
         atoms :
             enables only effectively looping a subset of the full geometry
         method : {'rand', 'sphere', 'cube'}
@@ -1150,7 +1150,7 @@ class Geometry(LatticeChild, _Dispatchs,
             yield from self.iter_block_rand(iR, R, atoms)
         elif method in ("sphere", "cube"):
             if R is None:
-                R = self.maxR() + 0.01
+                R = self.maxR() + 0.001
 
             # Create shapes
             if method == 'sphere':
@@ -1746,7 +1746,7 @@ class Geometry(LatticeChild, _Dispatchs,
             axis = _a.asarrayi(axis).ravel()
 
         if R is None:
-            R = self.maxR()
+            R = self.maxR() + 0.001
         if R < 0:
             R = 0.00001
             warn(self.__class__.__name__ +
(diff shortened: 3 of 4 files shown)


zerothi:
I have found the issue. In some rare cases the algorithm would search radii just crossing the boundaries which in the end could result in some atoms being part of the first neighbour circle, but its true neighbours would not be part of the 2nd circle.

This was a very subtle issue, and would generally only occur for somewhat large structures.

pfebrer:
Maybe moving to https://github.com/zerothi/sisl/pull/393 would fix it as well ;)

zerothi:
Damn, not fixed yet...
@pfebrer yeah, but...

zerothi:
I don't know if that will be stable given I have to prepare for the workshop...

zerothi:
I'll play a bit more tonight...

pfebrer:
I mean replacing the internals, while keeping the same `geom.close` API.

zerothi:
> I mean replacing the internals, while keeping the same `geom.close` API.

I know. I don't think that is smart to do just before a workshop... It's a big change.

tfrederiksen:
I am afraid this is not a complete fix. Try this:
```python
def test(g, runs=100):
    x = []
    for i in range(runs):
        H = sisl.Hamiltonian(g)
        H.construct([(0.1, 3), (1, -2.7)])
        x.append(H.nnz)
        if x[-1] != x[0]:
            print(f"Error encountered after {i+1} runs: {x}")
            return
    print(f"Success after {runs} runs")

g1 = sisl.geom.zgnr(10) * (20, 1, 1)
g2 = sisl.geom.sc(1, sisl.Atom('H', R=1))
g = g1.add(g2)
g.set_nsc([1, 1, 1])

test(g)
print(sisl.__version__)
```
which gives me something like this:
```
Error encountered after 6 runs: [4749, 4749, 4749, 4749, 4749, 4742]
0.14.3.dev30+g32b3be34
```

zerothi:
Yeah, I discovered, see my 2nd comment... :(

zerothi:
@tfrederiksen could you try now...

Please do note, that the problem arises because of the way the search is built. It does a batched search space made of 2 radii. One that is an integer number times the `self.maxR()` and one just one bigger. Thus when you use an R that is bigger than maxR, then you end up in situations where the edges are not great... :(
Now I am trying to catch this, but it will only do so with pure construct (as the ones you have). If people start doing their own `def construct_func`, then I can't get that information... And thus resulting in incorrect usage.

For now there is a warning raised when `sisl` thinks it is used incorrectly. This is why your above script *fails*. It is very hard for the batched search to correct this as it does not know `your R / maxR()` which is what `sisl` would need to figure it out correctly.

I have tested ~6 geometries with correct input, and they don't fail... :)
But... As always, there may be some corner case I haven't looked into... :(

tfrederiksen:
It looks good to me now. Despite several attempts I couldn't design a case with failure :)

Would it make sense to have a test with a critical comparison of `close` vs `construct` for some random structure? Perhaps the following script can serve as inspiration (although everything but the last `assert` would have passed before the fix)?

```python
import sisl
import numpy as np

def test(g):
    T = 1 + np.arange(na)
    # close call first
    H = sisl.Hamiltonian(g)
    for ia in g:
        idx = g.close(ia, R=R)
        for ja, t in zip(idx, T):
            H[ia, ja] = t
    x = [H.nnz + np.sum(H)]
    # now compare with construct calls
    for i in range(10):
        for method in ['rand', 'sphere', 'cube']:
            H = sisl.Hamiltonian(g)
            H.construct([R, T], method=method)
            x.append(H.nnz + np.sum(H))
            if x[-1] != x[0]:
                return False # inconsistency detected
    return x[0] # no inconsistency detected

np.random.seed(1234)

# build a random geometry with different orbitals
na = 7
R = 1e-4 + np.arange(na)
atoms = [sisl.Atom('HBCNOFPISK'[i], R=R[i]) for i in range(na)]
g = sisl.geom.fcc(2.5, atoms=atoms[0]) * (4, 3, 2)
g.xyz[:] += 10 * np.random.rand(g.na, 3)
for ia in g:
    g.atoms[ia] = atoms[ia % na]
g.reduce()

g.set_nsc([1, 1, 1])
assert test(g) == 1224.0
g.set_nsc([1, 1, 9])
assert test(g) == 4518.0
g.set_nsc([3, 3, 3])
assert test(g) == 18286.0

g2 = g.add(g.move(1e-2)) # put atoms almost on top of each other!
g2.set_nsc([1, 1, 1])
assert test(g2) == 4960.0
g2.set_nsc([1, 1, 3])
assert test(g2) == 13038.0

g3 = g.scale(1e3) # separate atoms by a lot!
g3.set_nsc([1, 1, 3])
assert test(g3) == 2 * g3.na

g4 = g.scale(1e-12) # collapse!
g4.set_nsc([1, 1, 3])
assert test(g4) == 6 * g4.na ** 2
```

Btw, I noticed some debug lines you may want to remove:
https://github.com/zerothi/sisl/blob/a6344c494b64ec89f3ec5cab5eaf79a7a72a5a26/src/sisl/geometry.py#L974-L975

zerothi:
@tfrederiksen thanks for the tests!! I'll see if some of that test can be added, currently there is a test that ensures symmetric construct, it is always hard to know when to cut the line here...

PR #635, opened 2023-10-31
added an as_supercell which returns geometry in supercell

zerothi:
This will tile the geometry and re-order the atoms such that they correspond to the supercell.

This is primarily useful for analysing real-space quantities or couplings between supercell sites.

 - [x] Tests added

@pfebrer as we discussed on discord, here is a method to create the supercell geometry. In the end I didn't go for returning the lattice as well. The coordinates are shifted to exactly match those of the supercell, in correct order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ we hit release version 1.0.0.
 ## [0.14.3] - YYYY-MM-DD
 
 ### Added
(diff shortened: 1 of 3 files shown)


pfebrer:
:+1:

> In the end I didn't go for returning the lattice as well.

You left the parameter description it in the docstring :)

zerothi:
Thanks, will merge!

PR #636, opened 2023-10-31
Honeycomb flakes

pfebrer:
Could a honeycomb/graphene flake geometry be useful? I'm using one currently and I think they are useful in general to create finite bilayers with arbitrary twist.

This PR is a proposal to add them.

**Examples**

```python
import sisl
import sisl.viz

sisl.geom.graphene_flake(shells=3).plot(axes="xy")
```

newplot (4)

```python
import sisl
import sisl.viz

sisl.geom.honeycomb_flake(shells=3, bond=1.6, atoms=["B", "N"]).plot(axes="xy")
```

newplot (5)

 - [ ] Closes #xxxx
 - [x] Tests added

### DIFF
--- a/src/sisl/geom/flat.py
+++ b/src/sisl/geom/flat.py
@@ -156,7 +156,7 @@ def honeycomb_flake(shells: int, bond: float, atoms, vacuum: float = 20.) -> Geo
     geom = geom.translate(geom.center(what="cell"))
 
     # Set boundary conditions
-    geom.lattice.set_boundary_condition(a="Unknown", b="Unknown", c="Unknown")
+    geometry_define_nsc(geom, [False, False, False])
 
     return geom
 
--- a/src/sisl/geom/flat.py
+++ b/src/sisl/geom/flat.py
@@ -74,6 +74,7 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
         atoms = Atom(Z=6, R=bond * 1.01)
     return honeycomb(bond, atoms, orthogonal)
 
+@set_module("sisl.geom")
 def honeycomb_flake(shells: int, bond: float, atoms, vacuum: float = 20.) -> Geometry:
     """Hexagonal flake of a honeycomb lattice, with zig-zag edges.
 
@@ -159,6 +160,7 @@ def honeycomb_flake(shells: int, bond: float, atoms, vacuum: float = 20.) -> Geo
 
     return geom
 
+@set_module("sisl.geom")
 def graphene_flake(shells: int, bond: float = 1.42, atoms=None, vacuum: float = 20.) -> Geometry:
     """Hexagonal flake of graphene, with zig-zag edges.
 
--- a/src/sisl/geom/flat.py
+++ b/src/sisl/geom/flat.py
@@ -8,7 +8,7 @@ from sisl._internal import set_module
 
 from ._common import geometry_define_nsc
 
-__all__ = ['honeycomb', 'graphene']
+__all__ = ['honeycomb', 'graphene', 'honeycomb_flake', 'graphene_flake']
 
 
 @set_module("sisl.geom")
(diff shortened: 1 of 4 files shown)


zerothi:
How about an API of `graphene_flakes` that resembles the `heteroribbon`? In that way attachments of various segments could be done?

pfebrer:
What would you attach to what in this case? :sweat_smile:

zerothi:
> What would you attach to what in this case? 😅

A new flake/ribbon, connecting ribbons and flakes could be quite interesting ;)

pfebrer:
Hmm yes, but the construction of the flake itself is not a composite geometry, no?

So I would imagine you can take `graphene_flake` and implement a section that knows how to connect to ribbons, which determines the arguments for `graphene_flake` and maybe rotates it, etc...

zerothi:
yes, that sounds cool :)
Ok, but anyways, I think this is fitting

pyproject.toml
```diff
@@ -11,7 +11,7 @@
 [build-system]
 requires = [
     "setuptools_scm[toml]>=6.2",
-    "scikit-build-core[pyproject]",
+    "scikit-build-core[pyproject]==0.5.1",
```
zerothi, line 14: please leave this out... It builds fine with later versions, just not editable installs :)

src/sisl/geom/flat.py
```diff
@@ -73,3 +73,107 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+    def _minimal_op(shells):
+
+        # The function is based on the horizontal lines of the hexagon,
+        # which are made of a pair of atoms.
+        # For each shell, we first need to complete the incomplete horizontal
+        # lines of the previous shell, and then branch them up and down to create
+        # the next horizontal lines.
+
+        # Displacement from the end of one horizontal pair to the beggining of the next
+        branch_displ_x = bond * np.cos(np.radians(60))
```
zerothi, line 105: `r = pi/3`, `np.cos(pi/3) == 0.5`w
`np.sin(pi/3) == 3 ** 0.5 / 2`

zerothi (changes requested):
great!

If adding 2 atoms, would it then alternate in each hexagon? That would be my ideal usecase, no?

Sorry, haven't tested, but looks good!

pfebrer:
> Ok, but anyways, I think this is fitting

Good :+1: I will then finish the PR some time this week probably.

To do the section stuff, it would be useful that the `graphene_flake` stored metadata on the `Geometry`, like `edge_axes` (the axes perpendicular to edges), `edge_R` (how far is the edge from the center) or `edge_W` (the number of C in the edge). But this requires the transformable attributes that we've discussed many times already :sweat_smile:

pyproject.toml
pfebrer: Yes, I will remove it :+1:

src/sisl/geom/flat.py
pfebrer: Ok!

pfebrer:
> If adding 2 atoms, would it then alternate in each hexagon? That would be my ideal usecase, no?

See the picture in the initial PR comment, is this how you expect it? Is there any other way?

zerothi:
> > If adding 2 atoms, would it then alternate in each hexagon? That would be my ideal usecase, no?
> 
> See the picture in the initial PR comment, is this how you expect it? Is there any other way?

Sorry, too fast ;) Beautiful! :)

pfebrer:
Ok, this PR is ready :+1:

src/sisl/geom/flat.py
```diff
@@ -73,3 +73,107 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+    Parameters
+    ----------
+    shells:
+        Number of shells in the flake. 0 means a single hexagon, and subsequent
+        shells add hexagon units surrounding the previous shell.
+    bond:
+        bond length between atoms (*not* lattice constant)
+    atoms:
+        the atom (or atoms) that the honeycomb lattice consists of
+    vaccuum:
```
zerothi, line 89: Spell, and all the way through

zerothi: Sorry, was on the phone, `vacuum`, and everywhere else :)

pfebrer: Fixed it :+1:

pfebrer:
Should I set the boundary conditions to `open`?

zerothi:
> Should I set the boundary conditions to `open`?

All you need to do is to call `geometry_define_nsc(geom, [False, False, False])` then return `geom`.

This will set the boundary conditions correctly, to UNKNOWN.

pfebrer:
Why do you say the correct bc is UNKNOWN and not OPEN? It is a 0D flake.

zerothi:
> Why do you say the correct bc is UNKNOWN and not OPEN? It is a 0D flake.

For this it would be better to use `Dirichlet`, `OPEN` is a reference to NEGF, perhaps `OPEN` should be bulk, or something. Typically OPEN is used in the nomenclature of NEGF BC.

pfebrer:
Ok, it should be fine now :+1:

src/sisl/geom/flat.py
```diff
@@ -73,3 +73,112 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+    max_x = np.max(geom.xyz[:,0])
+    geom.cell[0,0] = max_x * 2 + vacuum
+    geom.cell[1,1] = max_x * 2 + vacuum
+    geom.cell[2,2] = 20.
+
+    # Center the flake
+    geom = geom.translate(geom.center(what="cell"))
+
+    # Set boundary conditions
+    geom.lattice.set_boundary_condition(a="Unknown", b="Unknown", c="Unknown")
```
zerothi, line 158: No, call `geometry_set_nsc(geom, [False, False, False])`, `geometry_set_nsc` does other things as well. :)

src/sisl/geom/flat.py
```diff
@@ -73,3 +73,112 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
     if atoms is None:
         atoms = Atom(Z=6, R=bond * 1.01)
     return honeycomb(bond, atoms, orthogonal)
+
+def honeycomb_flake(shells: int, bond: float, atoms, vacuum: float = 20.) -> Geometry:
```
zerothi, line 77: missing `@set_module("sisl.geom")`

src/sisl/geom/flat.py
```diff
@@ -73,3 +73,112 @@ def graphene(bond=1.42, atoms=None, orthogonal=False):
...
+
+    # Center the flake
+    geom = geom.translate(geom.center(what="cell"))
+
+    # Set boundary conditions
+    geom.lattice.set_boundary_condition(a="Unknown", b="Unknown", c="Unknown")
+
+    return geom
+
+def graphene_flake(shells: int, bond: float = 1.42, atoms=None, vacuum: float = 20.) -> Geometry:
```
zerothi, line 162: missing `@set_module("sisl.geom")`

src/sisl/geom/flat.py
pfebrer: The only thing that it does is to set nsc to `[1, 1, 1]`, no? But to do that it calls first `geometry.optimize_nsc` which is a waste of time in this case...

zerothi: The point is that it we have a single reference function that sets up everything. All `geom` codes call this one. And so it should also.

pfebrer: Hmm ok but then I think `geometry_set_nsc` should check if all directions are not periodic and don't call `optimize_nsc` then

zerothi: True, I'll do that afterwards, call it and I'll get it in.

pfebrer: Done

zerothi (approved):

PR #639, opened 2023-11-03
Black linting

### LINKED ISSUE #413: Using black code formatting style
**Describe the feature**
I would like to use some kind of code formatter/style tool.

And while I don't agree with all of them on all things, I think it is better to just have consistency above all.

Any suggestions?

I have thought about using `black`, it is a tough formatter, but is really consistent.

zerothi:
I have long wanted to do some kind of linting.
Playing round turns out that there is no tool that really suits every wish.

I think it would be beneficial for sisl to use a linter to reduce the pain of going through code styles which is somewhat subjective, and also turns out to be too time-consuming.

With this PR we enforce the black coding style, whether we like it or not.

Comments would be greatly appreciated.

 - [x] Closes #413 
 - [x] Tests added


pfebrer:
Could you add this and `isort` to the PR CI? So that you don't need to run them locally and therefore we don't forget :)

pfebrer:
Maybe using `pre-commit`? https://pre-commit.com/#intro

pfebrer:
And actually if `pre-commit` is installed locally it's a great tool because it will run everything automatically when you do `git commit`.

tfrederiksen:
Looks good to me!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,6 +2,6 @@
 
  - [ ] Closes #xxxx
  - [ ] Tests added
- - [ ] Ranned `isort .` at top--level
+ - [ ] Ranned `isort .` and `black .` at top--level
```
tfrederiksen, line 5: Do you mean "ran"?

zerothi: yes! thanks!

tfrederiksen: I'd also change "reproducable" found in the bugs template. Perhaps "Code to reproduce the issue" or something similar would be more clear?

zerothi:
I have tried to make the test depend on the linter, this seems to not work since it is not part of the main branch.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

pfebrer:
I'm just checking `pre-commit` for another package and it works great with black, and probably `isort` too.

All you have to do is to have a `.pre-commit-config.yaml` file on the root folder which looks like:

```yaml
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.5.0
    hooks:
    -   id: check-yaml
    -   id: end-of-file-fixer
    -   id: trailing-whitespace
# Using this mirror lets us use mypyc-compiled black, which is about 2x faster
- repo: https://github.com/psf/black-pre-commit-mirror
  rev: 23.10.1
  hooks:
    - id: black-jupyter
```

Then run `pip install pre-commit` and `pre-commit install`. Then each time you run `git commit` everything runs and the fixes are added to the commit.

It is much better than handling each fixer manually, give it a try :)

Then the checkbox can simply be removed, and run `pre-commit` on the CI to check that everything is fine, optionally automatically commiting the fixes.

zerothi:
I'll merge and continue :)

> I'm just checking `pre-commit` for another package and it works great with black, and probably `isort` too.
> 
> All you have to do is to have a `.pre-commit-config.yaml` file on the root folder which looks like:
> 
> ```yaml
> # See https://pre-commit.com for more information
> # See https://pre-commit.com/hooks.html for more hooks
> repos:
> -   repo: https://github.com/pre-commit/pre-commit-hooks
>     rev: v4.5.0
>     hooks:
>     -   id: check-yaml
>     -   id: end-of-file-fixer
>     -   id: trailing-whitespace
> # Using this mirror lets us use mypyc-compiled black, which is about 2x faster
> - repo: https://github.com/psf/black-pre-commit-mirror
>   rev: 23.10.1
>   hooks:
>     - id: black-jupyter
> ```
> 
> Then run `pip install pre-commit` and `pre-commit install`. Then each time you run `git commit` everything runs and the fixes are added to the commit.
> 
> It is much better than handling each fixer manually, give it a try :)
> 
> Then the checkbox can simply be removed, and run `pre-commit` on the CI to check that everything is fine, optionally automatically commiting the fixes.

I think this is a great advice, if anything we could encourage users to do it, but since this package is still in its infancy int terms of adoption, I would be afraid that it scares people away. If people find this useful, then use it locally, it will help us, but it shouldn't be a barrier for new devs :)

pfebrer:
I think the separate fixers is more of a barrier than `pre-commit`, with the checkbox "Ran X, Y, ... at top level" being scary.

And the point is that in the end `pre-commit` should run on CI to make sure everything is fine, so the devs are not really required to use it.

pfebrer:
> I think this is a great advice, if anything we could encourage users to do it

By the way I think having both approaches is possible. You just have the `.pre-commit-config.yaml` file in the root of the repo in case anyone wants to do it with `pre-commit`, but it can still be done manually.

PR #647, opened 2023-11-07
Added a node to keep track of files

pfebrer:
This PR adds a **`FileNode` that watches the filesystem for changes** and gets marked as outdated if there is some change.

This is a **step towards having auto-updatable plots**, for example.

The way it works is that you simply initialize it with a path and then pass it to any argument that expects a path, and it will be like passing directly the path, except with the functionality of listening for updates.

```python
from sisl.nodes import Node, FileNode
import time

# Write something to file
with open("testfile", "w") as f:
    f.write("HELLO")

# Create a FileNode
n = FileNode("testfile")

# Define a file reader node that prints the contents of the file
@Node.from_func
def print_contents(path):
    print("printing contents...")
    with open(path, "r") as f:
        print(f.read())

# And initialize it by passing the FileNode as an input
printer = print_contents(path=n)

print("---RUNNING NODE")

# Run the printer node
printer.get()

print("--- SET AUTOMATIC UPDATING")

# Now set it to automatically update on changes to upstream inputs
printer.context.update(lazy=False)

print("--- APPENDING TO FILE")

# Append to the file which will trigger the update.
with open("testfile", "a") as f:
    f.write("\nBYE")

# Give some time for the printer to react before exiting
time.sleep(1)
```
```
---RUNNING NODE
printing contents...
HELLO
--- SET AUTOMATIC UPDATING
--- APPENDING TO FILE
printing contents...
HELLO
BYE
```

### DIFF
--- a/src/sisl/nodes/__init__.py
+++ b/src/sisl/nodes/__init__.py
@@ -1,4 +1,5 @@
 from .context import SISL_NODES_CONTEXT, NodeContext, temporal_context
+from .file_nodes import FileNode
 from .node import Node
 from .utils import nodify_module
 from .workflow import Workflow
--- a/src/sisl/nodes/file_nodes.py
+++ b/src/sisl/nodes/file_nodes.py
@@ -1,0 +1,176 @@
+from pathlib import Path
+
+try:
+    import watchdog.events
+    import watchdog.observers
+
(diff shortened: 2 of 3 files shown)


zerothi:
Linting failed

PR #648, opened 2023-11-08
Growth direction of zigzag heteroribbons

tfrederiksen:
The growth direction for heteroribbons should be the `x`-axis also for `kind=zigzag`. This PR produces expected structures, eg.:
```python
import sisl
import sisl.viz
g = sisl.geom.graphene_heteroribbon([(5, 2), (6, 3)], kind='zigzag')
g.plot(axes="xy", atoms_scale=0.6, nsc=[6, 2, 1])
```
newplot

 - [ ] Closes #x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ we hit release version 1.0.0.
 
 ## [0.14.4] - YYYY-MM-DD
 
+### Fixed
+- growth direction for zigzag heteroribbons
 
 
 ## [0.14.3] - 2023-11-07
--- a/src/sisl/geom/nanoribbon.py
+++ b/src/sisl/geom/nanoribbon.py
@@ -237,10 +237,8 @@ class _heteroribbon_section(CompositeGeometrySection):
     invert_first: bool = field(default=False, repr=False)
 
     def __post_init__(self):
-        if self.kind == "armchair":
+        if self.kind in ("armchair", "zigzag"):
             self.long_ax, self.trans_ax = 0, 1
-        elif self.kind == "zigzag":
-            self.long_ax, self.trans_ax = 1, 0
         else:
             raise ValueError("Unknown kind={kind}, must be one of zigzag or armchair")
 


zerothi:
I don't know if that is intended, but it creates periodic defects, and also the 4 square atoms has a non-conforming bond-length?
What should it do, this? Or?

tfrederiksen:
You're right! My example was not good, here is a better one:
```python
g = sisl.geom.graphene_heteroribbon([(5, 6), (7, 3), 5], kind='zigzag')
g.plot(axes="xy", atoms_scale=0.6, nsc=[2, 1, 1])
```
newplot
But yes, more generally the zigzag method seems a bit limited to produce desired structures. @pfebrer, how would do you for instance attach an extra hexagon to the edge of a zigzag ribbon? The following does not generate what I have in mind:
```python
g = sisl.geom.graphene_heteroribbon([(5, 6), (6, 3)], kind='zigzag')
g.plot(axes="xy", atoms_scale=0.6, nsc=[2, 1, 1])
```
newplot

pfebrer:
Hmm this last mistaken structure happens only after your commit or it also happens before the commit?

I guess there was a reason I set the growth axis different, maybe it doesn't work if you change it. If it worked before I guess we could rotate at the end when all coordinates are determined.

pfebrer:
But in this last example if you set the junction at the center properly the one at the boundary will be wrong, so there's really no right solution, no? Still I think when appending the second section the code doesn't know that it is the last one, so it's probably still a bug.

tfrederiksen:
> Hmm this last mistaken structure happens only after your commit or it also happens before the commit?

Before this commit there was no heteroribbon. The segments were stacked vertically with vacuum gaps in between.

pfebrer:
This is very strange, has the growth axis changed in `nanoribbon`?

zerothi:
So I guess the heteroribbon tests did not catch the change of the axis order, what should I do here? Let me know when I should merge!

pfebrer:
Hmm I guess we could merge this and fix the edge cases later. It is better to have something that produces good ribbons sometimes than something that creates parallel ribbons :sweat_smile:

zerothi:
when @tfrederiksen acknowledges, I'll merge, thanks!

tfrederiksen:
Green light from my side to merge. Should we open an issue with the aim to address the zigzag heteroribbons more generally later?

zerothi:
Ok, i'll merge

PR #682, opened 2024-02-14
Added Jmol colors as defaults for atoms.

pfebrer:
I also ran black and modified some unrelated files :sweat_smile:


zerothi:
Could you update your black version, and run it with the newest one, as you can see the lint action fails, I noticed in ab45ef4a766d6ac5c7e5f6a652953ec3d75ecd5d that black changed substantial things.

src/sisl/viz/data_sources/atom_data.py
```diff
@@ -66,15 +66,117 @@ def AtomNOrbitals(geometry, atoms=None):
 
 
 class AtomDefaultColors(AtomData):
+    # Taken from Jmol (https://jmol.sourceforge.net/jscolors/)
     _atoms_colors = {
```
zerothi, line 70: would it make sense to make this a `defaultdict`?

zerothi (commented):
Just a small question, I haven't looked at how you are using it, but perhaps a `defaultdict` would make this simpler?

pfebrer:
Ok, done. Again, pre-commit would be a good idea because it specifies the version :)

src/sisl/viz/data_sources/atom_data.py
pfebrer: Why?

pfebrer: It's used like this, right below the dictionary:

```python
def function(self, geometry, atoms=None):
        return np.array(
            [
                self._atoms_colors.get(atom.symbol, self._atoms_colors["else"])
                for atom in geometry.sub(atoms).atoms
            ]
        )
```

zerothi:
> Ok, done. Again, pre-commit would be a good idea because it specifies the version :)

Yes, lets do this. :)

src/sisl/viz/data_sources/atom_data.py
pfebrer: Hmm but a `defaultdict` is meant to initialize objects when the key is not there. Isn't it a bit overkill for this?

zerothi: Not a problem then, it would make its use simpler since one shouldn't care if the key is there or not. I'll merge.

pfebrer: I can convert it to something like `self._atoms_colors.get( atom.symbol, self._fallback_color)` if it looks better.

pfebrer: > Not a problem then, it would make its use simpler since one shouldn't care if the key is there or not. I'll merge.

Ok, anyway this can be modified whenever we want :)

PR #683, opened 2024-02-14
Added pre-commit config

pfebrer:
With `pre-commit` changes to versions of the checker tools (e.g. `isort` and `black`) or even addition/removal of tools will be easier to track. That is, if this config changes, `pre-commit` will automatically update the environment when a developer tries to add a new commit. This ensures that everyone is using exactly the same tools, and that they don't forget to run them.

If `pre-commit` is not set up, this file does nothing.

To set it up:
- Install pre-commit: `pip install pre-commit`
- Run install on root directory: `pre-commit install`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,27 @@
+# This config is used by pre-commit to run hooks on files before committing them to git
+# To use it install pre-commit with:
+#     `pip install pre-commit`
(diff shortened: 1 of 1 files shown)


zerothi:
Great, could you add a small comment in `CONTRIBUTING.md`, great!

zerothi (approved):

PR #692, opened 2024-02-26
Lazy loading of the viz module

### LINKED ISSUE #690: sisl.viz lazily loaded
**Describe the feature**

@pfebrer could we make it so sisl tries to import the necessary packages, if a user requests the plotting back-end?

Then we don't need the `import sisl.viz` stuff which shouldn't really be needed.

I am guessing some hooks to the `geometry.plot()` functions which then loads, and calls again? Or some `__get__` functionality on the class constructs?
What do you think?

pfebrer:
- [x] Closes #690 

As proposed in #690, now the sisl module is lazily loaded, so no need for explicitly importing `sisl.viz` or the existence of the `SISL_VIZ_AUTOLOAD` variable anymore. 

I updated the documentation accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ we hit release version 1.0.0.
 - `vacuum` is now an optional parameter for all ribbon structures
 - enabled `array_fill_repeat` with custom axis, to tile along specific
   dimensions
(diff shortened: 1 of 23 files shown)


zerothi:
Thanks!

PR #695, opened 2024-03-01
added typing in lots of io places, added SileSlicer

zerothi:
Added lots of typehints in the code, primarily in the io code base.

Changed the way the siesta.stdout siles handle data. They now rely on the SileSlicer and thus can be sliced upon calling. It makes the code a little simpler but also highligted some problems.

In particular the problem arise when the return function needs to do different things depending on whether the file is done reading, or not.

The defaults for stdoutSileSiesta has changed to read the next entry in the file. This is in contrast to the way it was. It will require users to adapt!
the reason for this can be found in the discussion in #586.

This streamlines the usage across the different parts of the IO handling.

@pfebrer I would be interested in your view on whether we could adapt these changes to the `read_scf` method, i.e. slicing would act on MD indices.

 - [ ] Closes #x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ we hit release version 1.0.0.
 - removed `Selector` and `TimeSelector`, they were never used internally
 
 ### Changed
+- `stdoutSileSiesta.read_*` now defaults to read the *next* entry, and not the last
+- `stdoutSileSiesta.read_*` changed MD output functionality, see #586 for details
 - `AtomNeighbours` changed name to `AtomNeighbor` to follow #393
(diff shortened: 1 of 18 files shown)


pfebrer:
> @pfebrer I would be interested in your view on whether we could adapt these changes to the read_scf method, i.e. slicing would act on MD indices.

I think it makes sense, but I won't be able to work on it for a long time. Next week I'm at a conference and then I go on a one month holiday 😅 Also on some other thread I think I commented that there could be two slicers in this case: one for MD step and one for scf step. I don't remember what was the conclusion.

zerothi:
@pfebrer no need, I just did it.

PR #710, opened 2024-03-15
removed txtSileOrca.no/info.no since it was never present

zerothi:
@tfrederiksen I checked the tests, in none (of the two examples) are the number of basis functions present, will they ever be?
This PR suggests to remove the attribute if it will never be present?
If this is not the case, then please let me know. :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ we hit release version 1.0.0.
   be done for assigning matrix elements (it fills with 0's).
 
 ### Removed
+- removed `txtSileOrca.info.no` since it was not present in any txt files
 - removed `Selector` and `TimeSelector`, they were never used internally
 
 ### Changed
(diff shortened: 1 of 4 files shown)


tfrederiksen:
Indeed this info is not included in those two simple examples, but with other flags I've encountered the number of basis functions also in `_property.txt`. I can supply another example.

zerothi:
ok, it would be great if we could have it in the tests.


If the `no` is not found, what should `sisl` do? Raise an error, issue a warning/info. If people need it, I would suspect raising an error would be appropriate?

tfrederiksen:
> ok, it would be great if we could have it in the tests.

OK, see https://github.com/zerothi/sisl-files/pull/8

> If the `no` is not found, what should `sisl` do? Raise an error, issue a warning/info. If people need it, I would suspect raising an error would be appropriate?

I agree, raising an error seems appropriate.

zerothi:
Ok, will amend this pr

zerothi:
Things should be fixed here.

PR #717, opened 2024-03-19
Fix for reading spinful CHG/CHGCAR

tfrederiksen:
I encountered a problem when reading spinful `CHG` and `CHGCAR` files. I'm unsure if the file formats could have changed, but as far as I can see, in VASP version 5.4.4 only the latter format contains "augmentation occupancies" blocks.

I also added a couple of new tests with a nitric oxide molecule in the cell.

 - [ ] Closes #x

### DIFF
--- a/src/sisl/io/vasp/chg.py
+++ b/src/sisl/io/vasp/chg.py
@@ -66,13 +66,19 @@ class chgSileVASP(carSileVASP):
         vals = []
         vext = vals.extend
 
+        is_chgcar = True
         i = 0
         while i < n * max_index:
-            vext(rl().split())
+            line = rl().split()
+            # CHG: 10 columns, CHGCAR: 5 columns
+            if is_chgcar and len(line) > 5:
+                # we have a data line with more than 5 columns, must be a CHG file
+                is_chgcar = False
+            vext(line)
             i = len(vals)
 
             if i % n == 0 and i < n * max_index:
-                if self.base_file.startswith("CHGCAR"):
+                if is_chgcar:
                     # Read over augmentation occupancies
                     line = rl()
                     while "augmentation" in line:
--- a/src/sisl/io/vasp/chg.py
+++ b/src/sisl/io/vasp/chg.py
@@ -72,11 +72,15 @@ class chgSileVASP(carSileVASP):
             i = len(vals)
 
             if i % n == 0 and i < n * max_index:
-                # Each time a new spin-index is present, we need to read the coordinates
-                j = 0
-                while j < geom.na:
-                    j += len(rl().split())
-
+                if self.base_file.startswith("CHGCAR"):
+                    # Read over augmentation occupancies
+                    line = rl()
+                    while "augmentation" in line:
+                        occ = int(line.split()[-1])
+                        j = 0
+                        while j < occ:
+                            j += len(rl().split())
+                        line = rl()
                 # one line of nx, ny, nz
                 rl()
 
--- a/src/sisl/io/vasp/tests/test_chg.py
+++ b/src/sisl/io/vasp/tests/test_chg.py
@@ -42,51 +42,46 @@ def test_graphene_chgcar_index_float(sisl_files):
     assert grid.grid.sum() / 2 == pytest.approx(gridh.grid.sum())
 
 
-def test_nitric_oxide_chg(sisl_files):
-    for spin in "unpol", "pol", "soi":
-        f = sisl_files(_dir, f"nitric_oxide/{spin}", "CHG.gz")
-        grid = chgSileVASP(f).read_grid()
-        gridf32 = chgSileVASP(f).read_grid(dtype=np.float32)
-        geom = chgSileVASP(f).read_geometry()
-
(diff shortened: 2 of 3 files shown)


src/sisl/io/vasp/chg.py
```diff
@@ -72,11 +72,15 @@ def read_grid(self, index=0, dtype=np.float64, **kwargs):
             i = len(vals)
 
             if i % n == 0 and i < n * max_index:
-                # Each time a new spin-index is present, we need to read the coordinates
-                j = 0
-                while j < geom.na:
-                    j += len(rl().split())
-
+                if self.base_file.startswith("CHGCAR"):
```
zerothi, line 75: how does this work with *other* files which are not a `CHGCAR`. And is this check even needed? I mean users could do: `mv CHGCAR run1.CHGCAR` and sisl will still get the correct type, but this check will fail. Is it necessary with this ending check?

src/sisl/io/vasp/tests/test_chg.py
```diff
@@ -40,3 +40,53 @@ def test_graphene_chgcar_index_float(sisl_files):
     gridh = chgSileVASP(f).read_grid(index=[0.5])
 
     assert grid.grid.sum() / 2 == pytest.approx(gridh.grid.sum())
+
+
+def test_nitric_oxide_chg(sisl_files):
```
zerothi, line 45: here you would be better of using fixtures to re-use the semantics.
```python
@pytest.fixture(scope="module", params=["unpol", "pol", "soi"])
def spin_conf(request):
    return request.param

def test_nitric_oxide_chg(sisl_files, spin_conf):
    #spin_conf will now be one of the above elements
```

untested, but should work

zerothi (commented):
Some minor points that isn't fully clear to me, and most likely this part of the code was never tested: https://app.codecov.io/gh/zerothi/sisl/blob/main/src%2Fsisl%2Fio%2Fvasp%2Fchg.py#L76 ;)
So it could be it should *always* be used!

src/sisl/io/vasp/chg.py
tfrederiksen: What is really needed is a check that distinguishes CHG and CHGCAR. As I zipped the files I decided to use `startswith`. But you're right, if the user changes the filename this is not a good way to do the check.

tfrederiksen:
> Some minor points that isn't fully clear to me, and most likely this part of the code was never tested: https://app.codecov.io/gh/zerothi/sisl/blob/main/src%2Fsisl%2Fio%2Fvasp%2Fchg.py#L76 ;) So it could be it should _always_ be used!

OK I see.

Could you bump sisl-files so the coverage tests will run after a rebase?

zerothi:
> > Could you bump sisl-files so the coverage tests will run after a rebase?
> 
> Can I actually do it? (would like to know)

you should do it in this branch :)

tfrederiksen:
> you should do it in this branch :)

Can you guide me how to do it?

zerothi:
something like this:
```shell
cd <root sisl folder>
git submodule init # if files doesn't already exist and is populated
cd files
git pull
cd ..
git add files
```
done

zerothi:
The idea is that the submodule should be understood as a file, the content being the commit hash. :)

src/sisl/io/vasp/chg.py
tfrederiksen: I've found another way to distinguish the two file types. CHG is 10 columns while CHGCAR has 5.

src/sisl/io/vasp/tests/test_chg.py
```diff
@@ -40,3 +37,36 @@ def test_graphene_chgcar_index_float(sisl_files):
...
+    geom = chgSileVASP(f).read_geometry()
+
+    assert grid.grid.sum() * grid.dvolume == pytest.approx(1, rel=1e-3)
+    assert geom == grid.geometry
+
+
+def test_nitric_oxide_soi(sisl_files, chg_type, np_dtype):
+    f = sisl_files(_dir, "nitric_oxide/soi", chg_type + ".gz")
+    s = 0
+    for i in range(1, 4):
```
zerothi, line 68: for SOI, shouldn-t it have 4 indices? This only uses 3 of them?

tfrederiksen: Correct, `index=0` is charge and the following three are x,y,z components (as far as I understand).

zerothi: ok, I see, so here you are checking spin. I'll have a look at this tomorrow. I am going to bed now! :bowing_man:

tfrederiksen: According to https://www.vasp.at/wiki/index.php/CHGCAR

`pol`:
- `index=0` is charge density (spin up + spin down)
- `index=1` is magnetization (spin up - spin down)

`soi`:
- `index=0` is charge density
- `index=1` is magnetization density in $\sigma_1$ direction
- `index=2` is magnetization density in $\sigma_2$ direction
- `index=3` is magnetization density in $\sigma_3$ direction

PR #722, opened 2024-03-21
Add `read_gtensor` and `read_hyperfine_coupling` to `txtSileORCA`

tfrederiksen:
This PR enables easy reading of the electronic g-tensor and the hyperfine tensors from ORCA.

### DIFF
--- a/src/sisl/io/orca/txt.py
+++ b/src/sisl/io/orca/txt.py
@@ -147,5 +147,119 @@ class txtSileORCA(SileORCA):
 
         return Geometry(xyz, atoms)
 
+    @sile_fh_open()
+    def read_gtensor(self):
+        """Reads electronic g-tensor data from the EPRNMR_GTensor block
+        in the ORCA property txt file
+
+        Returns
+        -------
+        PropertyDict : Electronic g-tensor
+        """
+        G = PropertyDict()
+        f, line = self.step_to("EPRNMR_GTensor", allow_reread=False)
+        if not f:
+            return None
+        for i in range(5):  # step 5 lines ahead
+            v = self.readline()
+        G["multiplicity"] = int(v.split()[-1])
+        self.readline()
+        self.readline()
+        tensor = np.empty([3, 3], np.float64)
+        for i in range(3):
+            v = self.readline().split()
+            tensor[i] = v[1:4]
+        G["tensor"] = tensor  # raw (total) tensor
+        self.readline()  # skip g tensor line
+        self.readline()  # skip header
+        vectors = np.empty([3, 3], np.float64)
+        for i in range(3):
+            v = self.readline().split()
+            vectors[i] = v[1:4]
+        G["vectors"] = vectors  # g-tensor eigenvectors
+        self.readline()  # skip eigenvalue line
+        self.readline()  # skip header
+        v = self.readline().split()
+        G["eigenvalues"] = np.array(v[1:4], np.float64)
+
+        return G
+
+    @sile_fh_open()
+    def read_hyperfine_coupling(self):
+        """Reads hyperfine couplings from the EPRNMR_ATensor block
+        in the ORCA property txt file
+
+        For a nucleus :math:`k`, the hyperfine interaction is usually
+        written in terms of the symmetric :math:`3\times 3` hyperfine
+        tensor :math:`A_k` such that
+
+        .. math::
+
+           H_\text{hfi} = \boldsymbol{S} \cdot A_k \boldsymbol{I}_k
+
+        where :math:`\boldsymbol{S}_k` and :math:`\boldsymbol{I}_k`
+        represent the electron and nuclear spin operators, respectively.
+
+        For a study of hyperfine coupling in nanographenes using ORCA
+        see :cite:`Sengupta2023`.
+
+        Note
+        ----
+        The hyperfine tensors are given in units of MHz.
+
+        Returns
+        -------
+        PropertyDict or list of PropertyDict : Hyperfine coupling data (in MHz)
+        """
+        f, line = self.step_to("EPRNMR_ATensor", allow_reread=False)
+        if not f:
+            return None
+
+        def read_A():
+            A = PropertyDict()
+            # Read the EPRNMR_ATensor block
+            f, line = self.step_to("Nucleus:", allow_reread=False)
+            if not f:
+                return None
+            v = line.split()
+            A["ia"] = int(v[1])
+            A["sa"] = v[2]
+            v = self.readline().split()
+            A["isotope"] = int(v[-1])
+            v = self.readline().split()
+            A["spin"] = float(v[-1])
+            v = self.readline().split()
+            A["prefactor"] = float(v[-1])
+            self.readline()  # skip Raw line
+            self.readline()  # skip header
+            tensor = np.empty([3, 3], np.float64)
+            for i in range(3):
+                v = self.readline().split()
+                tensor[i] = v[1:4]
+            A["tensor"] = tensor  # raw A_total tensor
+            self.readline()  # skip eigenvector line
+            self.readline()  # skip header
+            vectors = np.empty([3, 3], np.float64)
+            for i in range(3):
+                v = self.readline().split()
+                vectors[i] = v[1:4]
+            A["vectors"] = vectors
+            self.readline()  # skip eigenvalue line
+            self.readline()  # skip header
+            v = self.readline().split()
+            A["eigenvalues"] = np.array(v[1:4], np.float64)  # eigenvalues of A_total
+            v = self.readline().split()
+            A["iso"] = float(v[1])  # Fermi contact A_FC
+            return A
+
+        for i in range(4):
+            line = self.readline()
+        stored_nuclei = int(line.split()[-1])
+
+        if stored_nuclei == 1:
+            return read_A()
+
+        return [read_A() for k in range(stored_nuclei)]
+
 
 add_sile("txt", txtSileORCA, gzip=True)
--- a/src/sisl/io/orca/txt.py
+++ b/src/sisl/io/orca/txt.py
@@ -149,8 +149,7 @@ class txtSileORCA(SileORCA):
 
     @sile_fh_open()
     def read_gtensor(self):
-        """Reads electronic g-tensor data from the EPRNMR_GTensor block
-        in the ORCA property txt file
+        r"""Reads electronic g-tensor data from the ``EPRNMR_GTensor`` block
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


tfrederiksen:
I'm unsure about how we should handle units. ORCA reports the hfi-coupling in units of MHz. We could convert with a factor `eV2MHz = 2.417989242e8` or, perhaps better, utilize the unit functionality of sisl (where MHz is currently missing).

zerothi:
I just had a glance at the molecule3_property.txt output, some comments and suggestions:

- the `GTensor` seems to be something that can be calculated for several property indices (`geom`/`prop`):
   > $ EPRNMR_GTensor
   > description: Electronic g tensor
   > geom. index: 1
   > prop. index: 1
   > Source density: 1 SCF 
   > Spin multiplicity: 2 

  Would it make sense to extend this to allow reading more? Can ORCA do this?

  Great that this is an SCF quantity extracted via `SileBinder` :)
  How does this look in an MD + SCF? Our main usage with these indices is to refer to MD steps, and then scf steps are handled in the parsing method, so what does ORCA do? SCF-step output, or?

- the same thing can be said about the ATensor? Also, the ATensor is an atomic quantity, so basically it is indexed by the atom.
  I think the return here should be a complete array containing the data for all atoms.
  Perhaps using `xarray` would be useful here? It can be set-up in a tabulated fashion, which seems useful in an `xarray` context (also makes post-processing data much simpler!)

Thanks for this @tfrederiksen !

zerothi:
And, should `hfi` be named `hyperfine_atensor`? Would this be more used in the litt.? The name `hfi` seems very short (read non-descriptive)

tfrederiksen:
> And, should `hfi` be named `hyperfine_atensor`? Would this be more used in the litt.? The name `hfi` seems very short (read non-descriptive)

I think people working with hyperfine interactions are used to the abbreviation "HFI", but I'm also fine with choosing a more descriptive name, perhaps `read_hyperfine_coupling`.

zerothi:
> I think people working with hyperfine interactions are used to the abbreviation "HFI", but I'm also fine with choosing a more descriptive name, perhaps `read_hyperfine_coupling`.

I would be in favor of that :+1:

tfrederiksen:
> I just had a glance at the molecule3_property.txt output, some comments and suggestions:
> 
>   the `GTensor` seems to be something that can be calculated for several property indices (`geom`/`prop`):
>       > $ EPRNMR_GTensor
>       > description: Electronic g tensor
>       > geom. index: 1
>       > prop. index: 1
>       > Source density: 1 SCF
>       > Spin multiplicity: 2
>       
>       
>   Would it make sense to extend this to allow reading more? Can ORCA do this?

I don't quite see the relevance of that. Within the EPRNMR module there is (as far as I understand) just a single g-tensor corresponding to the given (global) electron spin state computed. In case of `molecule3` this is a spin doublet.

> Great that this is an SCF quantity extracted via `SileBinder` :)

Do you mean that the use of `SileBinder` for stepping through the A-tensors is not the intended usage (only SCF stuff)?

> How does this look in an MD + SCF? Our main usage with these indices is to refer to MD steps, and then scf steps are handled in the parsing method, so what does ORCA do? SCF-step output, or?

I am not sufficiently familiar with ORCA to answer this. I think in practice, the way we have used it, is to first determine a molecular geometry, then to switch to a specialized basis and to compute the g-tensor and HFIs.

> the same thing can be said about the ATensor? Also, the ATensor is an atomic quantity, so basically it is indexed by the atom.

In the case of `molecule3` indeed we've asked to compute the hyperfine couplings for all atoms through this specification in the input:

> %EPRNMR
>          GTENSOR   TRUE
>          NUCLEI    = ALL C {AISO, ADIP}
>          NUCLEI    = ALL H {AISO, ADIP}
> END

However, the user may choose to compute for just a subset of nuclei. According to the ORCA manual:

> % For example:
> % calculates the hyperfine coupling for all nitrogen atoms
> Nuclei = all N { aiso, adip, fgrad, rho};
> % calculates the spin spin coupling constants for all carbon atoms
> % assuming all carbon atoms are 13C
> Nuclei = all C { ssall, ist = 13};
> % etc

It is therefore not clear which nuclei to be found in the output.

> I think the return here should be a complete array containing the data for all atoms.
> Perhaps using `xarray` would be useful here? It can be set-up in a tabulated fashion, which seems useful in an `xarray` context (also makes post-processing data much simpler!)

I tend to think `xarray` is unnecessary here because at the end of the day, the tensors are just simple 3x3 matrices.

zerothi:
> I don't quite see the relevance of that. Within the EPRNMR module there is (as far as I understand) just a single g-tensor corresponding to the given (global) electron spin state computed. In case of `molecule3` this is a spin doublet.

It was more if ORCA enabled one to split stuff into molecules, and then this would return for `geom.index 2` or something similar.
But lets keep it. Thanks.

> 
> > Great that this is an SCF quantity extracted via `SileBinder` :)
> 
> Do you mean that the use of `SileBinder` for stepping through the A-tensors is not the intended usage (only SCF stuff)?
No, just a comment, but, see below.
> 
> > How does this look in an MD + SCF? Our main usage with these indices is to refer to MD steps, and then scf steps are handled in the parsing method, so what does ORCA do? SCF-step output, or?
> 
> I am not sufficiently familiar with ORCA to answer this. I think in practice, the way we have used it, is to first determine a molecular geometry, then to switch to a specialized basis and to compute the g-tensor and HFIs.

Ok, the point of the silebinder would be to do `MD` values. So `read_gtensor[1]` would yield the gtensor at MD step 2 (starting at 1).
So in this case if `gtensor` will always only be written once, then it shouldn't have the `SileBinder()` location.

> 
> > the same thing can be said about the ATensor? Also, the ATensor is an atomic quantity, so basically it is indexed by the atom.
> 
> In the case of `molecule3` indeed we've asked to compute the hyperfine couplings for all atoms through this specification in the input:
> 
> > %EPRNMR
> > GTENSOR   TRUE
> > NUCLEI    = ALL C {AISO, ADIP}
> > NUCLEI    = ALL H {AISO, ADIP}
> > END
> 
> However, the user may choose to compute for just a subset of nuclei. According to the ORCA manual:
> 
> > % For example:
> > % calculates the hyperfine coupling for all nitrogen atoms
> > Nuclei = all N { aiso, adip, fgrad, rho};
> > % calculates the spin spin coupling constants for all carbon atoms
> > % assuming all carbon atoms are 13C
> > Nuclei = all C { ssall, ist = 13};
> > % etc
> 
> It is therefore not clear which nuclei to be found in the output.
> 
> > I think the return here should be a complete array containing the data for all atoms.
> > Perhaps using `xarray` would be useful here? It can be set-up in a tabulated fashion, which seems useful in an `xarray` context (also makes post-processing data much simpler!)
> 
> I tend to think `xarray` is unnecessary here because at the end of the day, the tensors are just simple 3x3 matrices.

Ok, so disregard my *all atoms* return value. It should then return for all available atoms.

So here, a few comments. AFAIK, the current implementation would do: `read_hyperfine_interaction[1]` would yield the 2nd entry for whatever atom the user have requested.  
This goes against the main utilization of the `read_*[:]` method, which should be MD indexable things.

So I would suspect the `read_hyperfine_interaction` to return a list of elements for all atoms requested.
So line `205` should read: `self.step_to("EPRNMR_ATensor", ...)` and then loop over `Number of stored nuclei`.

I would still advocate on moving this to `xarray`. For instance to easily search for the atom with the lowest eigenvalue. Or some other query, but we could easily start with a list of dicts with values if you wish.

A question, is the pre-factor something that should be factored on the `Raw A tensor`? If so shouldn't this just be done and then let it be discarded?

tfrederiksen:
> > I am not sufficiently familiar with ORCA to answer this. I think in practice, the way we have used it, is to first determine a molecular geometry, then to switch to a specialized basis and to compute the g-tensor and HFIs.
> 
> Ok, the point of the silebinder would be to do `MD` values. So `read_gtensor[1]` would yield the gtensor at MD step 2 (starting at 1). So in this case if `gtensor` will always only be written once, then it shouldn't have the `SileBinder()` location.

OK I see. Will change this.

> Ok, so disregard my _all atoms_ return value. It should then return for all available atoms.
> 
> So here, a few comments. AFAIK, the current implementation would do: `read_hyperfine_interaction[1]` would yield the 2nd entry for whatever atom the user have requested. This goes against the main utilization of the `read_*[:]` method, which should be MD indexable things.
> 
> So I would suspect the `read_hyperfine_interaction` to return a list of elements for all atoms requested. So line `205` should read: `self.step_to("EPRNMR_ATensor", ...)` and then loop over `Number of stored nuclei`.

OK, will do.

> I would still advocate on moving this to `xarray`. For instance to easily search for the atom with the lowest eigenvalue. Or some other query, but we could easily start with a list of dicts with values if you wish.

A list of dicts would be easier for me at this stage (how we currently have it in our scripts).

> A question, is the pre-factor something that should be factored on the `Raw A tensor`? If so shouldn't this just be done and then let it be discarded?

To my understanding the raw tensor already contains "everything" (dipolar + Fermi contact contributions).  I am not sure why the output also provides a prefactor value (varies per nuclear species) but it is a factor that the code has used to get the raw tensor. Maybe some users will want it, I don't know.

zerothi:
Great! Thanks!

tfrederiksen:
Maybe ready now?

src/sisl/io/orca/txt.py
```diff
@@ -147,5 +147,119 @@ def read_geometry(self):
...
+            v = self.readline().split()
+            A["iso"] = float(v[1])  # Fermi contact A_FC
+            return A
+
+        for i in range(4):
+            line = self.readline()
+        stored_nuclei = int(line.split()[-1])
+
+        if stored_nuclei == 1:
+            return read_A()
```
zerothi, line 260: I will merge and change to always return a list. I find this more consistent, as users don't need to check in their scripts.

src/sisl/io/orca/txt.py
```diff
@@ -147,5 +147,119 @@ def read_geometry(self):
...
+        self.readline()  # skip eigenvalue line
+        self.readline()  # skip header
+        v = self.readline().split()
+        G["eigenvalues"] = np.array(v[1:4], np.float64)
+
+        return G
+
+    @sile_fh_open()
+    def read_hyperfine_coupling(self):
+        """Reads hyperfine couplings from the EPRNMR_ATensor block
```
zerothi, line 189: I will also adapt these small things:

1. when having math, one needs to use the `r"""` prefix to tell it to sustain the `\` in the code. `r` for `raw`.
2. I will change `\text` to `\mathrm` and `\boldsymbol` to `\mathbf`

zerothi (approved):

src/sisl/io/orca/txt.py
tfrederiksen: OK, you may then also take a look at the docstring in `sisl.physics.electron.berry_curvature` that I followed.

src/sisl/io/orca/txt.py
tfrederiksen: Ok

src/sisl/io/orca/txt.py
zerothi: Thanks! 
The difference between `boldsymbol` and `mathbf` is that `boldsymbol` can be used to bold a greek letter in a consistent manner. AFAIK boldsymbol on letters is just making them bold, but not upright. `mathbf` is bold *and* uprighting the letters, but does nothing for greek letters. So `\boldsymbol` shold be used for greek letters, otherwise, `\mathbf`. I just checked `berry_curvature`, it seems consistent with what I wrote (unless I missed something).

Anyway, thanks!

zerothi:
Many thanks @tfrederiksen !

PR #726, opened 2024-03-22
`units` as parameter in `read_*`

tfrederiksen:
Although sisl works with `eV` as the default energy unit, users may just use sisl as an interface and keep original units (or some other choice). 

Here a suggestion for ORCA. I could extend to the other interfaces too.

### DIFF
--- a/src/sisl/io/orca/stdout.py
+++ b/src/sisl/io/orca/stdout.py
@@ -260,12 +260,21 @@ class stdoutSileORCA(SileORCA):
 
     @SileBinder()
     @sile_fh_open()
-    def read_energy(self):
+    def read_energy(self, unit="eV"):
         """Reads the energy blocks
+
+        Parameters
+        ----------
+        unit :
+            what energy unit the data should be returned in
+
+        Note
+        ----
+        Energies written by ORCA have units of Ha.
 
         Returns
         -------
-        PropertyDict or list of PropertyDict : all energy data (in eV) from the "TOTAL SCF ENERGY" and "DFT DISPERSION CORRECTION" blocks
+        PropertyDict or list of PropertyDict : all energy data from the "TOTAL SCF ENERGY" and "DFT DISPERSION CORRECTION" blocks
         """
         f = self.step_to("TOTAL SCF ENERGY", allow_reread=False)[0]
         if not f:
@@ -274,28 +283,28 @@ class stdoutSileORCA(SileORCA):
         self.readline()  # skip ---
         self.readline()  # skip blank line
 
-        Ha2eV = units("Ha", "eV")
+        Ha2unit = units("Ha", unit)
         E = PropertyDict()
 
         line = self.readline()
         while "----" not in line:
             v = line.split()
             if "Total Energy" in line:
-                E["total"] = float(v[-4]) * Ha2eV
+                E["total"] = float(v[-4]) * Ha2unit
             elif "E(X)" in line:
-                E["exchange"] = float(v[-2]) * Ha2eV
+                E["exchange"] = float(v[-2]) * Ha2unit
             elif "E(C)" in line:
-                E["correlation"] = float(v[-2]) * Ha2eV
+                E["correlation"] = float(v[-2]) * Ha2unit
             elif "E(XC)" in line:
-                E["xc"] = float(v[-2]) * Ha2eV
+                E["xc"] = float(v[-2]) * Ha2unit
             elif "DFET-embed. en." in line:
-                E["embedding"] = float(v[-2]) * Ha2eV
+                E["embedding"] = float(v[-2]) * Ha2unit
             line = self.readline()
 
         if self.info.vdw_correction:
             self.step_to("DFT DISPERSION CORRECTION")
             v = self.step_to("Dispersion correction", allow_reread=False)[1].split()
-            E["vdw"] = float(v[-1]) * Ha2eV
+            E["vdw"] = float(v[-1]) * Ha2unit
 
         return E
 
--- a/src/sisl/io/orca/stdout.py
+++ b/src/sisl/io/orca/stdout.py
@@ -5,7 +5,8 @@ import numpy as np
 
 from sisl._internal import set_module
 from sisl.messages import deprecation
-from sisl.unit import units
+from sisl.unit import serialize_units_arg
+from sisl.unit import units as sunits
 from sisl.utils import PropertyDict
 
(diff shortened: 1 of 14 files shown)


zerothi:
When you brought up the MHz question, I was also thinking about this. 

I can't really figure out if this is a good idea or not. There are soo many places where values have units. And I am worried it might end up being confusing, *and* error-prone.

tfrederiksen:
I think the use case is quite clear for the HFI. Here some people will want to work with "MHz" as this is a more natural energy scale without having to worry about making conversion errors if we force it to "eV" at the interface.

tfrederiksen:
It also makes it explicit for the user what the units are (both in the source file as well as after import in sisl).

tfrederiksen:
OK, I see your point that it may not be relevant to implement this everywhere. But at least for `read_hyperfine_coupling`?

zerothi:
Yeah, probably we should do it. But... :)

If we are going to do this, we need to figure out how to do it for multiple return values, say it also returned some coordinate (length), how should we deal with that?

tfrederiksen:
Do you have an example of that? I guess we can stick to sisl defaults if it is about reading `nc` files etc with lots of different data inside.

zerothi:
> Do you have an example of that? I guess we can stick to sisl defaults if it is about reading `nc` files etc with lots of different data inside.

A simple example would be to read the Hamiltonian, it uses the coordinates and an energy.
So you might request `Ry` energy, and then typically one would use `Bohr` for the coordinates.
I know this is a contrieved example, but later on when some return functions returns complicated things multiple units could be valuable. E.g. `read_scf` in siesta, it basically returns all information in the SCF cycle, and there we might add dipole information etc.

tfrederiksen:
But reading this one gets a sisl object `Hamiltonian`. Here I would not enable unit conversion. I think it should be limited to "low-level" reading only. And maybe on a case-by-case consideration?

tfrederiksen:
Eventually, one could have multiple parameters: `unit_energy`, `unit_length` etc.

zerothi:
Yes, but I would consider `read_scf` as low-level. So it might be worthwhile to get the interface correct.

We could imagine a decorator enabling the `units` argument, which then processes it to a dict. Something like this:
```python
@units_handling # not the best word
def read_hyperfine_interaction(...)


sile.read_hyperfine_interaction(units=("MHz", "Ang"))
```
then internally the `units` would be a `dict` with entries:
```python
units = {"energy": "MHz", "length": "Ang"}
```
or similar?

tfrederiksen:
We could also always add the quantity to the parameter?

zerothi:
> We could also always add the quantity to the parameter?

you mean that users should do `units={"energy": "MH"}`?

tfrederiksen:
> > We could also always add the quantity to the parameter?
> 
> you mean that users should do `units={"energy": "MH"}`?

I mean that this PR should be changed to `unit_energy="eV"`.

zerothi:
I am not in favor of having `unit_*` for all variants. ;) That seems extremely verbose where a dict could handle it.

tfrederiksen:
OK, then what about accepting both str and dict as inputs for `unit`?

zerothi:
> OK, then what about accepting both str and dict as inputs for `unit`?

Yes, this was my idea, here I don't think we need anymore (other than changing it to `units` for future consistency). Then a decorator can change the way it is intercepted to streamline the conversions for arguments.

zerothi:
My idea of the decorator was to convert the `units` argument to a dict *always*

tfrederiksen:
> My idea of the decorator was to convert the `units` argument to a dict _always_

Should we also allow calls without the keyword, i.e., `read_energy("MHz")`? Then the decorator becomes a little complex, no?

zerothi:
> > My idea of the decorator was to convert the `units` argument to a dict _always_
> 
> Should we also allow calls without the keyword, i.e., `read_energy("MHz")`? Then the decorator becomes a little complex, no?

good question, no, I think it would be wise to forcefully ask the user to write `units=`, good point!

tfrederiksen:
I've given the decorator a try. What do you think?

src/sisl/io/orca/stdout.py
```diff
@@ -258,9 +259,10 @@ def read_block(step_to):
 
         return read_block(step_to)
 
+    @units2dict
```
zerothi, line 262: This decorator needs to be the first decorator added. `SileBinder` should be the last since it changes the function to a class.

src/sisl/unit/base.py
```diff
@@ -424,3 +426,20 @@ def __call__(self, *units):
...
+
+@set_module("sisl.unit")
+def units2dict(func):
+    "Decorator to force units parameter into dictionary"
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        # Check for units in kwargs, and adjust as needed
+        if "units" in kwargs and not isinstance(kwargs["units"], dict):
+            unit = kwargs["units"]
```
zerothi, line 439: it could still be a list, say 'units=("eV", "Ang")` there would be no ambiguity for the end-user.

tfrederiksen: OK, and what should happen if the user provides `units=("eV", "J")`?

zerothi: an error should be raised, then they should be explicit if they should belong to different dimensions.

tfrederiksen:
I'm facing some problems that I think is related to the other decorators. Wouldn't it be simpler with just a function like this?
```python
def read_hyperfine_coupling(units="eV"):
    units = sisl.unit.unit2dict(units)
```
Then it is perhaps also easier to see what is going on, as compared to
```python
@units2dict
def read_hyperfine_coupling(units="eV"):
    assert isinstance(units, dict)
```

zerothi:
fine for me, I'll probably rename `unit2dict` to something like: `serialize_units_arg` for clarity

zerothi:
The decorator requires an inspection pass. So lets just keep it as is. :)

zerothi:
Should we merge this now, and take it from there?

zerothi:
> OK, maybe the docstrings just need an update with `units : {str, dict, list, tuple}`?

Yeah, later we should change this to use typing constructs. :)

zerothi:
But I think this is fine

tfrederiksen:
OK, I think it is ready now.

zerothi:
Merged! Thanks a lot @tfrederiksen

PR #727, opened 2024-03-22
Resolve Issue #716

### LINKED ISSUE #716: Reading wannier Hamiltonian with out '_centres.xyz' present
When trying to read the Hamiltonian from `<>_tb.dat` or `<>_hr.dat` without `<>_centeres.xyz` present, we try to read the geometry from the Wannier input, which returns a geometry with the 'correct' atoms and positions. However, for the Hamiltonian, this geometry is not suited (the number of wannier functions in general is different from the number of atoms). 

To fix this problem I would suggest, implementing the following procedure:

In `sisl.io.wannier90.seedname.read_hamiltonian` we explicitly call read_geometry with `order=['centers']`. Should this call fail, we obtain the number of Wannier functions (`n_wann`) from the `<>.win` file or as a fallback from the Hamiltonian file and create a fake geometry with `n_wann` orbitals all centered at the origin.

This could be later improved by using the initial projections as the locations of the orbitals.

@zerothi, please let me know if this makes sense to you. If so, I'll create a PR for this.

nils-wittemeier:
This should resolve #716.

In this PR, I 
- changes the default behavior in which geometry is read: the geometry is only read from the `<>.win` file if explicitly requested.
- split the functionality of the `winSileWannier90` into different siles: `xyzSileWannier90`, `tbSileWannier90`, `hrSileWannier90`.
-
@zerothi, could you have a look at this and see if it matches what you had imagined?

### DIFF
--- a/src/sisl/io/wannier90/seedname.py
+++ b/src/sisl/io/wannier90/seedname.py
@@ -23,7 +23,7 @@ from ..sile import *
 from .sile import SileWannier90
 from ..xyz import xyzSile
 
-__all__ = ["winSileWannier90"]
+__all__ = ["winSileWannier90", "tbSileWannier90", "hrSileWannier90", "xyzSileWannier90"]
 
 
 def _construct_hamiltonian(geometry: Geometry, Hsc):
@@ -528,3 +528,6 @@ class hrSileWannier90(hamSileWannier90):
 
 
 add_sile("win", winSileWannier90, gzip=True)
+add_sile("_tb.dat", tbSileWannier90, gzip=True)
+add_sile("_hr.dat", hrSileWannier90, gzip=True)
+add_sile("_centres.xyz", xyzSileWannier90, gzip=True)
--- a/src/sisl/io/wannier90/seedname.py
+++ b/src/sisl/io/wannier90/seedname.py
@@ -314,6 +314,10 @@ class winSileWannier90(SileWannier90):
             # to ensure we get the correct orbital positions
             kwargs["geometry"] = self.read_geometry()
 
+        if "lattice" not in kwargs:
+            # to ensure we get the correct orbital positions
+            kwargs["lattice"] = self.read_lattice()
+
         kwargs["cutoff"] = cutoff
 
         for f in order:
@@ -392,7 +396,7 @@ class tbSileWannier90(hamSileWannier90):
         #  Lattice vectors [Ang]
         cell = _a.zerosf((3, 3))
         for i in range(3):
-            cell[i] = map(float, self.readline().split())
+            cell[i] = list(map(float, self.readline().split()))
 
         # Number of orbitals
         no = int(self.readline())
@@ -482,7 +486,8 @@ class hrSileWannier90(hamSileWannier90):
         # Number of orbitals
         no = int(self.readline())
         if geometry is None:
-            geometry = Geometry([0.0, 0.0, 0.0], lattice=Lattice([[0.0, 0.0, 0.0]] * 3))
+            lattice = kwargs.pop("lattice", Lattice(_a.zerosf((3, 3))))
+            geometry = Geometry([0.0, 0.0, 0.0] * no, lattice=lattice)
         elif no != geometry.no:
             raise ValueError(
                 f"{self.__class__.__name__}"
--- a/src/sisl/io/wannier90/seedname.py
+++ b/src/sisl/io/wannier90/seedname.py
@@ -12,7 +12,7 @@ import scipy.sparse as sps
 from scipy.sparse import lil_matrix
 
 import sisl._array as _a
-from sisl import Geometry, Lattice
+from sisl import Geometry, Lattice, Atom
 from sisl.messages import deprecate_argument
 from sisl.physics import Hamiltonian
 from sisl.unit import unit_convert
@@ -21,6 +21,7 @@ from ..sile import *
 
 # Import sile objects
 from .sile import SileWannier90
+from ..xyz import xyzSile
 
 __all__ = ["winSileWannier90"]
 
@@ -134,29 +135,17 @@ class winSileWannier90(SileWannier90):
 
         return self._r_lattice()
 
-    @sile_fh_open()
     def _r_geometry_centres(self, *args, **kwargs):
         """Defered routine"""
 
-        nc = int(self.readline())
-
-        # Comment
-        self.readline()
-
-        na = 0
-        sp = [None] * nc
-        xyz = np.empty([nc, 3], np.float64)
-        for ia in range(nc):
-            l = self.readline().split()
-            sp[ia] = l.pop(0)
-            if sp[ia] == "X":
-                na = ia + 1
-            xyz[ia, :] = [float(k) for k in l[:3]]
-
-        return Geometry(xyz[:na, :], atoms="H")
+        f = self.dir_file(self._seed + "_centres.xyz")
+        geometry = None
+        if f.exists():
+            geometry = xyzSileWannier90(f).read_geometry(*args, **kwargs)
+        return geometry
 
     @sile_fh_open()
-    def _r_geometry(self, lattice, *args, **kwargs):
+    def _r_geometry_win(self, lattice, *args, **kwargs):
         """Defered routine"""
 
         is_frac = True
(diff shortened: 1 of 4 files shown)


src/sisl/io/wannier90/seedname.py
```diff
@@ -297,69 +278,133 @@ def _r_wigner_seitz_weights(self):
...
+
+    def _r_hamiltonian_hr(self, *args, **kwargs):
+        """Reads a Hamiltonian model from the <>_hr.dat file"""
+        f = self.dir_file(self._seed + "_hr.dat")
+        H = None
+        if f.exists():
+            H = hrSileWannier90(f).read_hamiltonian(*args, **kwargs)
+        return H
+
     @sile_fh_open()
```
zerothi, line 296: no `sile_fh_open`, the file gets changed

src/sisl/io/wannier90/seedname.py
```diff
@@ -297,69 +278,133 @@ def _r_wigner_seitz_weights(self):
...
+        dtype: np.float64, optional
+            the default data-type used for the matrix.
+            Is mainly useful to check whether the TB model has imaginary
+            components (it should not since it is a Wannier model).
+        """
+        order = kwargs.pop("order", ["tb", "hr"])
+
+        if "geometry" not in kwargs:
+            # to ensure we get the correct orbital positions
+            kwargs["geometry"] = self.read_geometry()
```
zerothi, line 315: perhaps be explicit here, so it is clear in the code what we are doing? `order=centres`

src/sisl/io/wannier90/seedname.py
```diff
@@ -297,69 +278,133 @@ def _r_wigner_seitz_weights(self):
...
-            # Get super-cell, row and column
-            isc[0], isc[1], isc[2], r, c = map(int, l[:5])
+        ws = 1.0 / _a.arrayd(ws)
+        return ws
+class tbSileWannier90(hamSileWannier90):
 
-            # Update index for degeneracy, if required
-            if r + c == 2:
-                iws += 1
+    @sile_fh_open()
```
zerothi, line 378: do `sile_fh_open(True)` and remove `self.fh.seek(0)`, this will force a complete re-read of the file.

src/sisl/io/wannier90/seedname.py
```diff
@@ -297,69 +278,133 @@ def _r_wigner_seitz_weights(self):
...
-                h = hr + 1j * float(l[6]) * w
-            else:
-                h = hr
+        # Time of creation
+        self.readline()
 
-            if abs(h) > cutoff:
-                Hsc[tuple(isc)][r - 1, c - 1] = h
+        #  Lattice vectors [Ang]
+        cell = _a.zerosf((3,3))
```
zerothi: `zerosf` is 32 bit, please use 64 bit, `zerosd`.

src/sisl/io/wannier90/seedname.py
```diff
@@ -419,44 +463,71 @@ def _r_hamiltonian_tb(self, geometry, dtype=np.float64, **kwargs):
...
-    def read_hamiltonian(self, cutoff: float = 1e-5, *args, **kwargs):
-        """Read the electronic structure of the Wannier90 output by reading the <>_tb.dat, <>_hr.dat
+class hrSileWannier90(hamSileWannier90):
 
-        Parameters
-        ----------
-        cutoff:
-           the cutoff value for the zero Hamiltonian elements, default
-           to 0.00001 eV.
+    @sile_fh_open()
```
zerothi, line 471: again here

src/sisl/io/wannier90/seedname.py
```diff
@@ -419,44 +463,71 @@ def _r_hamiltonian_tb(self, geometry, dtype=np.float64, **kwargs):
...
-                H = getattr(self, f"_r_hamiltonian_{f.lower()}")(
-                    geom, *args, cutoff=cutoff, **kwargs
-                )
-            self._set_file()
-            if H is not None:
-                return H
+        # Number of orbitals
+        no = int(self.readline())
+        if geometry is None:
+            geometry = Geometry([0., 0., 0.], lattice=Lattice([[0.,0.,0.]]*3))
```
zerothi: should we be restrictive here, does it make sense to read a Hamiltonian without having the `geometry`?

zerothi (changes requested):
overall this looks very good!

Also add these classes to the `docs/api/io/wannier90.rst` file.

src/sisl/io/wannier90/seedname.py
nils-wittemeier: I found a better workaround. Just pushed the commit, using the lattice read from win file. I suppose other wise we should raise an error.

nils-wittemeier:
I think all of your comments should be addressed now.

src/sisl/io/wannier90/seedname.py
```diff
@@ -211,21 +199,13 @@ def read_geometry(self, *args, **kwargs):
...
-        self._set_file("_centres.xyz")
-        if self.file.is_file():
-            geom = self._r_geometry_centres()
-        else:
-            self._set_file()
-            geom = self._r_geometry(lattice, *args, **kwargs)
-
-        # Reset file
-        self._set_file()
+        kwargs["lattice"] = self.read_lattice()
```
zerothi, line 202: this won't allow users to put in their own lattice (for whatever reason).
Test for `lattice` in `kwargs` if not, read it, otherwise retain.

src/sisl/io/wannier90/seedname.py
```diff
@@ -211,21 +199,13 @@ def read_geometry(self, *args, **kwargs):
         """Reads a `Geometry` and creates the Wannier90 cell"""
```
zerothi, line 199: add doc-string related to the order argument, also clarify that it will try and read from `_centres.xyz`

src/sisl/io/wannier90/seedname.py
```diff
@@ -297,69 +277,136 @@ def _r_wigner_seitz_weights(self):
...
+            components (it should not since it is a Wannier model).
+        """
+        order = kwargs.pop("order", ["tb", "hr"])
+
+        if "geometry" not in kwargs:
+            # to ensure we get the correct orbital positions
+            kwargs["geometry"] = self.read_geometry(order=["centres"])
+
+        if "lattice" not in kwargs:
+            # to ensure we get the correct orbital positions
```
zerothi, line 317: comment is wrong ;)

src/sisl/io/wannier90/seedname.py
```diff
@@ -297,69 +277,136 @@ def _r_wigner_seitz_weights(self):
...
-            hr = float(l[5]) * w
-            if is_complex:
-                h = hr + 1j * float(l[6]) * w
-            else:
-                h = hr
+class tbSileWannier90(hamSileWannier90):
 
-            if abs(h) > cutoff:
-                Hsc[tuple(isc)][r - 1, c - 1] = h
+    @sile_fh_open(True)
```
zerothi, line 381: perhaps add a `read_lattice` and then let read_geometry use that, it gets more versatile like that.

zerothi (changes requested):
love it, minor details. Thanks!

zerothi:
And how is your experience with the TB from Wannier90, is it working like it should? I didn't test this extensively, so feedback would be great! ;)

nils-wittemeier:
> And how is your experience with the TB from Wannier90, is it working like it should? I didn't test this extensively, so feedback would be great! ;)

Personally, I am not using it for anything right now, I just tested it by calculating bands. 

I think @monxe might be better able to give some feedback soon.

src/sisl/io/wannier90/seedname.py
nils-wittemeier: functionality, should be split now

src/sisl/io/wannier90/seedname.py
nils-wittemeier: rephrased

src/sisl/io/wannier90/seedname.py
nils-wittemeier: should be fixed

src/sisl/io/wannier90/seedname.py
nils-wittemeier: changed the doc string here and updated most of them a bit

zerothi:
@monx-ne feedback would be most welcome! :)

zerothi:
In case you are interested, please have a look here, where I fixed some remaining issues (and I saw some generics that could be re-used).
https://github.com/zerothi/sisl/commit/5cac580720447578caf6b602eef7c521bcb96bb1
I added the ability to read the lattice from the tb file.

zerothi:
Thanks a lot for all this @nils-wittemeier

PR #733, opened 2024-03-26
preliminary test for 0-orbital Atoms

### LINKED ISSUE #449: Atoms without orbitals?
I am trying to build a geometry for which I would like to have zero orbitals associated to certain atoms. Can it be done?
As far as I can see, I get at least one orbital associated to an atom:
```python
print(sisl.Atom('H', orbitals=0))
Atom{H, Z: 1, mass(au): 1.00794, maxR: 0.00000,
 Orbital{R: 0.00000, q0: 0.0}
}
```
Motivation:  Think of a GNR structure with H-atoms at the edges (it could be a geometry obtained from DFT). Now, if we could simply set zero orbitals for the H and one orbital for C, we would readily arrive at the effective model for the pi-electrions only, without having to remove any physical atoms from the geometry.

zerothi:
@tfrederiksen @pfebrer 

this is a first stab at 449. I am starting to grow on the idea.

However, as you can see in this PR, there are many places in the code base where it assumes that there exists some orbitals on atoms.

Whether or not these should be defaulted to be present or not is something that needs to be discussed.
However, it seems that one should be explicit on defining orbitals if one wishes to use them.
Many places the code simply fails because it now requires some kind of explicitness.

But I would agree that this could be useful. If anybody cares to expand/fix/add tests for the remaining details, feel free :)

 - [x] Closes #449

### DIFF
--- a/src/sisl/_core/atom.py
+++ b/src/sisl/_core/atom.py
@@ -1076,7 +1076,7 @@ class Atom(
         if isinstance(orbitals, (tuple, list, np.ndarray)):
             if len(orbitals) == 0:
                 # This may be the same as only regarding `R` argument
-                pass
+                orbitals = None
             elif isinstance(orbitals[0], Orbital):
                 # all is good
                 self._orbitals = orbitals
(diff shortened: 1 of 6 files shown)


pfebrer:
I want to push this forward to use it for QM/MM geometries, should I start a new branch and create a separate PR?

zerothi:
> I want to push this forward to use it for QM/MM geometries, should I start a new branch and create a separate PR?

Yes, please do, branch this off would be ideal imho.

However, I think it would be very beneficial to discuss here, or reiterate the decisions made in #449. Eg how sparse matrices etc should handle things.

PR #737, opened 2024-04-03
added math notation convention

### LINKED ISSUE #731: Math in sisl documentation
**Describe the issue**

I would like to open up the discussion on how we should format math/equations in the documentation.

There are a few details that requires explicitness (IMHO).

- which values should we use for atomic indices, and for looping
  
  Suggestion: $i\in I$, where $i$ is an atomic index, and $I$ is an atomic index range
- which values should we use for orbital indices

  Suggestion: $\nu\in\alpha$ where $\alpha$ is an orbital range (e.g. $\alpha_i$ orbitals on atom $i$), and $\nu$ is a single index

- how should index ranges be written?

  Suggestion: above we just did $I$, but should it be $`\{I\}`$ or?
- how should matrices be written?

  Suggestion: $\mathbf A$ (capital bold-faced letters)
- how should vectors be written?

  Suggestion: $\vec a$
- how should scalares be written?

  Suggestion: $a$

With a convention, we can also formalize how math should be understood:

- $\vec a \cdot \vec b$ is dot-product
- $\vec a \mathbf A$ (no need for $\cdot$)
- $\mathbf A \mathbf B$, or $\mathbf A \times \mathbf B$, where I would argue $\times$ can be optional

Are there things forgotten?

this came about due to #722 where (I plead guilty) the math got changed after merge.

The floor is open to suggestions!

zerothi:
@tfrederiksen @nils-wittemeier could you please have a look here.

Are there any things that is missing?

Once agreed upon, we should start migrating documentation to follow these guidelines. Preferably against this branch, so we solve everything in one go!

 - [x] Closes #731  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ we hit release version 1.0.0.
 
 ### Changed
 - documentation build system on RTD is updated, #745
+- `gauge` arguments now accept 'cell' and 'orbital' in replacements for 'R' and 'r', respectively
 - `siesta.*.read_basis` now defaults to read an `Atoms` object with all atoms
 - `atoms.specie` changed to `atoms.species`, generally species is the singular form
 - `in_place` arguments changed to `inplace`
--- a/docs/api/typing.rst
+++ b/docs/api/typing.rst
@@ -36,6 +36,7 @@ The typing types are shown below:
    Coord
    CoordOrScalar
    FuncType
+   GaugeType
    GeometryLike
    GridLike
    KPoint
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -74,6 +74,7 @@ Since then it has expanded to accommodate a rich set of DFT code input/outputs.
    tutorials.rst
    scripts/index
    environment
+   math
    nodes/nodes_intro
 
 .. toctree::
--- a/docs/math.rst
+++ b/docs/math.rst
@@ -1,0 +1,42 @@
+.. _math_convention:
+
+Mathematical notation convention
+================================
+
+In `sisl` we strive to make the documentation as consistent
+as possible.
+This should make reading different parts of the documentation
+very simple to understand.
+
+Here is a list of rules that sisl will strive to adhere to.
+Please, if you find any inconsistencies in the documentation,
+please let us know!
+
+* upper case characters such as :math:`I` and :math:`J` refer
+  to atomic indices
+* lower case characters such as :math:`i` and :math:`j` refer
+  to orbital indices, e.g. :math:`i\in I`
+* scalars are represented via lower case italics, such
+  as :math:`a`
+* vectors are represented via lower case bold faced, such
+  as :math:`\mathbf a`
+
+  One may use :math:`\hat{\mathbf a}` to signal normal vectors
+
+  * dot-products between vectors should be explicit :math:`\mathbf a\cdot\mathbf b`
+
+* matrices are represented via upper case bold faced, such
+  as :math:`\mathbf A`
+
+  * vector-matrix products are implicit :math:`\mathbf a\mathbf B`
+
+  * matrix-matrix products are implicit :math:`\mathbf A\mathbf B`
+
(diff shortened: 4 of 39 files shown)


nils-wittemeier:
Looks good to me

tfrederiksen (approved):
Looks good to me too!

zerothi:
Thanks!

One final thing that occurred to me.
Generally we have used $\mathbf R$ as lattice vectors, and $\mathbf r$ as atomic coordinates. So now, what should we do? $\mathbf R_\alpha$ to signal lattice matrix, and an index, and then $\mathbf r_\alpha$ to signal coordinates?

tfrederiksen:
> One final thing that occurred to me. Generally we have used R as lattice vectors, and r as atomic coordinates. So now, what should we do? Rα to signal lattice matrix, and an index, and then rα to signal coordinates?

According to the above, $\mathbf R$ would be a matrix, no?

nils-wittemeier:
That's true. Maybe we are trying to make these definitions to rigid?!

nils-wittemeier:
We could return to Nick's initial suggestion of indicating vectors with arrows. That would remove the ambiguity

tfrederiksen:
What about avoiding $\mathbf R$? 

$\mathbf r_i$ and $\mathbf r_I$ could be vectors to an orbital and an atom, respectively? And $\mathbf r_\alpha$ a cell vector?

zerothi:
> What about avoiding R?
> 
> ri and rI could be vectors to an orbital and an atom, respectively? And rα a cell vector?

I think we can't really avoid $\mathbf R$ the arguments for `gauge` is `r` or `R`, so I think it would be smart if we could retain that notation. 

> > One final thing that occurred to me. Generally we have used R as lattice vectors, and r as atomic coordinates. So now, what should we do? Rα to signal lattice matrix, and an index, and then rα to signal coordinates?
> 
> According to the above, R would be a matrix, no?

But with indices $\mathbf R_\alpha$ it would be a column/vector for the matrix. We need to be able to use indices for matrices anyways, so I guess this is *ok*. 

Otherwise, I think we really should go the $\vector R$ way.

tfrederiksen:
What about reconsidering the inputs to `gauge`? Maybe something like `gauge='cell|orbital|atom'` would be more explicit than `r` and `R`?

zerothi:
indeed that would be a better choice. :)

zerothi:
I have (in this pr) changed `gauge=cell|orbital` for consistency.

zerothi:
> What about avoiding R?
> 
> ri and rI could be vectors to an orbital and an atom, respectively? And rα a cell vector?

I would still advocate we could use $\mathbf R_\alpha$ for indexing lattice vectors. 
Or, we should differentiate between indices, and labels.

For instance $\mathbf A_\alpha$ will be the $\alpha$ column of the matrix. In this case that will result in a vector.  
However, if you do $\mathbf a_\alpha$ it will actually be a scalar. because you are taking out a single component.  
So if we want to *name* a lattice vector with the $\alpha$ name we should do: $\mathbf r^\alpha$ or $\mathbf R_\alpha$, no?

zerothi:
I know I am being pedantic here. But it think this could be nice to enforce ;)

nils-wittemeier:
Upon reconsidering it, I think $\mathbf R_\alpha$ is a perfectly adequate way to name the lattice vectors. We can interpret $\mathbf R$ as a matrix where each column/row is a lattice vector.

tfrederiksen:
> Upon reconsidering it, I think Rα is a perfectly adequate way to name the lattice vectors. We can interpret R as a matrix where each column/row is a lattice vector.

OK, I like this too. So $\mathbf R$ would directly be the 3x3 matrix from `Geometry.cell`.

For some generic coordinate vector $\mathbf r$ we could introduce labels as in $\mathbf r ^{(i)}$ or $\mathbf r ^{(I)}$, with parentheses to avoid interpretation as exponent.

zerothi:
I think I got pretty much everything now, probably a bit too fast sometimes ;)

PR #738, opened 2024-04-03
Pre-commit hook to add license

pfebrer:
Added a pre-commit hook that makes sure the license header is at the beggining of the file.

The hook needs a file containing the header, which I named `LICENSE_header`, but can be renamed to anything else.


zerothi:
Merged! Thanks!

PR #742, opened 2024-04-10
Draft: add vectorsSileSIESTA for parsing vibra output

nils-wittemeier:
See title.

### DIFF
--- a/src/sisl/io/siesta/vibra.py
+++ b/src/sisl/io/siesta/vibra.py
@@ -3,6 +3,8 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 from __future__ import annotations
 
+import re
+
 import numpy as np
 
 import sisl._array as _a
@@ -150,7 +152,7 @@ class vectorsSileSiesta(SileSiesta):
         nmodes = 1
         while True:
             line = self.readline()
-            if line.startswith("k            ="):
+            if re.match("k *=", line):
                 break
             elif line.startswith("Eigenvector"):
                 nmodes += 1
--- a/src/sisl/io/siesta/vibra.py
+++ b/src/sisl/io/siesta/vibra.py
@@ -200,7 +200,7 @@ class vectorsSileSiesta(SileSiesta):
                 line = self.readline()
                 mode[imode, :, iatom].imag = list(map(float, line.split()))
 
-        info = dict(k=self._convert_k(k), parent=self._parent, gauge="r")
+        info = dict(k=self._convert_k(k), parent=self._parent, gauge="orbital")
         return EigenmodePhonon(mode.reshape(self._nmodes, -1), c * _cm1_eV, **info)
 
     def yield_eigenmode(self):
--- a/src/sisl/io/siesta/vibra.py
+++ b/src/sisl/io/siesta/vibra.py
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 import numpy as np
-from itertools import product
 
 from sisl import Geometry, Lattice, constant
 from sisl._internal import set_module
@@ -12,7 +11,7 @@ from sisl.messages import warn
 from sisl.physics.phonon import EigenmodePhonon
 from sisl.unit.siesta import unit_convert
 
-from ..sile import add_sile, sile_fh_open, sile_raise_write
+from ..sile import add_sile
 from .sile import SileSiesta
 
 
@@ -22,14 +21,27 @@ _Bohr2Ang = unit_convert("Bohr", "Ang")
 _J2eV = unit_convert("J", "eV")
 _hc_eV = constant.h * constant.c * _J2eV
 
+
 @set_module("sisl.io.siesta")
 class vectorsSileSiesta(SileSiesta):
-    """Phonon eigenmode file"""
+    """Phonon eigenmode file
+
+    Parameters
+    ----------
+    parent : obj, optional
+        a parent may contain a geometry, and/or a supercell
+    geometry : Geometry, optional
+        a geometry contains a cell with corresponding lattice vectors
+        used to convert k [1/Ang] -> [b]
+    lattice : Lattice, optional
+        a supercell contains the lattice vectors to convert k
+
+    """
 
     def _setup(self, *args, **kwargs):
         """Simple setup that needs to be overwritten
 
-        All _r_next_* methods expect the fortran file unit to be handled
+        All _r_next_* methods expect the file unit to be handled
         and that the position in the file is correct.
         """
         super()._setup(*args, **kwargs)
@@ -37,7 +49,6 @@ class vectorsSileSiesta(SileSiesta):
         # default lattice
         lattice = None
 
-        # In case the instantiation was called with wfsxSileSiesta("path", geometry=geometry)
         parent = kwargs.get("parent")
         if parent is None:
             geometry = None
@@ -112,28 +123,27 @@ class vectorsSileSiesta(SileSiesta):
         # Read the sizes relevant to the file.
         # We also read whether there's only gamma point information or there are multiple points
         if self._state == -1:
-            self._sizes = self._r_sizes()
+            self._sizes = self._r_next_sizes()
             self._state = 0
 
         if close:
             self._close()
 
-    def _r_sizes(self):
-        """ Determine the dimension of the data stored in the vectors file
-        """
+    def _r_next_sizes(self):
+        """Determine the dimension of the data stored in the vectors file"""
         pos = self.fh.tell()
         # Skip past header
-        self.readline() # Empty line
-        self.readline() # K point
-        self.readline() # Eigenmode index
-        self.readline() # Frequency
-        self.readline() # Eigenmode header
-        
+        self.readline()  # Empty line
+        self.readline()  # K point
+        self.readline()  # Eigenmode index
+        self.readline()  # Frequency
+        self.readline()  # Eigenmode header
+
         # Determine number of atoms (i.e. rows per mode)
         natoms = 0
         while True:
             line = self.readline()
-            if line.startswith('Eigenmode'):
+            if line.startswith("Eigenmode"):
                 break
             else:
                 natoms += 1
@@ -143,9 +153,9 @@ class vectorsSileSiesta(SileSiesta):
         nmodes = 1
         while True:
             line = self.readline()
-            if line.startswith('k            ='):
-                break 
-            elif line.startswith('Eigenvector'):
+            if line.startswith("k            ="):
+                break
+            elif line.startswith("Eigenvector"):
                 nmodes += 1
         self._nmodes = nmodes
 
@@ -155,11 +165,6 @@ class vectorsSileSiesta(SileSiesta):
     def _r_next_eigenmode(self):
         """Reads the next phonon eigenmode in the vectors file.
 
-        Parameters
-        ----------
-        ik: integer
-            the (python) k index of the next eigenstate.
-
         Returns
         --------
         EigenmodePhonon:
@@ -167,40 +172,37 @@ class vectorsSileSiesta(SileSiesta):
         """
         # Skip empty line at the head of the file
         self.readline()
-            
+
         k = list(map(float, self.readline().split()[2:]))
         if len(k) == 0:
             raise GeneratorExit
-        # Read first eigenvector index 
-        
+        # Read first eigenvector index
+
         # Determine number of atoms (i.e. rows per mode)
         state = np.empty((self._nmodes, self._natoms, 3), dtype=np.complex128)
         c = np.empty(self._nmodes, dtype=np.float64)
         for imode in range(self._nmodes):
-            line = self.readline()
-            if line.strip() == '':
-                # End of block for current k is indicated by empty line
-                break
-            
+            self.readline()
+
             # Read frequency
-            c[imode] = float(self.readline().split('=')[1].strip()) * _hc_eV
-            
+            c[imode] = float(self.readline().split("=")[1])
+
             # Read real part of eigenmode
             # Skip eigenmode header
             self.readline()
             for iatom in range(self._natoms):
-                line = self.readline().strip().lower()
-                state[imode,iatom,:].real = list(map(float, line.split()))
+                line = self.readline()
+                state[imode, iatom, :].real = list(map(float, line.split()))
 
             # Read imaginary part of eigenmode
             # Skip eigenmode header
             self.readline()
             for iatom in range(self._natoms):
-                line = self.readline().strip().lower()
-                state[imode,iatom,:].imag = list(map(float, line.split()))
+                line = self.readline()
+                state[imode, iatom, :].imag = list(map(float, line.split()))
 
-        info = dict(k=self._convert_k(k))
-        return EigenmodePhonon(state, c, **info)
+        info = dict(k=self._convert_k(k), parent=self.parent)
+        return EigenmodePhonon(state, c * _hc_eV, **info)
 
     def yield_eigenmode(self):
         """Iterates over the states in the vectors file
@@ -222,7 +224,7 @@ class vectorsSileSiesta(SileSiesta):
             # The loop in which the generator was used has been broken.
             pass
 
-    def read_eigenmode(self, k=(0,0,0), ktol=1e-4) -> EigenmodePhonon:
+    def read_eigenmode(self, k=(0, 0, 0), ktol=1e-4) -> EigenmodePhonon:
         """Reads a specific eigenstate from the file.
 
         This method iterates over the modes until it finds a match. Do not call
--- a/src/sisl/io/siesta/vibra.py
+++ b/src/sisl/io/siesta/vibra.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import numpy as np
 
+import sisl._array as _a
 from sisl import Geometry, Lattice, constant
 from sisl._internal import set_module
 from sisl.messages import warn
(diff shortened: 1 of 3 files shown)


nils-wittemeier:
@zerothi could you have a look at this and see if it makes sense to you?

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+        if rewind:
+            self.seek(0)
+        # Here we initialize the variables that will keep track of the state of the read.
+        # The process for identification is done on this basis:
+        #  _ik is the current (Python) index for the k-point to be read
+        #  _state is:
+        #        -1 : the file-descriptor has just been opened (i.e. in front of header)
+        #         0 : it means that the file-descriptor is in front of k point information
+        self._state = -1
+        self._ik = 0
```
zerothi, line 101: I don't know if this is even necessary, I guess the `read_eigenmode` could use the `SileBinder` method, in which case it isn't really necessary with all these details.

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+
+__all__ = ["vectorsSileSiesta"]
+
+_Bohr2Ang = unit_convert("Bohr", "Ang")
+_J2eV = unit_convert("J", "eV")
+_hc_eV = constant.h * constant.c * _J2eV
+
+@set_module("sisl.io.siesta")
+class vectorsSileSiesta(SileSiesta):
+    """Phonon eigenmode file"""
```
zerothi, line 27: Argument details so it is clear that `geometry`/`parent=Hessian` would be required.

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+        # Read the sizes relevant to the file.
+        # We also read whether there's only gamma point information or there are multiple points
+        if self._state == -1:
+            self._sizes = self._r_sizes()
+            self._state = 0
+
+        if close:
+            self._close()
+
+    def _r_sizes(self):
```
zerothi, line 121: probably this should be named `_r_next_sizes` since it does not seek to 0

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+
+        # Rewind
+        self.seek(pos)
+
+    def _r_next_eigenmode(self):
+        """Reads the next phonon eigenmode in the vectors file.
+
+        Parameters
+        ----------
+        ik: integer
```
zerothi, line 160: there is no `ik` parameter?

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+        state = np.empty((self._nmodes, self._natoms, 3), dtype=np.complex128)
+        c = np.empty(self._nmodes, dtype=np.float64)
+        for imode in range(self._nmodes):
+            line = self.readline()
+            if line.strip() == '':
+                # End of block for current k is indicated by empty line
+                break
+            
+            # Read frequency
+            c[imode] = float(self.readline().split('=')[1].strip()) * _hc_eV
```
zerothi, line 186: do the `_hc_eV` conversion in the end, faster.

Also `strip` shouldn't be needed

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+        # Skip empty line at the head of the file
+        self.readline()
+            
+        k = list(map(float, self.readline().split()[2:]))
+        if len(k) == 0:
+            raise GeneratorExit
+        # Read first eigenvector index 
+        
+        # Determine number of atoms (i.e. rows per mode)
+        state = np.empty((self._nmodes, self._natoms, 3), dtype=np.complex128)
```
zerothi, line 177: If I am not mistaken, then `state` should have final shape `nmodes, natoms * 3`, not a 3D array.

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+                break
+            
+            # Read frequency
+            c[imode] = float(self.readline().split('=')[1].strip()) * _hc_eV
+            
+            # Read real part of eigenmode
+            # Skip eigenmode header
+            self.readline()
+            for iatom in range(self._natoms):
+                line = self.readline().strip().lower()
```
zerothi, line 192: why `lower` and `strip`? Shouldn't do anything here since you do a split

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+                state[imode,iatom,:].real = list(map(float, line.split()))
+
+            # Read imaginary part of eigenmode
+            # Skip eigenmode header
+            self.readline()
+            for iatom in range(self._natoms):
+                line = self.readline().strip().lower()
```
zerothi, line 199: same here

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,257 @@
...
+                state[imode,iatom,:].imag = list(map(float, line.split()))
+
+        info = dict(k=self._convert_k(k))
```
zerothi, line 202: here `parent` would be nice, or you need a new internal method which creates the default info dict. I.e. in case where `parent` isn't defined, but `geometry` or `lattice` is

zerothi (changes requested):
Looks fine! Some comments here and there ;)

src/sisl/io/siesta/vibra.py
nils-wittemeier: Based on the way it is used in `DynamicalMatrix.eigenmode` I agree with you. What I am less sure about is the order should be. Based on the way the dynamical matrix is constructed from FC output, I understand that it should be [u_1x, u_1y, u_1z, u_2x, ...] right?

src/sisl/io/siesta/vibra.py
nils-wittemeier: removed

src/sisl/io/siesta/vibra.py
nils-wittemeier: removed

src/sisl/io/siesta/vibra.py
nils-wittemeier: I added parent. 

`parent` should always be defined unless neither `parent`, `geometry`, nor `lattice` are passed, because of:
```python
      def _setup(self, *args, **kwarg):
        # ...
        # ...   
        if parent is None:
            parent = geometry
        if parent is None:
            parent = lattice
 ```

File: src/sisl/io/siesta/vibra.py (hunk 1 above)
Comment (zerothi) [reply src/sisl/io/siesta/vibra.py:112]: yeah, not first priority, so disregard this comment. Thanks!

File: src/sisl/io/siesta/vibra.py (hunk 6 above)
Comment (zerothi) [reply src/sisl/io/siesta/vibra.py:182]: yes, you could create it like this, but when storing, simply do `state.reshape(nmodes, -1)`, then the correct size is passed.

**Comment by nils-wittemeier** [comment]
One thing that sill 'bothers' me, is that I am not 100% certain of the units of the eigenmodes stored by Vibra. I think it should be in Bohr. 

Once that is clarified and the units correctly converted to Angstrom in here.

**Comment by zerothi** [comment]
> One thing that sill 'bothers' me, is that I am not 100% certain of the units of the eigenmodes stored by Vibra. I think it should be in Bohr.
> 
> Once that is clarified and the units correctly converted to Angstrom in here.

I believe vibra uses default Siesta units, so it will be in Ry^2/Bohr, however the frequencies are in cm**-1. And everything is written *as-is*. The eigenmodes have unit-length, so I think they should be unit-less, no? They are basically the `zheev` values times `1/sqrt(mass_ia)`.

**Comment by zerothi** [comment]
Now all we need are tests ;)

According to https://github.com/zerothi/sisl-files/issues/14 feel free to add all files, I'll move them over to a new repo.

File: src/sisl/io/siesta/vibra.py [hunk 10]
```diff
@@ -0,0 +1,259 @@
...
+        self._parent = parent
+        self._geometry = geometry
+        self._lattice = lattice
+        if self._parent is None and self._geometry is None and self._lattice is None:
+
+            def conv(k):
+                if not np.allclose(k, 0.0):
+                    warn(
+                        f"{self.__class__.__name__} cannot convert stored k-points from 1/Ang to reduced coordinates. "
+                        "Please ensure parent=Hamiltonian, geometry=Geometry, or lattice=Lattice to ensure reduced k."
```
zerothi, line 86: this should be `DynamicalMatrix`

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+@set_module("sisl.io.siesta")
+class vectorsSileSiesta(SileSiesta):
+    """Phonon eigenmode file
+
+    Parameters
+    ----------
+    parent : obj, optional
+        a parent may contain a geometry, and/or a supercell
```
zerothi, line 32: mention that one would preferentially need a `DynamicalMatrix` structure

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+                break
+            else:
+                natoms += 1
+        self._natoms = natoms
+
+        # Determine number of modes
+        nmodes = 1
+        while True:
+            line = self.readline()
+            if line.startswith("k            ="):
```
zerothi, line 156: I suspect that newer versions would change the spacing, could we use a `regexp` with variyng space?

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+
+            # Read imaginary part of eigenmode
+            # Skip eigenmode header
+            self.readline()
+            for iatom in range(self._natoms):
+                line = self.readline()
+                state[imode, iatom, :].imag = list(map(float, line.split()))
+
+        info = dict(k=self._convert_k(k), parent=self.parent)
+        return EigenmodePhonon(state, c * _hc_eV, **info)
```
zerothi, line 205: the `c` unit from the file is cm**-1, will this bring it to `eV`? If so, could we name the variable `_cm1_eV` to make it a bit clearer?

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+            # Iterate over all eigenstates in the WFSX file, yielding control to the caller at
+            # each iteration.
+            while True:
+                yield self._r_next_eigenmode()
+            # We ran out of eigenstates
+        except GeneratorExit:
+            # The loop in which the generator was used has been broken.
+            pass
+
+    def read_eigenmode(self, k=(0, 0, 0), ktol=1e-4) -> EigenmodePhonon:
```
zerothi, line 227: could you use typing for `ktol` -> `ktol: float = 1e-4`, then you may delete the `float, optional` in the doc-str

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+            The threshold value for considering two k-points the same (i.e. to match
+            the query k point with the states k point).
+
+        See Also
+        --------
+        yield_eigenstate
+
+        Returns
+        -------
+        EigenstateElectron or None:
```
zerothi, line 247: eigenmode

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+        ----------
+        k: array-like of shape (3,), optional
+            The k point of the state you want to find.
+        ktol: float, optional
+            The threshold value for considering two k-points the same (i.e. to match
+            the query k point with the states k point).
+
+        See Also
+        --------
+        yield_eigenstate
```
zerothi, line 243: eigenmode

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+            # each iteration.
+            while True:
+                yield self._r_next_eigenmode()
+            # We ran out of eigenstates
+        except GeneratorExit:
+            # The loop in which the generator was used has been broken.
+            pass
+
+    def read_eigenmode(self, k=(0, 0, 0), ktol=1e-4) -> EigenmodePhonon:
+        """Reads a specific eigenstate from the file.
```
zerothi, line 228: eigenmode

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,259 @@
...
+
+        This method iterates over the modes until it finds a match. Do not call
+        this method repeatedly. If you want to loop eigenstates, use `yield_eigenstate`.
```
zerothi, line 231: eigenmodes

zerothi (changes requested):
minor comments

zerothi:
> > One thing that sill 'bothers' me, is that I am not 100% certain of the units of the eigenmodes stored by Vibra. I think it should be in Bohr.
> > Once that is clarified and the units correctly converted to Angstrom in here.
> 
> I believe vibra uses default Siesta units, so it will be in Ry^2/Bohr^2, however the frequencies are in cm**-1. And everything is written _as-is_. The eigenmodes have unit-length, so I think they should be unit-less, no? They are basically the `zheev` values times `1/sqrt(mass_ia)`.

Actually, this is a discussion point, should phonon modes always have the factor `1/sqrt(mass)` or should they just be normalized? What is your opinion here? Because we need to streamline this. :)

nils-wittemeier:
> > > One thing that sill 'bothers' me, is that I am not 100% certain of the units of the eigenmodes stored by Vibra. I think it should be in Bohr.
> > > Once that is clarified and the units correctly converted to Angstrom in here.
> > 
> > 
> > I believe vibra uses default Siesta units, so it will be in Ry^2/Bohr^2, however the frequencies are in cm**-1. And everything is written _as-is_. The eigenmodes have unit-length, so I think they should be unit-less, no? They are basically the `zheev` values times `1/sqrt(mass_ia)`.
> 
> Actually, this is a discussion point, should phonon modes always have the factor `1/sqrt(mass)` or should they just be normalized? What is your opinion here? Because we need to streamline this. :)

In most cases, I would think the scaled version with energies is needed. But, maybe it would make sense to implement this in a way that the EigenmodePhonon can return both? How could we handle those cases where the atomic masses are unknown? Maybe passing a parent object with masses has to be strictly mandatory?

nils-wittemeier:
Related to the discussion on the phase of the eigenvectors in vibra. Do we need to change the gauge when reading, to be consistent with the phonon modes obtained from the `DynamicalMatrix` object in sisl?

zerothi:
> Related to the discussion on the phase of the eigenvectors in vibra. Do we need to change the gauge when reading, to be consistent with the phonon modes obtained from the `DynamicalMatrix` object in sisl?

Not really (I think), the `gauge` variable is passed together with the object, so a simple `change_gauge` would be fine in any script, it simply is a noop if the gauge is correct.

zerothi:
> > Related to the discussion on the phase of the eigenvectors in vibra. Do we need to change the gauge when reading, to be consistent with the phonon modes obtained from the `DynamicalMatrix` object in sisl?
> 
> Not really (I think), the `gauge` variable is passed together with the object, so a simple `change_gauge` would be fine in any script, it simply is a noop if the gauge is correct.

Oh, you are correct, the gauge should be added to the `info` parameter! Yes, yes, yes! ;-)

src/sisl/io/siesta/vibra.py
```diff
@@ -14,12 +15,11 @@
 from ..sile import add_sile
 from .sile import SileSiesta
 
-
 __all__ = ["vectorsSileSiesta"]
 
 _Bohr2Ang = unit_convert("Bohr", "Ang")
 _J2eV = unit_convert("J", "eV")
-_hc_eV = constant.h * constant.c * _J2eV
+_cm1_eV = constant.h * constant.c * _J2eV
```
zerothi, line 22: You can do `unit_convert("cm^-1", "eV")`, clearer IMHO.

zerothi:
> Related to the discussion on the phase of the eigenvectors in vibra. Do we need to change the gauge when reading, to be consistent with the phonon modes obtained from the `DynamicalMatrix` object in sisl?

Related to the phase of the eigenvectors, what is your view on the phases in `sisl`?
I have had problems comparing vibra and `sisl`, but I think mainly because of the Wigner-Seitz reduction, which I don't have in sisl (it should be there but I don't have the competences and time to delve in to this).

zerothi:
What is missing here? Should I do something?

nils-wittemeier:
Sorry for going MIA. 

As far as I can tell, three things are missing:
- providing the correct gauge in the info dict 
- adding a test
- using your suggestion for the unit conversion

I'll update the PR tonight with missing bit.

nils-wittemeier:
Gauge was already mentioned in the info dict.

Units conversion is changed now, and I added a very basic test. 

Relevant files are added [here](https://github.com/zerothi/sisl-files/pull/16)

zerothi:
Example is in!

nils-wittemeier:
Should be ready now

src/sisl/io/siesta/vibra.py
```diff
@@ -0,0 +1,258 @@
...
+        elif isinstance(parent, Lattice):
+            lattice = parent
+        else:
+            geometry = parent.geometry
+
+        geometry = kwargs.get("geometry", geometry)
+        if geometry is not None and lattice is None:
+            lattice = geometry.lattice
+
+        lattice = kwargs.get("lattice", kwargs.get("sc", lattice))
```
zerothi, line 65: I'll merge and remove the `sc` argument

zerothi (approved):

PR #750, opened 2024-04-16
`norm2` as atom-resolved quantity

tfrederiksen:
Together with @shayanEdalatmanesh we thought it could be handy to have an option for `EigenstateElectron.norm2` to get this as an *atom-resolved* quantity. For instance, this may be relevant if one wishes to represent the probability density of a state across some structure with atoms hosting multiple orbitals (e.g. from a SIESTA calculation).

This PR makes a first suggestion in this direction.


src/sisl/physics/electron.py
```diff
@@ -1571,9 +1579,41 @@ def norm2(self, sum=True):
...
+        if projection == "sum":
             return self.inner(matrix=S)
-        return conj(self.state) * S.dot(self.state.T).T
+
+        elif projection == "orbital":
+            return conj(self.state) * S.dot(self.state.T).T
+
+        elif projection == "atom":
+            na = self.parent.na
+            no = self.parent.no
```
zerothi, line 1596: I guess this would break for `Spin(nc)`. since orbitals does not encompass spin.
A better estimate would be `len(self.parent)` which would work when `parent` is a `Hamiltonian` (still failing if the `parent=Geometry` and `state` is from an `NC` calculation. But...

src/sisl/physics/electron.py
```diff
@@ -1571,9 +1579,41 @@ def norm2(self, sum=True):
...
+
+            # build sparse matrix M to map from orbital-to-atom-resolved quantity
+            idx = self.parent.a2o(range(na))
```
zerothi, line 1599: I think it is faster to do:
```python
col_idx = np.arange(len(self.parent))
row_idx = self.parent.o2a(col_idx)
```
isn't this the same?

src/sisl/physics/electron.py
```diff
@@ -1571,9 +1579,41 @@ def norm2(self, sum=True):
...
+            data = np.ones(no)
```
zerothi, line 1600: use `len(parent)` here as well

zerothi (changes requested):
Looks good, some minor points that could be used to streamline it. And also moving/adding this functionality to the `State` object.

src/sisl/physics/electron.py
```diff
@@ -1551,17 +1553,23 @@ def Sk(self, format=None, spin=None):
...
 
-    def norm2(self, sum=True):
+    @deprecate_argument(
+        "sum",
+        "projection",
+        "argument sum has been deprecated in favor of projection",
+        "0.15",
+        "0.16",
+    )
+    def norm2(self, projection="sum"):
```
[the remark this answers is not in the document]
tfrederiksen: OK, I've added `projection` also to `state.py`. Why is `orbital` a wrong argument?

src/sisl/physics/electron.py
```diff
@@ -1571,9 +1579,41 @@ def norm2(self, sum=True):
...
+        if projection is True:
+            projection = "sum"
+        elif projection is False:
+            projection = "orbital"
+
+        if projection == "sum":
             return self.inner(matrix=S)
-        return conj(self.state) * S.dot(self.state.T).T
+
+        elif projection == "orbital":
```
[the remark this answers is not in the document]
tfrederiksen: done

src/sisl/physics/electron.py
tfrederiksen: changed

src/sisl/physics/electron.py
tfrederiksen: Yes, much simpler :)

src/sisl/physics/electron.py
tfrederiksen: This follows from the definition of `no` above.

docs/tutorials/tutorial_siesta_1.ipynb
```diff
@@ -250,7 +250,10 @@
     "    h2o.plot(axes=\"xy\", atoms_style={\"size\": normalize(n), \"color\": c})\n",
     "    for n, c in zip(\n",
     "        h2o.apply(\n",
-    "            es.norm2(sum=False), np.sum, mapper=partial(h2o.a2o, all=True), axis=1\n",
+    "            es.norm2(projection=\"orbital\"),\n",
+    "            np.sum,\n",
+    "            mapper=partial(h2o.a2o, all=True),\n",
```
tfrederiksen, line 255: This mapper is doing what could now also be achieved with `projection="atom"`, or?

tfrederiksen: Yes, these two plots can now simply be achieved with
```python
plots = [
    h2o.plot(axes="xy", atoms_style={"size": normalize(n), "color": c})
    for n, c in zip(es.norm2(projection="atom"), ("red", "blue"))
]
```

src/sisl/physics/electron.py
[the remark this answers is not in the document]
zerothi: `orbital` is correct in the `electron.py`, but the `State` object is generic for both electrons and phonons. So then `orbital` is meaningless.

docs/tutorials/tutorial_siesta_1.ipynb
zerothi: Great! Yes!

zerothi:
When you are done, I would like to fix one of my comments, which was wrong ;) (for NC case)

zerothi:
I have realized some inconsistencies, so I would like to extend your contribution in a 2nd PR. I'll merge and then work on extending it! I'll ask for your input there! ;)

Thanks Thomas!

tfrederiksen:
> I have realized some inconsistencies, so I would like to extend your contribution in a 2nd PR. I'll merge and then work on extending it! I'll ask for your input there! ;)

Great, thanks!

PR #751, opened 2024-04-17
made State.inner a very flexible matrix product

zerothi:
It can now be used to calculate state norms
and is basically the backbone of a lot of the methods now. It can work in 3 different ways:

- diag (diagonal inner state products)
- matrix (off-diagonal inner states)
- atoms|state|orbital (only diagonal inner states but atom resolved and/or orbital resolved.

I don't know if this complicates the usage a great deal. It makes much things simpler.

@tfrederiksen could you have a look here. I don't know if it is too much, or whether the names makes sense.

 - [x] Closes #x

### DIFF
--- a/src/sisl/physics/electron.py
+++ b/src/sisl/physics/electron.py
@@ -3,7 +3,7 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 from __future__ import annotations
 
-from typing import Literal
+from typing import Literal, Optional
 
 from sisl.messages import deprecate_argument
 
@@ -1505,28 +1505,26 @@ class _electron_State:
     # pylint: disable=E1101
     __slots__ = []
 
-    def __is_nc(self):
+    def __is_nc(self, spin: Optional[Spin] = None):
         """Internal routine to check whether this is a non-colinear calculation"""
+        if spin is not None:
+            return not spin.is_diagonal
         try:
             return not self.parent.spin.is_diagonal
         except Exception:
             return False
 
-    def Sk(self, format=None, spin=None):
+    def Sk(self, format=None):
         r"""Retrieve the overlap matrix corresponding to the originating parent structure.
 
         When ``self.parent`` is a Hamiltonian this will return :math:`\mathbf S(\mathbf k)` for the
-        :math:`\mathbf k`-point these eigenstates originate from
+        :math:`\mathbf k`-point these eigenstates originate from.
 
         Parameters
         ----------
         format : str, optional
            the returned format of the overlap matrix. This only takes effect for
            non-orthogonal parents.
-        spin : Spin, optional
-           for non-colinear spin configurations the *fake* overlap matrix returned
-           will have halve the size of the input matrix. If you want the *full* overlap
-           matrix, simply do not specify the `spin` argument.
         """
         if format is None:
             format = self.info.get("format", "csr")
@@ -1544,14 +1542,9 @@ class _electron_State:
                         opt[key] = val
                 return self.parent.Sk(**opt)
 
-        if self.__is_nc():
-            n = self.shape[1] // 2
-        else:
-            n = self.shape[1]
+        n = m = self.shape[1]
         if "sc:" in format:
             m = n * self.parent.n_s
-        else:
-            m = n
 
         return _FakeMatrix(n, m)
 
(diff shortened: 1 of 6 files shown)


src/sisl/physics/state.py
```diff
@@ -423,46 +440,24 @@ def norm(self):
         "0.15",
         "0.16",
     )
-    def norm2(self, projection: Literal["sum", "orbital", "atom", "none"] = "sum"):
+    def norm2(self, projection: Literal["sum", "atoms", "state"] = "sum"):
```
tfrederiksen, line 443: Plural `atoms` but singular `state`?

tfrederiksen (approved):
Elegant!

src/sisl/physics/state.py
```diff
@@ -423,46 +440,24 @@ def norm(self):
...
         r"""Return a vector with the norm of each state :math:`\langle\psi|\psi\rangle`
 
         Parameters
         ----------
         projection :
-           whether to compute the norm per state as a single number or as orbital-/atom-resolved quantity
+           whether to compute the norm per state as a single number, atom-resolved or none quantity
```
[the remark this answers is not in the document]
zerothi: fixed

src/sisl/physics/electron.py
```diff
@@ -1562,7 +1555,7 @@ def Sk(self, format=None, spin=None):
         "0.15",
         "0.16",
     )
-    def norm2(self, projection: Literal["sum", "orbital", "atom", "none"] = "sum"):
+    def norm2(self, projection: Literal["sum", "orbitals", "state", "atoms"] = "sum"):
```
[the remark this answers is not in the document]
zerothi: fixed

zerothi:
I think we are there now! Anything else?

src/sisl/physics/state.py
```diff
@@ -603,7 +598,19 @@ def outer(self, ket=None, matrix=None):
...
+        "projection",
+        "argument diag has been deprecated in favor of projection",
+        "0.15",
+        "0.16",
+    )
+    def inner(
+        self,
+        ket=None,
+        matrix=None,
+        projection: Literal["diag", "atoms", "state", "none"] = "diag",
```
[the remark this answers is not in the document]
zerothi: done

src/sisl/physics/state.py
```diff
@@ -614,17 +622,26 @@ def inner(self, ket=None, matrix=None, diag=True):
...
-           only return the diagonal matrix :math:`\mathbf A_{ii}`.
+        projection:
+            how to perform the final projection.
+            This can be used to sum specific sub-elements, return the diagonal, or the
+            full matrix.
+
+            * ``diag`` only return the diagonal of the inner product
+            * ``matrix`` return a matrix of inner products, also the off-diagonals
+            * ``atoms`` only do inner products for same states, summed over atoms on the state
+            * ``states`` only do inner products for same states
```
tfrederiksen, line 633: I find this a bit confusing. `matrix` would give the state-resolved quantity of shape `(ns, ns)`, where `ns = len(self)` while this mode returns a quantity of shape `(ns, no`). Shouldn't it rather be labeled `orbitals` (each state is resolved orbitally, no off-diagonals between states)?

zerothi:
I think we are there! I will merge! Thanks!

src/sisl/physics/state.py
zerothi: the problem is that the `state` code is generic, so no orbitals. For phonons, this is cartesian directions.
I don't really know what to name it? Suggestions?
- `states`
- `dof`
- ?

tfrederiksen: Maybe `basis` would be a better term to avoid confusion with the "states" themselves?

The docstring could read:
* `atoms` only do inner products for individual states, but return them atom-resolved 
* `basis` only do inner products for individual states, but return them basis-resolved

In the most general case, could one also have offdiagonals (*between* states) resolved per atom- or basis?

zerothi: much better, thanks! I'll amend!

For the general case, I think that would result in too much information, no? `(nstates, nbasis, nbasis)` ? For 1000 orbital systems, this results in a matrix of 14 GB! I'll leave it out for now. But thanks for the consideration.

PR #754, opened 2024-04-22
Reading spin densities from CHGCAR

tfrederiksen:
The spin densities were not correctly read from CHGCAR, because after the augmentation occupancies *an additional block of values* (apparently one per atom) is found on file.

I could not determine what this extra block refers to, as it is not explicitly mentioned in the documentation from the [VASP forum](https://www.vasp.at/wiki/index.php/CHGCAR), which just mentions:
> For magnetic calculations, the CHGCAR file contains additional data blocks for the magnetization. In particular, for spin-polarized calculations ([ISPIN](https://www.vasp.at/wiki/index.php/ISPIN)=2), the first set contains the total charge density (spin up + spin down) and the second one is the magnetization density (spin up - spin down):
> 
> - Structure
> - FFT-grid dimensions
> - Charge density times FFT-grid volume (spin up + spin down)
> - Augmentation occupancies
> - FFT-grid dimensions
> - Magnetization density (spin up - spin down)
> - Augmentation occupancies

This PR resolves this issue and adds some further checks.

### DIFF
--- a/src/sisl/io/vasp/chg.py
+++ b/src/sisl/io/vasp/chg.py
@@ -32,15 +32,28 @@ class chgSileVASP(carSileVASP):
         Parameters
         ----------
         index : int or array_like, optional
-           the index of the grid to read. For a spin-polarized VASP calculation 0 and 1 are
-           allowed, UP/DOWN. For non-collinear 0, 1, 2 or 3 is allowed which equals,
+           the index of the grid to read.
+           For spin-polarized calculations, 0 and 1 refer to the charge (spin-up plus spin-down) and
+           magnetitization (spin-up minus spin-down), respectively.
+           For non-collinear calculations, 0 refers to the charge while 1, 2 and 3 to
+           the magnetization in the :math:`\sigma_1`, :math:`\sigma_2`, and :math:`\sigma_3` directions, respectively.
            TOTAL, x, y, z charge density with the Cartesian directions equal to the charge
-           magnetization. For array-like they refer to the fractional
-           contributions for each corresponding index.
+           magnetization.
+           For array-like they refer to the fractional contributions for each corresponding index.
         dtype : numpy.dtype, optional
            grid stored dtype
         spin : optional
            same as `index` argument. `spin` argument has precedence.
+
+        Examples
+        --------
+        Read the spin polarization from a spin-polarized CHGCAR file
+
+        >>> fh = sisl.get_sile('CHGCAR')
+        >>> charge = fh.read_grid()
+        >>> spin = fh.read_grid(1)
+        >>> up_density = fh.read_grid([0.5, 0.5])
+        >>> assert np.allclose((charge + spin).grid, up_density.grid)
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


src/sisl/io/vasp/chg.py
```diff
@@ -90,8 +90,12 @@ def read_grid(self, index=0, dtype=np.float64, **kwargs) -> Grid:
                         while j < occ:
                             j += len(rl().split())
                         line = rl()
+                    # read over an additional block with geom.na entries???
+                    j = len(line.split())
+                    while j < geom.na:
+                        j += len(rl().split())
                 # one line of nx, ny, nz
-                rl()
+                assert len(rl().split()) == 3
```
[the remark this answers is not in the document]
tfrederiksen: Good idea! Will do.

src/sisl/io/vasp/chg.py
```diff
@@ -90,8 +90,12 @@ def read_grid(self, index=0, dtype=np.float64, **kwargs) -> Grid:
                         while j < occ:
                             j += len(rl().split())
                         line = rl()
+                    # read over an additional block with geom.na entries???
+                    j = len(line.split())
+                    while j < geom.na:
+                        j += len(rl().split())
```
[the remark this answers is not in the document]
zerothi: ok, that is fine then! :)

tfrederiksen:
> Could we add a link to the documentation or something, so it is clear what the indices of `grid` values refers too?

Where would you add a link? The docstring already describes the `index` and `spin` arguments, no?

tfrederiksen:
> It seems to me that line 35 is then wrong, no?

Oh, I see. Yes, indeed `index=1` would give the spin density and not the down-component as currently stated. By the way, why two keywords `index/spin` for the same thing? Wouldn't it not be better to have just one?

src/sisl/io/vasp/chg.py
```diff
@@ -36,7 +36,7 @@ def read_grid(self, index=0, dtype=np.float64, **kwargs) -> Grid:
            For spin-polarized calculations, 0 and 1 refer to the charge (spin-up plus spin-down) and
            magnetitization (spin-up minus spin-down), respectively.
            For non-collinear calculations, 0 refers to the charge while 1, 2 and 3 to
-           the magnetization in the :math:`\sigma_1`, :math:`\sigma_2`, and :math:`\sigma_3` directions, respectively.
+           the magnetization in the :math:`\sigma_x`, :math:`\sigma_y`, and :math:`\sigma_z` directions, respectively.
```
tfrederiksen, line 39: This is not generally correct, only when the user has `SAXIS = 0 0 1` in the VASP calculation (the default).

zerothi: ah! Thanks for always double checking my things! :) I will revert, and reference the `SAXIS` key.

zerothi: fixed!

tfrederiksen: Perfect!

PR #763, opened 2024-04-26
changed phonon displacement shape

zerothi:
@nils-wittemeier and @monx-ne
I am planning on changing this for consistencies.

What do you think of this re-ordering of the shapes? It seems more natural to me?


nils-wittemeier:
I am in favor of this change. It is much more natural.

zerothi:
That's good enough for me!

PR #767, opened 2024-05-06
fixed pbc usage across sisl, fixes #764

### LINKED ISSUE #764: `translate2uc` should use boundary conditions (?)
Here in `Geometry.translate2uc`, `nsc` is used to understand which axes to translate:

https://github.com/zerothi/sisl/blob/8d6da596972b77451376b5492c42ba490f1bec2c/src/sisl/_core/geometry.py#L1741

I think it should check for periodic boundary conditions instead, no?

zerothi:
Now pbc usage has been moved across the
code base (I hope everything is managed).
I have added a setter for pbc to enable
fast setting pbc. It will silently ignore
any dimensions not specified.
This may be used as a shorthand for
lattice.set_boundary_condition(...)

@pfebrer could you please review this? I hope it fixes the issue!

 - [x] Closes #764 

### DIFF
--- a/src/sisl/utils/misc.py
+++ b/src/sisl/utils/misc.py
@@ -10,11 +10,12 @@ import inspect
 import operator as op
 import re
 import sys
+from collections.abc import Callable, Iterable, Iterator
 from math import pi
 from numbers import Integral
-from typing import Union
+from typing import Any, Union
 
-__all__ = ["merge_instances", "str_spec", "direction", "angle"]
+__all__ = ["merge_instances", "str_spec", "direction", "listify", "angle"]
 __all__ += ["iter_shape", "math_eval", "allow_kwargs"]
 __all__ += ["import_attr", "lazy_import"]
 __all__ += ["PropertyDict", "NotNonePropertyDict"]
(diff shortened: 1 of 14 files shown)


pfebrer:
There are some failing tests, but I don't know if they are related to this PR.

Apart from that, could we set `nsc=[1,1,1]` in the tests that check the modified functions? E.g. here https://github.com/zerothi/sisl/blob/43f3c2904783af2c11932058496378113ad54bf7/src/sisl/_core/tests/test_geometry.py#L1812C1-L1816C44

zerothi:
Thanks! This actually uncovered a bit more problematic issues, I'll ask for another review once completed. Thanks!

zerothi:
Ok, it got complicated, but the pbc was never really copied over, and this might have inflected problems here and there. I think it should be done now.!

src/sisl/utils/misc.py
```diff
@@ -220,6 +221,71 @@ def str_spec(name):
...
+    >>> Listify(tuple)(1)
+    (1,)
+
+    It can greatly improve readability with `map` constructs:
+    >>> map(lambda x, [1]) | listify
+    [1]
+
+    Notes
+    -----
+    If using this to convert to `tuple` instances ``Iterfy(tuple)``,
```
pfebrer, line 253: `Iterfy` -> `Listify`

src/sisl/utils/misc.py
```diff
@@ -220,6 +221,71 @@ def str_spec(name):
...
+            cls = self._cls
+        if isinstance(arg, Iterable):
+            if isinstance(arg, cls):
+                return arg
+            return cls(arg)
+        return cls([arg])
+
+    def __ror__(self, arg: Any) -> IterableAny:
+        """Allow piping of function calls"""
+        return self(arg)
```
pfebrer, line 280: Is this a good idea? 

Is `|` used for piping elsewhere with python? 

In the description you say:

> It provides also an easy mechanism for "piping" content to
>     a function call (for better readability).

It is better readability if `|` would be used for piping in general in python, but since it is used for `or` operations I think it can be more confusing than readable :sweat_smile:

zerothi: Hmm, I think here it is quite clear, have you seen this package https://pypi.org/project/pipe/
I think the expressiveness without parenthesis makes it more clear, although I can see it a bit confusing if you are not familiar with pipes on Unix. In any case it is more used internally here.

pfebrer: I searched for pipes in python and I found that package. But calling a function with `|` is not something that is considered in the specification of the syntax.

Even if you are used to `|` meaning "pipe" in bash, it's still confusing to find it in python. I wouldn't think about it if I see it in python code. I would think that listify is some kind of fallback if the expression on the left is false 😅

zerothi: ha, I see. I think I'll try and keep it, the amount of parenthesis gets heavily reduced with this ;)

PR #772, opened 2024-05-20
Removed deprecated argument `diag->projection`

tfrederiksen:
Changed deprecated keyword `diag` with `projection=matrix`.


zerothi (approved):
Thanks!

PR #773, opened 2024-05-20
Dihedral angles

tfrederiksen:
The dihedral angle is a common quantity to characterize molecular conformations, see https://en.wikipedia.org/wiki/Dihedral_angle. This PR enables its computation for a geometry and 4 atom indices (or a list of 4-indices for multiple calculations).


src/sisl/_core/geometry.py
```diff
@@ -1685,6 +1685,85 @@ def angle(
...
+                f"{self.__class__.__name__}.dihedral requires atoms index of shape (4,) or (N, 4)"
+            )
+
+        angles = []
+        for idx in atoms:
+            # The 2 planes are defined by
+            #  r0, r1, r2
+            # and
+            #  r1, r2, r3
+            v = np.diff(self.xyz[idx], axis=0)
```
tfrederiksen, line 1747: Should we rename to `u` to be consistent with the presented formula?

zerothi: Yes, I have another proposition, I'll play a bit with it, but perhaps we should do it the other way around?

So one can do
```
a0 = [0, 1]
a1 = [2, 3]
a2 = [4, 5]
a3 = [6, 7]
geom.dihedral([a0, a1, a2, a3])
```
that seems a bit more natural to me, rather than the current way? What is your reasoning for the other way?

tfrederiksen: I thought about that too. I ended up with the current order because I wanted to calculate first one angle, then decided to add another one, and it was easier to write the next four indices together in a list. For me, getting `N` angles from an `(N, 4)` input array seems a natural ordering.

zerothi: yes, but how would you use it? The internal algorithm does not matter here ;) That is trivially fixed ;)

zerothi: the return dimension would look the same, regardless.

tfrederiksen: Sure, I don't have a strong opinion on the ordering. `(4, N)` input would be fine too, as long as the documentation is clear about it.

zerothi: hmm, I can also see that one would do this:
```python
d1 = [0, 2, 4, 6]
d2 = [1, 3, 5, 7]
geom.dihedral([d1, d2])
```
hmm, now I am in doubt, what do you think would be most natural?

tfrederiksen: You can see how I would use it in the tests.

zerothi: ok, lets keep it as is, I will amend the calculation to be done without the loop.

zerothi:
I'll merge once CI checks, let me know if there is anything else! Thanks!

Issue #340, opened 2021-05-23
Discussion on API to initialize MultiplePlot.


pfebrer:
Discussion so far on https://github.com/zerothi/sisl/pull/339 (with @zerothi, @tfrederiksen ):

**Pol:**

```python
grid.plot(represent=["imag", "real"], varying="represent")
```
This would draw both the imaginary and real contributions in the same plot, while

```python
grid.plot(represent=["imag", "real"], subplots="represent")
```
would create a subplot for each of them (warning, 3D subplots are not possible yet), and

```python
grid.plot(represent=["imag", "real"], animate="represent")
```
would create an animation where the first step displays the real part and the second step displays the imaginary part.

So I'd say it is unnecessary to implement it inside GridPlot. It would also make the code more complex. But I'm open to discuss it :)

PD: I don't like the keyword varying at all, if you have a better suggestion it would be great!

**Nick:**

Ok! :)

What about `variables`?

Don't know if there are multiple variables, or always 1, in the latter case `variable`?

**Pol:**

```python
grid.plot(setting1=[0,1], setting2=[2,3], varying=["setting1", "setting2"])
```
is allowed. It creates a plot with `(setting1, setting2) = (0, 2)` and another one where `(setting1, setting2) = (1, 3)`.

Also

```python
grid.plot(varying={"setting1": [0,1], "setting2": [2,3]})
```
is equivalent.

So, `variables`? I was looking for some word that makes it clear that they are going to be drawn in the same plot. I find `variables` better than `varying`, but still not ideal :) Also it is important that the name does not clash with names that people could give to settings, and in that respect `variables` seems more dangerous to me. Taking into account that I think I like `variable` better (as an adjective, not as a noun :smile:  ).

**Thomas:**

Have you considered having a single parameter here to control what kind of representation to be produced, e.g. style : `{"single", "multiple", "subplots", "animation"}`? To me this seems more intuitive from the user perspective, but perhaps it presents some limitations?

**Pol:**
Yes, but then it is extremely difficult to determine which are the settings that need to be iterated. Unless you have some idea about it that I didn't think of. You can't check for lists, because there might be settings that accept lists as a value.

**Nick:**

Perhaps it would be useful to use tuples or callable in `setting` which are then elements in the varying space.
This is a bit unintuitive but follows the pandas design where list and tuples have different meanings. It would save an argument.

I think @tfrederiksen has a point, something like:
```
[...]

PR #777, opened 2024-05-25
WIP: Facilitate the creation of combined plots

pfebrer:
Since #340, there was the discussion of how to make the API for multiple plots as convenient as possible. This PR attempts to solve it.

In `nodify`, I introduced the concept of a `Batch`. If you pass a `Batch` as an input of a workflow, the workflow will also return a `Batch`. The most important thing is that **only the parts of the workflow that depend on the batch are computed for each item**, the rest are shared, saving time and memory.

This is automatically applicable to the plots, since they are workflows:

```python
from nodify import Batch
import sisl

graphene = sisl.geom.graphene()

# Create a batch
scales = Batch(1, 2, 3)
# Then call the plotting function, but passing the batch as an input on several arguments
g_plot = graphene.plot(axes="xy", show_cell=False, atoms_scale=scales, bonds_scale=scales)
```
`g_plot.get()` will now return a batch, which contains the three figures.

Now the question is what is the best interface for sisl users. In this PR I have changed slightly the functions that merge plots so that they can support batches. Now one can do for example:

```python
sisl.viz.subplots(g_plot)
```
newplot (71)

One can also change the way in which batches interact with each other. This is defined by the node's context. In this case they were zipped, but one could ask to take the product as well:

```python
from nodify import temporal_context

with temporal_context(batch_iter="product"):
    sisl.viz.subplots(g_plot).show()
```

newplot (72)

I have decided that the user must use the merging functions, and `g_plot.show()` does not show anything, since there is no indication of how the user wants to merge the plots. Also if we were to automatically merge them it would not work in cases like:

```python
# Create a batch
backends = Batch("plotly", "matplotlib")

g_plot =  graphene.plot(axes="xy", backend=backends)
```

Where the backends are different. In that case one can do:
```python
for plot in g_plot.get():
    plot.show()
```

Screenshot from 2024-05-25 18-10-12

Do you think this is fine? Is there something that you would change to make it more convenient to use?

Thanks!

### DIFF
--- a/src/sisl/viz/plots/merged.py
+++ b/src/sisl/viz/plots/merged.py
@@ -35,8 +35,20 @@ def merge_plots(
         which controls the arrangement ("rows", "cols" or "square").
     """
 
+    figures_to_pass = []
+    for fig in figures:
+        if isinstance(fig, Plot):
+            fig = fig.get()
+            if isinstance(fig, Figure):
+                figures_to_pass.append(fig)
+            else:
+                # This has to be a batch
+                figures_to_pass.extend(list(fig))
+        else:
+            figures_to_pass.append(fig)
+
     plot_actions = combined(
-        *[fig.plot_actions for fig in figures],
+        *[fig.plot_actions for fig in figures_to_pass],
         composite_method=composite_method,
         **kwargs,
     )


pfebrer:
When we agree, I will change the docs that show how to combine plots.

zerothi:
I think most people will use grouped plots to combine different types of plots (i.e. not just a variable).

For instance, geometry plots + PDOS scaling of atomic sizes, vs. a fat-bands-plot to the right. Would the `Batch` method aid in this? I can however, see that the batch is a nice way to figure out a parameter size.

I think the `batch.show()` should indicate an error of some sorts, to signal that it can't accomblish it, possibly advocating the use of `is_batch_plot` and `sisl.viz.subplots` or similar.

Is there an easy way to query whether a plot is a `Batch` plot (by not `isinstance(plot, list)`), at least there should be something like `is_batch_plot` (or similar), just to make it easy to distinguish, if that is a wrapper for `isinstance(plot, list)` is fine.

pfebrer:
> I think most people will use grouped plots to combine different types of plots (i.e. not just a variable).
> 
> For instance, geometry plots + PDOS scaling of atomic sizes, vs. a fat-bands-plot to the right. Would the Batch method aid in this? I can however, see that the batch is a nice way to figure out a parameter size.

Yes, makes sense. In that case batches are not the right tool. You just build a list of plots and merge them (as it is done until now).

Batches just help in getting the result of a workflow for multiple inputs while sharing the common parts of the computation. I.e. they don't help if you want to execute different workflows.

There are some typical cases where one might want to do just batch an input. For example, plotting multiple wavefunctions:

```python
from nodify import Batch

H.plot.wavefunction(i=Batch(4, 5, 6, 9))
```

will create plots for the 4 wavefunctions doing one single diagonalization.

>  I think the batch.show() should indicate an error of some sorts, to signal that it can't accomblish it, possibly advocating the use of is_batch_plot and sisl.viz.subplots or similar.
> Is there an easy way to query whether a plot is a Batch plot (by not isinstance(plot, list)), at least there should be something like is_batch_plot (or similar), just to make it easy to distinguish, if that is a wrapper for isinstance(plot, list) is fine.

Makes sense. Something like `plot.is_batched` could work?.

pfebrer:
```python
from nodify import Batch, temporal_context

with temporal_context(batch_iter="product"):
    H.plot.wavefunction(i=Batch(4, 5, 6, 9), represent=Batch("real", "imag"))
```

would generate 8 plots with 1 diagonalization and 4 grid projections.

zerothi:
> Another example:
> 
> ```python
> from nodify import Batch, temporal_context
> 
> with temporal_context(batch_iter="product"):
>     H.plot.wavefunction(i=Batch(4, 5, 6, 9), represent=Batch("real", "imag"))
> ```
> 
> would generate 8 plots with 1 diagonalization and 4 grid projections.

what does `temporal_context` mean? Isn't temporal implicit in contexts? I would think you should name this differently, `nodify_context`... or?

Also, could one create product batches *and* zip batches. I.e.
```python
batch1 = Batch(1,2)
batch2 = Batch(2,3)
batch3 = Batch(*range(4))
with nodify_context(batch_iter={(batch1, batch2): "product", "*": "zip"})
```
probably bad names here, something that *binds* them together?

pfebrer:
> what does temporal_context mean? Isn't temporal implicit in contexts? I would think you should name this differently, nodify_context... or?

Hmm interesting, so `context` here is referring to the nodes context, which determines how the computation is performed. Each node has a context attached to it and that is permanently bound to the node (although you can update it). But still I agree that probably `temporal` is not needed here, since the `with` statement already indicates "temporality" as you say. Maybe just `context` would suffice:

```python
from nodify import context

with context(batch_iter="product"):
    ...

#or
import nodify

with nodify.context(batch_iter="product"):
    ...
```

> Also, could one create product batches and zip batches. I.e.
> 
> batch1 = Batch(1,2)
> batch2 = Batch(2,3)
> batch3 = Batch(*range(4))
> with nodify_context(batch_iter={(batch1, batch2): "product", "*": "zip"})
> probably bad names here, something that binds them together?

I don't want to complicate things for now. I always keep in mind that these things should be modifiable from the GUI :sweat_smile: 

I thought about it but then I realised that programatically one could always create complicated combinations and then zip the batches. E.g.:

```python
batch1 = Batch(1,1,2, 2)
batch2 = Batch(2,3,2,3)
batch3 = Batch(*range(4))

# Now when batches are zipped it will result in the same situation as what you were proposing
```

zerothi:
I don't see the final comment in your post? Maybe it got truncated, or something?

zerothi:
Is there something missing here? Should we get this in?

pfebrer:
I'll try to check if everything is ok. I recall that the thing missing was documentation and tests

Issue #775, opened 2024-05-22
Merging Atom/Orbital etc.


zerothi:
**Describe the feature**
It would be nice if one could deterministically *merge* two atoms/orbitals etc. in a consistent way.
The usecase would be when a users has a simple geometry, but wants to read in from other sources which contains more information (but not necessarily everything).

The difficult thing here is the orbitals, should one be able to merge an `Orbital` and a `AtomicOrbital`, and how to select these things.

However, having this centralized in the objects would make these things much simpler.

zerothi:
For orbital:
```python
    def merge(self, other) -> Orbital:
        """ Merges two orbitals into one, with a preference of `self` """
        cls = self.__class__
        s = self.__getstate__()
        o = other.__getstate__()

        def inst_merge(typ, op):
            for k, v in s.items():
                if isinstance(v, typ):
                    s[k] = op(v, o[k])
        def key_merge(key, op):
            for k, v in s.items():
                if k == key:
                    s[k] = op(v, o[k])
        
        if isinstance(other, cls):
            inst_merge(float, max)
            # tag is defined for this orbital, no change
            key_merge("tag", lambda s, o: v if v else o)
            return cls(**s)
```

for atom
```python
    def merge(self, other: Atom) -> Atom:
        """Merges two atoms into a single one, the calling atom will have precedence

        When merging there will be a couple of rules that will apply:

        1. When in doubt, the calling atom will have precedence.
        2. If the two atoms have the same `Z` value, they can be merged,
           otherwise, `self` will be returned *as-is*.
        3. The tag is only overwritten if not set.
        4. There is no way to decide which mass to choose. It will *always*
           take `self`.
        5. The orbitals requires explicit handling, the order will prefer
           `self`, but if there is more information in `other` those will
           be copied over.
        
        
        Parameters
        ----------
        other : Atom
            the other atom to merge into a single one
        """
        if self.Z != other.Z:
            return self
        mass = self.mass
        tag = self._tag or other._tag
        
        orbs = []
        for sorb in self:
            oorb = other[sorb.name()]
            if oorb is None:
                orbs.append(sorb)
                continue
            if isinstance(sorb, type(oorb)):
                # same type, now only check

```

PR #778, opened 2024-05-28
Fix the basis when reading from XV

pfebrer:
When reading the geometry from an XV file using `fdf.read_geometry(output=True)`, the basis was not correctly set. It was completely broken, I don't know why I (or anyone) hadn't noticed until now :sweat_smile:

### DIFF
--- a/src/sisl/io/siesta/_help.py
+++ b/src/sisl/io/siesta/_help.py
@@ -170,9 +170,8 @@ def _replace_with_species(basis, ref_basis):
     """
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
-        for atom, _ in basis.iter(True):
-            basis.replace(atom, ref_basis[atom.Z - 1])
-        basis.reduce(inplace=True)
+        for at in basis.atom:
+            basis.replace_atom(at, ref_basis.atom[at.Z - 1])
 
 
 def _fc_correct(fc, trans_inv=True, sum0=True, hermitian=True):
--- a/src/sisl/io/siesta/fdf.py
+++ b/src/sisl/io/siesta/fdf.py
@@ -1457,7 +1457,7 @@ class fdfSileSiesta(SileSiesta):
         f = self.dir_file(self.get("SystemLabel", default="siesta") + ".XV")
         _track_file(self._r_geometry_xv, f)
         if f.is_file():
-            basis = self.read_basis()
+            basis = self.read_basis(sort_by_fdf_appearance=False)
             if basis is None:
                 geom = xvSileSiesta(f).read_geometry(species_as_Z=False)
             else:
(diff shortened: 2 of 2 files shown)


pfebrer:
Minimal example:

[XV_fail.zip](https://github.com/zerothi/sisl/files/15475837/XV_fail.zip)

tfrederiksen:
I can confirm this issue. For one of my systems I find: 
```
# read from fdf
g = sisl.get_sile("./RUN.fdf").read_geometry()
print(g.atoms.atom)
[<sisl.Atom H.opt88, Z=1, M=1.01, maxR=2.8520000000000043, no=5>, 
 <sisl.Atom Au.blyp, Z=79, M=196.97, maxR=3.328400000000018, no=15>,
 <sisl.Atom Aus.blyp, Z=79, M=196.97, maxR=4.7660999999999865, no=16>,
 <sisl.Atom C.blyp, Z=6, M=12.01, maxR=2.954100000000009, no=13>]

# read from XV
g = sisl.get_sile("./RUN.fdf").read_geometry(True)
print(g.atoms.atom)
[<sisl.Atom H.opt88, Z=1, M=1.01, maxR=2.8520000000000043, no=5>,
 <sisl.Atom Li, Z=3, M=6.941, maxR=-1.0, no=1>]
```
The first case has the correct atoms (and input coordinates), the latter wrong atoms (but correct, relaxed coordinates).

pfebrer:
Actually the fix here isn't fully correct either. 

It seems that using `species_as_Z` is not a good idea. When reading from the fdf file the atoms list is not sorted by species index but by appearance.

So something like:

```
In the fdf file

Species:
1 Au
2 O

Coordinates:
X X X 2
X X X 1
```

```
In the XV file

Coordinates:
2 8   X X X
1 79 X X X
```

Will lead to reversed species if you read from `fdf.read_geometry(output=True)`

So either:
   - Species must be sorted when reading the basis from fdf.
   - Atoms from the XV file must be retreived from the basis based on their atomic number, not their index.

I think the second one makes more sense, but I don't know if other files that use `_replace_with_species` also contain the atomic numbers.

pfebrer:
Hmm getting the basis by Z won't work either if you have multiple atoms with the same Z, as Thomas' example :sweat_smile: 

So the only option is that `fdf.read_basis()` returns the atoms sorted by species.

pfebrer:
Ok, now everything should work fine. You might want to change the name of the argument that I introduced on `read_basis` (`sort_by_fdf_appearance`) :sweat_smile:

zerothi:
> Ok, now everything should work fine. You might want to change the name of the argument that I introduced on `read_basis` (`sort_by_fdf_appearance`) 😅

I'll probably revert that one, it should be handled differently, the whole idea of using species indices like that would be wrong...
Other parts of sisl uses `atoms` argument to pass a reference basis, here I would do the same (wait for it ;))

Ideally this issues should be resolved with some generic methods that resolves merges of geometries, atoms, orbitals, in a very strict manner, see e.g. #775.
This kind of procedure is used all-over, so it could be streamlined quite a bit I would assume.

zerothi:
Thanks for discovering this!

Could you both try out the #779 branch? It should solve this problem (and more!). Also added some more tests.

zerothi:
Closed via #779

PR #780, opened 2024-05-29
Allow dash to be set in bands plots.

pfebrer:
This commit also makes line dash work for the matplotlib backend

One can now use the`"dash"` key in `bands_style`, as in all other line style specifications in the viz module:

```python
sisl.get_sile("file.bands").plot(bands_style={"dash": "dot"})
```

newplot (73)

### DIFF
--- a/src/sisl/viz/figure/matplotlib.py
+++ b/src/sisl/viz/figure/matplotlib.py
@@ -233,6 +233,13 @@ class MatplotlibFigure(Figure):
     def show(self, *args, **kwargs):
         return self.figure.show(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


zerothi (approved):

zerothi:
@pfebrer this one fails CI, bands.py: L132, KeyError.

pfebrer:
Can you add `"dash": "solid"` to the default styles here? https://github.com/zerothi/sisl/blob/58f4758438a66f598e6140826a87571c3e047c12/src/sisl/viz/processors/bands.py#L10

I forgot, that should solve it.

PR #797, opened 2024-06-23
766 shc

### LINKED ISSUE #766: Transport routines within the kubo approach
Dear developers, I'm sharing some routines to calculate transport properties using SIESTA and sisl. I'm not sure if this submit is done the properly nor I'm an actual developer, so I've tried to comment the lines within the codes to highlight how they work.

This routines allow to calculate some transport quantities using an approach base on the Kubo formalism. The velocities are constructed using both derivatives of the Hamiltonian and Overlap matrices. Within this formalism, there are three examples showing how to calculate the spin Hall conductivity in Pt, the Orbital Hall conductivity in MoS2 and the Orbital Rashba Edelstein effect in Cu/O.

**Example cases**
SHC in bulk Pt (reference https://arxiv.org/abs/1810.07637) - routine -> shc.py
OHC in 2d MoS2 - routine -> ohc.py
Orbital Edelstein effect in oxidized Cu (Cu/O) (used in https://pubs.aip.org/aip/apm/article/12/5/051105/3287821/Quantifying-the-large-contribution-from-orbital) - routine -> job_surface_orbital ( In this case the quantity is calculated as a projection on the layers forming the system)

Comments:
[routines_sisl_kubo.zip](https://github.com/zerothi/sisl/files/15201018/routines_sisl_kubo.zip)

*The main advantage of using sisl as a post-processing tool to siesta is that we don't need to rely in any wannierization procedure since the orbitals are already localized. 
* The SHC and OHC are the intrinsic contributions to the Hall response and there's an issue appearing when Spin is set to SOC+onsite. I still don't know if this is related to the version of Siesta under which the SCF simulation was performed.
* Finally, if the routines aren't well explained through the comments, I'll be more than happy to help with any doubt.

- [x] Author : Armando Pezo
- [x] Which version of sisl it was originally developed and tested with sisl version: 0.0.0.dev+eacb8351ff416ad9f7f93732777c3d7931c5f711
[routines_sisl_kubo.zip](https://github.com/zerothi/sisl/files/15201029/routines_sisl_kubo.zip)

zerothi:
- `berry_curvature` is changed to only live on
  the `StateCElectron`.
  Old scripts relying on `berry_curvature`
  will not work now.
- `conductivity` has been deprecated, in favor
  of explicit calculations, `ahc` and `shc`.
- `axes` and `operator` arguments has been added
  for `derivative`
  They allow to determine which axes to extract the derivatives
  of, and to change the velocity operator to something else.

The above changes are required to enable the `ahc` and `shc` calculations.

ahc is basically the same as the old `conductivity` with changed API.
shc is the ahc where one of the velocity operators is changed
to the spin current operator.

Lots of the methods has changed their API's to enable faster parallel
calculations by explicitly allowing arguments passed to apply
methods.

However, the units of these methods are still wrong.
That needs subsequent fixing.

 - [x] Closes #766

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ we hit release version 1.0.0.
 ## [0.15.0] - YYYY-MM-DD
 
 ### Added
+- `axes` argument added to `derivative` to only calculate on a subset
+  of directions (can greatly improve performance for some systems)
+- `operator` argument added to `derivative` to apply an operator
+  to `dHk` and `dSk` matrices.
+  of directions (can greatly improve performance for some systems)
 - added `apply_kwargs` to methods which uses a `BrillouinZone` object.
   This enables one to leverage parallel processing for calculations.
 - `SISL_PAR_CHUNKSIZE=25`, new default parameter for parallel processing.
@@ -104,6 +109,8 @@ we hit release version 1.0.0.
 - removed `Selector` and `TimeSelector`, they were never used internally
 
 ### Changed
+- `conductivity` is deprecated, use `ahc` and `shc` instead
+- `berry_curvature` has completely changed, checks it API
 - BZ apply methods are now by default parallel (if ``SISL_NUM_PROCS>1``)
 - `hsxSileSiesta.read_hamiltonian` now implicitly shifts Fermi-level to 0 (for newer HSX versions)
 - deprecated `periodic` to `axes` argument in `BrillouinZone.volume`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,7 @@ we hit release version 1.0.0.
 - removed `Selector` and `TimeSelector`, they were never used internally
 
 ### Changed
+- units of `conductivity` has changed to S / Ang
 - `conductivity` is deprecated, use `ahc` and `shc` instead
 - `berry_curvature` has completely changed, checks it API
 - BZ apply methods are now by default parallel (if ``SISL_NUM_PROCS>1``)
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -20,6 +20,7 @@ sisl objects.
    add
    append
    apply
+   berry_curvature
    center
    copy
    insert
@@ -29,11 +30,13 @@ sisl objects.
    rotate
    scale
    sort
+   spin_berry_curvature
    sub
    swap
    swapaxes
    tile
    translate
+   velocity
    unrepeat
    untile
    write
--- a/docs/api/physics.electron.rst
+++ b/docs/api/physics.electron.rst
@@ -21,6 +21,8 @@ One may also plot real-space wavefunctions.
    COP
    berry_phase
    berry_curvature
+   ahc
+   shc
    conductivity
    wavefunction
    spin_moment
--- a/docs/api/typing.rst
+++ b/docs/api/typing.rst
@@ -13,7 +13,9 @@ The typing types are shown below:
 .. autosummary::
    :toctree: generated/
 
+   CellAxisStrLiteral
    CellAxisLiteral
+   CartesianAxisStrLiteral
    CartesianAxisLiteral
    CellAxis
    CellAxes
--- a/docs/math.rst
+++ b/docs/math.rst
@@ -40,3 +40,6 @@ please let us know!
   is an atomic index range and :math:`\{\alpha\}` refers
   to some *other* range which should be inferred from
   the context
+
+* the imaginary number is generally referred to as :math:`i` in
+  physics, its meaning should be implicit from the context.
--- a/docs/references.bib
+++ b/docs/references.bib
@@ -1,5 +1,34 @@
 %comment{This file was created with betterbib v4.3.11.}
 
+@article{PhysRevB.98.214402,
+  title = {Calculation of intrinsic spin Hall conductivity by Wannier interpolation},
+  author = {Qiao, Junfeng and Zhou, Jiaqi and Yuan, Zhe and Zhao, Weisheng},
+  journal = {Phys. Rev. B},
+  volume = {98},
+  issue = {21},
+  pages = {214402},
+  numpages = {10},
+  year = {2018},
+  month = {Dec},
+  publisher = {American Physical Society},
+  doi = {10.1103/PhysRevB.98.214402},
+  url = {https://link.aps.org/doi/10.1103/PhysRevB.98.214402}
+}
+
+@article{Ji2022,
+  title = {Spin Hall conductivity and anomalous Hall conductivity in full Heusler compounds},
+  volume = {24},
+  ISSN = {1367-2630},
+  url = {http://dx.doi.org/10.1088/1367-2630/ac696c},
+  DOI = {10.1088/1367-2630/ac696c},
+  number = {5},
+  journal = {New Journal of Physics},
+  publisher = {IOP Publishing},
+  author = {Ji,  Yimin and Zhang,  Wenxu and Zhang,  Hongbin and Zhang,  Wanli},
+  year = {2022},
+  month = may,
+  pages = {053027}
+}
 
 @article{Sengupta2023,
   author    = {Sengupta, Sanghita and Frederiksen, Thomas and Giedke, Geza},
--- a/docs/references.bib
+++ b/docs/references.bib
@@ -1,5 +1,14 @@
 %comment{This file was created with betterbib v4.3.11.}
 
+@article{gan2021calculation,
+      title={Calculation of Berry curvature using nonorthogonal atomic orbitals},
+      author={Jin Gan and Daye Zheng and Lixin He},
+      year={2021},
+      eprint={2105.14662},
+      archivePrefix={arXiv},
+      primaryClass={id='cond-mat.mtrl-sci' full_name='Materials Science' is_active=True alt_name='mtrl-th' in_archive='cond-mat' is_general=False description='Techniques, synthesis, characterization, structure.  Structural phase transitions, mechanical properties, phonons. Defects, adsorbates, interfaces'}
+}
+
 @article{PhysRevB.98.214402,
   title = {Calculation of intrinsic spin Hall conductivity by Wannier interpolation},
   author = {Qiao, Junfeng and Zhou, Jiaqi and Yuan, Zhe and Zhao, Weisheng},
--- a/src/sisl/physics/__init__.py
+++ b/src/sisl/physics/__init__.py
@@ -133,6 +133,7 @@ from ._brillouinzone_apply import *
 from ._ufuncs_brillouinzone import *
 from ._ufuncs_densitymatrix import *
 from ._ufuncs_dynamicalmatrix import *
+from ._ufuncs_electron import *
 from ._ufuncs_energydensitymatrix import *
 from ._ufuncs_hamiltonian import *
 from ._ufuncs_overlap import *
--- a/src/sisl/physics/_ufuncs_electron.py
+++ b/src/sisl/physics/_ufuncs_electron.py
@@ -115,13 +115,21 @@ def berry_curvature(
     See Also
     --------
     derivative: method for calculating the exact derivatives
+    spin_berry_curvature : calculate the spin Berry curvature
+    ahc: anomalous Hall conductivity
+    shc: spin Hall conductivity
 
     Returns
     -------
-    numpy.ndarray
-        Berry flux with final dimension ``(3, 3, *)``. The dtype will be imaginary.
-        The Berry curvature is in the real part of the values.
-        The unit is :math:`\mathrm{Ang}^2`
+    bc: numpy.ndarray
+        Berry curvature returned in certain dimensions ``bc[:, :]``.
+        If `sum` is False, it will be at least a 3D array with the 3rd dimension
+        having the contribution from state `i`.
+        If one passes `axes` to the `derivative_kwargs` argument one will get
+        dimensions according to the number of axes requested, by default all
+        axes will be used (even if they are non-periodic).
+        The dtype will be imaginary.
+        The unit is :math:`\mathrm{Ang}^2`.
     """
     # cast dtypes to *any* complex valued data-type that can be expressed
     # minimally by a complex64 object
@@ -158,6 +166,8 @@ def berry_curvature(
     # to the states.
     # Hence we transpose it for direct usage below.
     dist_e = distribution(energy)
+    # Since we are using the double counting dist_e[si] - dist_e
+    # we have to omit the factor 2 here.
 
     if sum:
         dsigma_shape = (len(dA), len(dB)) + (1,) * (dist_e.ndim - 1)
@@ -168,7 +178,7 @@ def berry_curvature(
 
         for si, ei in enumerate(energy):
             de = (ei - energy) ** 2 + ieta2
-            np.divide(-2, de, where=(de != 0), out=de)
+            np.divide(-1, de, where=(de != 0), out=de)
             # the order of this term can be found in:
             # 10.21468/SciPostPhysCore.6.1.002 Eq. 29
             dd = dist_e[..., [si]] - dist_e
@@ -188,7 +198,7 @@ def berry_curvature(
 
         for si, ei in enumerate(energy):
             de = (ei - energy) ** 2 + ieta2
-            np.divide(-2, de, where=(de != 0), out=de)
+            np.divide(-1, de, where=(de != 0), out=de)
             dd = dist_e[..., [si]] - dist_e
 
             for iA in range(len(dA)):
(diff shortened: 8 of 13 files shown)


zerothi:
@apezol

This is now ready for you testing this...

The units are troubling me, quite a bit ;)

I have tried a great variety of combinations.

Here is a small snippet that can reproduce your results:
```python
import numpy as np

import sisl as si

### Spin Hall Conductivity
pool = True
format = "array"
k = [35] * 3
k = [5] * 3
E = np.linspace(-10, 5, 451)

sile = si.get_sile("lead.fdf")
geom = sile.read_geometry()
H = sile.read_hamiltonian()

bz = si.MonkhorstPack(H, k, trs=False)  # BZ sampling

distribution = si.physics.get_distribution("step", x0=E.reshape(-1, 1))

shc = (
    si.physics.electron.shc(
        bz,
        # This allows parallel computation
        apply_kwargs={"pool": pool},
        # For these small systems, this will speed up details...
        # eta=True == show progress
        eigenstate_kwargs={"format": format, "dtype": np.complex128, "eta": True},
        eta=1e-2,
        # To speed-up calculation, only calculate the cross-terms for xy directions.
        derivative_kwargs={"degenerate": None, "axes": "xy"},
        distribution=distribution,
    )[1, 0]
)

with si.io.tableSile("shc3_Pt.txt", "w") as fh:
    fh.write(E, -shc * 1e4 / 2)
```
You'll note that I have to multiply by `-1e4/2` in order to get the same results as you have.
You can see some of what I have tried from the discussion in the code, see here: https://github.com/zerothi/sisl/blob/0531a494d7340acb5a82f694b67a67f41bafc725/src/sisl/physics/electron.py#L853

Also, the `berry_curvature` term is calculated [here](https://github.com/zerothi/sisl/blob/0531a494d7340acb5a82f694b67a67f41bafc725/src/sisl/physics/_ufuncs_electron.py#L155).
As you can see there is only a factor of -2 (since hbar should cancel out from the veloctiy operators). So at that point, it should have the unit Ang^2. 

So I have two questions:
1. How can I ensure that the units of the anomalous Hall conductivity is correct? And what unit would you recommend?
2. and the same for the spin Hall conductivity.

In order for a fast calculation, launch the script like this:

```shell
SISL_PAR_CHUNKSIZE=100 SISL_NUM_PROCS=6 OMP_NUM_THREADS=1 python shc2.py 
```
This should give a very fast execution, even for the `35^3` calculation.

If you have any questions, please don't hesitate to ask! Thanks for sharing your link.

Note, the `ohc` is very easily done by creating a temporary operator and calling `ahc`.

zerothi:
And... What happens when you only have a 2D structure, in this case the volume is not well-defined, should `V` then only be area, in which case the conductivity is in units `S`?

apezol:
Thanks a lot @zerothi , I'll test it asap!. 

1._How can I ensure that the units of the anomalous Hall conductivity is correct? And what unit would you recommend?_

Regarding the factor, there appears the same '10e4' as in the code I shared and the factor '1/2' which seems to come from the definition of the  spin matrices ($\vec{s}$), that is, $\frac{\hbar}{2}\vec{\sigma}$, which on my case only carry the Pauli matrices $\sigma$. So it totally make sense, I guess. As I recall, the factor I multiply by in my routines came after transforming the Ang to $cm$.

2._and the same for the spin Hall conductivity._

As you pointed out, the Berry curvature should be giving units of Ang**2, and after integration the units for the SHC should be the same as those for the AHC , if we consider the expression

Selection_036

from reference:https://arxiv.org/pdf/1810.07637

That is, in order to have the units of standard SHC ($\frac{\hbar}{e}S/lenght$), one should multiply by -$\hbar/2e$ which should be coming from the factor of the spin matrices in the Spin velocity operator $1/2$  $\\{v_y,s^z\\} $. I'll check this out with some simulations I have for other materials.

zerothi:
> Thanks a lot @zerothi , I'll test it asap!.
> 
> 1._How can I ensure that the units of the anomalous Hall conductivity is correct? And what unit would you recommend?_
> 
> Regarding the factor, there appears the same '10e4' as in the code I shared and the factor '1/2' which seems to come from the definition of the spin matrices (s→), that is, ℏ2σ→, which on my case only carry the Pauli matrices σ. So it totally make sense, I guess. As I recall, the factor I multiply by in my routines came after transforming the Ang to cm.

Yes, the factor 2 I think makes sense, but the `1e4`, `Ang -> cm` is `1e-8`, so I don't see the `1e4`?
> 
> 2._and the same for the spin Hall conductivity._
> 
> As you pointed out, the Berry curvature should be giving units of Ang**2, and after integration the units for the SHC should be the same as those for the AHC , if we consider the expression
> 
> Selection_036
> 
> from reference:https://arxiv.org/pdf/1810.07637
> 
> That is, in order to have the units of standard SHC (ℏeS/lenght), one should multiply by -$\hbar/2e$ which should be coming from the factor of the spin matrices in the Spin velocity operator 1/2 {vy,sz}. I'll check this out with some simulations I have for other materials.

However, internally there is not any `e` or `hbar` entering, so what exactly should we provide.

Do you agree the best unit for AHC would be S/cm? Or would S/Ang make sense? Generally everything in sisl is in Ang.. So while it might not be what is typically reported in literature, it might be more straightforward to retain S/Ang for clarity?

apezol:
> And... What happens when you only have a 2D structure, in this case the volume is not well-defined, should `V` then only be area, in which case the conductivity is in units `S`?

That's right, the 'Volume' parameter should be an 'Area' in the 2d case, since the unit cell always has the three unit cell vectors. However, you can still divide the Volume by the corresponding  vector on which there is no periodicity, usually the 'z' component and then the units will correspond to that of a 2D system.

apezol:
@zerothi , I was re-checking my convention of units and there is an extra 1e-4 for having them in 'S cm^-1', therefore this leads to a final factor of 1e4. On the other hand, I did some tests in FM and it looks nice:

Screen Shot 2024-06-25 at 8 05 08 AM

The bottom part comes from : https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001

But, I've found that there's a minus sign for the final result. From that, the Nernst coefficient can be also implemented easily =D.

zerothi:
> @zerothi , I was re-checking my convention of units and there is an extra 1e-4 for having them in 'S cm^-1', therefore this leads to a final factor of 1e4. On the other hand, I did some tests in FM and it looks nice:

But I don't understand, sisl is in Ang, and Ang -> cm is 1e-8...? Where does the last 1e-4 factor come from?

> Screen Shot 2024-06-25 at 8 05 08 AM
> 
> The bottom part comes from : https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001
> 
> But, I've found that there's a minus sign for the final result. From that, the Nernst coefficient can be also implemented easily =D.

I think my latest commits fixed the `-` sign. Now it will always return AHC/SHC as imaginary values, where the real value is the actual result. Couldn't the imaginary parts be useful in some cases?

zerothi:
PS, thanks for testing!! :)

apezol:
> > @zerothi , I was re-checking my convention of units and there is an extra 1e-4 for having them in 'S cm^-1', therefore this leads to a final factor of 1e4. On the other hand, I did some tests in FM and it looks nice:
> 
> But I don't understand, sisl is in Ang, and Ang -> cm is 1e-8...? Where does the last 1e-4 factor come from?
> 
> > Screen Shot 2024-06-25 at 8 05 08 AM
> > The bottom part comes from : https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001
> > But, I've found that there's a minus sign for the final result. From that, the Nernst coefficient can be also implemented easily =D.
> 
> I think my latest commits fixed the `-` sign. Now it will always return AHC/SHC as imaginary values, where the real value is the actual result. Couldn't the imaginary parts be useful in some cases?

Indeed, there are some cases where the imaginary part is useful. Particularly the imaginary part of the longitudinal conductivity but at finite frequencies, closely related to opto-electronic properties. But I think the current version only supports the static limit, i.e. vanishing frequency.

apezol:
> > @zerothi , I was re-checking my convention of units and there is an extra 1e-4 for having them in 'S cm^-1', therefore this leads to a final factor of 1e4. On the other hand, I did some tests in FM and it looks nice:
> 
> But I don't understand, sisl is in Ang, and Ang -> cm is 1e-8...? Where does the last 1e-4 factor come from?
> 
> > Screen Shot 2024-06-25 at 8 05 08 AM
> > The bottom part comes from : https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001
> > But, I've found that there's a minus sign for the final result. From that, the Nernst coefficient can be also implemented easily =D.
> 
> I think my latest commits fixed the `-` sign. Now it will always return AHC/SHC as imaginary values, where the real value is the actual result. Couldn't the imaginary parts be useful in some cases?

As for the units, that's correct, the factors isn't coming from the length conversion but rather from the e^2/h -> S (siemens) conversion, although, there's a factor close to '1' which need to be adapted for accuracy in my case. For the current implementation it's perfect!.

zerothi:
> @zerothi , I was re-checking my convention of units and there is an extra 1e-4 for having them in 'S cm^-1', therefore this leads to a final factor of 1e4. On the other hand, I did some tests in FM and it looks nice:
> Screen Shot 2024-06-25 at 8 05 08 AM
> 
> The bottom part comes from : https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001
> 
> But, I've found that there's a minus sign for the final result. From that, the Nernst coefficient can be also implemented easily =D.

By the way, how do you calculate the spin-Hall conductivity for a simple polarized calculation? I have put in a restriction saying you can't do that unless the system is non-colinear.

apezol:
> > @zerothi , I was re-checking my convention of units and there is an extra 1e-4 for having them in 'S cm^-1', therefore this leads to a final factor of 1e4. On the other hand, I did some tests in FM and it looks nice:
> > Screen Shot 2024-06-25 at 8 05 08 AM
> > The bottom part comes from : https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001
> > But, I've found that there's a minus sign for the final result. From that, the Nernst coefficient can be also implemented easily =D.
> 
> By the way, how do you calculate the spin-Hall conductivity for a simple polarized calculation? I have put in a restriction saying you can't do that unless the system is non-colinear.

In principle you cannot do that, at least for the intrinsic part which comes from the Berry curvature, since it involves inter-band transitions: <n| Operator| m>. This is only possible once SOC is considered, and also within non-collinear systems as well( as you pointed out), so it might appear in these new systems so-called altermagnets. So, taking advantage of this, is it possible to run SIESTA considering SOC+U (Hubbard correction) in the latest versions?.

zerothi:
@apezol I think I have fixed everything, and the units seems to be consistent.

Now you should do this:

```python
import numpy as np

import sisl as si

### Spin Hall Conductivity
pool = True
format = "array"
k = [35] * 3
k = [5] * 3
E = np.linspace(-10, 5, 451)

sile = si.get_sile("lead.fdf")
geom = sile.read_geometry()
H = sile.read_hamiltonian()

bz = si.MonkhorstPack(H, k, trs=False)  # BZ sampling

distribution = si.physics.get_distribution("step", x0=E.reshape(-1, 1))

shc = (
    si.physics.electron.shc(
        bz,
        # This allows parallel computation
        apply_kwargs={"pool": pool},
        # For these small systems, this will speed up details...
        # eta=True == show progress
        eigenstate_kwargs={"format": format, "dtype": np.complex128, "eta": True},
        eta=1e-2,
        # To speed-up calculation, only calculate the cross-terms for xy directions.
        derivative_kwargs={"degenerate": None, "axes": "xy"},
        distribution=distribution,
    )[0, 1].real
)

with si.io.tableSile("shc3_Pt.txt", "w") as fh:
    fh.write(E, shc / si.units("Ang", "cm"))
```
and you'll get the correct things.
Could you please have a last check, and if you wouldn't mind (you are the expert here), could you check the AHC calculation which is equivalenty called. I think everything should be correct now.

The units returned are:
- AHC: S/Ang
- SHC: \hbar / e  S/Ang
which seems consistent with litterature (except the Ang -> cm, which is trivially handled).

Let me know what you think here! :)

zerothi:
@apezol oh, and note I added the *correct* pre-factor, $G_0 \pi \approx 2.424e-4$, so direct values with your script should divide by the factor, and multiplied by $2e-4$. :)

zerothi:
> > By the way, how do you calculate the spin-Hall conductivity for a simple polarized calculation? I have put in a restriction saying you can't do that unless the system is non-colinear.
> 
> In principle you cannot do that, at least for the intrinsic part which comes from the Berry curvature, since it involves inter-band transitions: <n| Operator| m>. This is only possible once SOC is considered, and also within non-collinear systems as well( as you pointed out), so it might appear in these new systems so-called altermagnets. So, taking advantage of this, is it possible to run SIESTA considering SOC+U (Hubbard correction) in the latest versions?. Perhaps this should correspond to the SIESTA forum, but still, transport calculations (as this new one implemented in sisl) are very interesting to look at currently.

I see, that was also how I understood it. I am pretty sure SOC+U is there. So you can play with it! :)

apezol:
> @apezol I think I have fixed everything, and the units seems to be consistent.
> 
> Now you should do this:
> 
> ```python
> import numpy as np
> 
> import sisl as si
> 
> ### Spin Hall Conductivity
> pool = True
> format = "array"
> k = [35] * 3
> k = [5] * 3
> E = np.linspace(-10, 5, 451)
> 
> sile = si.get_sile("lead.fdf")
> geom = sile.read_geometry()
> H = sile.read_hamiltonian()
> 
> bz = si.MonkhorstPack(H, k, trs=False)  # BZ sampling
> 
> distribution = si.physics.get_distribution("step", x0=E.reshape(-1, 1))
> 
> shc = (
>     si.physics.electron.shc(
>         bz,
>         # This allows parallel computation
>         apply_kwargs={"pool": pool},
>         # For these small systems, this will speed up details...
>         # eta=True == show progress
>         eigenstate_kwargs={"format": format, "dtype": np.complex128, "eta": True},
>         eta=1e-2,
>         # To speed-up calculation, only calculate the cross-terms for xy directions.
>         derivative_kwargs={"degenerate": None, "axes": "xy"},
>         distribution=distribution,
>     )[0, 1].real
> )
> 
> with si.io.tableSile("shc3_Pt.txt", "w") as fh:
>     fh.write(E, shc / si.units("Ang", "cm"))
> ```
> 
> and you'll get the correct things. Could you please have a last check, and if you wouldn't mind (you are the expert here), could you check the AHC calculation which is equivalenty called. I think everything should be correct now.
> 
> The units returned are:
> 
>     * AHC: S/Ang
> 
>     * SHC: \hbar / e  S/Ang
>       which seems consistent with litterature (except the Ang -> cm, which is trivially handled).
> 
> 
> Let me know what you think here! :)

@zerothi  sisl works perfectly now! =D, the only example for testing the AHC I found is this calculation in bcc Fe. Siesta gives ~ 680 Scm-1 (left) while they report 750 Scm-1, close enough. 

Selection_039

Taken from https://www.aflowlib.org/src/paoflow/ , example 04. One last thing is that, and I guess this is my fault, in the previous calculations of the SHC there was a minus sign right at the printing .txt file, so that 's  why it was inverted, sorry for that. Then in the current version you only need to take off this minus sign. Apart from that, it's working perfect.  This new implementation for distributing the BZ summation is really nice!.

zerothi:
> ... One last thing is that, and I guess this is my fault, in the previous calculations of the SHC there was a minus sign right at the printing .txt file, so that 's why it was inverted, sorry for that. Then in the current version you only need to take off this minus sign. 

@apezol I don't fully understand this? When I compare with your scripts, I get the same sign, so I guess everything is fine, no?

Your original `shc` had this at the end:
```python

ene = np.linspace(-10, 5, 251)

with ProcessPoolExecutor(max_workers=35) as executor:
    shc = list(executor.map(spin_conductance, ene))

for i in zip(ene, shc):
    print(i[0], i[1])
```
no minus sign?

zerothi:
Or, maybe you are correct, I don't have the same sign... I'll do a last comparison with shc...

zerothi:
I must admit, I am a bit at a loss now, how exactly did you do the AHC test?

zerothi:
But I don't understand why there shouldn't be an extra factor `-1` on SHC compared to AHC?

If you look at the discussion just after Eq. 4 in [1810.07637](https://arxiv.org/pdf/1810.07637), it does seem to me that there is missing a factor `-1`. It seems to me that SHC has a factor on `-1` compared to AHC, what am I missing?

But you seem to agree on AHC, so I can't multiply both with `-1`, what am I missing?

zerothi:
I will now merge, to keep other developments, but if you have any clues, it would be really appreciated! :)

zerothi:
I looked a bit more into the `-1` sign, I think it is because your code calculates $\sigma_{yx}$, and that is $-\sigma_{xy}$, hence the minus sign seem off.

I took your Pt system, and did `shc[0, 1]`, at $E_F$ I get a value $~2200$ which I guess is close enough to Fig. 1 of ~2000 for Pt in https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001 (also the rest of the curve looks quite identical).

So, I'll re-insert the `-` sign, if you have further inputs, please don't hesitate to return!

apezol:
> I looked a bit more into the `-1` sign, I think it is because your code calculates σyx, and that is −σxy, hence the minus sign seem off.
> 
> I took your Pt system, and did `shc[0, 1]`, at EF I get a value  2200 which I guess is close enough to Fig. 1 of ~2000 for Pt in https://journals.aps.org/prmaterials/pdf/10.1103/PhysRevMaterials.6.095001 (also the rest of the curve looks quite identical).
> 
> So, I'll re-insert the `-` sign, if you have further inputs, please don't hesitate to return!

@zerothi , so that might be the reason!. I'll certainly use it in another set of materials and share any inconsistency (if any). Thanks a lot for implementing this new feature =). I'll be working in other quantities relevant for transport, sisl really provides several advantages with respect to other tools!.

PR #801, opened 2024-06-26
First try at automated data CLI

pfebrer:
As discussed with Nick, this is a minimal first attempt for the automatic generation of a CLI to postprocess data.

The aim is to organize things in a way that new methods to generate data can be simply registered to data classes and automatically added to the CLI, without needing to maintain CLI options all across sisl.

The automatic generation of a CLI requires of course some boilerplate code, but the amount of boilerplate code is almost constant when we add new functionality.

The approach taken:
   - Created data classes in `sisl.data` that should help gathering all the possible ways in which the same data can be generated. For now I only implemented `DOSData`.
   - Then functions to create that data should be registered. For now I have only registered the function that creates `DOSData` from `eigSileSiesta`. This functions should be type annotated to be automatically supported by the CLI.
   - The CLI is created independently in `sisl.cli.sdata`. For now it can be ran with `python -m sisl.cli.sdata`. That file is hyper documented, and there I explain all the parts that play a role in generating the CLI.
 
This standarization should not only help to create sisl's CLI but also facilitate the creation of other interfaces to sisl down the line (including the GUI).

To use it, for now, you need to install `nodify`'s dev version, `click` and `rich_click`: 
```
pip install git@github.com:pfebrer/nodify.git click rich_click 
```

The CLI can work in two ways:

**Specify the type of data that you want and how to get it.**

Main:
Screenshot from 2024-06-26 23-21-16

DOS data:
Screenshot from 2024-06-26 23-20-46

DOS data from the EIG file:
Screenshot from 2024-06-26 23-20-56

**Pass a file, then only specify which data you want to get**

Main:
Screenshot from 2024-06-26 23-24-25

DOS data:
Screenshot from 2024-06-26 23-25-38

### DIFF
--- a/src/sisl/cli/sdata.py
+++ b/src/sisl/cli/sdata.py
@@ -1,0 +1,370 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+"""Automatic CLI generation for sisl data classes.
+
+This file contains the functions to automatically create a CLI to
+generate and postprocess sisl's data classes.
+
+There are multiple ways in which you might want to extend the CLI:
+
+ - By adding a way of generating data from a certain file type. This is
+ the most likely scenario. In that case, all the modifications required
+ are OUTSIDE of this file. You need to:
+
+    1. Create a function that generates the data from the file. Add type
+    annotations to it. E.g.:
+
+    ```python
+    import DOSData
+
+    def dos_from_bandsSile(
+        sile: sisl.io.bandsSileSiesta,
+        Erange: Tuple[float, float] = (-2, 2)
+    ):
+        # Get energies and DOS somehow
+        ...
+        E = ...
+        DOS = ...
+
+        return DOSData.new(E, DOS)
+    ```
+
+    2. Register your function as a way of creating DOSData:
+    ```python
+    DOSData.new.register("siesta-bands", dos_from_bandsSile)
+    ```
+
+    3. This data generating function should be automatically added to the CLI as
+    'siesta-bands'. Enjoy!
+
+  - By adding a post processing method to a data class. In this case, all you need to
+    do is to add the method to the data class and make sure it is type annotated. The
+    method should then automatically show up in the CLI.
+
+  - By adding a new type of data. In this case, you need to create the data class and
+    make sure that it is added to the CLI in this file. You need to add a new line to
+    the `main` function that calls `add_data_group` with the new data class.
+
+Alternatively, it is also possible that you might want to change the way the CLI works.
+This is of course more complex, but here is a sketch of how the CLI works to help
+pinpoint where you might want to make changes:
+
+    - The CLI is a click CLI.
+    - The CLI is created on `main`.
+    - The main app is created and then we add a group for each data class. Each group
+      contains:
+        - Commands to generate data:
+          We automatically gather all the possible sources from the `new` dispatcher
+          of the data class. We then decorate them automatically with
+          `decorate_data_gen_function` to be used in click based on type annotations.
+        - Commands to post process data:
+          We automatically gather all the methods of the data class that are not private
+          and decorate them with `decorate_data_method` to be used in click based on type
+          annotations.
+
+    - The way we can generate and post process data in the same CLI is by allowing commands
+      to be chained. However, since commands execute independently, we need a global storage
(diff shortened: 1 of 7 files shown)


src/sisl/cli/sdata.py
```diff
@@ -0,0 +1,370 @@
...
+    # All sile classes. If a CLI has an argument that is annotated as a sile
+    # then the value provided to the CLI will be parsed simply as sile_cls(value),
+    # i.e. a sile will be created from the user provided path.
+    **{sile_cls: {"parser": sile_cls} for sile_cls in sisl.get_siles()},
+}
+
+# List that keeps track of the data that the CLI has generated.
+# When chaining commands, the first command will generate data and
+# the subsequent commands will be methods to apply on that data.
+GENERATED = []
```
zerothi, line 104: I think this is sub-optimal, imagine one parses a grid, and a geometry (from two sources) and then writes them out, then one has to figure out the order.

I think click contexts are better here, with a dict or something, then sub-commands can also change data to other data-types.
I guess we just need some way of distinguishing data, and also enable the possibility of adding more of the same data-type. E.g. comparing two DOS is not that unreasonable, and here it would be, perhaps better, to have a data-container that contains origin of data. So one can query where it came from.

src/sisl/cli/sdata.py
```diff
@@ -0,0 +1,370 @@
...
+      need to define how those types will be generated from the values (strings) passed in
+      the `CLI`. We do that in the `custom_type_kwargs` dictionary.
+
+"""
+import functools
+import inspect
+import sys
+from typing import Callable, Optional
+
+import rich_click as click
```
zerothi, line 82: I am probably more in favor of importing click, and if available importing rich-click.

zerothi (changes requested):
I'll have to play a bit more with this after vacation. Thanks!

zerothi:
Some general comments here.

I don't particularly fancy the *special* data-type specification. I can see why it may be useful, but I kind of find it a bit duplicate because it builds on the current state of things and *adds a layer between simple numpy arrays*. And it is never know exactly when you might need what....

Your `dos_from_eig` is also duplication, I had imagined something more generic for something that reads an eigenvalue spectrum. For instance when you also do a `dos_from_eig(vasp_eig)` then the arguments needs to be duplicated. I guess the only way to do this generically would be to have some kind of `dos_from_eig(eig, weights, Erange, dE, distribution, smearing)` and then do `dos_from_vasp(sile, **<extend interface with dos_from_eig>)`.

How, in this framework, would you envision how things comes out of generic stuff.

For instance, I could imagine reading a Hamiltonian, and then post-processing it for a DOS, but also PDOS and then plotting it?

What I do like about this is that it does not become depending on the sile it self... Perhaps I should play a bit more with this.

I will have a look at how this could be generalized a bit, towards simpler data-structures. E.g. the `DOSdata` seems like it will quickly explode in terms of how things gets implemented, we need specific classes for a huge number of things... Perhaps it wuold be smarter to be *data-type* agnostic in some form, but rely on some simpler metadata.

zerothi:
@pfebrer would you mind rebasing this one? It should not interfere!

pfebrer:
I have some time now to be able to work on this. Would you be available to review and merge?

I think it is really important so that the plotting CLI can be build, which would be really useful.

> Your dos_from_eig is also duplication, I had imagined something more generic for something that reads an eigenvalue spectrum.

Duplication of what? I just took your current argument parser. Once there are multiple entry points that do something similar one can always generalize, of course.

> E.g. the DOSdata seems like it will quickly explode in terms of how things gets implemented, we need specific classes for a huge number of things... Perhaps it wuold be smarter to be data-type agnostic in some form, but rely on some simpler metadata.

How would you centralize all the ways of computing a density of states?

DOSData can also be private, simply a tool for centralizing all the functionality.

> How, in this framework, would you envision how things comes out of generic stuff.
> 
> For instance, I could imagine reading a Hamiltonian, and then post-processing it for a DOS, but also PDOS and then plotting it?

I think initially the CLI doesn't need to do very complex stuff, just a CLI that allows doing the same analysis (e.g. DOS) from different entry points is useful enough I think.

An advanced user can always write their own code to do custom stuff if needed. From my perspective, most users on a daily basis need simpler things than a generic interface to compute many properties at once. Having a more specific CLI for a particular popular analysis I think it is good for widespread adoption. As an example, you can just take the confusion of @ialcon with the `MonkhorstPack` object.

We can always add later another command to the CLI to do more generic stuff...

PR #805, opened 2024-07-15
Fix bug with numpy v2.0

pfebrer:
The repr of `numpy` 0D arrays changed from `0` to `np.int64(0)` in numpy 2.0, which caused a bug when creating queries for orbital filtering, making tests fail: https://github.com/zerothi/sisl/actions/runs/9922030378/job/27410643845.


zerothi:
Thanks, I would not have been able to track this down! 🎆

PR #809, opened 2024-07-24
enabled vacuum for all geometries with non-pbc directions

zerothi:
This is a long saught solution to fix vacuum lengths by enabling users to control them.

@tfrederiksen @pfebrer I hope this covers everything, without breaking stuff...
I don't know if this should be moved to the `_common.py` file, perhaps it should... Well...

Let me know if this is ok, as is, or something is not covered.


pfebrer:
Looks good to me!

PR #819, opened 2024-08-24
Make sure wheels are uploaded to pypi

pfebrer:
Only the source distribution was being uploaded to Pypi because of a mistake in the workflow file: https://pypi.org/project/sisl/#files

There was also an unnecessary artifact download.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -164,11 +164,6 @@ jobs:
         os: [ubuntu-latest, macos-13, macos-14]
 
     steps:
-      - uses: actions/download-artifact@v4
-        with:
-          name: artifact
-          path: dist
-
       - name: Python installation
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/wheels.yaml
```diff
@@ -164,11 +164,6 @@ jobs:
         os: [ubuntu-latest, macos-13, macos-14]
 
     steps:
-      - uses: actions/download-artifact@v4
-        with:
-          name: artifact
-          path: dist
-
```
zerothi, line 171: Ok, can see why this isn't necessary, at least if we want to check that test-pypi is working.

.github/workflows/wheels.yaml
```diff
@@ -219,7 +214,7 @@ jobs:
     steps:
       - uses: actions/download-artifact@v4
         with:
-          name: artifact
```
zerothi, line 222: Thanks for catching this!

zerothi (approved):

PR #823, opened 2024-09-02
Fix orbital values

### LINKED ISSUE #821: Atoms that contribute to the density not correctly found
**Describe the bug**

This part of the code:
https://github.com/zerothi/sisl/blob/115479f6e54591de16196986fe01526380497ca7/src/sisl/physics/densitymatrix.py#L913-L926

which finds atoms with an orbital that reaches the unit cell, does not work properly.

This is critical because it gives wrong densities, since atoms that should contribute are not taken into account.

**Code to reproduce problem**

The simplest reproducing code that I could come up with is the following. I build a cell with two atoms inside and no periodic boundary conditions. Clearly, it should find both atoms to contribute to the unit cell.

With a cell `[10, 10, 20]`, if the second atom is at `z=10` it is correctly found, while if it is at `z=15` it is not found.

I have no idea what causes this problem, but it is pretty nasty :(

(I didn't change anything in this part when I implemented the new algorithm for the density)

```python
import sisl
import numpy as np
from sisl import Lattice

# This is what is being used to find atoms that contribute to the density
def get_contributing_atoms(geom):
    self = geom
    # Instead of looping all atoms in the supercell we find the exact atoms
    # and their supercell indices.
    add_R = np.full(3, self.maxR())
    # Calculate the required additional vectors required to increase the fictitious
    # supercell by add_R in each direction.
    # For extremely skewed lattices this will be way too much, hence we make
    # them square.
    o = self.lattice.to.Cuboid(True)
    lattice = Lattice(o._v + np.diag(2 * add_R), origin=o.origin - add_R)
    
    # Retrieve all atoms within the grid supercell
    # (and the neighbours that connect into the cell)
    IA, XYZ, ISC = self.within_inf(lattice, periodic=self.pbc)
    XYZ -= self.lattice.origin.reshape(1, 3)
    
    return IA, XYZ, ISC

# With the second atom at z=10, it determines that it contributes. 
geom = sisl.Geometry([[1, 1, 2], [1, 1, 10]], lattice=[10, 10, 20], atoms=sisl.Atom("C", R=3))
geom.lattice.set_boundary_condition("")

print("THIS IS FINE")
print(get_contributing_atoms(geom))

# With the second atom at z=15, it determines that it doesn't contribute
geom = sisl.Geometry([[1, 1, 2], [1, 1, 15]], lattice=[10, 10, 20], atoms=sisl.Atom("C", R=3))
geom.lattice.set_boundary_condition("")

print("THIS IS NOT")
print(get_contributing_atoms(geom))
```
```
THIS IS FINE
(array([0, 1], dtype=int32), array([[ 1.,  1.,  2.],
       [ 1.,  1., 10.]]), array([[0, 0, 0],
       [0, 0, 0]], dtype=int32))
THIS IS NOT
(array([0], dtype=int32), array([[1., 1., 2.]]), array([[0, 0, 0]], dtype=int32))
```

pfebrer:
Fixes #821

### DIFF
--- a/src/sisl/_core/geometry.py
+++ b/src/sisl/_core/geometry.py
@@ -3721,7 +3721,7 @@ class Geometry(
         # supercell by add_R in each direction.
         # For extremely skewed lattices this will be way too much, hence we make
         # them square.
-        o = self.lattice.to.Cuboid(True)
+        o = self.lattice.to.Cuboid(orthogonal=True)
         lattice = Lattice(o._v + np.diag(2 * add_R), origin=o.origin - add_R)
 
         # Retrieve all atoms within the grid supercell


zerothi (approved):
thanks, that is definitely a left-over! Nice catch!

PR #836, opened 2024-10-04
Fix orbital values when an orbital is completely outside of the unit cell

pfebrer:
The orbital values computation was failing if there were orbitals that were completely outside of the unit cell.

### DIFF
--- a/src/sisl/_core/geometry.py
+++ b/src/sisl/_core/geometry.py
@@ -3752,9 +3752,9 @@ class Geometry(
             cmin = np.maximum(corners_i.min(axis=0), 0)
             cmax = np.maximum(corners_i.max(axis=0) + 1, 0)
 
-            rx = slice(cmin[0], min(cmax[0], grid.shape[0]))
-            ry = slice(cmin[1], min(cmax[1], grid.shape[1]))
-            rz = slice(cmin[2], min(cmax[2], grid.shape[2]))
+            rx = slice(min(cmin[0], grid.shape[0]), min(cmax[0], grid.shape[0]))
+            ry = slice(min(cmin[1], grid.shape[1]), min(cmax[1], grid.shape[1]))
+            rz = slice(min(cmin[2], grid.shape[2]), min(cmax[2], grid.shape[2]))
 
             indices = np.mgrid[rx, ry, rz].reshape(3, -1).T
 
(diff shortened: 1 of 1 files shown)


pfebrer:
E.g. I tested one case where `rx,ry,rz` was `slice(354, 300, None), slice(0, 4, None), slice(8, 12, None)`

which resulted in:

```
File ~/webDevelopement/sislGUI/sisl/src/sisl/_core/geometry.py:3825, in Geometry._orbital_values(self, grid_shape)
   3822     warn(f"Atom '{atom}' does not have a wave-function, skipping atom.")
   3823     continue
-> 3825 idx = sphere_grid_index(grid, ia_xyz, R)
   3827 if len(idx) == 0:
   3828     continue

File ~/webDevelopement/sislGUI/sisl/src/sisl/_core/geometry.py:3765, in Geometry._orbital_values.<locals>.sphere_grid_index(grid, center, R)
   3761 rz = slice(cmin[2], min(cmax[2], grid.shape[2]))
   3763 print(rx, ry, rz)
-> 3765 indices = np.mgrid[rx, ry, rz].reshape(3, -1).T
   3767 return indices

File ~/miniconda3/envs/simulationspy311/lib/python3.11/site-packages/numpy/lib/index_tricks.py:173, in nd_grid.__getitem__(self, key)
    170     nn = [_nx.arange(_x, dtype=_t)
    171           for _x, _t in zip(size, (typ,)*len(size))]
    172 else:
--> 173     nn = _nx.indices(size, typ)
    174 for k, kk in enumerate(key):
    175     step = kk.step

File ~/miniconda3/envs/simulationspy311/lib/python3.11/site-packages/numpy/core/numeric.py:1764, in indices(dimensions, dtype, sparse)
   1762     res = tuple()
   1763 else:
-> 1764     res = empty((N,)+dimensions, dtype=dtype)
   1765 for i, dim in enumerate(dimensions):
   1766     idx = arange(dim, dtype=dtype).reshape(
   1767         shape[:i] + (dim,) + shape[i+1:]
   1768     )

ValueError: negative dimensions are not allowed
```

zerothi:
Removed `np.maximum` I think they were too heavy for such small arrays. Will merge once coveraged.

zerothi:
Thanks @pfebrer

Issue #838, opened 2024-10-04
Document plot handlers.


pfebrer:
Methods like `Hamiltonian.plot.pdos()` are not documented anywhere, and the only way of getting information about them is if your editor can show you the arguments and documentation of the function.

I want to document them, but I don't know how exactly. I have thought of modifying the class template for autosummary (https://github.com/zerothi/sisl/blame/main/docs/_templates/autosummary/class.rst) to check for the plot attribute and then add some documentation, but I don't know exactly how or if this is the best approach.

pfebrer:
I don't know if it is possible to make sphinx document things like `plot.pdos` as a normal method of the `Hamiltonian`. I think this could be a reasonable approach.

Another idea that I have is to gather all plot handlers and create a page in the `sisl.viz` documentation like:

```
List of objects that can be used to generate each plot:

AtomicMatrixPlot
================
  - SparseCSR
  - SparseOrbital
  - SparseAtom
  - Hamiltonian

GeometryPlot
================
  - Geometry
  - fdfSileSiesta
  - pdosSileSiesta
  - tbtncSileTBtrans
  - cubeSile
  - pdbSile
  - xsfSile
  - xyzSile
  - asciiSileBigDFT
  - inSileFHIaims
  - gotSileGULP
  - hamiltonianSile
  - mdSileOpenMX
  - omxSileOpenMX
  - txtSileORCA
  - refSileScaleUp
  - restartSileScaleUp
  - aniSileSiesta
  - tshsSileSiesta
  - onlysSileSiesta
  - hsxSileSiesta
  - ncSileSiesta
  - structSileSiesta
  - xvSileSiesta
  - stdoutSileSiesta
  - deltancSileTBtrans
  - tbtavncSileTBtrans
  - phtncSilePHtrans
  - phtavncSilePHtrans
  - tbtprojncSileTBtrans
  - phtprojncSilePHtrans
  - tbtsencSileTBtrans
  - phtsencSilePHtrans
  - carSileVASP
  - chgSileVASP
  - locpotSileVASP
  - winSileWannier90
  - tbSileWannier90
  - centresSileWannier90

GridPlot
================
  - Grid
  - fdfSileSiesta
  - cubeSile
  - ncSileSiesta
  - chgSileVASP
  - locpotSileVASP
  - rhoSileSiesta
  - ldosSileSiesta
  - rhoinitSileSiesta
  - rhoxcSileSiesta
  - drhoSileSiesta
  - baderSileSiesta
[...]

PR #839, opened 2024-10-05
Improving documentation of plot handlers

pfebrer:
This is related to #838 

Most changes are just general improvements of the automatic documentation of plot handlers (regardless of who renders it).

The changes related to sphinx are only in `docs/conf.py` and `docs/build_docs.sh`, where I followed the approach explained in https://github.com/zerothi/sisl/issues/838#issuecomment-2394113346. I also made `conf.py` document the `Geometry.to` and `BrillouinZone.apply` dispatchers, as the difficulty to document dispatchers also affects them.

I have tried to use sphinx events, but I haven't succeeded. If you come up with a better way to document nested methods inside dispatchers then great, but I think that the current method, although hacky, it produces nice documentation results:

Plotting methods in the summary:
Screenshot from 2024-10-05 18-17-52

Plotting methods documentation:
Screenshot from 2024-10-05 18-18-11

Geometry.to dispatchs:
Screenshot from 2024-10-05 18-18-43

(this PR is not ready yet because I want to further improve the docstrings of the plotting methods, it's just so that you can see the approach that I'm taking and provide better ideas)

### DIFF
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -7,7 +7,11 @@
     .. autosummary::
     {% for item in attributes %}
         {% if not item.startswith('_') %}
+        {% if '.' in item %}
+            {{ name }}.{{ item }}
+        {% else %}
           ~{{ name }}.{{ item }}
+        {% endif %}
 	{% endif %}
     {%- endfor %}
     {% endif %}
@@ -26,7 +30,11 @@
                           'step_either', 'step_to',
 			  'isDataset', 'isDimension', 'isGroup',
 			  'isRoot', 'isVariable']) %}
-           ~{{ name }}.{{ item }}
+        {% if '.' in item %}
+            {{ name }}.{{ item }}
+        {% else %}
+          ~{{ name }}.{{ item }}
+        {% endif %}
       {% endif %}
    {%- endfor %}
    {% endif %}
--- a/docs/api/viz/index.rst
+++ b/docs/api/viz/index.rst
@@ -26,6 +26,7 @@ Plot classes are workflow classes that implement some specific plotting.
     GridPlot
     WavefunctionPlot
     PdosPlot
+    AtomicMatrixPlot
 
 Utilities
 ---------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -471,6 +471,8 @@ nbsphinx_prolog = r"""
 
 nbsphinx_thumbnails = {}
 
+nbsphinx_allow_errors = True
+
 import inspect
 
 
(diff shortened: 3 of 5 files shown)


pfebrer:
In this PR the automatic docstrings for plot handlets depend on `numpydoc`(https://github.com/numpy/numpydoc), which I think is not super bad to add as a dependency to `sisl.viz` as it avoids plenty of boilerplate code and it is a lightweight pure python package.

zerothi:
Hmm.. I have looked a bit into this.

And it seems this is the wrong way to go about this.

- `xarray`
- `pandas`
- etc.

are all explicitly creating their `api.rst` documents. Perhaps we should too?

This would also enable accessor documentation as we would like it:

```rst
Plotting
--------

.. autosummary::
   :toctree: generated/
   :template: autosummary/accessor_method.rst

   Dataset.plot.scatter
   Dataset.plot.quiver
   Dataset.plot.streamplot
```
above is an excerpt from `xarray`.

Only thing that is needed is the package `sphinx-autosummary-accessors`.

I think this could solve all issues, albeit it would be a somewhat hurdle to get right for the first time *and* it requires us to manually add functions to the documentation. However, perhaps that could solve other issues, such as grouping, proper documentation style etc.

What would be your thoughts on this transition?

zerothi:
It might require a bit of adjustment and play with it. But it seems to be easier to manage in the long run, and I don't like the hacky way of building a project with errors, then patching it and *hoping* nothing broke... It is too fragile IMHO.

pfebrer:
I think it could make sense to create our custom `.rst` files but doing it manually seems like a very bad idea, because it is just a matter of time that something will get outdated.

We could have some scripts to automatically generate the rst files. 

> and I don't like the hacky way of building a project with errors, then patching it and hoping nothing broke... It is too fragile IMHO.

Users are never going to build the documentation, it is always built in a controlled environment, so it is not so clear to me that the approach is that bad. But yeah if you have an idea for automatizing the custom rst files it is better of course. I can do it if you have a clear picture of what you want.

Otherwise if this issue is going to stall because of lack of time I think having the documentation built with the hack is much better than having undocumented parts of the code.

pfebrer:
The replacement of the placeholder could be done also within sphinx I think, but I haven't found the exact way to do it yet. That would be less hacky in the sense that it doesn't need external scripts so it would work every time

zerothi:
> I think it could make sense to create our custom `.rst` files but doing it manually seems like a very bad idea, because it is just a matter of time that something will get outdated.


It is only manual for the *overview*, not for the actual documentation. I.e. see [here](https://github.com/pydata/xarray/blob/main/doc/api.rst). 
It is definitely not everything, but I think it is doable.

> We could have some scripts to automatically generate the rst files.
Sure, that could work, but to allow more complicated things, perhaps this is fine.

> 
> > and I don't like the hacky way of building a project with errors, then patching it and hoping nothing broke... It is too fragile IMHO.
> 
> Users are never going to build the documentation, it is always built in a controlled environment, so it is not so clear to me that the approach is that bad. But yeah if you have an idea for automatizing the custom rst files it is better of course. I can do it if you have a clear picture of what you want.

Just like we had an issue with the notebooks not fully working (which required us to actually see it) this could pose problems. I agree users don't interwene, but these things always bite back at some point...

pfebrer:
> It is only manual for the overview, not for the actual documentation. I.e. see [here](https://github.com/pydata/xarray/blob/main/doc/api.rst).
> It is definitely not everything, but I think it is doable.

That page is generated automatically, no?

zerothi:
> > It is only manual for the overview, not for the actual documentation. I.e. see [here](https://github.com/pydata/xarray/blob/main/doc/api.rst).
> > It is definitely not everything, but I think it is doable.
> 
> That file is generated automatically, no?

I am pretty sure its hand-made... But rather simple, right, the `Dataset` methods/attributes are not here.
The way `xarray` does this is to *not* document the `plot` method, that one is documented [here](https://docs.xarray.dev/en/latest/api.html#plotting) so they are kind of separating these things. Annoying, but perhaps makes sense, since it

pfebrer:
Actually I find `xarray`'s documentation hard to navigate. Each method has its own page and therefore to see the documentation for another method of the same class you have to go back to the main page, look at the list of methods and click on the other one that you think might be what you want. The navigation that they provide at the bottom saying "Previous", "Next" is useless in my opinion:

Screenshot from 2024-10-07 14-15-13

In any case, I think `xarray` is quite different because it mainly has two important classes: `DataArray` and `Dataset`. The rest of classes are very rarely used. In `sisl` there are lots of classes that are important and I think it is a good idea to keep the methods for each class contained on their own page.

I have been playing with `sphinx-autosummary-accessors` and I don't get what I want because it only creates a new page for the accessor method, it doesn't add the method documentation's in the page of the class as all other methods. Also I haven't managed to make it document the signature and the help message for each parameter, I just got this:

Screenshot from 2024-10-07 14-22-43

zerothi:
I agree xarray is horrible in may ways... :(

But, I don't think we should put it in like this, we will never fix it... And I am afraid we end up with bugs later on that are hard to deal with...

pfebrer:
Ok, so I discovered that autoclass actually has no problem with attributes that contain a dot and will document them.

The only problem is with autosummary, which interprets a dot as a "get attribute" and can only show either the full name or the last word after a dot. E.g. for the pdos plot handler either `pdos` or `Hamiltonian.plot.pdos`.

So I removed the placeholder thing and updated the class template for autosummary so that it shows this:

Screenshot from 2024-10-08 14-22-22

This is a step, but I think it will be not entirely clear for users. What I have thought also is to separate the dispatchers in a different autosummary. Maybe then the long path is not so confusing...

zerothi:
Yeah, I think we need to separate it in some way. `plot` would also benefit from having its own section.

PR #843, opened 2024-10-10
Fix sparse grid tests

pfebrer:
This PR handles the previous problems that the sparse grid had when reducing orbital products because of a mismatch in `nsc` between the orbital values and weights (e.g. DM). This resulted in trying to access arrays outside of their memory. This PR:

- Introduces a check to see if the shape of the weights array provided is consistent with the `nsc` of the sparse grid.
- When computing the density from the DM, adds a check to set `nsc` of the matrix to the sparse grid one.
- Added the possibility of truncating the orbital values in the sparse grid with the `nsc` of the geometry. This is done by an extra `truncate_with_nsc` argument in `Geometry._orbital_values`.
- Change the `nsc` in sparse grid tests so that it works.


src/sisl/_core/geometry.py
```diff
@@ -3733,6 +3737,11 @@ def _orbital_values(self, grid_shape: tuple[int, int, int]):
         # supercell indices. Here we revert that
         ISC -= np.floor(self.fxyz[IA]).astype(int32)
 
+        # Don't consider atoms that are outside of the geometry's auxiliary cell.
+        if truncate_with_nsc:
+            mask = ~(abs(ISC) > self.nsc // 2).any(axis=1)
```
zerothi, line 3742: why not `mask = (abs(ISC) <= self.nsc // 2).any(axis=1)`?

zerothi (changes requested):

src/sisl/_core/geometry.py
pfebrer: Because that would mean "there is an axis that is within the constrains of `nsc`", and what I want is "there is no axis that is outside of the constraints of `nsc`".

Maybe I should use parenthesis to clarify it:

```python
~((abs(ISC) > self.nsc // 2).any(axis=1))
```

pfebrer: So actually I could do `(abs(ISC) <= self.nsc // 2).all(axis=1)`

zerothi: sorry, `(abs(ISC) <= self.nsc // 2).all(axis=1)`?

pfebrer: changed

zerothi: same code, same time, nice ;)

zerothi (approved):

Issue #728, opened 2024-03-23
Documentation of SileBinder methods


tfrederiksen:
The documentation is not very informative for the `SileBinder` methods, e.g., in `txtSileORCA`:
https://zerothi.github.io/sisl/api/io/generated/sisl.io.orca.txtSileORCA.html#sisl.io.orca.txtSileORCA.read_energy just shows:
```
read_energy = <sisl.io._multiple.SileBound object> [source]
```
Here, for instance, the optional `units` argument is not shown.

Can this be improved?

zerothi:
It is a known issue, see #632. I have yet to figure out why it exactly happens...
I'll close this as it is a duplicate of #632

tfrederiksen:
Indeed 😅

PR #845, opened 2024-10-10
Accessors and grouping documentation entries

zerothi:
This is an extended try of @pfebrer try in #839.

I took the details there, and played with it, and extended it a bit more.

So now basically all is done in the template files.
It ain't pretty, but is self-contained.

I kind-of-don't like it. But it seems much simpler than hacking in the python code, and extending what autodoc does.

The advantages are:
- we have manual control of how the layout should be
- we can also double document a few details, if needed, e.g. the `to`/`plot` could both be in their designated sections, but also in the `Methods` section.
- it is very easy to add another section
- right now everything is automatic, I just check whether an attribute of a class is an instance of a `sisl._dispatch.AbstractDispatcher`, and patches the attributes based on that.

Disadvantages:
- we don't have access to the actual documentation, so can't really do anything there in the jinja templates.
- the way we *get* the documentation is very hacky.... It overwrites attributes in the classes, which it shouldn't do, but
  it gives the effect. It just means that tool-tips might not work... :(
- fine-grained control is limited to a rather horrible jinja templating engine, which is extremely verbose, and sometimes limited.

@pfebrer let me know what you think?

Of course, the actual docs are still missing... ;)

Here are a few pics:

Type 1:
image

Type 2:
image

Type 3:
image

 - [x] Closes #839

### DIFF
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -1,33 +1,191 @@
-{% extends "!autosummary/class.rst" %}
+{{ fullname | escape | underline}}
 
-    {% block attributes %}
-    {% if attributes %}
-    .. rubric:: {{ _('Attributes') }}
+{# Global variables specified for use further down #}
 
-    .. autosummary::
-    {% for item in attributes %}
-        {% if not item.startswith('_') %}
-          ~{{ name }}.{{ item }}
-	{% endif %}
-    {%- endfor %}
-    {% endif %}
-    {% endblock %}
+{% set check_ns = namespace(
+      plot=false,
+      to=false,
+      new=false,
+      apply=false,
+      )
+%}
+{% set specials = [
+      'plot',
+      'apply',
+      'to',
+      'new',
+   ]
+%}
 
-   {% block methods %}
-   {% if methods %}
+{% macro check_sections() %}
+{# Check for variables that should be in specific sections #}
+
+{% for things in varargs %}
+{% for item in things %}
+
+   {# Determine whether we should do any Plotting section #}
+   {% if item.startswith('plot.') %}
+      {% set check_ns.plot = true %}
+   {%- endif %}
+
+   {% if item.startswith('to.') %}
+      {% set check_ns.to = true %}
+   {%- endif %}
+
+   {% if item.startswith('new.') %}
+      {% set check_ns.new = true %}
+   {%- endif %}
+
+   {# Determine whether we should do any Dispatch section #}
+   {% if item.startswith('apply.') %}
+      {% set check_ns.apply = true %}
+   {%- endif %}
+
+{%- endfor %}
+{%- endfor %}
+{%- endmacro %}
+
+
+{% macro extract_startswith(start, short) %}
+{#
+
+Write out a list of items that starts with `start`
+
+Parameters
+ start :
+      what to check if the item starstwith. If yes, then keep it.
+ short :
+      whether it should be formatted like ~{{name}}.{{item}} (true)
+      or without the ~.
+
+#}
+
+{% for things in varargs %}
+{% for item in things %}
+
+   {% if item.startswith(start) %}
+      {% if short %}
+      ~{{ name }}.{{ item }}
+      {% else %}
+      {{ name }}.{{ item }}
+      {%- endif %}
+   {%- endif %}
+
+{%- endfor %}
+{%- endfor %}
+{%- endmacro %}
+
+
+{% macro accepted_methods() %}
+{% set skip_methods =
+                   ['ArgumentParser', 'ArgumentParser_out',
+                    'is_keys', 'key2case', 'keys2case',
+                    'line_has_key', 'line_has_keys', 'readline',
+                    'step_either', 'step_to',
+                    'isDataset', 'isDimension', 'isGroup',
+                    'isRoot', 'isVariable'] + specials
+%}
+
+{# Loop on the arguments #}
+{% for things in varargs %}
+{% for item in things %}
+
+   {% set tmp = namespace(keep=true) %}
+
+   {#
+
+   Use the namespace variable to pass down
+   contexts. This is needed because variables
+   in Jinja are hard scoped.
+   #}
+
+   {% if item.startswith('_') %}
+      {% set tmp.keep = false %}
+   {%- endif %}
+   {% if item in skip_methods %}
+      {% set tmp.keep = false %}
+   {%- endif %}
+
+   {% for starts in specials %}
+   {% if item.startswith(starts + '.') %}
+      {% set tmp.keep = false %}
+   {%- endif %}
+   {%- endfor %}
+
+   {% if tmp.keep %}
+   {# Call the parent caller with the item that has
+      passed all tests for eligibility
+   #}
+   {{ caller(item) }}
+   {%- endif %}
+
+{%- endfor %}
+{%- endfor %}
+
+{%- endmacro %}
+
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+
+{#
+   Just call the macro, this won't write anything,
+   only set some variables in the check_ns namespace variable
+#}
+{{ check_sections(attributes, methods) }}
+
+{% block attributes %}
+{% if attributes %}
+   .. rubric:: {{ _('Attributes') }}
+
+   .. autosummary::
+      {% call(item) accepted_methods(attributes) %}
+      ~{{ name }}.{{ item }}
+      {%- endcall %}
+{%- endif %}
+{%- endblock %}
+
+{% if check_ns.to or check_ns.new %}
+   .. rubric:: {{ _('Conversions') }}
+
+   .. autosummary::
+      {% if check_ns.new %}
+      {{ name }}.new
+      {% endif %}
+      {{ extract_startswith('new.', false, attributes, methods) }}
+      {% if check_ns.to %}
+      {{ name }}.to
+      {% endif %}
+      {{ extract_startswith('to.', false, attributes, methods) }}
+
+{%- endif %}
+
+{% if check_ns.plot %}
+   .. rubric:: {{ _('Plotting') }}
+
+   .. autosummary::
+      {{ name }}.plot
+      {{ extract_startswith('plot.', false, attributes, methods) }}
+
+{%- endif %}
+
+{% if check_ns.apply %}
+   .. rubric:: {{ _('Dispatching') }}
+
+   .. autosummary::
+      {{ name }}.apply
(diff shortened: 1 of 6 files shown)


pfebrer:
Wow, first of all let me say that what you have managed to do inside the jinja template is very impressive 😅 

Also, I love the automatic finding of the dispatchers to ensure everything is automatically documented. This is something that I would do, but I didn't think you would like :)

I have two suggestions:

1. Out of your disadvantages, the one that I find most important is that putting all the logic in a jinja template looks horrible and makes the template hard to edit with confidence for future developers that are not you or me. I have seen that there is a variable that you can define in `conf.py`, `autosummary_context`, that is a dictionary of variables that become available to the template (docs [here](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#confval-autosummary_context), and is used [like this](https://github.com/sphinx-doc/sphinx/blob/2f1d775dfda9e4f81dfff6cfbe9edf7731e32a97/sphinx/ext/autosummary/generate.py#L295)). This is not per-class, it is a global context. But still, while we assign the nested attributes in `conf.py` we could build a table that is available in the template, where you would just query the table using the name of the class. This would move most of the logic out of the jinja template, which could be made fairly simple again.
2. We really need to find a way to show `apply.iterate` instead of `BrillouinZone.apply.iterate`. Things would look much cleaner and would be easier to interpret for the user. There is no configuration available to do this, see [here](https://github.com/sphinx-doc/sphinx/blob/2f1d775dfda9e4f81dfff6cfbe9edf7731e32a97/sphinx/ext/autosummary/__init__.py#L320-L323) how autosummary determines the display name. So I see only to options: either we wrap the `get_items()` method of autosummary to postprocess the display names at our will or we don't use the autosummary directive and build the table ourselves. I think that although the first option is a bit hacky it is probably the best (we are already in a hacky mood, so who cares about one more little hack 😅 )

pfebrer:
For suggestion (2) we could also contribute to `sphinx.ext.autosummary` to provide a config to modify the display name. I don't know how easy/hard it is to put things in there.

pfebrer:
Well actually we could also contribute the thing about creating a per-class context for the template.

pfebrer:
I decided to take a chance https://github.com/sphinx-doc/sphinx/issues/13003, let's see :)

zerothi:
> Wow, first of all let me say that what you have managed to do inside the jinja template is very impressive 😅
> 
> Also, I love the automatic finding of the dispatchers to ensure everything is automatically documented. This is something that I would do, but I didn't think you would like :)
> 
> I have two suggestions:
> 
>     1. Out of your disadvantages, the one that I find most important is that putting all the logic in a jinja template looks horrible and makes the template hard to edit with confidence for future developers that are not you or me. I have seen that there is a variable that you can define in `conf.py`, `autosummary_context`, that is a dictionary of variables that become available to the template (documented [here](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#confval-autosummary_context), and is used [like this](https://github.com/sphinx-doc/sphinx/blob/2f1d775dfda9e4f81dfff6cfbe9edf7731e32a97/sphinx/ext/autosummary/generate.py#L295)). This is not per-class, it is a global context. But still, while we assign the nested attributes in `conf.py` we could build a table that is available in the template, where you would just query the table using the name of the class. This would move most of the logic out of the jinja template, and the template could be made fairly simple again.

Great, I didn't find this doc, only `html_context` which isn't passed to the `autosummary` templates, thanks, I'll have a stab at that!

> 
>     2. We really need to find a way to show `apply.iterate` instead of `BrillouinZone.apply.iterate`. Things would look much cleaner and would be easier to interpret for the user. There is no configuration available to do this, see [here](https://github.com/sphinx-doc/sphinx/blob/2f1d775dfda9e4f81dfff6cfbe9edf7731e32a97/sphinx/ext/autosummary/__init__.py#L320-L323) how autosummary determines the display name. So I see only two options: either we wrap the `get_items()` method of `Autosummary` to postprocess the display names at our will, or we don't use the autosummary directive and build the table ourselves instead. I think that although the first option is a bit hacky it is probably the best (we are already in a hacky mood, so who cares about one more little hack 😅 )

Yes, that is really required, but so far still manageable if `BrillouinZone` is shown (not pretty... But...).

zerothi:
I can't figure out how to do namespace retrieval of variables...

```jinja
{% set check = namespace() %}
{% for item in ["foo", "bar"] %}
{% set check.{{item}} = false %} <- does not work
{% set check[{{item}}] = false %} <- does not work
{% endfor %}
```
so, some variables just needs to be hard-coded... :(

pfebrer:
It should be `check[item]`, no?

pfebrer:
And probably without the `set`:

```
check[item] = false
```

But I don't understand why you need to set things inside the template.

zerothi:
I tried both, none work...

The idea was to have the *special* fields defined in the `conf.py`, and then dynamically create the `namespace` to check if the section should be there or not. I can't find a way around this... :(

pfebrer:
Why do you use `namespace`? Can't you create a normal dictionary?

pfebrer:
But I don't understand why you need to set variables inside the template, can't you set them all in conf.py?

zerothi:
No, because I want to change them in the template. I have to discover the methods in the template. I.e. if `plot.*` exists, then create `Plotting` section, else don't. And I can't figure out a way to do this programmatically.

I found a way, it ain't pretty.. Dynamic lists are the way to go...

pfebrer:
What I had in mind was that in `conf.py` you could build a context like:

```python
autosummary_context = {
   "info": {
       "Hamiltonian": {...info for the Hamiltonian},
       "BrillouinZone": {...info for the BrillouinZone}
    }
}
```
And then in the template do something like:
```
class_info = info[objname]
```

So you would compute all the needed variables for each class in `conf.py`

zerothi:
> What I had in mind was that in `conf.py` you could build a context like:
> 
> ```python
> autosummary_context = {
>    "info": {
>        "Hamiltonian": {...info for the Hamiltonian},
>        "BrillouinZone": {...info for the BrillouinZone}
>     }
> }
> ```
> 
> And then in the template do something like:
> 
> ```
> class_info = info[objname]
> ```
> 
> So you would compute all the needed variables for each class in `conf.py`

Lets, take that as a 2nd step, lets start by getting this to work, which is already hard enough ;)

pfebrer:
Ok, I thought it would be much easier. I don't understand what you want to use the `autosummary_context` for then, but let's see.

zerothi:
@pfebrer I think this is ready for a quick review, just whether you like the overall details?

Some issues remaining:

The method that is documented is controlled [here](https://github.com/zerothi/sisl/blob/55ca9a0f9240cb8381da8c87504fdb5a73f2486c/docs/conf.py#L655)
As you can see, one can finetune these details.
You can try and building this in two cases, one as it is, and one where it has:
```python
        dispatcher_name = (dispatcher_name, None)
```
then you'll find:
1. building it with no changes uses the *docstring* from the `dispatch` methods, which actually does the lifting. So here you get the parameter list from that dispatch method, which isn't correct..
2. building it with the change uses the docstring for the class that holds the `dispatch` method. Here you get the `__call__` method documented, which isn't great either.

In both cases are the signatures wrong. And generally the signatures change depending on what you request provide.
So how to solve this is not exactly clear to me...


You'll also note that the latest commits fixes #728 ! :)
- I tr

pfebrer:
I think signatures should display exactly the way in which the functions can be called. This is what I will do with the plotting in #839 once this gets in.

I guess logically what should be documented is the `__call__` method, because this is also what code editors (e.g. jupyter notebook) will show. The approach I follow with the plot handlers is this:

https://github.com/zerothi/sisl/blob/0af33dbe913c84cce900bb57035699b4568fced1/src/sisl/viz/_plotables.py#L80-L88

That is, I create a subclass for each dispatch to be able to document them individually.

pfebrer:
Are you sure that you want to display apply dispatchs as functions? I think across sisl documentation the displayed usage is as an attribute:

```python
bz.apply.average.eigenstate()
``` 
instead of:

```python
bz.apply.average("eigenstate")
``` 

Isn't it?

pfebrer:
While we can not modify upstream `autosummary`, here is a workaround for display names:

This will use "|" as a marker for where the display name starts, as you were proposing:

```python

from sphinx.ext.autosummary import Autosummary

class CustomAutosummary(Autosummary):
    """Wrapper around the autosummary directive to allow for custom display names.
    
    In particular, | is interpreted as the mark to indicate where the display name
    should begin.
    """

    def get_items(self, names):

        # Find all names that have a | marker, and store their index along with the name that we want to display
        display_names = [(i, name.split("|")[-1]) for i, name in enumerate(names) if "|" in name ]
        # Clean the names so that Autosummary can process them
        names = [name.replace("|", "") for name in names]

        # Process names with Autosummary
        items = super().get_items(names)

        # Substitute the display names that we determined. This is only possible if the number of
        # items is the same as the number of names, i.e. no import errors have ocurred for any item.
        # Otherwise the indices will not match.
        if len(names) == len(items):
            for i, name in display_names:
                items[i] = (name, *items[i][1:])

        return items
```

This custom class must be registered as the autosummary directive:

```python
def setup(app):
    app.add_directive("autosummary", CustomAutosummary, override=True)
```

Then you add the "|" marks to the template and it looks beautiful :)

pfebrer:
Another approach to display names that I think is much more robust:

```python
from sphinx.ext.autosummary import Autosummary
from docutils.parsers.rst import directives

class CustomAutosummary(Autosummary):
    """Wrapper around the autosummary directive to allow for custom display names.
    
    Adds a new option `:removeprefix:` which removes a prefix from the display names.
    """

    option_spec = {
        **Autosummary.option_spec,
        "removeprefix": directives.unchanged
    }

    def get_items(self, names):
        items = super().get_items(names)

        remove_prefix = self.options.get("removeprefix")
        if remove_prefix is not None:
            items = [(item[0].removeprefix(remove_prefix), *item[1:]) for item in items]

        return items

def setup(app):
    app.add_directive("autosummary", CustomAutosummary, override=True)
```

Then you add the option to the autosummary directives for the dispatchers, e.g.:

```rst
.. autosummary::
    :removeprefix: {{ name }}.

    {{ name }}.plot
    {{ extract_startswith('plot.', false, attributes, methods) }}
```

zerothi:
@pfebrer thanks, I have implemented your last suggestion!!!

zerothi:
@pfebrer perhaps we should merge this, and then continue on the documentation stuff in later PR's?

pfebrer:
Yes, sounds good! It is already much better than it was :+1:

Now it is just a matter of making sure all the dispatchs are properly documented in general, and sphinx is ready to show the documentation.

zerothi:
This superseeds #839.

PR #848, opened 2024-10-17
Allow for any type of integer when requesting neighbor items

pfebrer:
This solves the problem of not being able to use numpy integers to index neighbor lists


zerothi (changes requested):
Missing `isort` on sources

pfebrer:
I did this so fast that I really thought I had sorted the imports manually, but apparently the order of the alphabet is not so clear to me :smile:

PR #849, opened 2024-10-18
Improve plotables documentation

pfebrer:
Improves the docstrings and signatures of plotting functions so that they show correctly in the documentation.

It introduces a new dependency for `sisl.viz` to have nice docstrings on the plotting dispatchers: `numpydoc`. However, since `numpydoc` depends on `sphinx` I preferred to not make it a hard requirement so `sisl.viz` can work without it.


pfebrer:
Hi @zerothi, remember to merge this one please :)

zerothi:
Yeah, sorry, I have been on holidays.

I have to say, that putting `numpydoc` and thus `sphinx` is too much. I really don't like that. I thought it was a slim dependency, but I think this is the wrong way to go.

What exactly is it doing (other than merging the `Parameters` section)? Could a simple solution just be to put the *amended* documentation parameters into an [`Other Parameters`](https://numpydoc.readthedocs.io/en/latest/format.html#other-parameters) section?
And, it seems you are also removing things from the parameters section... 

I know you like the tooltip that it handles, but can't we solve this in a simpler way?

Note, if I do a simple venv with sisl and one with only numpydoc I get:
```shell
du -hs venv-*                                                                                                                                                                
102M	venv-numpydoc        
377M	venv-sisl
```
so numpydoc is adding 102M of additional requirements (the overlap is minimal disk-space).

pfebrer:
I don't think a section with Other Parameters makes sense, users won't understand it.

The other way to solve this is to define a function to parse docstrings inside `sisl`. We could simply copy this file from `numpydoc`: https://github.com/numpy/numpydoc/blob/main/numpydoc/docscrape.py, as it is the only thing I'm using from `numpydoc`.

zerothi:
Ok, that sounds viable, see e.g. how scipy is doing it https://github.com/scipy/scipy/pull/21215/files

They do exactly what you propose, I would be fine with that!

Put it somewhere at the top `sisl._lib/utils/_docscrabe` or something.

pfebrer:
Done!

zerothi:
ha, I meant `/` as different options, I'll push to move it to `sisl._lib`

PR #850, opened 2024-10-23
trying to fix the within_inf problem

### LINKED ISSUE #649: `within_inf` returns wrong supercell indices?
**Describe the bug**

The supercell indices that `within_inf` returns are for the atoms translated to the unit cell, is this intended?

**Code to reproduce problem**
```python
import numpy as np

# Create a geometry with an atom outside of the unit cell
geometry = sisl.Geometry([-0.5, 0, 0], lattice=np.diag([2, 10, 10]))

# Search for atoms
search = sisl.Lattice(np.diag([3, 10, 10]))
ia, xyz, isc = geometry.within_inf(search, periodic=[True, True, True])

print("ia", ia)
print("xyz", xyz)
print("isc", isc)
```
```
ia [0]
xyz [[1.5 0.  0. ]]
isc [[0 0 0]]
```

As you can see, it says that the image of atom 0 on sc `(0,0,0)` is inside the search lattice, which is a lie. The supercell index should be `(1, 0, 0)`, unless the function specifies that it will first translate atoms to the unit cell.

I understand even less the output if I change the search to `[2, 10, 10]`:

```python
import numpy as np
geometry = sisl.Geometry([-0.5, 0, 0], lattice=np.diag([2, 10, 10]))

search = sisl.Lattice(np.diag([2, 10, 10]))
ia, xyz, isc = geometry.within_inf(search, periodic=[True, True, True])

print("ia", ia)
print("xyz", xyz)
print("isc", isc)
```
```
ia []
xyz []
isc []
```

It says there are no atoms in the search lattice, this is clearly wrong, isn't it? Regardless of whether the first one is intended.

zerothi:
Basically the isc did not respect the initial placement of the atomic coordinates, so if atoms were already outside of the unit-cell, then isc would be wrong.

@pfebrer I think this corrects your remarks in #649.
Comments?

 - [x] Closes #649

### DIFF
--- a/src/sisl/_core/_ufuncs_geometry.py
+++ b/src/sisl/_core/_ufuncs_geometry.py
@@ -1014,9 +1014,10 @@ def translate(
     """
     g = geometry.copy()
     if atoms is None:
-        g.xyz += np.asarray(v, g.xyz.dtype)
+        g.xyz += v
+        np.asarray(v, g.xyz.dtype)
     else:
-        g.xyz[geometry._sanitize_atoms(atoms).ravel(), :] += np.asarray(v, g.xyz.dtype)
+        g.xyz[geometry._sanitize_atoms(atoms).ravel(), :] += v
     return g
 
 
--- a/src/sisl/_core/geometry.py
+++ b/src/sisl/_core/geometry.py
@@ -3631,11 +3631,11 @@ class Geometry(
         tile_max = ceil(idx.max(0)).astype(dtype=int32)
 
         # Intrinsic offset (when atomic coordinates are outside primary unit-cell)
-        fxyz = self.move(1e-8).fxyz
+        fxyz = np.round(self.fxyz, decimals=5)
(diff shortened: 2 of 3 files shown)


pfebrer:
This fixes the first code snippet in #649, but not the second one :sweat_smile: 

By the way, before getting this in, we have to remove these lines: https://github.com/zerothi/sisl/blob/614aadbbb352e53b4c8b3d677014b3281eff6648/src/sisl/_core/geometry.py#L3736-L3738

zerothi:
> This fixes the first code snippet in #649, but not the second one 😅
> 
> By the way, before getting this in, we have to remove these lines:
> 
> https://github.com/zerothi/sisl/blob/614aadbbb352e53b4c8b3d677014b3281eff6648/src/sisl/_core/geometry.py#L3736-L3738

I think I tried that, but I'll double check! Thanks for the heads up of the blob, I'll add this to this PR then!

PR #851, opened 2024-10-24
major restructure of the documentation

zerothi:
New sections, and moved the viz section to a more
sensible name.
Perhaps it should be changed altogether.
Currently the ASE viz has been removed, is it even required? It might be easier to just document ASE conversion.

@pfebrer what do you say to this restructuring?

I think we should strive to move the sections a bit closer so the navigation is more intuitive.

The current layout looks something like this:

image

Suggestions?


pfebrer:
Looks good to me! ASE visualization was just something that was there before `sisl.viz`, I think indeed it can be removed.

PR #859, opened 2024-11-06
Use cutoff from basis files (xml, nc)

### LINKED ISSUE #858: Cutoff radius from ion.xml files is ignored.
When reading a basis from an `ion.xml` file, the cutoff radius is ignored and instead is computed with the `radial_minimize_range` function here:

https://github.com/zerothi/sisl/blob/6ef44002aa6b41118e93272fb08a081ad70a6d2a/src/sisl/_core/orbital.py#L340

Wouldn't it make more sense to use the cutoff that is specified in the file?

pfebrer:
Closes #858 

It considers the cutoff defined in the `ion.xml` and `ion.nc` files when the basis is read.


zerothi:
THANKS!

PR #860, opened 2024-11-06
New wheels

zerothi:
Enables python 3.13 wheels.


zerothi:
This changes, quite drastically, the wheel creation.

The idea is that the `cibuildwheel` process should handle everything. And this should enable a simpler workflow. And also more easily a customization for specific versions (as is currently needed for 3.13 with free-threading).

PR #861, opened 2024-11-06
Cython perf simplification

### LINKED ISSUE #104: Pk on complex matrices
**Describe the bug**
I need to fix the pyx codes to handle typedef's for different data-types

Currently Pk for complex matrices does not work... :(

zerothi:
- [ ] Closes #104

Here are some details that needs to be sorted out before merging:

- [ ] Adding tests for complex spin-orbit calculations, this includes post-processing like PDOS etc.
      The idea is that instead of 8 numbers, we can now do 4, `[uu, dd, ud, du]`, this has the problem of
      overlap being stored as a complex number, but that might be *ok*.
- [x] Document the SOC TB handling a bit better, right now the order of elements are not intuitive.


zerothi:
@pfebrer you won't like this...

Sadly, pure-python mode does not allow cimport statements.

This code-restructuring is a first step to reduce the amount of Cython code.

zerothi:
> > Sadly, pure-python mode does not allow cimport statements.
> 
> It does, see here:
> 
> https://github.com/zerothi/sisl/blob/6bd13954ecae5a8fb83a70342335e76f8e2b5bf9/src/sisl/geom/_neighbors/_operations.py#L8-L11
> 
> Do you mean that it doesn't allow to `cimport` sisl files?

True, I misunderstood this here...

You're right, but for this to work for sisl intrinsics, you would have to duplicate interfaces in `pxd` files, *and* you wouldn't be able to run that code in pure python mode, which kind of defeats the purpose. :S

See here: http://docs.cython.org/en/latest/src/tutorial/pure.html#cimports

pfebrer:
Well, of course it depends on how complex things are, but you could always mock the `cimports` inside an `if cython.compiled:` clause so that the file runs in the python interpreter.

But apart from files running out of the box in normal python (which is nice), it is a question of readability and the feeling that the code can be ported to something else much more easily if there is a better accelerating framework than `cython`.

Not saying we should translate everything to pure python, just giving some arguments in favour of the pure python syntax :)

Issue #799, opened 2024-06-25
sisl.viz.FatBandsPlot error in fatbands_mode = 'line/scatter'


rreho:
**Describe the bug**
From issue #790 I created a test with PbTe trying to test if `sisl.viz.FatBandsPlot` works with `fatbands_mode` different than `area_line`
I share here a zip folder with input files and databases. You will also find a jupyter notebook : ``pp_pbte_test.ipynb' with comments and questions inside.
[test.zip](https://github.com/user-attachments/files/15972700/test.zip)

Briefly, I reproduced the bug related to `fatbands_mode`. Moreover, I tried to visualize the projected  (fatbands) band structure starting from the `fatbands.nodes['bands_data']` and it seems to be work

ing. It would be great if you can have a look at the notebook to check that I do it properly and that it reproduce the expected results.
 
I am not sure I interpreted correctly the values stored in `data.norm2`. Are those normalized to 1? I find the maximum value to be around 0.92 

**Version details**
3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:50:58) [GCC 12.3.0]
0.1.dev4752+g1fa01ad

zerothi:
> Thanks! I will try to look at it during this week.
> 
> I think that all `norm2` from an orbital should sum to 2 (if the calculation is spin unpolarized)

No, `norm2` should also give 1 for eigenstates, it just doesn't take the square root of the inner product.

pfebrer:
Ok, so both:

```python
norm2.sum("band")
# and
norm2.sum("orb")
```

should give an array of ones for a spin unpolarized calculation, no?

I recall that in one case summing over bands didn't result into 1. With non-colinear spin maybe?

rreho:
In my case: 
``` norm2.sum("band") = all 2
      norm2.sum("orb") =all 1
```
I also tried:
``` np.sum(np.sum(data.norm2[:,:,:],axis=2),axis=1) = all 72```
where 72 is two times the number of orbitals

zerothi:
`norm2` is equivalent to $\langle \psi | \mathbf S | \psi \rangle$, so for $\psi$ being eigenstates it should always be 1 (otherwise there is an error! ;)). Even for non-colinear.

In @rreho case it seems like you are doing a non-colinear calculation, since you have 2 times number of orbitals.

pfebrer:
[...]

PR #862, opened 2024-11-07
Fixed matplotlib fatbands and enabled opacity

pfebrer:
#790 had not been fully fixed by #799. There were some bugs regarding `fatbands_mode="scatter"` and `fatbands_mode="line"` for the matplotlib backend.

Also enabled the possibility to use `"opacity"` for all variants of fatbands, to avoid some fatbands getting hidden by the overlap with the next ones.

With the example that @rreho provided using "scatter" "and matplotlib", and setting `"opacity": 0.2` for the red fatbands:

Screenshot from 2024-11-07 12-04-55

### DIFF
--- a/src/sisl/viz/figure/matplotlib.py
+++ b/src/sisl/viz/figure/matplotlib.py
@@ -238,6 +238,7 @@ class MatplotlibFigure(Figure):
         return {
             "dash": "dashed",
             "dot": "dotted",
+            None: "solid",
         }.get(dash, dash)
 
     def _sanitize_colorscale(self, colorscale):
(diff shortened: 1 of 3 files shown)


pfebrer:
Here with "area_line", where previously the red area would have hidden the blue one in the valence bands. Now it is clear that both contribute:

Screenshot from 2024-11-07 12-09-06

src/sisl/viz/figure/matplotlib.py
```diff
@@ -380,6 +381,8 @@ def draw_multisize_line(
         # Set the values used for colormapping
         lc.set_linewidth(line.get("width", 1))
         lc.set_linestyle(self._plotly_dash_to_matplotlib(line.get("dash", "solid")))
+        lc.set_color(line.get("color"))
+        lc.set_alpha(line.get("opacity"))
```
zerothi, line 385: a bit pedantic here, but matplotlib users are used to using `alpha=`, could it allow *both* things?

pfebrer: I think it would become a complete chaos if I try to support all keywords of all backends :sweat_smile: 
I took the decision to stick to plotly terminology in general to make things easier to maintain and document.

zerothi: ok, fair...

zerothi (approved):

PR #863, opened 2024-11-07
enabled ellipsis here and there

zerothi:
Should be improved so None fully gets removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@ we hit release version 1.0.0.
       sisl.geom.graphene
 
 - added Nambu spin configuration, this is still experimental
+- enabled `...` when extracting slices of MD steps in siesta output
+  files.
+  Here it is the same as `:`. But it also allows
+  inline arguments: `read_scf(imd=...)` where `imd=:` is not
+  allowed, partly fixes #835
+- enabled `...` for `atoms=` arguments. Selects all atoms.
 
 ### Fixed
 - `projection` arguments of several functions has been streamlined
@@ -29,10 +35,6 @@ we hit release version 1.0.0.
   with *many* edges in the sparse matrix, but a perf. hit for very
   small TB matrices.
 - dtype removed from `Spin` class
-- enabled `...` when extracting slices of MD steps.
-  Here it is the same as `:`. But it also allows
-  inline arguments: `read_scf(imd=...)` where `imd=:` is not
-  allowed, partly fixes #835
 
 
 ## [0.15.2] - 2024-11-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ we hit release version 1.0.0.
   with *many* edges in the sparse matrix, but a perf. hit for very
   small TB matrices.
 - dtype removed from `Spin` class
+- enabled `...` when extracting slices of MD steps.
+  Here it is the same as `:`. But it also allows
+  inline arguments: `read_scf(imd=...)` where `imd=:` is not
+  allowed, partly fixes #835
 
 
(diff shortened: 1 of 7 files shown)


pfebrer:
Are you sure you want to do this? This is not standard behavior. Numpy for example uses ellipsis to indicate, "for all other axes do nothing". I.e. it is not something that applies for a single axis.

zerothi:
> Are you sure you want to do this? This is not standard behavior. Numpy for example uses ellipsis to indicate, "for all other axes do nothing". I.e. it is not something that applies for a single axis.

But right now `read_scf(iscf=None)` quite unintuitively means *all* iscf elements, I can see that it isn't standard, but on the other hand, `iscf` is definitely (and will always be) a single dimension, so `...` is clear IMHO. Using `Opt.ALL` is also annoying, using some weird constant for obtaining this.

So before one would do:
```
read_scf(iscf=None)
read_scf(iscf=slice(None))
```
Why would `read_scf(iscf=...)` be so far from standard behaviour?

The standard says:
> The same as the ellipsis literal “...”. Special value used mostly in conjunction with extended slicing syntax for user-defined container data types. 

which I kind of think is ok here? But I would be happy to hear why not! ;)

pfebrer:
I have to admit when I say standard I mean `numpy`, but I would assume people follows numpy's conventions when it comes to array slicing.

In particular:

```python
array[..., 0]
```
means: 
- `numpy`: Give me the first element of the last dimension, regardless of how many dimensions there are.
- This PR: Give me the first element of the first dimension.

It is true that for now we don't support multiple dimensions, but this is where it seems the implementation would go if we supported them.

I am ok with using the ellipsis as arguments of the function call (although I don't see the problem with using None), but I would not support it in `__getitem__`, where `[:]` is much more natural.

zerothi:
> I have to admit when I say standard I mean `numpy`, but I would assume people follows numpy's conventions when it comes to array slicing.
> 
> In particular:
> 
> ```python
> array[..., 0]
> ```
> 
> means:
> 
>     * `numpy`: Give me the first element of the last dimension, regardless of how many dimensions there are.
> 
>     * This PR: Give me the first element of the first dimension.

I see, I think mainly this would be useful in the `iscf` arguments, so probably I should just remove it from the `SileBinder` as you say.

> 
> It is true that for now we don't support multiple dimensions, but this is where it seems the implementation would go if we supported them.
> 
> I am ok with using the ellipsis as arguments of the function call (although I don't see the problem with using None), but I would not support it in `__getitem__`, where `[:]` is much more natural.

Agreed, `getitem` is simpler with `:`. I can remove that here, but for `iscf=...` I think this reads more natural than `iscf=None`, no?

Let me fix it...

zerothi:
So, would only for `iscf` arguments be fine?

pfebrer:
> Agreed, getitem is simpler with :. I can remove that here, but for iscf=... I think this reads more natural than iscf=None, no?

I guess it depends on the person, I am used to `None` meaning "all". For example, whenever there is an `atoms` argument, `atoms=None` means "all atoms". But it could be that `...` feels more natural for some people. It does not hurt to support `...` if you think it is more natural, but I would keep both if possible (?)

zerothi:
Yeah, for sure we need to have them both.
So would you think adding `atoms=...` would be bad? :thinking: :stuck_out_tongue_winking_eye:

pfebrer:
(no one has ever complained about it)

zerothi:
To contest the `numpy` default.

This means *everything*:
```python
array[...] = 0.
```
so I would think that using ellipsis for *everything* in these arguments would make sense.

I have now amended so `atoms` and `orbitals` arguments can be `...` to mean *all*.

PR #870, opened 2024-11-13
Cleanup of the plotables module

pfebrer:
This fixes a bug: one could not pass **kwargs to the data generators when plotting through plot handlers.

Now:

```python

# Passing kwargs to data from the plot handler
object.plot(data_kwargs={"arg2": 2})

# Is equivalent to
data = DataClass.new(obj, arg2=2)
plot = PlotClass(data)
```

in other words, `data_kwargs` is correctly expanded to pass the **kwargs for the creation of the data.

**Most importantly**, this PR rewrites the `_plotables.py` with plenty of documentation so that it is easier to modify in the future.

### DIFF
--- a/src/sisl/viz/_plotables.py
+++ b/src/sisl/viz/_plotables.py
@@ -4,18 +4,52 @@
 from __future__ import annotations
 
 """
-This file provides tools to handle plotability of objects
+This file provides tools to handle plotability of objects.
+
+Registering `ClassA` as a plotable means that given an object of `ClassA`,
+one can plot it like:
+
+... code-block:: python
+
+    object.plot()
+    # or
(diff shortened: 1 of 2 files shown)


zerothi:
Thanks @pfebrer, perhaps we should have some tests that reflects what is possible, and how they should work. I'll merge this, and work on my other tbtrans branch. Thanks!

pfebrer:
Yes, definitely, and we should probably also discuss the behavior of **kwargs.

Right now, `nodify` workflows don't support `**kwargs` (nodes do support them). Therefore plots are not allowed to have **kwargs. So perhaps the best idea is to just pass the kwargs to the data function if they are declared. 🤷

Issue #873, opened 2024-11-21
Support for SIESTA-BdG in sisl


ahkole:
**Describe the feature**

As was mentioned in a discussion between @zerothi and @rreho https://github.com/zerothi/sisl/issues/869, it would be good to have support for the BdG Hamiltonian in sisl that is used/produced by SIESTA-BdG (which will be merged into main SIESTA sometime in the future) (see https://journals.aps.org/prb/abstract/10.1103/PhysRevB.110.134505 for implementation paper). Here we will supply some details about how the BdG Hamiltonian is implemented in SIESTA so that we can discuss what is necessary to add this support to sisl.

First of all, we will explain the structure of the Hamiltonian. The structure of the BdG Hamiltonian is (see paper for more details):

```
   !        | H^ee           Delta |   | H^ee          Delta     |
   ! HBdG = |                      | = |                         |
   !        | Delta^dagger   H^hh  |   | Delta^dagger  -(H^ee)^* |
```

Within SIESTA-BdG there are two matrices, one for `H^ee` and one for `Delta`, both with 8 spin components (same as SOC). The spin components have the following meaning in the SIESTA-BdG code:

```
   ! 1.  The electron-electron part uses the same structure as in the SOC case
   !     with an additional shift of the chemical potential to zero:
   !
   !                     | H_{jo,iuo}^{u,u}  H_{jo,iuo}^{u,d} |
   !     H^ee_{jo,iuo} = |                                    |
   !                     | H_{jo,iuo}^{d,u}  H_{jo,iuo}^{d,d} |
   !
   !                                            | S_{jo,iuo}  0          |
   !                                 - mu       |                        |
   !                                            | 0           S_{jo,iuo} |
   !
   !         | H(ind,1) + i H(ind,5) - mu S(ind),  H(ind,3) - i H(ind,4)             |
   !       = |                                                                       |
   !         | H(ind,7) + i H(ind,8)            ,  H(ind,2) + i H(ind,6) - mu S(ind) |
   !
   !
   !
   ! 2.  The hole-hole part is -conjg(H^ee) and hence very similar
   !
   !     H^hh_{jo,iuo} = -(H^ee)^*
   !
   !         | - H(ind,1) + i H(ind,5) + mu S(ind),              - H(ind,3) - i H(ind,4) |
   !       = |                                                                           |
   !         | - H(ind,7) + i H(ind,8)            ,  - H(ind,2) + i H(ind,6) + mu S(ind) |
   !
   !
[...]

PR #877, opened 2024-11-27
initial commit of nambu code

### LINKED ISSUE #676: Nambu spin box
**Describe the feature**
Enable nambu spin components. 16 elements.

This would also require that density matrices etc uses a generic method for converting matrix elements to spin directions etc. This is already duplicated in many places.
- add nambu
- add method to split matrix elements into total, xyz, or spherical elements.

zerothi:
Added matrix codes to create the Nambu
Hamiltonian.
This required restructuring the diag
fold_csr matrix codes. Now they can be given
a number to determine the number of elements
that are added. It generalizes the code a bit.

Currently the matrix creation does not implement
the no-phase code-path. So Gamma-point is not functional.

Changed some of the cython codes to use preprocessors at the comment level at the top of the file.
It makes it much simpler to debug.

Enabled siesta routines to read the matrices
with Nambu spin configuration.

 - [x] Closes #676 and #873

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ we hit release version 1.0.0.
       import sisl
       sisl.geom.graphene
 
+- added Nambu spin configuration, this is still experimental
+
 ### Fixed
 - `projection` arguments of several functions has been streamlined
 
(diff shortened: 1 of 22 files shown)


zerothi:
I am a little inclined to merge this, it holds many other details that are necessary. For now I'll add a warning when using Nambu spin-configuration, and then users needs to be aware, but I would still like you, the experienced users, to bug-track this!

PR #888, opened 2025-01-24
Fixes for native hamiltonian format

### LINKED ISSUE #887: Problems with writing native `.ham` files
**Describe the bug**

1) When exporting a hamiltonian to the native `.ham` file format, some intercell matrices may be missing if one uses the default `hermitian=True`.

2) For orthogonal hamiltonians, the `.ham` file includes incorrect intercell overlaps (should be zero).

**Code to reproduce problem**
```python
import sisl

g = sisl.geom.graphene()
H = sisl.Hamiltonian(g)

# build a 3rd nearest neighbor model
H.construct([[0.1, 1.6, 2.6, 3.1], [0, -2.7, -0.20, -0.18]])

H.write("test1.ham")
H.write("test2.ham", hermitian=False)
```
Here, the file `test1.ham` only contains 3 intercell matrices,
```
begin matrix 1 0 0
 0 0 -0.2 1
 1 0 -2.7
 1 1 -0.2 1
end matrix 1 0 0

begin matrix 0 1 0
 0 0 -0.2 1
 1 0 -2.7
 1 1 -0.2 1
end matrix 0 1 0

begin matrix 1 1 0
 1 0 -0.18
end matrix 1 1 0
```
while `test2.ham` contains four (ignoring here for simplicity the additional four that are related by transpose):
```
begin matrix 1 0 0
 0 0 -0.2 1
 1 0 -2.7
 1 1 -0.2 1
end matrix 1 0 0

begin matrix -1 1 0
 0 0 -0.2 1
 0 1 -0.18
 1 1 -0.2 1
 1 0 -0.18
end matrix -1 1 0

begin matrix 0 1 0
 0 0 -0.2 1
 1 0 -2.7
 1 1 -0.2 1
end matrix 0 1 0

begin matrix 1 1 0
 1 0 -0.18
end matrix 1 1 0
```

**Summary**

In short,

1) The `matrix -1 1 0` is missing if one uses the default `hermitian=True`.

2) Intercell matrices contain (incorrectly) overlap values of 1

tfrederiksen:
Closes #887

This PR produces the following `test1.ham` that looks complete and correct:
```
begin cell
  2.13000000 -1.22975607 0.00000000
  2.13000000 1.22975607 0.00000000
  0.00000000 0.00000000 20.00000000
end cell

supercell 3 3 1

begin atoms
  6 0.00000000 0.00000000 0.00000000
  6 1.42000000 0.00000000 0.00000000
end atoms

begin matrix 0 0 0
 0 1 -2.7
end matrix 0 0 0

begin matrix 1 0 0
 0 0 -0.2
 1 0 -2.7
 1 1 -0.2
end matrix 1 0 0

begin matrix -1 1 0
 0 0 -0.2
 0 1 -0.18
 1 1 -0.2
 1 0 -0.18
end matrix -1 1 0

begin matrix 0 1 0
 0 0 -0.2
 1 0 -2.7
 1 1 -0.2
end matrix 0 1 0

begin matrix 1 1 0
 1 0 -0.18
end matrix 1 1 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ we hit release version 1.0.0.
   The old values are still respected.
 
 ### Fixed
+- `hamiltonianSile` can now handle skewed lattices with 6 input parameters
+  for the cell block
+- `hamiltonianSile` wrote wrong overlap and supercell connections for
+  more than 1 supercell direction, #887
 - `projection` arguments of several functions has been streamlined
 - `orbitals` arguments with slices without ends returned up to `geometry.na`,
   now it correctly returns up to the maximum number of orbitals.
--- a/src/sisl/io/ham.py
+++ b/src/sisl/io/ham.py
@@ -344,9 +344,9 @@ class hamiltonianSile(Sile):
 
             if hermitian and is_primary:
                 # only write upper right of primary unit-cell connections
-                Hsub = triu(Hsub)
+                Hsub = triu(Hsub, format="csr")
                 if not is_orthogonal:
-                    Ssub = triu(Ssub)
+                    Ssub = triu(Ssub, format="csr")
 
             # Ensure that when h[i,i] == 0, we still write something
             # Since S != 0.
--- a/src/sisl/io/ham.py
+++ b/src/sisl/io/ham.py
@@ -318,7 +318,7 @@ class hamiltonianSile(Sile):
                     o = o - geom.a2o(a)
                     if not H.orthogonal:
                         s = Ssub[jo, io]
-                    elif jo == io:
+                    elif jo == io and i == 0:
                         s = 1.0
                     else:
                         s = 0.0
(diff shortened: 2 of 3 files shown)


tfrederiksen:
One should perhaps take this opportunity to also include some tests....

zerothi:
Thanks for this, I can see why this fixes the issue, I will have to amend this to ensure it works for all `sc_off` cases (in fact, users can specify their own ordering of supercells, so I'll have to check that (shouldn't be difficult!

tfrederiksen:
> Thanks for this, I can see why this fixes the issue, I will have to amend this to ensure it works for all `sc_off` cases (in fact, users can specify their own ordering of supercells, so I'll have to check that (shouldn't be difficult!

Yes, indeed this algorithm assumes a specific ordering.

zerothi:
@tfrederiksen I have fixed lots of small problems. With added tests.
Please have a look, once the tests pass, I'll merge!

Thanks for the fixes and report!

zerothi:
Thanks!

PR #889, opened 2025-01-27
Test with nonorthogonal hamiltonian

tfrederiksen:
Following up on #888, what about also having a check for a nonorthogonal model and more intercell connections?


zerothi:
Thanks!

PR #907, opened 2025-04-02
Git ignoring debug_info.py

pfebrer:
The `_debug_info.py` file (generated on installation) was messing with my git control :smile:


zerothi:
Indeed! Thanks!

Issue #735, opened 2024-03-27
Change CLI to click


zerothi:
**Describe the feature**

@pfebrer did #608 in response to #606.

I have been looking around on `typer` and trying to get a grasp of where the project is heading.

I must admit, that I am worried about its future commitment. The last many months of commits are (almost) purely maintenance.
It does seem that it might catch on, but...

While `click` is a bit verbose, it is probably much less than what we have right now with argparse.

The main advantage of using `click` is that it is well maintained, and has a clear goal.
`typer` is really lacking behind, but might be more susceptible to changes along the way, but due to its infancies its restriction on the `Type` system means that one might need to overwrite many types in order to get behaviour that looks coherent for the users.
I am also worried about the complexity one has to go about to show *nice* types for the CLI. I.e. `Union` types (which is not supported).

Typer has a nice `rich` usage, but `rich-click` seems like a drop-in replacement for viewing things in `rich`.

I would also argue that once `click` is in use, having smaller portions of the code use `typer` is not that much of a deal since `click` is the underlying machinery. I guess that they can be combined and used together?

@pfebrer I am sorry I have pushed you into the other details, I think it was worth it to get an idea of how it looked, but the above worries me.

pfebrer:
Hmm ok, but could we have some kind of wrapper where we get the type annotations and help messages from the function definition so that we don't have to redefine all the arguments? That is very annoying to me, and that's why I like typer.

We could have the automatic detection, but allow normal click customization on top.

PR #908, opened 2025-04-02
Formatting the argparse CLIs

pfebrer:
This PR adds the possibility to use formatting from `rich-argparse` if the package is installed.

It formats all of the CLIs in sisl. 

I think this is a very good compromise that could unblock some nice features (https://github.com/zerothi/sisl/pull/801, basis optimization CLI) that haven't gone in due to discrepancies on what the CLI should look like :) (see https://github.com/zerothi/sisl/pull/608,  https://github.com/zerothi/sisl/issues/735).

The formatting could be gradually made nicer and nicer by extending the formatter, and nothing about the implementation of the CLIs needs to change.

I don't know whether `rich-argparse` should be a dependency, optional dependency or what. But if it is not a dependency it should be advertised, because it makes things look nicer :sparkles:

### DIFF
--- a/src/sisl/_core/geometry.py
+++ b/src/sisl/_core/geometry.py
@@ -45,6 +45,7 @@ from sisl._indices import (
     list_index_le,
 )
 from sisl._internal import set_module
+from sisl._lib._argparse import SislHelpFormatter
 from sisl._math_small import cross3, is_ascending, xyz_to_spherical_cos_phi
 from sisl._namedindex import NamedIndex
 from sisl.messages import SislError, deprecate_argument, deprecation, info, warn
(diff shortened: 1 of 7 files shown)


zerothi:
Thanks! Looks good! I think this is a nice step forward!

pfebrer:
Thanks! Have you seen that tests are failing because of debug_info? 😅

zerothi:
> Thanks! Have you seen that tests are failing because of debug_info? 😅

yeah, but why? I don't see how your gitignore thing should trigger this... Ideas?

pfebrer:
No :(

zerothi:
Damn, I am bit at a loss, can't replicate this locally... :(

zerothi:
Fixed it, got no clue why this turned up just now... I couldn't find information on whether scikit-build-core should omit gitignore files (or setuptools for that matter). So I am a bit puzzled. But, now its inclusion is hardcoded in the `pyproject.toml` file, as it apparently should have been done! :)

pfebrer:
Nice!

PR #910, opened 2025-04-14
Basis optimization CLI

pfebrer:
Here is the PR with the basis optimization CLI.

Probably I should add something to the documentation. 

@zerothi could you test if it works fine for you? You just need to have the inputs for a calculation in a directory and then run:

```
stoolbox basis optim --geometry file.fdf
```

The only constraint is that the fdf file can't be named `RUN.fdf`.

### DIFF
--- a/src/sisl_toolbox/siesta/minimizer/_minimize.py
+++ b/src/sisl_toolbox/siesta/minimizer/_minimize.py
@@ -9,6 +9,7 @@ from abc import abstractmethod
 from hashlib import sha256
 from numbers import Real
 from pathlib import Path
+from typing import Callable
 
 import numpy as np
 from scipy.optimize import dual_annealing, minimize
@@ -22,7 +23,9 @@ __all__ = [
     "BaseMinimize",
     "LocalMinimize",
     "DualAnnealingMinimize",
+    "BADSMinimize",
     "MinimizeToDispatcher",
+    "ParticleSwarmsMinimize",
 ]
 
 _log = logging.getLogger(__name__)
@@ -54,12 +57,15 @@ class BaseMinimize:
     # Basic minimizer basically used for figuring out whether
     # to use a local or global minimization strategy
 
-    def __init__(self, variables=(), out="minimize.dat", norm="identity"):
+    def __init__(
+        self, variables=(), out="minimize.dat", norm="identity", out_fmt="20.17e"
+    ):
         # ensure we have an ordered dict, for one reason or the other
         self.variables = []
         if variables is not None:
             for v in variables:
                 self.add_variable(v)
+        self.out_fmt = out_fmt
         self.reset(out, norm)
 
     def reset(self, out=None, norm=None):
@@ -221,7 +227,9 @@ class BaseMinimize:
             comment += f" The first {len(self.data)} lines contains prior content."
             data = np.column_stack((self.data.x, self.data.y))
             self._fh = tableSile(self.out, "w").__enter__()
-            self._fh.write_data(data.T, comment=comment, header=header, fmt="20.17e")
+            self._fh.write_data(
+                data.T, comment=comment, header=header, fmt=self.out_fmt
+            )
             self._fh.flush()
 
         return self
@@ -288,15 +296,18 @@ class BaseMinimize:
             pass
 
         # Else we have to call minimize
-        metric = np.array(self(variables, *args))
-        # add the data to the output file and hash it
-        self._fh.write_data(
-            variables.reshape(-1, 1), metric.reshape(-1, 1), fmt="20.17e"
-        )
-        self._fh.flush()
-        self.data.x.append(variables)
-        self.data.y.append(metric)
-        self.data.hash.append(current_hash)
+        metric = self(variables, *args)
+
+        if metric is not None:
+            metric = np.array(metric)
+            # add the data to the output file and hash it
+            self._fh.write_data(
+                variables.reshape(-1, 1), metric.reshape(-1, 1), fmt=self.out_fmt
+            )
+            self._fh.flush()
+            self.data.x.append(variables)
+            self.data.y.append(metric)
+            self.data.hash.append(current_hash)
 
         return metric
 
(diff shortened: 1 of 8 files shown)


src/sisl_toolbox/siesta/minimizer/_minimize.py
```diff
@@ -54,12 +57,15 @@ class BaseMinimize:
     # Basic minimizer basically used for figuring out whether
     # to use a local or global minimization strategy
 
-    def __init__(self, variables=(), out="minimize.dat", norm="identity"):
+    def __init__(
+        self, variables=(), out="minimize.dat", norm="identity", out_fmt="20.17e"
```
zerothi, line 61: should we rename this to `fmt`, isn't it clear enough? This is what we do for other siles?

src/sisl_toolbox/siesta/minimizer/_minimize.py
```diff
@@ -288,15 +296,18 @@ def _minimize_func(self, norm_variables, *args):
...
-        self._fh.write_data(
-            variables.reshape(-1, 1), metric.reshape(-1, 1), fmt="20.17e"
-        )
-        self._fh.flush()
-        self.data.x.append(variables)
-        self.data.y.append(metric)
-        self.data.hash.append(current_hash)
+        metric = self(variables, *args)
+
+        if metric is not None:
```
zerothi, line 301: when would `metric` be none? Would it even be viable?

Shouldn't something break if the metric is none?

src/sisl_toolbox/siesta/minimizer/_minimize.py
```diff
@@ -332,6 +343,77 @@ def run(self, *args, **kwargs):
...
+    It uses the pybads package to perform the minimization.
+    """
+
+    def run(self, options={}, get_hard_bounds: Callable = lambda x: x, **kwargs):
+        from pybads.bads import BADS
+
+        bounds = self.normalize_bounds()
+        bounds = np.array(bounds)
+
+        bounds[bounds == 0] = 0.5
```
zerothi, line 359: Why is this?

What happens when one has a bounds `[0, 0.2]` then it would be changed to `[0.5, 0.2]`?

src/sisl_toolbox/siesta/minimizer/_minimize.py
```diff
@@ -332,6 +343,77 @@ def run(self, *args, **kwargs):
...
 
 
+@set_module("sisl_toolbox.siesta.minimizer")
+class BADSMinimize(BaseMinimize):
+    """Bayesian Adaptive Direct Search (BADS) minimizer.
+
+    It uses the pybads package to perform the minimization.
+    """
+
+    def run(self, options={}, get_hard_bounds: Callable = lambda x: x, **kwargs):
```
zerothi, line 353: I would rename `get_hard_bounds` to `func_conv_bounds_to_hard` or something more descriptive.

src/sisl_toolbox/siesta/minimizer/_minimize.py
```diff
@@ -332,6 +343,77 @@
...
+                plausible_upper_bounds=bounds[:, 1],
+                options=options,
+            )
+
+            # They do a deep copy of the arguments at the end, when generating the result.
+            # This might be a problem if the arguments are not pickleable. E.g. the minimizer function
+            # is associated with the minimizer object, which might contain siles.
+            try:
+                opt = bads.optimize()
+            except TypeError:
```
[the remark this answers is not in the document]
zerothi: is there a reason for `try ... pass`? Don't we want it to scream at the user if it fails?

src/sisl_toolbox/siesta/minimizer/basis_optimization.py
```diff
@@ -0,0 +1,1031 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
zerothi, line 1: how much of this file is duplicating the yaml conf file? Seems to have some overlap?

zerothi (changes requested):
I'll play with it at its current state, but some comments here and there would be beneficial to be cleared out!

zerothi:
Seems like we should move the basis optim writer to the `_yaml_reader` and rename it.

src/sisl_toolbox/siesta/minimizer/_minimize.py
pfebrer: Yes, I guess, but who has 0.2 as the maximum cutoff for the basis? I agree that this is not general, And I don't remember what happened if one of the bounds was 0, maybe something nasty :sweat_smile:

src/sisl_toolbox/siesta/minimizer/_minimize.py
pfebrer: I think None meant that it has not been able to compute the metric, e.g. the calculation has failed. It shouldn't break because the optimizer can set variable values that result in a failed calculation.

src/sisl_toolbox/siesta/minimizer/_minimize.py
pfebrer: Whatever you prefer

src/sisl_toolbox/siesta/minimizer/_minimize.py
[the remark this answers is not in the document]
pfebrer: This is explained in the comment above, basically their final saving or whatever (I don't remember exactly what it was) uses pickle and it can fail, but I guess we didn't need it because we do our own saving of the optimization.

src/sisl_toolbox/siesta/minimizer/_minimize.py
zerothi: but if you want 0.5 to be the minimum, then we should have `bounds[bounds < 0.5] = 0.5`, right?

Otherwise we should select a *sensible* minimum, and let users control everything.

pfebrer:
Hey, I have answered them, sorry I think I was too busy when I saw the review

src/sisl_toolbox/siesta/minimizer/_minimize.py
zerothi: hmm, I see. :)

src/sisl_toolbox/siesta/minimizer/_minimize.py
pfebrer: Yes clearly something should be done if there is someone who wants to use this for something other than basis optimization.

src/sisl_toolbox/siesta/minimizer/_minimize.py
pfebrer: I recall having tested with all the minimizers and they all handled a None, but I can't say 100%

src/sisl_toolbox/siesta/minimizer/_register_cli.py
```diff
@@ -29,10 +29,11 @@ def basis_cli(subp=None, parser_kwargs={}):
     subp = p.add_subparsers(title="Commands")
 
     get_argparse_parser(
-        optimize_basis, name="optim", subp=subp, parser_kwargs=parser_kwargs
+        write_basis_to_yaml, name="build", subp=subp, parser_kwargs=parser_kwargs
     )
+    parser_kwargs["aliases"] = ("optim",)
     get_argparse_parser(
-        write_basis_to_yaml, name="build", subp=subp, parser_kwargs=parser_kwargs
+        optimize_basis, name="optimize", subp=subp, parser_kwargs=parser_kwargs
```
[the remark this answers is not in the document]
zerothi: both are valid, but yes, I should have done that (will do!)

src/sisl_toolbox/siesta/minimizer/_minimize.py
zerothi: Just read through my old yaml conf, we have to remove this, because there are soft-confinement things that uses negative bounds, so you would have to know which kind of boundary it attaches to...

This check was only done for the bads, but in principle, it should be there for all, so I'll remove it...

zerothi:
I have some questions, the charge-confinement scheme is not only targeted polarization orbitals. AFAIK, it's use is for non-populated orbitals, and so could be either regular *empty* orbitals or something else.

I will also add the soft-confinement to play with. Do you have any comments on this before I make changes?

src/sisl_toolbox/siesta/minimizer/basis_optimization.py
```diff
@@ -0,0 +1,1031 @@
...
+
+    - ...
+
+    - Then it optimizes the cutoff radii for the last zeta shell.
+
+    - Finally, it optimizes the polarization orbitals (if requested).\n\n
+
+    Appropiate optimization bounds are set for each parameter.\n\n
+
+    At each step, all corresponding parameters are optimized simultaneously.\n\n
```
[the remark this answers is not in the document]
pfebrer: The simultaneously means that for example when optimizing the first zeta shell, all cutoff radii (for all atoms) are optimized at the same time. And same for the polarization orbitals

src/sisl_toolbox/siesta/minimizer/basis_optimization.py
```diff
@@ -257,30 +264,48 @@ def generate_basis_config(
...
             return 2
         else:
             return 2.5
 
     def _orbs_from_row_and_block(table_row, atom_block):
         """To be used if the orbitals to optimize can't be inferred from the pseudopotential."""
+        warn(
+            "The orbitals are pre-filled, but likely they are not correct. "
+            "Please check the basis before doing excessive optimizations!"
+        )
```
pfebrer, line 308: Why likely they are not correct? :sweat_smile:

src/sisl_toolbox/siesta/minimizer/basis_optimization.py
```diff
@@ -220,18 +225,20 @@ def generate_basis_config(
...
+
+        For multiple arguments, each can be fine-tuned on/off.
+
+        This will only optimize charge-confinement for *empty* orbitals (:math:`q=0`).
+    optimize_soft_conf : bool, tuple of bool, optional
+        Whether all orbitals will be optimized for soft-confinement.
+        If true, it will optimize, the potential (:math:`V_0`), and the inner
+        radius :math:`r_i`.
+
+        For multiple arguments, each can be fine-tuned on/off.
```
pfebrer, line 241: Does this work from the CLI? If so, how?

Doesn't it make the CLI harder to understand? 
`--optimize-soft-conf false` will optimize soft confinement.

For the charge confinement, are you sure optimizing the three parameters at the same time will result in a good optimization/ is worth it? I also don't understand to which orbitals it applies, Federico said that the charge optimization should only be done for polarization orbitals. So again make sure to talk to Federico because I am not sure that this added flexibility is good.

PR #914, opened 2025-04-27
Added support for I/O from zipfiles and buffers

pfebrer:
#### Motivation

When dealing with very big datasets (e.g. ML) you often need to compress it into a zip or tar file to move it around. Needing to then decompress it for reading the files is a pain, because:

1. Decompressing the full dataset might take a very long time
2. File number limitations in HPC clusters. (also recently I've used a cluster that removes unused files every 30 days, which can basically destroy the parts of the dataset that you have not used).

#### Why zip and not tar?

Because tar doesn't keep an index of the contained files, so it can be very slow to access files within it when the tarfile gets big.

Therefore I have determined that putting effort into zip files is much more worth it.

#### Approach taken in this PR

Turns out that the `zipfile` library already had a `zipfile.Path` object that mimics `pathlib.Path`. Therefore, `zipfile.Path` can be used within the sile framework. It just needed some extra functionality to fully work, so I created an extension `ZipPath`, which:

- Implements missing `pathlib.Path` functions that the sile framework was using, e.g. `with_suffix`.
- Wraps the `open` method of the returned file handles so that on close we can close also the root zip file. This is specially important when writing.

**What works at the moment:**
- Reading and writing text based files.
- Reading cdf files. This is done through the `memory` argument of `netCDF4.Dataset`, which accepts the in memory bytes.
- Reading binary files. This is done by writing on a temporary file. I need some clever way of knowing when the temporary file can be removed though.
- Searching for other files in the directory. E.g. when reading from an fdf inside a zipfile, it can find other files (other fdfs, basis files, matrices) that are also inside the zipfile.

**What doesn't work for now:**
- Writing CDF files. This can also be done through the `memory` argument.
- Writing binary files. This can also be done through temporary files.

#### Comments

I will implement the missing parts and also add tests when I get feedback on whether this approach is fine or it is preferrable to create separate sile classes as with `BufferSile`.

The handling of CDF and binary files I think it should be generalized to handle generic python filehandles/objects (e.g. `bytes`/`BytesIO`) and not just contents of a zip file.

#### How to test

Here is a zip file that you can use for testing, which contains a SIESTA run of a water molecule:

[run.zip](https://github.com/user-attachments/files/19929287/run.zip)

You can read the Hamiltonian like:

```python
import sisl

sisl.get_sile("/path/to/run.zip/run/RUN.fdf").read_hamiltonian()
```

The sile framework will automatically detect that the path contains a zip file and do the necessary adjustments.

**Hope you like it!**

### DIFF
--- a/src/sisl/io/_zipfile.py
+++ b/src/sisl/io/_zipfile.py
@@ -2,8 +2,11 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 import functools
+import sys
 import zipfile
 from pathlib import Path
+
+_is_python_old = sys.version_info < (3, 10)
 
 
 class ZipPath(zipfile.Path):
@@ -50,6 +53,12 @@ class ZipPath(zipfile.Path):
 
     def __init__(self, *args, close_zipfile: bool = False, **kwargs):
         """Initialize the ZipPath with a zipfile object and a path"""
+        if _is_python_old:
+            raise RuntimeError(
+                "Zip file functionality in sisl requires Python 3.10 or newer. "
+                "Upgrade your Python version if you want to use it."
+            )
+
         super().__init__(*args, **kwargs)
 
         self.close_zipfile = close_zipfile
--- a/src/sisl/io/_zipfile.py
+++ b/src/sisl/io/_zipfile.py
@@ -1,0 +1,131 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at https://mozilla.org/MPL/2.0/.
+import functools
+import zipfile
+from pathlib import Path
+
+
+class ZipPath(zipfile.Path):
+    """Extension of the zipfile.Path class to mimic the pathlib.Path class.
+
+    The main goal is this extension of zipfile.Path can be used within the
+    sile framework as a drop in replacement for pathlib.Path. By enabling
+    that, all the reading/writing functionality also works with zipfiles.
+
+    The extension has two purposes:
+
+    1. Some methods were missing in the original zipfile.Path class if we
+    wanted to use it as a Path object:
+
+        - ``with_suffix``
+
+    2. When writing a file in a zipfile, it is important to not only close the
+    file handle for the written file, but also to close the zipfile itself, so
+    that it writes the changes to disk.
+
+    To this end, when `open` is called, we wrap the `close` method of the returned
+    file handle so that we can close the zipfile if necessary.
+
+    Parameters
+    ----------
+    *args :
+        Arguments passed to the zipfile.Path constructor.
+    close_zipfile :
+        Whether to close the root zipfile when closing an open file handle
+        produced by this path.
+        Closing the zipfile is necessary for the changes to be written to disk.
+        However, if the zipfile is passed by the user, we do not want to close
+        the zipfile, because we don't know if the user wants to continue writing
+        to it.
+        Therefore, the default is ``False``, but if the ``ZipPath`` is created
+        from a normal path (i.e. the user didn't input the zipfile, it was created
+        on the fly), we set it to ``True`` (see ``ZipPath.from_path``).
+    **kwargs :
+        Keyword arguments passed to the zipfile.Path constructor.
+    """
+
+    #: Whether to close the root zipfile when closing the file handles.
+    close_zipfile: bool = False
+
+    def __init__(self, *args, close_zipfile: bool = False, **kwargs):
+        """Initialize the ZipPath with a zipfile object and a path"""
+        super().__init__(*args, **kwargs)
+
+        self.close_zipfile = close_zipfile
+
+    @property
+    def suffix(self):
+        """Override the suffix property to ensure it returns a string"""
+        return Path(str(self)).suffix
+
+    def with_suffix(self, *args, **kwargs):
+        """Override the with_suffix method to ensure it returns a ZipPath"""
+
+        file_path = Path(str(self)).relative_to(self.root.filename)
+
+        return self.__class__(self.root, str(file_path.with_suffix(*args, **kwargs)))
+
+    def open(self, *args, **kwargs):
+        """Override the open method to patch the file handle if necessary"""
+
+        fh = super().open(*args, **kwargs)
+
+        if self.close_zipfile:
+            # The root zipfile needs to be closed when closing the file handle,
+            # so we wrap the close method of the file handle
+            close = fh.close
+
+            @functools.wraps(close)
+            def _close(*args, **kwargs):
+                close(*args, **kwargs)
+                self.root.close()
+
+            fh.close = _close
+
+        return fh
+
+    @classmethod
+    def from_zipfile_path(cls, zipfile_path: zipfile.Path, close_zipfile: bool = False):
+        """Create a ZipPath from a zipfile.Path object"""
+
+        file_path = Path(str(zipfile_path)).relative_to(zipfile_path.root.filename)
+
+        return cls(zipfile_path.root, str(file_path), close_zipfile=close_zipfile)
+
+    @classmethod
+    def from_path(cls, path: Path, mode: str = "r", close_zipfile: bool = True):
+        """Create a ZipPath from a Path object.
+
+        Given a path object, scans the path to find the first zip file in the path.
+        If a zip file is found, a ZipPath object is created.
+
+        This function initializes a new ``zipfile.ZipFile`` object to use as the root.
+
+        Parameters
+        ----------
+        path :
+            The path to the file or directory.
+        mode :
+            The mode in which you will want to open paths. This
+            determines the mode in which the zipfile is opened.
+        close_zipfile :
+            Whether to close the zipfile when closing the file handle.
+            This is important because if you write to a zipfile, you need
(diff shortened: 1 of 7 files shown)


zerothi:
Gould you rebase on main, to remove basis cli?

Looks very interesting!

pfebrer:
The basis commits are already removed :+1:

src/sisl/io/_zipfile.py
```diff
@@ -0,0 +1,94 @@
...
+
+        return cls(zipfile_path.root, str(file_path))
+
+    @classmethod
+    def from_path(cls, path: Path, mode: str = "r"):
+        """Create a ZipPath from a Path object.
+
+        Given a path object, scans
+        """
+        for i, part in enumerate(path.parts[:-1]):
```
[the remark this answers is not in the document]
zerothi: It seems the CI is correct here, the `else` is wrongly formatted, will never occur, no break is present.

src/sisl/io/sile.py
```diff
@@ -497,6 +500,33 @@ class BaseSile:
...
+            filename = Path(filename)
+
+            # Try to convert to a ZipPath, which will only succeed if there
+            # is a .zip file in the middle of the path
+            try:
+                filename = ZipPath.from_path(filename, self._mode)
+            except FileNotFoundError:
+                pass
+
+        return filename
```
zerothi, line 528: it should convert to `Path` here, to follow the signature.

src/sisl/io/sile.py
```diff
@@ -1574,13 +1614,45 @@ class SileBin(BaseSile):
...
 
         # Must call setup-methods
         self._base_setup(*args, **kwargs)
 
+    def __getattribute__(self, name):
+        # If the file is inside a zip fortran can't read directly from it.
+        # We therefore store the contents in a temporary file and set this
+        # as the file to be read from.
+        if (
+            name.startswith("read_")
```
zerothi, line 1634: what about write? Should that even be allowed?

zerothi (commented):
I'll likely have some more comments, but it looks very clean, simple and effective. Major Q is whether `write` works, otherwise an error should be raised.

pfebrer:
Yes, I think there will be no problem in making write work :+1:

src/sisl/io/sile.py
pfebrer: It is already converted to `Path` at the beggining of the `elif` and everything should go through that `elif` except the case when filename is a `ZipPath`

src/sisl/io/sile.py
pfebrer: Yes, I think so. It is a bit more scary because if we don't manage the write correctly it can end up in a corrupted zip file, that's why I didn't want to implement it fast, I need to put a bit of thought into it.

src/sisl/io/sile.py
zerothi: true, missed that!

src/sisl/io/sile.py
zerothi: Otherwise, doing an `NotImplementedError` is sufficiently at the moment, it doesn't seem like it is something that's very useful ATM.

src/sisl/io/_zipfile.py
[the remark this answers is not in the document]
pfebrer: Hmm yeah ok I guess it is not 100% correct structuring. The `else` will happen though, because it happens if the end of the loop is reached.

There is no break but I exit the loop with a `return`. It is true that `else` is not needed then, we can just have the raise after the loop.

In other words, the code works but it is not the best way to do it :sweat_smile:

src/sisl/io/sile.py
pfebrer: Yes but on the other hand if I do it now it will be faster than if I come back at it 2 years later :sweat_smile:

pfebrer:
By the way, what I've done for CDF and binary files can work really with any buffer, its not like it just works with zipfiles. So I think it would be worth it now that we are at it to make it support arbitrary buffers, which is something that has come up in the past.

zerothi:
> By the way, what I've done for CDF and binary files can work really with any buffer, its not like it just works with zipfiles. So I think it would be worth it now that we are at it to make it support arbitrary buffers, which is something that has come up in the past.
> 
> Any ideas of how things could be structured to support the buffers? Just extend the `BufferSile` to handle more buffer types, create a new class...?

If the `BufferSile` could be extended as a baseclass for binary + text files, just like the `sile`, then I think that would be the most optimal way of doing things.

So:
```mermaid
classDiagram
    BaseSile <|-- Sile
    BaseSile <|-- SileCDF 
    BaseBufferSile
    BaseBufferSile <|-- BufferSile
    BaseBufferSile <|-- BufferCDF 
```

pfebrer:
Ok, will try to do that :+1:

By the way, where is the documentation for generating this diagrams? hahah

zerothi:
> Ok, will try to do that 👍
Nice!

> 
> By the way, where is the documentation for generating this diagrams? hahah

They are [mermaid charts](https://www.mermaidchart.com/) ;)

zerothi:
I think this is great!

Only one comment.

I am not sure I like the `get_sile("hello/test.zip/...")` should automatically create the `test.zip`. For reading it is fine, because you can detect whether it is a zip file, but for creation, I am a bit worried that it might be ambiguous, should it be a directory, or a zip file?
I think for creating files we shouldn't allow that nomenclature... Could one extend the `ZipPath` to have a *base* Zip-path so you could do something like:

```python
from sisl.io import ZipPath

zipfile = ZipPath("/hello/test.zip") / "dir_in_zip_file"
si.get_sile(zipfile, mode="w")
```
With the above, there should be no ambiguity.
I

pfebrer:
Hmm ok I understand but for writing the zip file will often also exist already, you just want to add files to It. So what we could do maybe is that `sisl` will never create a new zip file.

If you want to open the zip file externally I think the best thing to do is to use the `zipfile` classes. I think `ZipPath` should remain internal to sisl. This is what I show in the last example of the changes rst.

zerothi:
> Hmm ok I understand but, for writing, the zip file will often also exist already, you just want to add files to It. So what we could do maybe is that `sisl` will never create a new zip file.
If it exists, there will be no ambiguity, unless you specify `mode='w'`, in which case you want to overwrite it...

> If you want to open the zip file externally I think the best thing to do is to use the `zipfile` classes. I think `ZipPath` should remain internal to sisl. This is what I show in the last example of the changes rst.
Ok!

pfebrer:
Yes, when the zipfile is opened inside sisl it is always with mode "r" or "a", never "w".

zerothi:
> Yes, when the zipfile is opened inside sisl it is always with mode "r" or "a", never "w".

what happens when the user requests `mode='w'`?

src/sisl/io/_zipfile.py
```diff
@@ -0,0 +1,131 @@
...
+            if part.endswith(".zip"):
+                zip_path = Path(*path.parts[: i + 1])
+                if zip_path.is_file():
+                    zipfile_mode = "r" if mode.startswith("r") else "a"
+                    root_zip = zipfile.ZipFile(zip_path, zipfile_mode)
+                    return cls(
+                        root_zip,
+                        str(path.relative_to(zip_path)),
+                        close_zipfile=close_zipfile,
+                    )
```
pfebrer, line 129: @zerothi we are already checking that the zip file exists before opening it, I didn't remember :sweat_smile:

zerothi:
> The `mode="w"` that the user requests is for the file, not for the container zip file.
> 
> It is the same as if we wanted to write to a folder, we want to add things to the folder, never overwrite the full folder to write the file.

Hmm... Yeah, I see... :)

pfebrer:
Ok, so this is now super ready, I'm very happy with the final result, everything works:

**Read/write from/to buffers and zipfiles for ALL siles.** :partying_face: 

Plus all write/read operations for buffers and zipfiles are tested (both text, binary and CDF). So things are quite robust and I'm confident it should work fine.

I just have to figure out why the tests with temporary files fail in the CI (only with ubuntu). Could you try the tests in your computer and see if everything is fine?

I also refactored the I/O documentation to include information on how to do I/O with zipfiles and buffers, and added a link to it in the introduction section, because it seemed to me that it was not visible enough and it is an important part of the package :sweat_smile:

pfebrer:
So it is quite a mess to support zipfiles both in python 3.9 and python>3.9 because there was one key change in between: https://github.com/python/cpython/issues/84744#issuecomment-1093871456.

This means that we have to have many extra checks and code for a python version that we are going to drop in October :( The easiest fix also involves requiring an extra dependency `zipp`.

So I suggest it is not worth it to support zipfiles in python 3.9, we can just raise a meaningful error to update to python >= 3.10 if the user wants to use zipfiles. Is this ok?

zerothi:
> So it is quite a mess to support zipfiles both in python 3.9 and python>3.9 because there was one key change in between: [python/cpython#84744 (comment)](https://github.com/python/cpython/issues/84744#issuecomment-1093871456).
> 
> This means that we have to have many extra checks and code for a python version that we are going to drop in October :( The easiest fix also involves requiring an extra dependency `zipp`.
> 
> So I suggest it is not worth it to support zipfiles in python 3.9, we can just raise a meaningful error to update to python >= 3.10 if the user wants to use zipfiles. Is this ok?

Yes, that is fine. :)
Thanks, sorry, this week is busy up to the workshop, so I'll likely first have time to get things in after the workshop...

zerothi (approved):

PR #916, opened 2025-05-07
Some additions to the development documentation

ialcon:
- [x] Documentation for functionality in `docs/dev`

So, just added some additional information in the `docs/dev` to highlight some additional packages that need to be installed and, more substantially, including a new section explaining how one may contribute to the `docs`. Something that I have not done is to touch the `git clone https://github.com/<your-username>/sisl.git` bit we discussed. As I said, I would either remove that part or, preferably, add also the corresponding line for the SSH protocol (`git clone git@github.com:<your-username>/sisl.git`). Leaving only one of the two, in my opinion, may create some confusion for beginners. But anyways, please go on with the option you find more appropriate.

Finally, I also think it might be worth giving a bit more details in all this `changes/README.rst`, but maybe for that is sufficient to modify this file - so again, your call.

Finally, feel free to remove or modify anything that I added, it might well be that I was over-explaining things.

### DIFF
--- a/docs/dev/index.rst
+++ b/docs/dev/index.rst
@@ -32,8 +32,23 @@ Summary of development process
 
 Here is a short summary of how to do developments with `sisl`.
 
+1. You may need to install some packages system-wide, in case you do not have these:
 
-1. If you are a first time contributor, you need to clone your forked repository
+   *  For any type of contribution, ``gfortran`` :
+
+      .. code:: bash
+
+         sudo apt-get install gfortran
+
+   *  And, in case you want to specifically contribute to the docs, `pandoc` (version between
+      2.9.2 and 4.0.0). For example:
+
+      .. code:: bash
+
+         wget https://github.com/jgm/pandoc/releases/download/3.2/pandoc-3.2-1-amd64.deb
+         sudo dpkg -i pandoc-3.2-1-amd64.deb
+
+2. If you are a first time contributor, you need to clone your forked repository
    and setup a few things.
 
    The procedure enables one to follow the upstream changes, while simultaneously
@@ -64,7 +79,7 @@ Here is a short summary of how to do developments with `sisl`.
      This will run specific checks before you commit things to the repository.
      It ensures consistency in the project.
 
-2. Installing the project in development mode.
+3. Installing the project in development mode.
 
    It is advised to install the project in *editable* mode for faster
    turn-around times.
@@ -76,7 +91,7 @@ Here is a short summary of how to do developments with `sisl`.
    For further details, see
    :ref:`the editable|pip instructions <installation-pip>`.
 
-3. Developing your contribution.
+4. Developing your contribution.
 
    First start by ensuring you have the latest changes on the ``main``
    branch.
@@ -100,7 +115,7 @@ Here is a short summary of how to do developments with `sisl`.
    Be sure to document your changes, and write sensible documentation
    for the API.
 
-4. To submit your contribution:
+5. To submit your contribution:
 
    * Push your changes back to your fork on GitHub:
 
@@ -117,7 +132,7 @@ Here is a short summary of how to do developments with `sisl`.
      Please add one in ``/changes/`` by following the instructions found in
      the ``README.rst`` there.
 
-5. Review process.
+6. Review process.
 
    The maintainers of `sisl` will do their best to respond as fast as possible.
    But first ensure that the CI runs successfully, if not, maintainers will likely
@@ -147,3 +162,44 @@ Contribute additional tests
 Additional test files should be added to `this repository <sisl-files_>`_.
 Please follow the guidelines there, or open up an issue at that repository
 for specific details.
+
+
+Contribute to the docs
+----------------------
+
+To contribute to the documentation one needs to install `pandoc` first (see point
+**1.** in **Summary of development process** above). Then follow these steps:
+
+1. Sitting inside the `sisl` tree, install the `sisl` documentation via:
+
+   .. code:: bash
+
+      pip install -e .[docs]
+
+2. Download tutorial files outside the `sisl` repository:
+
+   .. code:: bash
+
+      git submodule init
+      git submodule update
+
+3. **OPTIONAL** : If your are not contributing specifically to the notebooks,
+   you may consider deactivating their compilation by commenting out (or eliminating)
+   this line ``"nbsphinx",`` in ``sisl/docs/conf.py`` file. This will enormously speed up
+   building times (see below).
+
+4. Within the docs folder (``sisl/docs``) do:
+
+   .. code:: bash
+
+      make html
+
+   This will build the documentation in the ``sisl/docs/build/html`` folder. Open any
+   **.html** file sitting there in your browser to visualize the built docs.
+
+5. Now you can contribute to the documentation by modifying any **.rst** file (reStructuredText
+   , or reST) sitting in ``sisl/docs``. Then build again (``make html``) and check your changes
+   in the browser.
+
+6. Once happy with your changes, *push* them to your fork and create a PR following the
+   instructions **4.** and **5.** in **Summary of development process**.


docs/dev/index.rst
```diff
@@ -32,8 +32,23 @@ Summary of development process
 
 Here is a short summary of how to do developments with `sisl`.
 
+1. You may need to install some packages system-wide, in case you do not have these:
 
-1. If you are a first time contributor, you need to clone your forked repository
+   *  For any type of contribution, ``gfortran`` :
```
pfebrer, line 37: What you need is a fortran compiler, not `gfortran` ;)

Then for the instructions, the installing command is specific to Ubuntu / Debian. I would add some tabs like the ones in the installation page (https://zerothi.github.io/sisl/quickstart/install.html).

See here for what is the "code" to add tabs: https://zerothi.github.io/sisl/quickstart/install.html

I would at least add a tab for `conda` (conda install -c conda-forge compilers), one for ubuntu/debian and one for MacOS

docs/dev/index.rst
```diff
@@ -32,8 +32,23 @@ Summary of development process
...
+
+      .. code:: bash
+
+         sudo apt-get install gfortran
+
+   *  And, in case you want to specifically contribute to the docs, `pandoc` (version between
```
pfebrer, line 43: Same for pandoc, the instructions are specific to linux. I would add the installation with `conda` and then for others just link to the pandoc installation page: https://pandoc.org/installing.html

Actually I think the best thing we can do is probably to recommend begginners to use `conda`, then we don't have to specify instructions for each platform.

I would also remove the version numbers, it will eventually get outdated and no one will fix it :)

docs/dev/index.rst
```diff
@@ -147,3 +162,44 @@ Contribute additional tests
 Additional test files should be added to `this repository <sisl-files_>`_.
 Please follow the guidelines there, or open up an issue at that repository
 for specific details.
+
+
+Contribute to the docs
+----------------------
+
+To contribute to the documentation one needs to install `pandoc` first (see point
+**1.** in **Summary of development process** above). Then follow these steps:
```
pfebrer, line 171: I would probably add a link instead of saying "above". See here for how to add links to sphinx: https://stackoverflow.com/questions/4385315/how-to-make-an-internal-hyperlink-in-sphinx-documentation

docs/dev/index.rst
```diff
@@ -147,3 +162,44 @@ Contribute additional tests
...
+
+   .. code:: bash
+
+      git submodule init
+      git submodule update
+
+3. **OPTIONAL** : If your are not contributing specifically to the notebooks,
+   you may consider deactivating their compilation by commenting out (or eliminating)
+   this line ``"nbsphinx",`` in ``sisl/docs/conf.py`` file. This will enormously speed up
+   building times (see below).
```
pfebrer, line 189: We need to just provide an option like ``--no-notebooks``, the current way is too hacky to get documented :)

docs/dev/index.rst
```diff
@@ -147,3 +162,44 @@ Contribute additional tests
...
+
+4. Within the docs folder (``sisl/docs``) do:
+
+   .. code:: bash
+
+      make html
+
+   This will build the documentation in the ``sisl/docs/build/html`` folder. Open any
+   **.html** file sitting there in your browser to visualize the built docs.
```
pfebrer, line 198: I would specify that the `index.html` is the "home page".

docs/dev/index.rst
```diff
@@ -147,3 +162,44 @@ Contribute additional tests
...
+
+5. Now you can contribute to the documentation by modifying any **.rst** file (reStructuredText
+   , or reST) sitting in ``sisl/docs``. Then build again (``make html``) and check your changes
+   in the browser.
```
pfebrer, line 202: You can contribute in any way :) Perhaps we could phrase it like: "The easiest thing that you can do is to modify one of the rst files"

docs/dev/index.rst
```diff
@@ -147,3 +162,44 @@ Contribute additional tests
...
+
+6. Once happy with your changes, *push* them to your fork and create a PR following the
+   instructions **4.** and **5.** in **Summary of development process**.
```
pfebrer, line 205: Again, probably better to link to the Summary of development process. Having explicit numbers is bound to the reference getting outdated at some point in the future :)

pfebrer (changes requested):
Here's your first review of sisl changes :)

The first reviews are the ones with most comments, don't worry :smile:

docs/dev/index.rst
ialcon: > See here for what is the "code" to add tabs: https://zerothi.github.io/sisl/quickstart/install.html

You mean to look at the corresponding `sisl/docs/quickstart/install.rst` to see how the tabs are implemented, no?

> I would at least add a tab for conda (conda install -c conda-forge compilers), one for ubuntu/debian and one for MacOS

So the command you wrote `conda install -c conda-forge compilers` is the one we need? I guess this is a command that will install a number of compilers, including those for fortran, no?

docs/dev/index.rst
ialcon: > Actually I think the best thing we can do is probably to recommend begginners to use conda, then we don't have to specify instructions for each platform.

So maybe it is enough to mention that a fortran compiler and `pandoc` need to be installed, and recommend to use `conda` to install these, but without providing the specific command to do so? (e.g. `conda install -c conda-forge compilers`, etc)

docs/dev/index.rst
ialcon: Should I remove **1.**? Also, the link should point to the **Summary of development process** or to **1.** subsection?

docs/dev/index.rst
ialcon: I guess this option should be for the building (`make html`) of the docs, no? So, until then, I remove this paragraph, correct?

docs/dev/index.rst
ialcon: So the link should point to **Summary of...** and not to **4.**, **5.**...

docs/dev/index.rst
pfebrer: Let's keep it here to remind ourselves that it needs to be implemented, and we implement it before merging this PR

docs/dev/index.rst
pfebrer: Yes, I think just specifying the dependencies is better. If someone wants to develop in sisl, they will find how to install them.

The dev dependencies are actually:
- C and fortran compilers
- CMake
- Pandoc (for documentation)

You just happened to already have valid versions of a C compiler and CMake.

pfebrer: Maybe you could add them here: https://zerothi.github.io/sisl/quickstart/install.html as development dependencies. Then they would be more visible. And in the dev/page you simply link to them.

I think it could be worth it to in this dev instructions have the full instructions for setting things up with conda, and then you could have something like:

------------------------------

1. Install the development dependencies. They can be found here (link). If you are in a conda environment, installing them is as simple as:

```
conda install -c conda-forge compilers cmake pandoc
```

----------------------------------------

At the beggining of the instructions you could have something like:

------------------------------------

You are free to choose your development environment, but we recommend using a conda virtual environment because things are very easy to set up. To create one, install `miniforge` (https://github.com/conda-forge/miniforge) and then create a new python environment. E.g. to create an environment named `sisl_dev` with the latest python version:

```
conda create -n sisl_dev python
```

Then activate it like:

```
conda activate sisl_dev
```

docs/dev/index.rst
pfebrer: Yes I think so

docs/dev/index.rst
```diff
@@ -25,15 +25,36 @@ In the following of this document you will find information related to the speci
 of making a development workflow for `sisl`.
 
 
-
+.. _devSummary:
 Summary of development process
 ------------------------------
 
+You are free to choose your development environment, but we recommend using a conda virtual
```
zerothi, line 32: do we recommend a `conda`? I am pretty agnostic here... Of course, conda makes things easier in terms of compilers. But it could be tabbed entries?

What's your thought here?

docs/dev/index.rst
```diff
@@ -25,15 +25,36 @@ In the following of this document you will find information related to the speci
...
+environment because things are very easy to set up. To create one, install `miniforge`_
+and then create a new python environment. E.g. to create an environment named ``sisl_dev``
+with the latest python version do:
+
+.. code:: bash
+
+   conda create -n sisl_dev python
```
zerothi, line 39: Why not have everything installed in this line? At least for the conda based solution?

docs/dev/index.rst
zerothi: Actually, we already have this implemented:

Do `_SISL_DOC_SKIP=notebook make`.

We can't put stuff on the `make` command line as its intercepted by `make`, and otherwise used as a target.

docs/dev/index.rst
```diff
@@ -147,3 +168,45 @@ Contribute additional tests
...
+   building times (see below).
+
+4. Within the docs folder (``sisl/docs``) do:
+
+   .. code:: bash
+
+      make html
+
+   This will build the documentation in the ``sisl/docs/build/html`` folder. Open any
+   **.html** file sitting there in your browser to visualize the built docs. Note that
```
zerothi, line 204: Probably, use `.html`, instead of boldfacing it.

docs/dev/index.rst
```diff
@@ -147,3 +168,45 @@ Contribute additional tests
...
+   `index.html` is the "home page" of the documentation.
+
+5. The easiest thing that you can do now is to modify one of the **.rst** files
```
zerothi, line 207: same here.

docs/dev/index.rst
zerothi: I would also recommend to only open the `index.html` since that is more explicit and what users would expect.

docs/quickstart/install.rst
```diff
@@ -26,6 +26,12 @@ Optional dependencies:
 - `matplotlib`_
 - `plotly`_ (for advanced visualization)
 
+Dependencies for development:
+
+- ``C`` and ``fortran`` compilers
+- `cmake`_
```
zerothi, line 32: while we're at it, we need cmake at least 3.21

zerothi (changes requested):
minor comments. I can change many of them, but please comment ;)

Thanks!

docs/dev/index.rst
ialcon: you mean to have different entries for different `conda` version? or for different ways to create the environment, including `conda`? (e.g. `pip`)? @pfebrer proposed the use of `conda` so his opinion is more relevant here

docs/dev/index.rst
ialcon: But what is the issue of opening another file, like the ones I modified, if you intend to contribute to other parts of the documentation? Or you mean to open the `index.html` and from there navigate to the document that one has modified?

docs/quickstart/install.rst
ialcon: Ok, so that should be mentioned in that line

docs/dev/index.rst
ialcon: also... further down we are recommending to use `pip` for other tasks such as installing `pre-commit`:
```bash
python -m pip install pre-commit
pre-commit install
```
or installing `sisl` in editable mode:
```bash
python -m pip install -e .
```
Isn't all this kind of contradictory with the above lines recommending `conda`?

ialcon:
So, @zerothi, do you want to proceed applying these changes or I do it? Also, I guess we should wait for @pfebrer comments, specially on the `conda` thing

docs/dev/index.rst
zerothi: Well, not exactly, conda doesn't allow installing editably, however, the best practice would of course be to install dependencies through conda, then only sisl with pip. But...

pfebrer:
Regarding the conda discussion, I think the easiest recommendation for begginers is:
- Setup environment with `conda`
- Install sisl with `pip -e`

zerothi:
Ok, agreed. That is what we have now! :)

ialcon:
Ok, that was confusing me because in my mind if one sets up a `conda` environment, then the packages need to be installed with `conda` (and the other way around with `pip`). But what you are saying (if I get it right) is to install `sisl` in editable mode with `pip` within the `conda` environment... right?

zerothi:
Agreed, one should always strive to do it like that! But, one can't do editable installs using conda... That's the catch...

pfebrer:
Actually `conda` tracks the packages that `pip` installs. So, even if it is not the most purist thing, it is not super bad to install things with `pip` in a conda environment. After all, it is not the end of the world if you break your environment because of untracked dependencies, you just destroy it and create a new one :) That's what environments are for, so that you don't have to destroy your OS when you make a mistake haha

pfebrer:
By the way, I thought that we could just keep an `environment.yaml` file somewhere with the dependencies for development, then we could tell people to just:

```
conda env create -f environment.yaml
```

Maybe it is possible to do something like:

```
conda env create -f https://sisl.hosted/environment.yaml
```

I don't know.

ialcon:
> By the way, I thought that we could just keep an `environment.yaml` file somewhere with the dependencies for development, then we could tell people to just:
> 
> ```
> conda env create -f environment.yaml
> ```
> 
> Maybe it is possible to do something like:
> 
> ```
> conda env create -f https://sisl.hosted/environment.yaml
> ```
> 
> I don't know.

That would be quite powerful... In fact, couldn't this make sense also for regular `sisl` installations?? So that you had different `environment.yaml` files for each type of installation? That would definitely facilitate not only the general access to *use* `sisl`, but also the access to new contributors (at least, in part)

pfebrer:
Yes but in principle for normal sisl installations that should be just:

```
conda install sisl
```

and all the dependencies are handled, so not much point in having a file. The main problem is that you can't do something like

```
conda install sisl[dev]
```

in conda, so you have to rely on environment files. This is linked to the issue I opened in conda-forge, where the recommendation is to keep these files hosted somewhere: https://github.com/conda-forge/conda-forge.github.io/issues/2043

ialcon:
> Yes but in principle for normal sisl installations that should be just:
> 
> ```
> conda install sisl
> ```
> 
> and all the dependencies are handled, so not much point in having a file. The main problem is that you can't do something like

Ok, but this is not yet in place right? I mean, all the dependencies specified in the installation page need to be installed separately, no? Or are these already included upon doing `conda install sisl`?

pfebrer:
No, if you install sisl it installs all the dependencies that you need to run sisl. A different thing is the dependencies that you need to develop sisl.

zerothi:
> does this also apply if you install `sisl` with `pip`? (be it a standard release or the development version in editable mode)? I ask this because, if that was the case, I think it should be noted somewhere that one does not need to install all the dependencies specified here _https://sisl.readthedocs.io/en/latest/quickstart/install.html_, because these are automatically installed with `sisl`. I am pretty sure this is not specified anywhere...

You're right, we could make this clearer in the docs.

ialcon:
I actually installed these separately... :tired_face:

zerothi:
I moved some things around :) thanks for noticing.

PR #917, opened 2025-05-13
changed Point -> Coord in neighbor lists

zerothi:
I just stumpled on this, @pfebrer comments?

### DIFF
--- a/docs/api/geom/neighbors.rst
+++ b/docs/api/geom/neighbors.rst
@@ -39,21 +39,21 @@ neighbor list.
 
 The following table summarizes the properties of each type of neighbor list:
 
-+----------------------+------------------------+---------------------+-----------------------+
-| Class                | Neighbors for          | :math:`I < J`       |  Items                |
-+======================+========================+=====================+=======================+
-| `FullNeighborList`   | All atoms              | :fas:`circle-xmark` |  `AtomNeighborList`   |
-+----------------------+------------------------+---------------------+-----------------------+
-| `UniqueNeighborList` | All atoms              | :fas:`circle-check` |         N/A           |
-+----------------------+------------------------+---------------------+-----------------------+
-| `PartialNeighborList`| Selected atoms         | :fas:`circle-xmark` |   `AtomNeighborList`  |
-+----------------------+------------------------+---------------------+-----------------------+
-| `AtomNeighborList`   | One atom               | :fas:`circle-xmark` |         N/A           |
-+----------------------+------------------------+---------------------+-----------------------+
-| `CoordsNeighborList` | Coordinates n space    | :fas:`circle-xmark` |   `CoordNeighborList` |
-+----------------------+------------------------+---------------------+-----------------------+
-| `CoordNeighborList`  | Coordinate in space    | :fas:`circle-xmark` |         N/A           |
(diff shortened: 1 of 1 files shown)


src/sisl/geom/_neighbors/_neighborlists.py
```diff
@@ -344,10 +344,10 @@ class AtomNeighborList(Neighbors):
         The lists that, when iterated, return `AtomNeighborList` objects.
     """
 
-    #: The atom for which the neighbors are stored.
     atom: int
+    """The atom for which the neighbors are stored."""
```
pfebrer, line 348: Why do you want it like this? It looks worse to me

pfebrer:
It is just a matter of preference, so if you prefer it like this it is fine for me. Personally I don't have a preference for one or the other, but I do think that changing names in the public API should not be done lightly. So if you think it really makes much more sense, then OK, but if not I would keep it as is.

zerothi:
> It is just a matter of preference, so if you prefer it like this it is fine for me. Personally I don't have a preference for one or the other, but I do think that changing names in the public API should not be done lightly. So if you think it really makes much more sense, then OK, but if not I would keep it as is.

Thats why I would do it rather soonish, it hasn't been picked up yet. So better now than in some months/years.

src/sisl/geom/_neighbors/_neighborlists.py
zerothi: Agreed, but when building the docs, I get errors on missing attribute docs.... This is the PEP suggested way of adding docs-strings to attributes.

pfebrer: Hmm this is very strange, last time I checked sphinx took the documentation from the previous line's comment if it started with `#:`. This is no longer true?

docs/api/geom/neighbors.rst
```diff
@@ -39,21 +39,21 @@ neighbor list.
...
++======================+========================+=====================+=======================+
+| `FullNeighborList`   | All atoms              | :fas:`circle-xmark` |  `AtomNeighborList`   |
++----------------------+------------------------+---------------------+-----------------------+
+| `UniqueNeighborList` | All atoms              | :fas:`circle-check` |         N/A           |
++----------------------+------------------------+---------------------+-----------------------+
+| `PartialNeighborList`| Selected atoms         | :fas:`circle-xmark` |   `AtomNeighborList`  |
++----------------------+------------------------+---------------------+-----------------------+
+| `AtomNeighborList`   | One atom               | :fas:`circle-xmark` |         N/A           |
++----------------------+------------------------+---------------------+-----------------------+
+| `CoordsNeighborList` | Coordinates n space    | :fas:`circle-xmark` |   `CoordNeighborList` |
```
pfebrer, line 53: Missing "i"

zerothi: thanks!

src/sisl/geom/_neighbors/_neighborlists.py
zerothi: Perhaps `#:` only works for module attributes (constants), and not class attributes. It isn't fully clear to me either, because in `constant.py` I do `#:`, and I don't get errors there!

Issue #947, opened 2025-09-16
EllipticalCylinder


zerothi:
**Describe the bug**

**Code to reproduce problem**
```python
ellipticalcylinder_1 = sisl.shape.EllipticalCylinder(v=50., h=100., center=(0,0,0))
print(ellipticalcylinder_1.within([0., 0., 100.]))   # True ---> WEIRD (should be ok only up to 50.)
print(ellipticalcylinder_1.within([0., 0., 101.]))   # False ---> (seems like 100. is the upper limit)  
print(ellipticalcylinder_1.within([0., 0., -100]))   # True ---> WEIRD (should be ok only up to -50.)
print(ellipticalcylinder_1.within([0., 0., -1000000000.]))   # True ---> SUPER WEIRD (seems like there is no lower limit!)
```

zerothi:
Fixed by #948

PR #948, opened 2025-09-17
Fixed bugs in EllipticalCylinder and indices_in_cylinder.

tanocalogero:
In indices_in_cylinder():
* there was no check on lower bounds along the cylinder axis;
* the height of the cylinder was doubled with respect to the value initialized in EllipticalCylinder class;
* the input R was defaulted to be a single float. This fails when radius has shape (2,) or (2,3).

In _cylinder.py:
* every call to EllipticalCylinder.__class__ was setting "axes" argument to "self.center".


zerothi:
Done! Fixes #947 

Thanks @tanocalogero !!!

PR #955, opened 2025-09-30
954 modified `Atoms.index`

### LINKED ISSUE #954: Improve `Atoms.index` usability
Currently, `geom.atoms.index("H")` raises a `KeyError("Could not find atom in the list of atoms.")` if no atoms of that species exist. This makes workflows like subsetting by element more cumbersome, since users must wrap calls in try/except.

It would be more intuitive if:
- `Atoms.index(symbol)` returned an empty array `[]` when no atoms are found.
- `Atoms.index` accepted multiple species in one call, e.g.:
```python
geom.atoms.index("H", "C")
geom.atoms.index(["H", "C"])
```
returning the combined indices.

Both changes would simplify common tasks like extracting subsets (`CH`, `CN`, etc.) while remaining backwards-compatible.

Happy to open a PR if this direction is welcome.

tfrederiksen:
- [x] Closes #954
 - [x] Added tests for changed function
 - [x] Changes documented in `changes/<pr-num>.<type>.rst`

### DIFF
--- a/src/sisl/_core/atom.py
+++ b/src/sisl/_core/atom.py
@@ -770,7 +770,21 @@ class Atoms:
         Returns
         -------
         numpy.ndarray
-            Indices of all matching atoms. Returns an empty array if no match is found.
+            Unique (sorted) indices of all matching atoms.
+            Returns an empty array if no match is found.
+
+        Examples
+        --------
+        >>> atoms = Atoms(["C", "H", "Au"])
+        >>> idx_C = atoms.index("C")
+        >>> idx_C = atoms.index(0)
+        >>> idx_CH = atoms.index(["C", "H"])
+        >>> idx_Au = atoms.index(Atom(79))
+
+        This can be useful to get a subset of a geometry, e.g.,
+        >>> geom = Geometry(...)
+        >>> idx_CH = geom.atoms.index(["C", "H"])
+        >>> geom_CH = geom.sub(idx_CH)
 
         Notes
         -----
--- a/src/sisl/_core/atom.py
+++ b/src/sisl/_core/atom.py
@@ -758,13 +758,14 @@ class Atoms:
         uZ = _a.arrayi([a.Z for a in self.atom])
         return uZ[self.species]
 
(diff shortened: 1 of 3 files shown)


src/sisl/_core/atom.py
```diff
@@ -758,9 +758,40 @@ def Z(self):
         uZ = _a.arrayi([a.Z for a in self.atom])
         return uZ[self.species]
 
-    def index(self, atom):
-        """Return the indices of the atom object"""
-        return (self._species == self.species_index(atom)).nonzero()[0]
+    def index(self, *atoms):
```
zerothi, line 761: Having seen this interface, I think it's better to have `atoms` as a single argument. In this way, we can enable keyword arguments later (well, your current API does not limit that, but it's very inconsistent with the rest of sisl).

Instead it should just be:
```python
def index(self, atoms)
```

src/sisl/_core/atom.py
```diff
@@ -758,9 +758,40 @@ def Z(self):
...
+        """Return indices of atoms of the given atom type.
+
+        Parameters
+        ----------
+        atoms : str, Atom, or lists hereof
```
zerothi, line 766: Since we should still use `species_index` (for improved usage), it could also be an int.

Maybe it needs typing, but that code hasn't got much attention, so we'll leave that.

tfrederiksen:
> I can adjust the PR if you'd like? Then I'll ping you for comments? Thanks!

Maybe I can first give it another try taking into account your comments?

By the way, I've always been a bit confused about `species` vs `atom`. What's the distinction?

src/sisl/_core/atom.py
zerothi: So to use that, you'll have to wrap it in a try-catch... :(

tfrederiksen:
Should we also add an example in the docstring how to use this?

zerothi:
@tfrederiksen do you know why it stopped? I think it's good, so I'll just merge, unless you object?

zerothi:
Thanks!

Issue #995, opened 2026-06-08
Non-standard object identity test in `SparseOrbital.iter_nnz`


ahkole:
While browsing through the implementation of `SparseOrbital.iter_nnz` I noticed that one of the object identity tests was the non-standard `not object is None` instead of `object is not None`, see https://github.com/zerothi/sisl/blob/fc153f0363a2fd14001641c8b6ca370b626495de/src/sisl/_core/sparse_geometry.py#L1500. This is very minor and doesn't really affect the logic, but I was wondering if this should be changed nonetheless to match the standard.

zerothi:
sure, feel free to make a PR :) thx!

zerothi:
fixed via #996

PR #996, opened 2026-06-08
Refactor non-standard object identity test

ahkole:
- [x] Closes https://github.com/zerothi/sisl/issues/995
 - [x] Changes documented in `changes/996.dev.rst`
 

It also refactors all other occurrences of `not [object_name] is None` to `[object_name] is not None`.

### DIFF
--- a/changes/996.change.rst
+++ b/changes/996.change.rst
@@ -1,0 +1,1 @@
+Fix non-standard object identity test in `SparseOrbital.iter_nnz`.
--- a/changes/996.change.rst
+++ b/changes/996.change.rst
@@ -1,1 +1,0 @@
-Fix non-standard object identity test in `SparseOrbital.iter_nnz`.
--- a/changes/996.dev.rst
+++ b/changes/996.dev.rst
@@ -1,0 +1,3 @@
+Refactor non-standard object identity tests.
+
+Refactors `not object is None` to `object is not None` to improve readability and adhere to the standard.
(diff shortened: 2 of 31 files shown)


zerothi:
If you think this should be fixed, then there are plenty of other places where this is done:
```
grep "is None" **/*.py | grep not | grep -v "is not None" 
```

ahkole:
> If you think this should be fixed, then there are plenty of other places where this is done:
> 
> ```
> grep "is None" **/*.py | grep not | grep -v "is not None" 
> ```

Ah okay, I see now yes. I personally think `object is not None` is easier to read than `not object is None`. If you agree maybe I could just replace all of them?

zerothi:
yeah, it would make sense to catch them all, while at it. Also, I wouldn't call it a fix, but a refactor/maintenance. So your changes should be named `<>.dev.rst`, so users don't think it will affect their code.

ahkole:
I have used `sed` to replace all occurrences of `not [object_name] is None` to `[object_name] is not None`. Hopefully that should catch all of them.

I also checked after the replacement with your suggested `grep` command and the only remaining matches are lines that I agree should not be changed, either because they were false positives or the current expression is actually more readable.

Finally I also updated the PR title and the news fragment to call this a refactor instead of a fix.

zerothi (approved):

zerothi:
thank!

PR #1002, opened 2026-07-20
BLD: Fix build failure with scikit-build-core >= 0.10

hoodmane:
`pyproject.toml` sets `cmake.verbose = true`. It was renamed in scikit-build-core 0.10 to `build.verbose`.


zerothi:
Thanks!

